PR #3, opened 2022-11-18
Add automatic PR labels

bennettgoble:
Automatically label PRs with a handful of code-area based groups.

### DIFF
--- a/.github/labeler.yaml
+++ b/.github/labeler.yaml
@@ -1,0 +1,78 @@
+llappearance:
+  - indra/llappearance/**/*
+  
+llaudio:
+  - indra/llaudio/**/*
+
+llcharacter:
+  - indra/llcharacter/**/*
+
+llcommon:
+  - indra/llcommon/**/*
+
+llcorehttp:
+  - indra/llcorehttp/**/*
+
+llcrashlogger:
+  - indra/llcrashlogger/**/*
+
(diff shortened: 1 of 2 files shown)


.github/labeler.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - indra/llrender/**/*
+
+llui:
+  - indra/llui/**/*
+
+llwindow:
+  - indra/llwindow/**/*
+
+llxml:
+  - indra/llxml/**/*
```
nat-goodspeed, line 59: Woof, my impulse here is to commit a script that will generate (most of) the `labeler.yaml` file, possibly appending the lines below from, say, a `labeler.yaml.in` file. This is a file I will _certainly_ forget to modify if in future I add a new `indra` subdirectory.

That said, we don't need to block this PR waiting for that.

nat-goodspeed (approved):

bennettgoble:
> There is directories missing. At least
> maybe some more. Is this intended?

Yes, this is a fast first pass. More can be added as appropriate as a follow-up.

bennettgoble:
> Please explain in a bit more detail what this accomplishes?

This change will result in PRs receiving a set of automatic labels based on the files that have been changed. This is useful for sifting through PRs after dozens/hundreds/thousands have been merged.

PR #4, opened 2022-11-18
Close stale PRs

bennettgoble:
Mark PRs as stale after 60 days, close them automatically if no response after 7 days.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,24 @@
+name: Stale PRs 
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 0 * * *
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v6
+        id: stale
+        with:
+          stale-pr-message: This pull request is stale because it has been open 60 days with no activity. Remove stale label or comment or it will be closed in 7 days
+          days-before-stale: 60
+          days-before-close: 7 
+          exempt-pr-labels: blocked,must,should,keep
+          stale-pr-label: stale
+      - name: Print outputs
+        run: echo ${{ join(steps.stale.outputs.*, ',') }}


Nicky-D:
Signal, are you sure 60 days is enough for LL to react? I suggest to bump this up to at least 365 days.
Okay, my bad jokes aside: LL can be really slow to react, I know you guys care but are swamped with the normal work. I do understand.
But it is very often LL not reacting, sometimes month. To this day I still have some PRs on bitbucket that are so stale I cannot even delete them (because Bitbucket is dense or whatever).
Or the Linux PR that sat so long, it got killed by the HG>GIT switch. Waiting on LL input.

On the other hand I think some automatic cleanup would be nice, again look at Bitbucket with some PRs that are years old and not even the original author pinging it.

LL should be careful introducing this though. Because then you need to live up to your own standards. Or people will get rightfully annoyed having to ping their PRs or getting them closed. Though it would be fantastic if turnaround gets faster.
And also to not just sound negative, things got a better lately, Andrey for example has been very reposive

bennettgoble:
@nat-goodspeed I think that's exactly the point: this automated behavior should spur LL to be more responsive in getting PRs merged. All participants of a PR will receive notifications when the bot tags PRs as stale and people can, rightful, yelp "hey: I'm waiting on LL, here." We can do a lot better, and I'd like to adopt tools and practices assuming good intent and outcome.

bennettgoble:
> Signal, are you sure 60 days is enough for LL to react? I suggest to bump this up to at least 365 days.

Last time I checked there's PRs on BitBucket that have been open for over two years 😉

Nicky-D:
> > Signal, are you sure 60 days is enough for LL to react? I suggest to bump this up to at least 365 days.
> 
> Last time I checked there's PRs on BitBucket that have been open for over two years 😉
image

There was also that one from Jonathan Yap (spelling) that been loitering there for I think 2 years but it finally did disappear.

bennettgoble:
If two months (60 days) is too ambitious I could bump it to 90.

Nicky-D:
> If two months (60 days) is too ambitious I could bump it to 90.

From a contributors view 60 days of radio silence if already way too high. I don't think there is a "right" value here. Probably a life and learn over time.
Probably it's best to start with a time LL employees are comfortable with and everyone trying their best to be as responsive and fast as the can

bennettgoble:
I received a begrudging go-ahead in slack. Let's try this out: if it's painful it should be removed.

PR #5, opened 2022-11-19
Fix a thread safety issue in the GL image worker.

sldevel:
LLViewerTexture::mNeedsCreateTexture needs to be an attomic bool since it is written both in the main thread and in the GL image worker thread.

We can now enable threaded bump maps creation as a result of this fix.


sldevel:

sldevel:
recheck

PR #8, opened 2022-11-21
Update message template URL after move to GitHub

Ansariel:
Master message template URL wasn't updated during the transition from Bitbucket to GitHub.


Ansariel:

bennettgoble:
Thanks for the PR, @Ansariel .There's also a reference to the template URL in Variables.cmake, which would be good to update as well.

Ansariel:
Changed the default URL in the python script as well.

PR #15, opened 2022-11-29
more puppetry fixes and attempt to add Puppetry support for per-joint scale

AndrewMeadows:
This PR fixes some bugs that were preventing correct harvest of Puppetry data for streaming through the server.  Also adds Puppetry support for modifying per-joint scale.

### DIFF
--- a/indra/llappearance/llavatarappearance.cpp
+++ b/indra/llappearance/llavatarappearance.cpp
@@ -720,8 +720,6 @@ BOOL LLAvatarAppearance::allocateCharacterJoints( U32 num )
 //-----------------------------------------------------------------------------
 BOOL LLAvatarAppearance::buildSkeleton(const LLAvatarSkeletonInfo *info)
 {
-    LL_DEBUGS("BVH") << "numBones " << info->mNumBones << " numCollisionVolumes " << info->mNumCollisionVolumes << LL_ENDL;
-
 	// allocate joints
 	if (!allocateCharacterJoints(info->mNumBones))
 	{
--- a/indra/llappearance/llavatarappearance.cpp
+++ b/indra/llappearance/llavatarappearance.cpp
@@ -720,13 +720,15 @@ BOOL LLAvatarAppearance::allocateCharacterJoints( U32 num )
 //-----------------------------------------------------------------------------
 BOOL LLAvatarAppearance::buildSkeleton(const LLAvatarSkeletonInfo *info)
 {
+    LL_DEBUGS("BVH") << "numBones " << info->mNumBones << " numCollisionVolumes " << info->mNumCollisionVolumes << LL_ENDL;
+
 	// allocate joints
 	if (!allocateCharacterJoints(info->mNumBones))
 	{
 		LL_ERRS() << "Can't allocate " << info->mNumBones << " joints" << LL_ENDL;
 		return FALSE;
 	}
-	
+
 	// allocate volumes
 	if (info->mNumCollisionVolumes)
 	{
--- a/indra/llcharacter/CMakeLists.txt
+++ b/indra/llcharacter/CMakeLists.txt
@@ -104,6 +104,11 @@ if (LL_TESTS)
     PROPERTIES
     LL_TEST_ADDITIONAL_LIBRARIES "${BOOST_THREAD_LIBRARY}"
   )
+  set_source_files_properties(
+    llik.cpp
+    PROPERTIES
+    LL_TEST_ADDITIONAL_SOURCE_FILES lljoint.cpp
+  )
   LL_ADD_PROJECT_UNIT_TESTS(llcharacter "${llcharacter_TEST_SOURCE_FILES}")
 endif (LL_TESTS)
 
--- a/indra/llcharacter/llik.cpp
+++ b/indra/llcharacter/llik.cpp
@@ -2274,7 +2274,7 @@ void LLIK::Solver::rebuildAllChains()
             }
             if (flags & CONFIG_FLAG_LOCAL_SCALE)
             {
-                joint->setLocalScale(config.getLocalPos());
+                joint->setLocalScale(config.getLocalScale());
             }
             continue;
         }
--- a/indra/llcharacter/llik.cpp
+++ b/indra/llcharacter/llik.cpp
@@ -35,6 +35,7 @@
 
 #include "llik.h"
 
+#include "lljoint.h"
 #include "llmath.h"
 #include "v3math.h"
 #include "llquaternion.h"
@@ -147,27 +148,33 @@ F32 compute_clamped_angle(F32 angle, F32 min_angle, F32 max_angle)
 void LLIK::Joint::Config::setLocalPos(const LLVector3& pos)
 {
     mLocalPos = pos;
-    mFlags |= FLAG_LOCAL_POS;
+    mFlags |= CONFIG_FLAG_LOCAL_POS;
 }
 
 void LLIK::Joint::Config::setLocalRot(const LLQuaternion& rot)
 {
     mLocalRot = rot;
     mLocalRot.normalize();
-    mFlags |= FLAG_LOCAL_ROT;
+    mFlags |= CONFIG_FLAG_LOCAL_ROT;
+}
+
+void LLIK::Joint::Config::setLocalScale(const LLVector3& scale)
+{
+    mLocalScale = scale;
+    mFlags |= CONFIG_FLAG_LOCAL_SCALE;
 }
 
 void LLIK::Joint::Config::setTargetPos(const LLVector3& pos)
 {
     mTargetPos = pos;
-    mFlags |= FLAG_TARGET_POS;
+    mFlags |= CONFIG_FLAG_TARGET_POS;
 }
 
 void LLIK::Joint::Config::setTargetRot(const LLQuaternion& rot)
 {
     mTargetRot = rot;
     mTargetRot.normalize();
-    mFlags |= FLAG_TARGET_ROT;
+    mFlags |= CONFIG_FLAG_TARGET_ROT;
 }
 
 void LLIK::Joint::Config::updateFrom(const Config& other_config)
@@ -182,19 +189,23 @@ void LLIK::Joint::Config::updateFrom(const Config& other_config)
         // find and apply all parameters in other_config
         if (other_config.hasLocalPos())
         {
-            setLocalPos(other_config.getLocalPos());
+            setLocalPos(other_config.mLocalPos);
         }
         if (other_config.hasLocalRot())
         {
-            setLocalRot(other_config.getLocalRot());
+            setLocalRot(other_config.mLocalRot);
         }
         if (other_config.hasTargetPos())
         {
-            setTargetPos(other_config.getTargetPos());
+            setTargetPos(other_config.mTargetPos);
         }
         if (other_config.hasTargetRot())
         {
-            setTargetRot(other_config.getTargetRot());
+            setTargetRot(other_config.mTargetRot);
+        }
+        if (other_config.hasLocalScale())
+        {
+            setLocalScale(other_config.mLocalScale);
         }
         if (other_config.constraintIsDisabled())
         {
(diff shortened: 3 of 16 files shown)


indra/llcharacter/CMakeLists.txt
```diff
@@ -104,6 +104,11 @@ if (LL_TESTS)
     PROPERTIES
     LL_TEST_ADDITIONAL_LIBRARIES "${BOOST_THREAD_LIBRARY}"
   )
+  set_source_files_properties(
+    llik.cpp
+    PROPERTIES
+    LL_TEST_ADDITIONAL_SOURCE_FILES lljoint.cpp
+  )
```
AndrewMeadows, line 111: This was required to add `lljoint.cpp` as a dependency of `llik_test.cpp`.

indra/llappearance/llavatarappearance.cpp
```diff
@@ -720,8 +720,6 @@ BOOL LLAvatarAppearance::allocateCharacterJoints( U32 num )
 //-----------------------------------------------------------------------------
 BOOL LLAvatarAppearance::buildSkeleton(const LLAvatarSkeletonInfo *info)
 {
-    LL_DEBUGS("BVH") << "numBones " << info->mNumBones << " numCollisionVolumes " << info->mNumCollisionVolumes << LL_ENDL;
-
```
AndrewMeadows, line 724: Oops, this was wrong.  I will revert this change.

indra/llcharacter/llik.cpp
```diff
@@ -147,27 +148,33 @@ F32 compute_clamped_angle(F32 angle, F32 min_angle, F32 max_angle)
...
     mLocalRot.normalize();
-    mFlags |= FLAG_LOCAL_ROT;
+    mFlags |= CONFIG_FLAG_LOCAL_ROT;
+}
+
+void LLIK::Joint::Config::setLocalScale(const LLVector3& scale)
+{
+    mLocalScale = scale;
+    mFlags |= CONFIG_FLAG_LOCAL_SCALE;
 }
```
AndrewMeadows, line 165: We always intended to add the ability to adjust the scale of a joint via Puppetry.  I'm adding it now, mostly because I had to fix some scale-related math for avatar models that use a non-unitary scale at the **LLJoint** context, and once I started digging into that I figured out what it would take to support Puppetry scale.

indra/llcharacter/llik.cpp
```diff
@@ -856,11 +867,22 @@ void LLIK::DoubleLimitedHinge::dumpConfig() const
...
+void LLIK::Joint::resetFromInfo()
+{
+    LLVector3 scale = mInfoPtr->getScale();
+    mLocalPos = mInfoPtr->getPosition().scaledVec(scale);
+    mBone = mInfoPtr->getEnd().scaledVec(scale);
     mLocalPosLength = mLocalPos.length();
+    // This is Correct: we do NOT store info scale in mLocalScale.
+    // mLocalScale represents Puppetry's tweak on  top of whatever is set in the info.
+    mLocalScale.set(1.0f, 1.0f, 1.0f);
 }
```
AndrewMeadows, line 886: `LLIK::Joint::resetFromInfo()` replaces the old `LLIK::Joint::reconfigure()` method below.

indra/llcharacter/llik.cpp
```diff
@@ -1299,20 +1313,24 @@ void LLIK::Joint::shiftPos(const LLVector3& shift)
...
+    mConfigFlags = 0;
+    // root Joint always has IK_FLAG_LOCAL_ROT_LOCKED set
+    mIkFlags = mParent ? 0 : IK_FLAG_LOCAL_ROT_LOCKED;
 }
 
 void LLIK::Joint::lockLocalRot(const LLQuaternion& local_rot)
 {
     mLocalRot = local_rot;
+    mIkFlags |= IK_FLAG_LOCAL_ROT_LOCKED;
     activate();
```
AndrewMeadows, line 1334: I had previously stored this "local_rot_locked" bit in its own bool data member, then I moved it to another recycled flag.  Here I'm moving it out into its own bit in `mIkFlags` data member: it is almost the same as when an **LLIK::Joint::Confi**g has specified the local rotation of the **LLIK::Joint**... but is slightly distinct because we also want the root **Joint** to be locked during the IK calculations, even when it hasn't been configured for a specific local rotation.  This detail matters when we harvest the data and stream (through server) only the joint properties that have changed.

indra/llcharacter/llik.cpp
```diff
@@ -1384,7 +1402,8 @@ void LLIK::Joint::setWorldPos(const LLVector3& pos)
 // this should only be called once before starting IK algorithm iterations
 void LLIK::Joint::setLocalPos(const LLVector3& pos)
 {
-    mLocalPos = pos;
+    mLocalPos = pos.scaledVec(mLocalScale);
+    mLocalPosLength = mLocalPos.length();
```
AndrewMeadows, line 1406: Here, and in a few other spots, we apply `mLocalScale `to `mLocalPos `(when either of them changes) before starting IK calculations.  This way we don't have to constantly recompute `mLocalPos` during IK.

indra/llcharacter/llik.cpp
```diff
@@ -1407,6 +1426,46 @@ void LLIK::Joint::setLocalRot(const LLQuaternion& new_local_rot)
...
+    LLVector3 inv_scale = mInfoPtr->getScale();
+    constexpr F32 MIN_INVERTABLE_SCALE = 1.0e-15f;
+    for (S32 i = 0; i < 3; ++i)
+    {
+        // verify scale component to avoid introducing NaN
+        inv_scale[i] = (inv_scale[i] > MIN_INVERTABLE_SCALE) ?  1.0f / inv_scale[i] : 0.0f;
+    }
+    pos.scaleVec(inv_scale);
+    return pos;
+}
```
AndrewMeadows, line 1467: At harvest time we want the local position in the **LLJoint** frame, so we need to back out any **LLJoint** scaling we applied before IK calculations were started.

indra/llcharacter/llik.h
```diff
@@ -55,22 +57,27 @@ namespace LLIK
...
-constexpr U8 MASK_POS = FLAG_TARGET_POS | FLAG_LOCAL_POS;
-constexpr U8 MASK_ROT = FLAG_TARGET_ROT | FLAG_LOCAL_ROT;
+// config flags
+constexpr U8 CONFIG_FLAG_LOCAL_POS = 1 << 0;
+constexpr U8 CONFIG_FLAG_LOCAL_ROT = 1 << 1;
+constexpr U8 CONFIG_FLAG_LOCAL_SCALE = 1 << 2;
+constexpr U8 CONFIG_FLAG_DISABLE_CONSTRAINT = 1 << 3;
+constexpr U8 CONFIG_FLAG_TARGET_POS = 1 << 4;
+constexpr U8 CONFIG_FLAG_TARGET_ROT = 1 << 5;
+constexpr U8 CONFIG_FLAG_HAS_DELEGATED = 1 << 6; // EXPERIMENTAL
```
AndrewMeadows, line 67: There are two sets of flags.  Those for properties set by the **LLIK::Joint::Config** class (above), and those that are the results of IK calculations (below).  Both sets of flags are combined to figure out what data needs to be harvested (see `LLIK::Joint::getHarvestFlags()` and where it is called).

indra/llcharacter/llik.h
```diff
@@ -384,15 +395,16 @@ class Joint
     using ptr_t = std::shared_ptr<Joint>;
     using child_vec_t = std::vector<ptr_t>;
 
-    Joint(S16 id, const LLVector3& local_pos, const LLVector3& bone);
+    Joint(LLJoint* info);
+    void resetFromInfo();
```
AndrewMeadows, line 399: I was loath to add this **LLJoint** dependency on **LLIK::Joint** class however I finally decided it was the way to go, rather than duplicate the data (e.g. `mDefaultLocalPos `and `mDefaultLocalScale`) in **LLIK::Joint**.  I felt the code was a bit more readable this way: the source of the default values is more explicit (and as a bonus we can reduce wasted memory).

indra/llcharacter/llik.h
```diff
@@ -523,7 +551,11 @@ class Solver
     LLVector3 computeReach(S16 to_id, S16 from_id) const;
 
     // Add a Joint to the skeleton.
-    void addJoint(S16 joint_id, S16 parent_id, const LLVector3& local_pos, const LLVector3& bone, const Constraint::ptr_t& constraint);
+    void addJoint(
+            S16 joint_id,
+            S16 parent_id,
+            LLJoint* info_ptr,
+            const Constraint::ptr_t& constraint);
```
AndrewMeadows, line 558: addJoint() now takes a raw` LLJoint*`.  The `local_pos `and `bone `(and scale!) are extracted from the **LLJoint** instance.

indra/llcharacter/lljoint.cpp
```diff
@@ -329,7 +329,7 @@ void LLJoint::removeAllChildren()
 //--------------------------------------------------------------------
 // getNumChildren()
 //--------------------------------------------------------------------
-U32 LLJoint::getNumChildren()
+U32 LLJoint::getNumChildren() const
```
AndrewMeadows, line 332: Use `const` when possible.

indra/llcharacter/lljoint.cpp
```diff
@@ -1050,27 +1050,5 @@ void LLJoint::setSkinOffset( const LLVector3& offset )
...
-		LLJoint* joint = *iter;
-		if (joint->isAnimatable())
-		{
-			main_axis = joint->getPosition();
-			main_axis.normVec();
-			// only care about first animatable child
-			break;
-		}
-	}
-}
```
AndrewMeadows, line 1073: Remove unused cruft.

indra/newview/llpuppetmotion.cpp
```diff
@@ -803,24 +797,26 @@ void LLPuppetMotion::solveIKAndHarvestResults(const LLIK::Solver::joint_config_m
...
 	LLPuppetEvent broadcast_event;
     const LLIK::Solver::joint_list_t& active_joints = mIKSolver.getActiveJoints();
-    LLVector3 pos;
-    LLQuaternion rot;
     for (auto joint: active_joints)
     {
         S16 id = joint->getID();
-        U8 flags = joint->getConfigFlags();
-        pos = joint->getLocalPos();
-        rot = joint->getLocalRot();
```
AndrewMeadows, line 813: Below we are figuring out what exact properties have changed so we can apply them to our avatar (`local_puppetry`) and/or stream to the server (`is_sending`).

indra/llcharacter/llik.cpp
```diff
@@ -1407,6 +1426,46 @@ void LLIK::Joint::setLocalRot(const LLQuaternion& new_local_rot)
...
+    // compute final scale adustment to applly to mLocalPos and mBone
+    constexpr F32 MIN_INVERTABLE_SCALE = 1.0e-15f;
+    LLVector3 re_scale;
+    for (S32 i = 0; i < 3; ++i)
+    {
+        // verify mLocalScale component to avoid introducing NaN
+        re_scale[i] = (mLocalScale[i] > MIN_INVERTABLE_SCALE) ?  scale[i] / mLocalScale[i] : 0.0f;
+    }
+    // We remember the final scale adjustment for later...
+    mLocalScale = scale;
```
simon-linden, line 1442: should this log a warning or check if mLocalScale == scale?   is that worth checking?

indra/llcharacter/llik.h
```diff
@@ -439,16 +454,17 @@ class Joint
...
     const LLVector3& getLocalPos() const { return mLocalPos; }
+    const LLVector3& getLocalScale() const { return mLocalScale; }
     F32 getBoneLength() const { return mBone.length(); }
     F32 getLocalPosLength() const { return mLocalPosLength; }
 
     ptr_t getParent() { return mParent; }
-    void activate() { mIsActive = true; }
-    bool isActive() const { return mIsActive; }
-    bool hasDisabledConstraint() const { return (mConfigFlags & FLAG_DISABLE_CONSTRAINT) > 0; }
+    void activate() { mIkFlags |= IK_FLAG_ACTIVE; }
```
simon-linden, line 462: are these ever deactivated ?

indra/llcharacter/llik.cpp
AndrewMeadows: No, I don't think we need a WARNING.  Eventually when we implement the "config once --> lasts forever" behavior we can add code that checks for non-changing config parameters.

indra/llcharacter/llik.h
AndrewMeadows: **Joints** are not deactivated during the IK calculations, but all activations are cleared before the IK starts its next `solve()`.  See `LLIK::Joint::resetFlags()` and where it is called.

simon-linden (approved):

indra/llcharacter/llik.cpp
```diff
@@ -2213,6 +2272,10 @@ void LLIK::Solver::rebuildAllChains()
                 joint->setLocalPos(p);
                 joint->activate();
             }
+            if (flags & CONFIG_FLAG_LOCAL_SCALE)
+            {
+                joint->setLocalScale(config.getLocalPos());
```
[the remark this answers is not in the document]
AndrewMeadows: Thank  you for spotting that cut-n-paste error.

PR #17, opened 2022-11-30
SL-18648 and SL-18742

akleshchev:
Just a cleanup (followup of SL-18648) and added a missing menu option SL-18742

### DIFF
--- a/indra/newview/llmaterialeditor.cpp
+++ b/indra/newview/llmaterialeditor.cpp
@@ -391,13 +391,9 @@ BOOL LLMaterialEditor::postBuild()
     {
         // Only allow fully permissive textures
         mBaseColorTextureCtrl->setImmediateFilterPermMask(PERM_ITEM_UNRESTRICTED);
-        mBaseColorTextureCtrl->setNonImmediateFilterPermMask(PERM_ITEM_UNRESTRICTED);
         mMetallicTextureCtrl->setImmediateFilterPermMask(PERM_ITEM_UNRESTRICTED);
-        mMetallicTextureCtrl->setNonImmediateFilterPermMask(PERM_ITEM_UNRESTRICTED);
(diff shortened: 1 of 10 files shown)


akleshchev:
SL-18727 should already be in D559, not sure why it's in the list...

maxim-productengine (approved):

indra/newview/llmaterialeditor.cpp
```diff
@@ -1816,58 +1821,95 @@ void LLMaterialEditor::loadLive()
...
+            me->setEmissiveColor(render_material->mEmissiveColor);
+            me->setDoubleSided(render_material->mDoubleSided);
+            me->setAlphaMode(render_material->getAlphaMode());
+            me->setAlphaCutoff(render_material->mAlphaCutoff);
 
             // most things like colors we can apply without verifying
             // but texture ids are going to be different from both, base and override
             // so only apply override id if there is actually a difference
-            if (local_mat->mBaseColorId != cmp_mat->mBaseColorId)
+            if (local_material->mBaseColorId != render_material->mBaseColorId)
```
cosmic-linden, line 1895: I think that the render material should be the correct combination of all material properties, so ideally these checks should not be needed.

cosmic-linden: Well, actually, there is one exception: The render material has transforms, but the saved material should not.

cosmic-linden (approved):

indra/newview/llmaterialeditor.cpp
akleshchev: In this case, it's a 'file' texture, if local material's texture and render texture match, viewer needs to upload said texture, if they do not much, no upload is needed because user replaced 'file' texture with something else. Admitedly this can be handled better and I probably should do some kind of is_texture_local(render_material->mBaseColorId) even if texture doesn't match one supplied by a gltf file, but that's a low priority.

PR #24, opened 2022-12-02
SL-18486 Complete Avatars floater is blank.

maxim-productengine:


indra/newview/llfloateravatar.cpp
```diff
@@ -40,16 +40,20 @@
 LLFloaterAvatar::LLFloaterAvatar(const LLSD& key)
 	:	LLFloater(key)
 {
+    mAvatarPicker = findChild<LLMediaCtrl>("avatar_picker_contents");
```
akleshchev, line 43: This seems very wrong to me. findChild will happen before floater will have a chance to construct UI. This should happen in postBuild(), after xml was parsed.

akleshchev (changes requested):

indra/newview/llfloateravatar.cpp
maxim-productengine: You're right, thanks.

akleshchev (approved):

PR #28, opened 2022-12-07
SL-14399: Ditch overflow queue LLViewerAssetStorage::mCoroWaitList.

nat-goodspeed:
mCoroWaitList was introduced to prevent an assertion failure crash: LLCoprocedureManager never expects to fill its LLCoprocedurePool::mPendingCoprocs queue. The queue limit was arbitrarily set to 4096 some years ago, but in practice LLViewerAssetStorage can post way more requests than that.

LLViewerAssetStorage checked whether the target LLCoprocedureManager pool's queue looked close to full, and if so posted the pending request to its mCoroWaitList instead. But then it had to override the base LLAssetStorage method checkForTimeouts() to continually check whether pending tasks could be moved from mCoroWaitList to LLCoprocedureManager.

A simpler solution is to enlarge LLCorpocedureManager::DEFAULT_QUEUE_SIZE, the upper limit on mPendingCoprocs. Since mCoroWaitList was an unlimited queue, making DEFAULT_QUEUE_SIZE "very large" does not increase the risk of runaway memory consumption.

### DIFF
--- a/indra/llmessage/llcoproceduremanager.cpp
+++ b/indra/llmessage/llcoproceduremanager.cpp
@@ -47,7 +47,10 @@ static const std::map<std::string, U32> DefaultPoolSizes{
 };
 
 static const U32 DEFAULT_POOL_SIZE = 5;
-const U32 LLCoprocedureManager::DEFAULT_QUEUE_SIZE = 4096;
+// SL-14399: When we teleport to a brand-new simulator, the coprocedure queue
+// gets absolutely slammed with fetch requests. Make this queue effectively
+// unlimited.
+const U32 LLCoprocedureManager::DEFAULT_QUEUE_SIZE = 1024*1024;
 
 //=========================================================================
 class LLCoprocedurePool: private boost::noncopyable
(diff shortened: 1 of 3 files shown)


akleshchev (approved):

indra/llmessage/llcoproceduremanager.cpp
```diff
@@ -47,7 +47,10 @@ static const std::map<std::string, U32> DefaultPoolSizes{
 };
 
 static const U32 DEFAULT_POOL_SIZE = 5;
-const U32 LLCoprocedureManager::DEFAULT_QUEUE_SIZE = 4096;
+// SL-14399: When we teleport to a brand-new simulator, the coprocedure queue
+// gets absolutely slammed with fetch requests. Make this queue effectively
+// unlimited.
+const U32 LLCoprocedureManager::DEFAULT_QUEUE_SIZE = 1024*1024;
```
akleshchev, line 53: The biggest offenders probably aren't teleports but landmarks and gestures. Viewer needs all landmark assets loaded to let search and landmarking work and there can be thousands of landmarks in some inventories.

Rider-Linden (approved):

PR #35, opened 2022-12-13
Integrate LLAgentIOModule into Puppetry

nat-goodspeed:
Creating PR for review purposes.


indra/newview/llagentiomodule.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include "llsdutil_math.h"
+#include "llagent.h"
+#include "llagentcamera.h"
+#include "llphysicsmotion.h"
+#include "llvoavatar.h"
+#include "llvoavatarself.h"
+
+static LLAgentIORegistrar AgentIOReg;
+static LLAgentIOModule* AgentIOModule;
+
```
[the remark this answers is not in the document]
nat-goodspeed: I realize I haven't been following puppetry protocol discussions closely, but implementing a generic get / set operation requiring sub-dispatching feels to me like a mistake.

If you published each of the individual operations as its own `"command"`, then the `LLLeap` introspection API would let a puppetry client script take inventory of supported commands, query their metadata and so forth.

If the puppetry LEAP protocol has been locked in to require get and set with subcommands, then maybe the right implementation would be to add an implementation-detail `LLEventAPI`to manage subcommand metadata and perform subcommand dispatching. Then the top-level `LLEventAPI` could supplement get and set with, say, list and describe operations to enumerate subcommands and describe any particular one.

indra/newview/llagentiomodule.cpp
```diff
@@ -0,0 +1,279 @@
...
+            AgentIOModule->setCamera(data["data"]);
+        }
+    }
+}
+
+void LLAgentIOModule::setCamera(const LLSD& data)
+{
+    //Camera and target position are expected as
+    //world-space coordinates in the viewer-standard
+    //coordinate frame.
```
nat-goodspeed, line 130: To me, this comment belongs as queryable metadata on a `"setCamera"` command.

indra/newview/llagentiomodule.cpp
```diff
@@ -0,0 +1,279 @@
...
+    mModuleName = module_name;
+};
+
+
+LLAgentIOModule::module_ptr_t LLAgentIOModule::getLeapModule() const
+{
+    // lock it
+    return mLeapModule.lock();
+}
+
```
nat-goodspeed, line 225: Instead of a `sendCommand()` method, how about instantiating [a `Response` object](https://github.com/secondlife/viewer/blob/aura/DRTVWR-558-dbljump/indra/llcommon/lleventapi.h#L72-L153) in each method that wants to send a response? Or -- the heart of `Response` -- just call [sendReply()](https://github.com/secondlife/viewer/blob/aura/DRTVWR-558-dbljump/indra/llcommon/llevents.h#L771-L783).

In my mind, an `LLEventPump` alias such as `"agentio.controller"` is more for sending unsolicited data to a plugin than it is for replying to an explicit request.

indra/newview/llagentiomodule.h
```diff
@@ -0,0 +1,94 @@
...
+    };
+
+	/*
+    void setAgentIOOptions(LLSD options);
+    static void SetAgentIOOptionsCoro(const std::string& capurl, LLSD options);
+	*/
+    mutable std::weak_ptr<LLLeap> mLeapModule; // weak pointer to module
+    std::string mModuleName;
+    LLBoundListener mListener;
+    LLTempBoundListener mPlugin;        //For event pump to send leap updates to plug-ins
```
nat-goodspeed, line 83: I'm dubious about the use cases for `mLeapModule`, `mModuleName`, `mListener` (already implemented in `LLEventAPI`'s base class `LLDispatchListener`), `mPlugin`, `module_ptr_t`, `setLeapModule()`, `getLeapModule()`, `haveAgentIOModule()`, `clearLeapModule()`, `getModuleName()`. All the above feel extraneous. I could be convinced otherwise, but I'd want actual usage explained.

indra/newview/llagentiomodule.h
```diff
@@ -0,0 +1,94 @@
...
+#include "llevents.h"
+#include "lleventapi.h"
+#include "lazyeventapi.h"
+#include "llsingleton.h"
+
+class LLLeap;
+using namespace LL;
+
+class LLAgentIOModule : public LLSingleton<LLAgentIOModule>,
+	public LLEventAPI
```
nat-goodspeed, line 47: n.b. `LLEventAPI` already isa `LLInstanceTracker` by its name, so it doesn't need to be an `LLSingleton` to find it.

indra/llcommon/llsys.cpp
```diff
@@ -454,14 +454,14 @@ LLOSInfo::LLOSInfo() :
 
 #ifndef LL_WINDOWS
 // static
-S32 LLOSInfo::getMaxOpenFiles()
+long LLOSInfo::getMaxOpenFiles()
```
AndrewMeadows, line 457: You prefer `long` here and not `S64` or `int64_t`?

AndrewMeadows (commented):
Wow, this was a lot of work!  In fact, it includes so much cleanup (changing int to auto etc) across 100 files that it guarantees the D-558 branch will be a pile of conflicts when it tries to return to the main branch.

Also, if we leave this major cleanup in the D-558 branch then it will be dependent on the completion of some puppetry deliverable before it ever sees the light of the main branch.  I'm wondering... maybe the general EventAPI cleanup should be ported to the main branch and D-558 should obtain it via a merge from main?  I kinda hate to suggest it because it would be so much work but... there is a work crunch no matter how it gets into main and at least this way it might get there faster.

indra/llcommon/lldefs.h
```diff
@@ -168,80 +169,79 @@ const U32	MAXADDRSTR		= 17;		// 123.567.901.345 = 15 chars + \0 + 1 for good luc
 //   llclampb(a)     // clamps a to [0 .. 255]
 //   				   
 
-template <class LLDATATYPE> 
-inline LLDATATYPE llmax(const LLDATATYPE& d1, const LLDATATYPE& d2)
+template <typename T1, typename T2> 
+inline auto llmax(T1 d1, T2 d2)
 {
 	return (d1 > d2) ? d1 : d2;
 }
```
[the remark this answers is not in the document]
nat-goodspeed: The intention there is that the compiler will choose the same type for C that it would choose for:

    auto C = (A > B)? A : B;

or in fact for:

    std::cout << "The larger one is " << ((A > B)? A : B) << std::endl;

I changed the `llmax()` template to allow different datatypes for `d1` and `d2` because the `size_t` changes made many existing `llmax()` expressions fail.

indra/llcommon/lldefs.h
```diff
@@ -168,80 +169,79 @@ const U32	MAXADDRSTR		= 17;		// 123.567.901.345 = 15 chars + \0 + 1 for good luc
...
 
 template <class LLDATATYPE> 
 inline void llswap(LLDATATYPE& lhs, LLDATATYPE& rhs)
 {
-	LLDATATYPE tmp = lhs;
-	lhs = rhs;
-	rhs = tmp;
+	std::swap(lhs, rhs);
 }
```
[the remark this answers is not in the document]
nat-goodspeed: Not a bad idea, but this is far less pervasive, and I have faith in the compiler's ability to inline a trivial function like this.

indra/llcommon/lleventdispatcher.cpp
```diff
@@ -164,20 +98,26 @@ void LLSDArgsSource::done() const
...
+                   const LLSD& names, const LLSD& defaults);
 
     /// Given arguments map, return LLSD::Array of parameter values, or
     /// trigger error.
     LLSD map(const LLSD& argsmap) const;
 
 private:
     static std::string formatlist(const LLSD&);
+    template <typename... ARGS>
+    void callFail(ARGS&&... args) const;
```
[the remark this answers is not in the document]
nat-goodspeed: Did you get a look at the `narrow_holder` class in stdtypes.h? It has a templated conversion operator whose template parameter is the inferred destination type -- with two implementations, conditionally chosen depending on the relationship of the destination type with the source type.

indra/llcommon/llsys.cpp
nat-goodspeed: I changed to `long` because the compiler scolded me for trying to return specifically `long` (the return type of the system function) through `S32`. I was intentionally matching the type.

indra/llcommon/lleventdispatcher.cpp
```diff
@@ -399,43 +337,64 @@ std::string LLSDArgsMapper::formatlist(const LLSD& list)
...
-            LLTHROW(DispatchError(stringize(desc, ": bad request: ", mismatch)));
+            return callFail(desc, ": bad request: ", mismatch);
         }
         // Event syntax looks good, go for it!
-        mFunc(event);
+        return mFunc(event);
     }
 
-    virtual LLSD addMetadata(LLSD meta) const
+    LLSD addMetadata(LLSD meta) const override
```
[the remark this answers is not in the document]
nat-goodspeed: LLSD uses copy-on-write functionality, yes.

You're right that the way `addMetadata()` is used, it could indeed modify its parameter. Call it a reflexive flinch from having been burned before: modifying a passed map pops up red flags in my head.

I observe that both my jitters and your efficiency note could be addressed by changing this to `getMetadata()` and making its _caller_ add the other two entries after the call, rather than before.

But is this branch even live any more?

PR #36, opened 2022-12-14
more correct LLPose::usage in LLPuppetMotion

AndrewMeadows:
This PR supplies a theoretical fix for an assert in LLJointState::setPosition() in the Puppetry experimental viewer.

Internal bug is SL-18818.

### DIFF
--- a/indra/newview/llpuppetmotion.cpp
+++ b/indra/newview/llpuppetmotion.cpp
@@ -652,7 +652,6 @@ void LLPuppetMotion::setAvatar(LLVOAvatar* avatar)
 
 void LLPuppetMotion::clearAll()
 {
-    //mJointStates.clear();   This kills puppetry - need a correct way to reset joint data when module restarts
     mEventQueues.clear();
     mOutgoingEvents.clear();
     mJointStateExpiries.clear();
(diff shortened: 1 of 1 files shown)


indra/newview/llpuppetmotion.cpp
```diff
@@ -991,7 +989,6 @@ void LLPuppetMotion::applyBroadcastEvent(const LLPuppetJointEvent& event, Timest
         {
             joint_state->setScale(event.getScale());
         }
-        joint_state->setUsage((U32)(flags & LLIK::MASK_JOINT_STATE_USAGE));
```
AndrewMeadows, line 994: This is the actual fix: we need to LLJointState::setUsage() BEFORE trying to update those fields.

I think the only reason this didn't ALWAYS crash was because we were not clearing the Usage flags elsewhere, so once the flags were set they stayed set for future joint updates.

Note: this is a theoretical fix since I haven't been able to reproduce the problem, but I'm confident this the fix: this is the only place I could find where we were trying to update the properties before declaring the usage.

indra/newview/llpuppetmotion.cpp
```diff
@@ -1241,6 +1239,10 @@ void LLPuppetMotion::onDeactivate()
     // this motion from its active list.
     mPose.removeAllJointStates();
     mJointsToRemoveFromPose.clear();
+    for (auto& data_pair : mJointStates)
+    {
+        data_pair.second->setUsage(0);
+    }
```
AndrewMeadows, line 1245: Clear the usage on deactivation (and when removing the pose further up).  The reason we want to clear it when possible is because the animation blending checks the usage flags to figure out what math needs to be done: leaving extraneous usage flags around would make the blending calculations do unnecessary work.

simon-linden (approved):

PR #38, opened 2023-01-01
Clean up for loops in llappearance to use C++11 range based for loops

Fawrsk:
I've also combined three loops into one within LLWearables::revertValues()

PR #41, opened 2023-01-04
Clean up for loops in llaudio to use C++11 range based for loops

Fawrsk:
I plan on doing this to all of the projects, if that's alright. I'm not sure if you'd rather them done in one single pull request, or multiple smaller ones for each project.

I was asked in #https://github.com/secondlife/viewer/pull/38#issuecomment-1370254701 if I wanted to be added to the contributors as Fawrsk, and yes, that is fine.

Also, happy new year to you too!

### DIFF
--- a/indra/llaudio/llaudioengine.cpp
+++ b/indra/llaudio/llaudioengine.cpp
@@ -123,18 +123,16 @@ void LLAudioEngine::shutdown()
 	cleanupWind();
 
 	// Clean up audio sources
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

akleshchev:
>I'm not sure if you'd rather them done in one single pull request, or multiple smaller ones for each project.
Loop changes are trivial to verify, so it should be fine either way.

PR #46, opened 2023-01-10
Fixes for pull requests #38, #41, and #42

Fawrsk:
Eliminate unnecessary copies, and remove uses of auto


akleshchev:
Thank you.
Wasn't aware it generated a copy, I thought C++11's 'for' loop was smarter than that.

PR #48, opened 2023-01-11
Sl 18820: Build floater should preserve GLTF transforms when changing the PBR material

cosmic-linden:
Requires server changes as well in order to take effect (merge order is unimportant).

This changes the data LLGLTFMaterialList sends over the wire. When LLGLTFMaterialList::queueApply is called, it will send both a material ID and a non-default material override JSON. Previously, it sent an empty string.

Viewer material editing will now always preserves GLTF texture transforms in overrides whenever the render material ID changes, even if the new render material ID is null. This is the current LSL script behavior as well. The code removes overrides when they become sDefault.

LLTextureEntry::setBaseMaterial modifies the material override in-place. Seems reasonable for now, but is there a case in the future where it won't be?
- Ideally overrides would use std::unique_ptr, but that has caveats for how LLRefCount objects are typically used

### DIFF
--- a/indra/llprimitive/llgltfmaterial.cpp
+++ b/indra/llprimitive/llgltfmaterial.cpp
@@ -392,9 +392,21 @@ bool LLGLTFMaterial::setBaseMaterial()
 {
     const LLGLTFMaterial old_override = *this;
     *this = sDefault;
-    // Preserve the texture transforms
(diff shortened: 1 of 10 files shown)


cosmic-linden:

RunitaiLinden (approved):

PR #49, opened 2023-01-11
SL-18946: Adding message from viewer to simulator.  Tracks attachment…

Rider-Linden:
… point position and rotation post animation/movement.

Transmit the  attachment point information up to the simulator for your own avatar after animation has completed.  This allows the simulator to react to things like hand position in world.


Rider-Linden:

Rider-Linden:
recheck

indra/newview/llviewerjointattachment.h
```diff
@@ -98,6 +98,17 @@ class LLViewerJointAttachment :
 	typedef std::vector<LLPointer<LLViewerObject> > attachedobjs_vec_t;
 	attachedobjs_vec_t mAttachedObjects;
 
+    bool    hasChanged(const LLVector3 &pos, const LLQuaternion &rot) const
+    {
+        return (!(pos - mLastTrackedPos).isNull() || !LLQuaternion::almost_equal(rot, mLastTrackedRot));
```
Rider-Linden, line 103: A thought experiment here... Right now I'm looking for ANY change in the position or rotation. 
Suppose, instead we look at the magnitude of the change instead. For position the magnitude of the change is greater than X% or the angle between the vectors is greater than some value. (This would mean that the further away from the center the less resolution, unless magnitude was taken into account.)  For rotation the angle between the two rotations is greater than some amount. (say 5degrees) ...

indra/newview/llvoavatarself.cpp
```diff
@@ -1288,6 +1288,50 @@ U32 LLVOAvatarSelf::getNumWearables(LLAvatarAppearanceDefines::ETextureIndex i)
 	return gAgentWearables.getWearableCount(type);
 }
 
+void LLVOAvatarSelf::postMotionUpdate() 
+{
```
Rider-Linden, line 1292: The viewer is sending every animation frame... in crowded regions this could add up. We could send every Nth frame instead.

indra/newview/llviewerjointattachment.h
```diff
@@ -98,16 +98,40 @@ class LLViewerJointAttachment :
...
+        LLVector3 unit_vec(LLVector3::x_axis * rot);
+        F32 unit_cos(0);
+        if (!mLastTrackedUnit.isNull())
+        {
+            unit_cos = unit_vec * mLastTrackedUnit;
+        }
+        mLastTrackedUnit = unit_vec;
+
+        return (((pos - mLastTrackedPos).magVecSquared() > SMALL_CHANGE) || (unit_cos < 0.9));
+    }
```
AndrewMeadows, line 114: It looks like you're using `mLastTrackedUnit` to figure out how much the rotation has changed but there is a better way to do it:

`LLQuaternion::almost_equal(rot, mLastTrackedRot, minimum_angle)`.

indra/newview/llvoavatarself.cpp
AndrewMeadows: Perhaps an expiry logic flow would be more tunable.  Add `mNextPostMotionUpdateExpiry`, set it to a future time, and only send an update when expired.  The period between each expiry could be adjusted at runtime if necessary.  Note, the server would be the most susceptible to overload in a crowded situation and would need protection.  It should: enforce its own dynamic version of the update rate it will accept and discard client updates that exceed it.  Bonus points if the server could tell the viewer the update rate it is enforcing on its end.

indra/newview/llvoavatarself.h
```diff
@@ -55,11 +55,11 @@ class LLVOAvatarSelf :
...
 	LLVOAvatarSelf(const LLUUID &id, const LLPCode pcode, LLViewerRegion *regionp);
 	virtual 				~LLVOAvatarSelf();
-	virtual void			markDead();
-	virtual void 			initInstance(); // Called after construction to initialize the class.
+	void			        markDead() override;
+	void 			        initInstance() override; // Called after construction to initialize the class.
 	void					cleanup();
 protected:
-	/*virtual*/ BOOL		loadAvatar();
+	BOOL		            loadAvatar() override;
```
AndrewMeadows, line 62: I approve the use of the the `override` keyword!

indra/newview/llvoavatarself.h
```diff
@@ -299,6 +299,17 @@ class LLVOAvatarSelf :
...
 
+/********************************************************************************
+ **                                                                            **
+ **                    ANIMATIONS
+ **/
+    void                postMotionUpdate() override;  // from llCharacter
+
+ /**                   ANIMATIONS
+  **                                                                            **
+  *******************************************************************************/
```
AndrewMeadows, line 310: I don't like excessive use of comment delimiters.  When used they should NOT use block-comments, only single-line comments: e.g. `//----------------------------------------------------`.

AndrewMeadows (changes requested):

indra/newview/llviewerjointattachment.h
```diff
@@ -98,16 +98,40 @@ class LLViewerJointAttachment :
...
-	BOOL			mIsHUDAttachment;
-	S32				mPieSlice;
+	BOOL			    mVisibleInFirst;
+	LLVector3		    mOriginalPos;
+	S32				    mGroup;
+	BOOL			    mIsHUDAttachment;
+	S32				    mPieSlice;
+    LLVector3           mLastTrackedPos;
+    LLQuaternion        mLastTrackedRot;
+    mutable LLVector3   mLastTrackedUnit;
```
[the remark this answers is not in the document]
Rider-Linden: Spaces are preferred, I was resistant to forcing every line in the file to be marked as changed.

indra/newview/llvoavatarself.h
Rider-Linden: The catch is that once you use it once, you have to make sure you use it everywhere. (Bit me on the mac build)

indra/newview/llviewerjointattachment.h
Rider-Linden: Nice.  I'll switch to that.

indra/newview/llviewerjointattachment.h
```diff
@@ -100,17 +100,14 @@ class LLViewerJointAttachment :
...
-        F32 unit_cos(0);
-        if (!mLastTrackedUnit.isNull())
-        {
-            unit_cos = unit_vec * mLastTrackedUnit;
-        }
-        mLastTrackedUnit = unit_vec;
-
-        return (((pos - mLastTrackedPos).magVecSquared() > SMALL_CHANGE) || (unit_cos < 0.9));
+        static constexpr F32 SMALL_CHANGE_DIST2(0.05 * 0.05);
+        static constexpr F32 SMALL_CHANGE_ANGL(0.5);    // about 28.6 degrees
```
AndrewMeadows, line 104: `0.05` --> `0.05f` to remove implicit conversion from `double` to `float`.

`0.5` --> `0.5f`.  Half a radian is rather large for a "small change" threshold.  Are you sure you want `SMALL_CHANGE_ANGL` to have that value?

AndrewMeadows (changes requested):
Please comment on whether you really want 0.5 radians to be the "small angle" threshold.  If yes, then this is mergeable: the rest of my comment are suggestions or can be fixed later.

AndrewMeadows (approved):

PR #50, opened 2023-01-13
Add missing panel_* files in Polish translation

PanteraPolnocy:
With this commit all missing files in /pl/ should be in place.


akleshchev:
Thank you!

akleshchev (approved):

PR #54, opened 2023-01-23
DRTVWR-558: Introduce support for batch LEAP operations

nat-goodspeed:
LLEventAPI's base class LLDispatchListener now supports a new way to write wrapper functions: when the incoming request contains a "reply" key whose value is the name of the LLEventPump on which to reply, and the target function returns a type convertible to LLSD, LLDispatchListener now implicitly sends a reply to the "reply" LLEventPump. So instead of registering a void function and coding that function to explicitly send a reply, either by calling sendReply() or by instantiating LLEventAPI::Response, now you can register a function that simply returns its intended response.

The benefit is that LLDispatchListener now supports batched invocations. If the value of the incoming request's operation key ("op" or "command" or whatever was specified to the constructor) is an LLSD array or map rather than LLSD::String, LLDispatchListener will attempt to call each of the functions named in the container (details in LLDispatchListener Doxygen comments). It aggregates the results and sends a single composite reply event.


nat-goodspeed:
The subject branch dispatch-return of this PR was based on aura/DRTVWR-558-dbljump. Since the PR targets the same aura/DRTVWR-558-dbljump, I expected to see only my own commits. Instead I find many commits by others mixed in. Was aura/DRTVWR-558-dbljump force-reset at some point to exclude those commits? If so, I should create a new branch based on some appropriate D558 branch and rebase my intended commits into that.

nat-goodspeed:
Closing due to unintended commits on the source branch; see #58 instead.

PR #56, opened 2023-01-24
remove server-side cruft from LLParcel

AndrewMeadows:
While examining the code I noticed there some legacy server code in the `LLParcel` class from when the two code bases were split.  This PR shows what I could find after a quick pass.

### DIFF
--- a/indra/llinventory/llparcel.cpp
+++ b/indra/llinventory/llparcel.cpp
@@ -112,8 +112,6 @@ static const std::string PARCEL_ACTION_STRING[LLParcel::A_COUNT + 1] =
 
 
 
-//const char* revert_action_to_string(LLParcel::ESaleTimerExpireAction action);
-//LLParcel::ESaleTimerExpireAction revert_string_to_action(const char* s);
 const std::string& category_to_ui_string(LLParcel::ECategory category);
 LLParcel::ECategory category_ui_string_to_category(const std::string& s);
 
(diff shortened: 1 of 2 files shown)


indra/llinventory/llparcel.cpp
```diff
@@ -166,16 +164,9 @@ void LLParcel::init(const LLUUID &owner_id,
...
-	//mBuyerID.setNull();
-	//mJoinNeighbors = 0x0;
-	mSaleTimerExpires.setTimerExpirySec(0);
-	mSaleTimerExpires.stop();
 	mGraceExtension = 0;
-	//mExpireAction = STEA_REVERT;
-	//mRecordTransaction = FALSE;
 
 	mAuctionID = 0;
-	mInEscrow = false;
```
AndrewMeadows, line 178: Notice: I'm not adding any code in this PR.  Everything is a delete: either class date members in this **ctor**, entire class methods, or in some places I only delete extra whitespace.

indra/llinventory/llparcel.cpp
```diff
@@ -947,128 +772,6 @@ const std::string& LLParcel::getActionString(LLParcel::EAction action)
...
-        mMediaResetTimer.stop();
-    }
-    return expired;
-}
-
-
-void LLParcel::startSale(const LLUUID& buyer_id, BOOL is_buyer_group)
-{
-	// TODO -- this and all Sale related methods need to move out of the LLParcel 
-	// base class and into server-side-only LLSimParcel class
```
AndrewMeadows, line 982: This **TODO** comment is relevant: the server-only logic in `LLParcel` was supposed to be removed a long time ago.

PR #58, opened 2023-01-24
 DRTVWR-558: Introduce support for batch LEAP operations

nat-goodspeed:
LLEventAPI's base class LLDispatchListener now supports a new way to write wrapper functions: when the incoming request contains a "reply" key whose value is the name of the LLEventPump on which to reply, and the target function returns a type convertible to LLSD, LLDispatchListener now implicitly sends a reply to the "reply" LLEventPump. So instead of registering a void function and coding that function to explicitly send a reply, either by calling sendReply() or by instantiating LLEventAPI::Response, now you can register a function that simply returns its intended response.

The benefit is that LLDispatchListener now supports batched invocations. If the value of the incoming request's operation key ("op" or "command" or whatever was specified to the constructor) is an LLSD array or map rather than LLSD::String, LLDispatchListener will attempt to call each of the functions named in the container (details in LLDispatchListener Doxygen comments). It aggregates the results and sends a single composite reply event.


nat-goodspeed:
This PR supersedes rejected #54 : this one rebases the requisite commits onto the current target branch.

auralinden (approved):
No specific comments.   Will be merging into dbljmp and restructuring the command/response syntax there but it looks like the unit test cases well cover everything I can think of.

PR #64, opened 2023-01-29
Fast hashing classes for use in place of the slow LLMD5, where speed matters

sldevel:
This commit adds the HBXXH64 and HBXXH128 classes for use as a drop-in replacement for the slow LLMD5 hashing class, where speed matters and backward compatibility (with standard hashing algorithms) and/or cryptographic hashing qualities are not required.
It also replaces LLMD5 with HBXXH* in a few existing hot (well, ok, just "warm" for some) paths meeting the above requirements, while paving the way for future use cases, such as in the [DRTVWR-559](https://github.com/secondlife/viewer/tree/DRTVWR-559) and sibling branches where the slow LLMD5 is used (e.g. to hash materials and vertex buffer cache entries), and could greatly benefit from such a (way) faster algorithm, without negative impact.

Here is the comment I added in indra/llcommon/hbxx.h:

```
// HBXXH* classes are to be used where speed matters and cryptographic quality
// is not required (no "one-way" guarantee, though they are likely not worst in
// this respect than MD5 which got busted and is now considered too weak). The
// xxHash code they are built upon is vectorized and about 50 times faster than
// MD5. A 64 bits hash class is also provided for when 128 bits of entropy are
// not needed. The hashes collision rate is similar to MD5's.
// See https://github.com/Cyan4973/xxHash#readme for details.
```

These hash classes are already in use in the Cool VL Viewer starting with v1.30.0.36 (stable branch) and v1.30.1.19 (experimental branch) with no side effect beside higher fps rates. ;-)

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -2816,6 +2816,36 @@ Copyright (c) 2012, 2014, 2015, 2016 nghttp2 contributors</string>
         <key>version</key>
         <string>0.54.1.555529</string>
       </map>
+      <key>xxhash</key>
+      <map>
+        <key>copyright</key>
+        <string>Copyright 2012-2020 Yann Collet</string>
+        <key>description</key>
+        <string>xxHash Extremely fast hash algorithm</string>
+        <key>license</key>
+        <string>bsd</string>
+        <key>license_file</key>
+        <string>LICENSES/xxhash.txt</string>
+        <key>name</key>
+        <string>xxhash</string>
+        <key>platforms</key>
+        <map>
+          <key>common</key>
+          <map>
+            <key>archive</key>
+            <map>
+              <key>hash</key>
+              <string>f54f21dda4ce25b112f0cc7b4ce38bba</string>
+              <key>url</key>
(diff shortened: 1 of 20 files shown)


akleshchev (approved):
Running xxhash package trough LL's command chain (I know it's BSD, but have to check, have a build of our own ready), otherwise looks good, thank you.

sldevel:
> Running xxhash package trough LL's command chain (I know it's BSD, but have to check, have a build of our own ready), otherwise looks good, thank you.
The xxhash package is just the xxhash.h header + the license file... I do not know anything about the "LL's command chain"... But if you want the "build script" (that will work just fine when ran from a Linux or Cygwin terminal), you can get it [from here](http://sldev.free.fr/libraries/sources/cvlv-xxhash-20230124.tar.bz2).

akleshchev:
> But if you want the "build script" (that will work just fine when ran from a Linux or Cygwin terminal) for that package, you can get it [from here](http://sldev.free.fr/libraries/sources/cvlv-xxhash-20230124.tar.bz2).
Thank you, already made that and have made the build (as much as a single header can be called a build), 3p-xxhash repo is just not public yet.

PR #68, opened 2023-02-04
SL-17761: Replace 'Grid emergency' message with generic error

bennettgoble:
The viewer currently presents a startling "Grid emergency" warning if an unrecognized error is returned from login. Let's tone this down a bit and present the error as it is: an unrecognized login failure rather than SLearth exploding.


bennettgoble:
Let me know if this should be rebased against a DRT branch. Thx.

marchcat-pe (approved):

PR #71, opened 2023-02-06
Faster and simpler inventory category hashing.

sldevel:
This commit changes inventory category hashing from slow LLMD5 to fast HBXX128 hashing, and allows to use a simple LLUUID for the hash, instead of an LLMD5 instance. It also removes some old cruft dealing with unused baked texture hashing.

### DIFF
--- a/indra/newview/llinventorymodel.cpp
+++ b/indra/newview/llinventorymodel.cpp
@@ -452,7 +452,7 @@ void LLInventoryModel::getDirectDescendentsOf(const LLUUID& cat_id,
 	items = get_ptr_in_map(mParentChildItemTree, cat_id);
 }
 
-LLUUID LLInventoryModel::hashDirectDescendentNames(const LLUUID& cat_id) const
+LLInventoryModel::digest_t LLInventoryModel::hashDirectDescendentNames(const LLUUID& cat_id) const
 {
 	LLInventoryModel::cat_array_t* cat_array;
 	LLInventoryModel::item_array_t* item_array;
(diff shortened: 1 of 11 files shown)


marchcat-pe (approved):

indra/newview/llinventorymodel.cpp
```diff
@@ -451,17 +452,16 @@ void LLInventoryModel::getDirectDescendentsOf(const LLUUID& cat_id,
 	items = get_ptr_in_map(mParentChildItemTree, cat_id);
 }
 
-LLMD5 LLInventoryModel::hashDirectDescendentNames(const LLUUID& cat_id) const
+LLUUID LLInventoryModel::hashDirectDescendentNames(const LLUUID& cat_id) const
```
akleshchev, line 455: I suggest using some alias instead of an LLUUID here to make sure it will be more obvious at first glance that this returns not an actual UUID.

sldevel: I could add a digest_t type to HBXXH* classes (i.e. 'typedef LLUUID digest_t' for HBXXH128, in this case), and change that method apparent signature for HBXXH128::digest_t 'LLInventoryModel::hashDirectDescendentNames(const LLUUID& cat_id) const', but then it would mean including the hbxxh.h header in llinventorymodel.h (meaning indirectly as well in a load of other modules including llinventorymodel.h)
Alternatively, digest_t could be made a typedef in LLInventoryModel only (avoiding the compile-time-costly hbxxh.h header inclusion)...

sldevel: I opted for the second (simpler) solution.

See the added commit.

akleshchev: Looks good, thank you.

I was about to suggest just adding typedef into lluuid.h, while not the best place, it's the same thing so won't be too out of place.

PR #72, opened 2023-02-06
Make HBXXH* classes no-copy.

sldevel:
These classes are not trivially copyable because of the mState pointer on an internal XXH3 state that would have to be explicitely copied. While it would be possible to add custom copy constructor and operator for them, it does not really make sense to allow copying an instance of these classes, since all we care about storing and copying is the digest (which is either an U64 or an LLUUID).


akleshchev:
Thank you.

PR #73, opened 2023-02-09
SL-19080: GLTF Material asset consistency: Part 2: Update viewer GLTF Material asset upload to v1.1

cosmic-linden:
This version of the viewer uploads the new v1.1 material assets, with more rigorous server validation requirements. Most notably:

- The occlusion and metallic roughness texture IDs must match
- No transforms allowed

There are also some new unit tests to verify that the serialization of GLTF materials works as expected, and some consolidation of code related to asset uploading.

This viewer code should be deployed after the server code is deployed to the beta testing regions, so that material uploading will continue to work.

### DIFF
--- a/indra/llprimitive/llgltfmaterial.h
+++ b/indra/llprimitive/llgltfmaterial.h
@@ -49,6 +49,11 @@ public:
     // default material for reference
     static const LLGLTFMaterial sDefault;
 
+    static const char* ASSET_VERSION;
+    static const char* ASSET_TYPE;
+    static const std::array<char*, 2> ACCEPTED_ASSET_VERSIONS;
+    static bool isAcceptedVersion(const std::string& version) { return std::find(ACCEPTED_ASSET_VERSIONS.cbegin(), ACCEPTED_ASSET_VERSIONS.cend(), version) != ACCEPTED_ASSET_VERSIONS.cend(); }
+
     struct TextureTransform
     {
         LLVector2 mOffset = { 0.f, 0.f };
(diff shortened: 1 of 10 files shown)


indra/llprimitive/llgltfmaterial.h
```diff
@@ -49,6 +49,11 @@ class LLGLTFMaterial : public LLRefCount
     // default material for reference
     static const LLGLTFMaterial sDefault;
 
+    static const char* ASSET_VERSION;
+    static const char* ASSET_TYPE;
```
brad-linden, line 53: these should possibly be `static const char* const` (realizing belatedly that I probably should have noticed this on the server side review already)

indra/llprimitive/tests/llgltfmaterial_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+    void ensure_gltf_material_trimmed(const std::string& material_json, const std::string& must_not_contain)
+    {
+        ensure("LLGLTFMaterial serialization trims property '" + must_not_contain + "'", material_json.find(must_not_contain) == std::string::npos);
+    }
+
+    // Test that GLTF material fields have not changed since these tests were written
+    template<> template<>
+    void llgltfmaterial_object_t::test<1>()
+    {
+        if (sizeof(void*) > 4) // Don't bother running this test for 32-bit systems
```
brad-linden, line 142: we have a preprocessor definition ADDRESS_SIZE for this kind of thing.  see `indra/cmake/00-Common.cmake:30` for where it is set

brad-linden (approved):
looks good, lets get this out there

PR #75, opened 2023-02-10
EMERGENCY FIX: revert LLUUID::combine() to old algorithm to match server code.

sldevel:
As it happens, the change in the LLUUID::combine() algorithm introduced by one of my previous commits (af65650) is causing invalid assets creation (seen with some clothing items, such as Shape and Universal types); obviously, the server is using the old algorithm for UUID validation purpose of these assets.

This commit reverts LLUUID::combine() code to use LLMD5.

### DIFF
--- a/indra/llcommon/lluuid.cpp
+++ b/indra/llcommon/lluuid.cpp
@@ -44,6 +44,7 @@
 #include "lltimer.h"
 #include "llthread.h"
 #include "llmutex.h"
+#include "llmd5.h"
 #include "hbxxh.h"
 
 const LLUUID LLUUID::null;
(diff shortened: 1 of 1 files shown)


indra/llcommon/lluuid.cpp
```diff
@@ -400,11 +401,16 @@ LLUUID LLUUID::operator^(const LLUUID& rhs) const
...
 void LLUUID::combine(const LLUUID& other, LLUUID& result) const
 {
-	HBXXH128 hash((const void*)mData, 16, false);	// false = do not finalize
-	hash.update((const void*)other.mData, 16);
-	hash.digest(result);
+	LLMD5 md5_uuid;
+	md5_uuid.update((unsigned char*)mData, 16);
+	md5_uuid.update((unsigned char*)other.mData, 16);
+	md5_uuid.finalize();
+	md5_uuid.raw_digest(result.mData);
```
akleshchev, line 413: Won't it be better to revert whole lluuid class? De we use uuid hashing often per second?

akleshchev:
>my previous commits (https://github.com/secondlife/viewer/commit/af656500b1a8ae0253abbb0da4a2f407bd62b79e)
Wrong changeset, this change was done in initial xxHash commit, not in the "std and boost containers keys"

indra/llcommon/lluuid.cpp
sldevel: Nah, the rest is safe (random UUID creation is safe, since... random).

sldevel:
> Wrong changeset, this change was done in initial xxHash commit, not in the "std and boost containers keys"

Oops, sorry... mixed the references... It was indeed in commit d75dbd8

PR #76, opened 2023-02-11
SL-19121: Add some override tests

cosmic-linden:
These tests confirm material overrides behave as expected for a few select cases.

For merge convenience, these are built on top of [SL-19080](https://jira.secondlife.com/browse/SL-19080), which is delayed by server changes. This PR will reviewed/merged after SL-19080 .

### DIFF
--- a/indra/llprimitive/tests/llgltfmaterial_test.cpp
+++ b/indra/llprimitive/tests/llgltfmaterial_test.cpp
@@ -339,4 +339,31 @@ namespace tut
             }
         }
     }
+
+    // Test non-persistence of default value flags in overrides
+    template<> template<>
+    void llgltfmaterial_object_t::test<11>()
+    {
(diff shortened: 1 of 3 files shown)


indra/llprimitive/tests/llgltfmaterial_test.cpp
```diff
@@ -0,0 +1,255 @@
...
+        {
+            LLGLTFMaterial metallic_factor_material;
+            metallic_factor_material.setMetallicFactor(0.5);
+            const std::string metallic_factor_material_json = metallic_factor_material.asJSON();
+            ensure_gltf_material_trimmed(metallic_factor_material_json, "baseColorTexture");
+            ensure_gltf_material_trimmed(metallic_factor_material_json, "metallicRoughnessTexture");
+        }
+    }
+
+    // Test that a GLTF material preserves values on serialization
```
akleshchev, line 229: If you are testing serialiation, I suggest enshuring that 'default' values do not get serialized after variable was set to something then reset to default.

brad-linden (approved):
looks good to me

PR #81, opened 2023-02-13
Add multi item properties floater support

Ansariel:
Re-adding long lost functionality of multi item properties floater that was available in V1 days, but got lost in the transition to V2.


akleshchev:
Is there a way to trigger this? "Properties" appear to be disabled for inventory multi selections.

Ansariel:
The "Properties" option was disabled in LLInvFVBridge in case of multi-selections. The second commit re-enables it.

akleshchev (approved):

PR #84, opened 2023-02-14
Fix for crash on startup for M1 macs

Geenz:
Looks like on M1 (and potentially Apple Silicon generally) - texture when sampling a shadow map implicitly returns a float causing the GLSL compiler to choke on startup on shadowUtil.


Geenz:

brad-linden:
this fixed it for me,  looks like we duplicated a bit of effort, 🤷

PR #85, opened 2023-02-14
SL-19002: Handle edge case in client networking which could cause material ID and overrides to become out-of-sync

cosmic-linden:


akleshchev (approved):

akleshchev:
Looks good, but pelase make sure this does not cause [SL-18583](https://jira.secondlife.com/browse/SL-18583) to reappear or add  it to test plan.

cosmic-linden:
> Looks good, but pelase make sure this does not cause [SL-18583](https://jira.secondlife.com/browse/SL-18583) to reappear or add scenario from [SL-18583](https://jira.secondlife.com/browse/SL-18583) to test plan.

Thanks, I'll double check and also add it to the test plan.

cosmic-linden:
@akleshchev SL-18583 appears to still not reproduce. I will merge this PR now.

PR #89, opened 2023-02-23
SL-19228: Fix GLTF texture transform rotation + add UV debug

cosmic-linden:
UV debug is PBR only. See textureUtilV.glsl for UV coordinate comments

### DIFF
--- a/indra/llprimitive/llgltfmaterial.cpp
+++ b/indra/llprimitive/llgltfmaterial.cpp
@@ -54,8 +54,8 @@ LLMatrix3 LLGLTFMaterial::TextureTransform::asMatrix()
     const F32 cos_r = cos(mRotation);
     const F32 sin_r = sin(mRotation);
     rotation.mMatrix[0][0] = cos_r;
-    rotation.mMatrix[0][1] = sin_r;
-    rotation.mMatrix[1][0] = -sin_r;
+    rotation.mMatrix[0][1] = -sin_r;
+    rotation.mMatrix[1][0] = sin_r;
     rotation.mMatrix[1][1] = cos_r;
 
     LLMatrix3 offset;
--- a/indra/newview/app_settings/shaders/class1/deferred/textureUtilV.glsl
+++ b/indra/newview/app_settings/shaders/class1/deferred/textureUtilV.glsl
@@ -1,0 +1,55 @@
+/** 
+ * @file class1/deferred/textureUtilV.glsl
+ *
+ * $LicenseInfo:firstyear=2023&license=viewerlgpl$
+ * Second Life Viewer Source Code
+ * Copyright (C) 2023, Linden Research, Inc.
+ * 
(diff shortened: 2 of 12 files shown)


indra/llprimitive/llgltfmaterial.cpp
```diff
@@ -54,8 +54,8 @@ LLMatrix3 LLGLTFMaterial::TextureTransform::asMatrix()
     const F32 cos_r = cos(mRotation);
     const F32 sin_r = sin(mRotation);
     rotation.mMatrix[0][0] = cos_r;
-    rotation.mMatrix[0][1] = sin_r;
-    rotation.mMatrix[1][0] = -sin_r;
+    rotation.mMatrix[0][1] = -sin_r;
+    rotation.mMatrix[1][0] = sin_r;
```
cosmic-linden, line 58: This makes the transform matrix identical to the one in the KHR_texture_transform spec.

indra/newview/app_settings/shaders/class1/deferred/textureUtilV.glsl
```diff
@@ -0,0 +1,55 @@
...
+
+    // Convert to left-handed coordinate system. The offset of 1 is necessary
+    // for rotations to be applied correctly.
+    // In the future, we could bake this coordinate conversion into the uniform
+    // that khr_gltf_transform comes from, since it's applied immediately
+    // before.
+    texcoord.y = 1.0 - texcoord.y;
+    texcoord = (khr_gltf_transform * vec3(texcoord, 1.0)).xy;
+    // Convert back to right-handed coordinate system
+    texcoord.y = 1.0 - texcoord.y;
```
cosmic-linden, line 48: Second part of the fix: Apply the GLTF texture transform in the correct UV coordinate space. This can be baked into the transform matrix later if we want.

indra/newview/skins/default/xui/en/menu_viewer.xml
```diff
@@ -2843,6 +2843,16 @@ function="World.EnvPreset"
            function="Advanced.ToggleInfoDisplay"
            parameter="texture anim" />
         </menu_item_check>
+        <menu_item_check
+         label="Texture Coordinates"
+            name="RenderDebugTexcoord">
```
cosmic-linden, line 2848: Everything else here is debug code/refactoring

RunitaiLinden (changes requested):
Kindly remove all the #if DEBUG_TEXCOORD code.  This is an anti-pattern that I've had to remove once before to make the shaders legible again, it creates a TON of clutter once you get more than a few debug displays implemented this way.  The proper way to provide data like that to end users is to use the LLPipeline::renderDebug framework and push all the geometry again using a dedicated debug shader that outputs the data you want to visualize. 
 Keeps the runtime code clean.  For developers, they can get at the data using GPU tracing tools or editing the shaders.

PR #90, opened 2023-02-24
Stream all animations through puppetry channel

AndrewMeadows:
This PR introduces an experiment where the local Viewer computes the final animation state and streams ALL joints through the puppetry relay of the server.  Remote Viewers will receive the stream and just play it, instead of mixing various animation assets.

### DIFF
--- a/indra/newview/llpuppetevent.h
+++ b/indra/newview/llpuppetevent.h
@@ -40,6 +40,11 @@
 #include "v3math.h"
 #include "llframetimer.h"
 
+namespace LLIK {
+    size_t pack_vec3(U8* wptr, LLVector3 vec);
+    size_t pack_quat(U8* wptr, LLQuaternion quat);
+};
+
 class LLPuppetJointEvent
 {
     //Information about an expression event that we want to broadcast
@@ -81,7 +86,7 @@ private:
     LLQuaternion mRotation;
     LLVector3 mPosition;
     LLVector3 mScale;
-    U16 mJointID = -1;
+    S16 mJointID = -1;
     U8 mMask = 0x0;
 };
 
--- a/indra/newview/llpuppetmotion.cpp
+++ b/indra/newview/llpuppetmotion.cpp
@@ -544,29 +544,6 @@ LLIK::Constraint::ptr_t get_constraint_by_joint_id(S16 joint_id)
 // END HACK
 
 
-void LLPuppetMotion::DelayedEventQueue::addEvent(
-        Timestamp remote_timestamp,
-        Timestamp local_timestamp,
-        const LLPuppetJointEvent& event)
-{
-    if (mLastRemoteTimestamp != -1)
-    {
-        // dynamically measure mEventPeriod and mEventJitter
-        constexpr F32 DEL = 0.1f;
-        Timestamp this_period = remote_timestamp - mLastRemoteTimestamp;
-        mEventJitter = (1.0f - DEL) * mEventJitter + DEL * std::abs(mEventPeriod - (F32)this_period);
-
-        // mEventPeriod is a running average of the period between events
-        mEventPeriod = (1.0f - DEL) * mEventPeriod + DEL * this_period;
-    }
-    mLastRemoteTimestamp = remote_timestamp;
-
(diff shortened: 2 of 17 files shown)


indra/newview/llpuppetevent.h
```diff
@@ -40,6 +40,11 @@
 #include "v3math.h"
 #include "llframetimer.h"
 
+namespace LLIK {
+    size_t pack_vec3(U8* wptr, LLVector3 vec);
+    size_t pack_quat(U8* wptr, LLQuaternion quat);
+};
```
AndrewMeadows, line 46: The `pack_*()` methods were originally helper methods in **llpuppetevent.cpp**, but they are being used in **llvoavatarself.cpp** so I moved them to the `LLIK` namespace to make them available, but now I think they need a better home than `LLIK`.

indra/newview/llpuppetmotion.cpp
```diff
@@ -544,29 +544,6 @@ LLIK::Constraint::ptr_t get_constraint_by_joint_id(S16 joint_id)
...
-        // mEventPeriod is a running average of the period between events
-        mEventPeriod = (1.0f - DEL) * mEventPeriod + DEL * this_period;
-    }
-    mLastRemoteTimestamp = remote_timestamp;
-
-    // we push event into the future so we have something to interpolate toward
-    // while we wait for the next
-    Timestamp delayed_timestamp = local_timestamp + (Timestamp)(mEventPeriod + mEventJitter);
-    mQueue.push_back({delayed_timestamp, event});
-}
```
AndrewMeadows, line 568: `LLPuppetMotion` used to both apply IK results (to the local avatar) and also play the puppetry stream (on other avatars) however the "play puppetry stream" duties have been moved to `LLStreamingMotion`.

indra/newview/llstreamingmotion.cpp
```diff
@@ -0,0 +1,332 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Linden Research, Inc., 945 Battery Street, San Francisco, CA  94111  USA
+ * $/LicenseInfo$
+ */
```
AndrewMeadows, line 25: LLStreamingMotion provides the receiving logic for other avatars.  When active it should be the ONLY active LLMotion in the LLMotionController.

indra/newview/llstreamingmotion.cpp
```diff
@@ -0,0 +1,332 @@
...
+        // mEventPeriod is a running average of the period between events
+        mEventPeriod = (1.0f - DEL) * mEventPeriod + DEL * this_period;
+    }
+    mLastRemoteTimestamp = remote_timestamp;
+
+    // we push event into the future so we have something to interpolate toward
+    // while we wait for the next
+    Timestamp delayed_timestamp = local_timestamp + (Timestamp)(mEventPeriod + mEventJitter);
+    mQueue.push_back({delayed_timestamp, event});
+}
```
AndrewMeadows, line 75: Much of this logic was moved out of `LLPuppetMotion`.

indra/newview/llvoavatarself.cpp
```diff
@@ -1350,6 +1364,260 @@ void LLVOAvatarSelf::updateMotions(LLCharacter::e_update_t update_type)
     }
 }
```
AndrewMeadows, line 1366: Sending the animation stream is now done by `LLVOAvatarSelf`, not by `LLPuppetMotion`.  All other avatars will only receive animation streams and use `LLStreamingMotion` to play them.

Rider-Linden:
I just thought of a gatcha that we are going to have to resolve. 
If we are sending the animations through the puppetry stream, the simulator will not send puppetry data for avatars more than 25 meters away from each other.  

It isn't hard to change or remove that limit... it is just something we need to be aware of and remember to do.

PR #91, opened 2023-02-24
Send binary LLSD to LEAP plugins; extend LLSD compatibility

AndrewMeadows:
This PR ports work from #29 (which was put in the **contribute** branch) for the **DRTVWR-558** branch (**puppetry**).  It is not clear if the work piling up in **contribute** will ever be merged, and enhancing the LEAP channel is of particular interest to the puppetry project, so this port is being submitted in case we decide to use it.

Note: with the **puppetry** project experiencing a long development cycle on its own branch... finally merging it into the **main** branch will be a messy operation and the work in this PR will contribute to that mess significantly.  However, perhaps that will be the best path for the final delivery of this work.


AndrewMeadows:
Closing this PR: the wrong set of changes were included in the port.  I'll try again and either re-open this PR or create a new one.

PR #95, opened 2023-02-27
SL-19209 Switch MS Bing to MS Azure

akleshchev:


marchcat-pe (approved):

marchcat-pe (approved):

maxim-productengine (approved):
Looks good!

indra/newview/lltranslate.cpp
```diff
@@ -194,8 +201,10 @@ void LLTranslationAPIHandler::translateMessageCoro(LanguagePair_t fromTo, std::s
         LL_INFOS("Translate") << "No translation URL" << LL_ENDL;
         return;
     }
+    LL_INFOS() << "Message: " << msg << LL_ENDL;
```
maxim-productengine: Probably we don't want to log all translated messages.

PR #99, opened 2023-03-01
SL-18330: Address further code review comments

nat-goodspeed:


indra/llcommon/lltracerecording.cpp
```diff
@@ -811,7 +811,7 @@ F64 PeriodicRecording::getPeriodMin( const StatType<EventAccumulator>& stat, siz
 			: NaN;
 }
 
-F64 PeriodicRecording::getPeriodMax( const StatType<EventAccumulator>& stat, size_t num_periods /*= S32_MAX*/ )
+F64 PeriodicRecording::getPeriodMax( const StatType<EventAccumulator>& stat, size_t num_periods /*= std::numeric_limits<size_t>::max()*/ )
 {
     LL_PROFILE_ZONE_SCOPED_CATEGORY_STATS;
 	num_periods = llmin(num_periods, getNumRecordedPeriods());
```
AndrewMeadows, line 817: I'm worried about how `getNumRecordedPeriods()` now automatically returns `size_t` type.  In particular, if you look in `PeriodicRecording::nextPeriod()` you'll see logic like this:

`mNumRecordedPeriods = llmin((S32)mRecordingPeriods.size() - 1, mNumRecordedPeriods + 1);`

So, what happens when mRecordingsPeriods is empty?   Does `mNumRecordedPeriods` get set to the unsigned equivalent to a negative number?  If that happens  ... then consider what happens in the loop below:
`for (S32 i = 1; i < num_periods; ++i)`
Does `num_periods` get converted to `S32` for the comparison?  Or does `i` get convertedd to `size_t`?  Is it possible for that for loop to effectively spinlock when `num_periods = (size_t)(-1)` ?

In short: I think `mNumRecordedPeriods`, and associated math, really should be `int32_t` instead of `size_t`.

AndrewMeadows: Alternatively... maybe the logic in PeriodicRecording::nextPeriod() should be fixed to never rollunder.  I don't know why the math subtracts 1 from the size.

If `nextPeriod()` is fixed... there remains a bunch of code that uses `mNumRecordedPeriods` and is casting to `S32`.  Maybe that should be audited too and converted to use `size_t`.

AndrewMeadows (changes requested):

AndrewMeadows (approved):
Looks good!

indra/llcommon/lltracerecording.cpp
AndrewMeadows: The new fixes look good.

PR #110, opened 2023-03-06
SL-18721: Faster viewer shutdown time since performance improvements can lead to perceived inventory loss due to cache corruption

LLGuru:
Make `LLWindowWin32::LLWindowWin32Thread::close()` to be called from the class destructor

### DIFF
--- a/indra/llcommon/threadpool.cpp
+++ b/indra/llcommon/threadpool.cpp
@@ -39,6 +39,11 @@ void LL::ThreadPool::start()
                 run(tname);
             });
     }
+
+    // Special workflow for LLWindowWin32Thread - it's close() should be called explicitly
+    if (mExplicitShutdown)
+        return;
+
     // Listen on "LLApp", and when the app is shutting down, close the queue
     // and join the workers.
     LLEventPumps::instance().obtain("LLApp").listen(
--- a/indra/llcommon/threadpool.h
+++ b/indra/llcommon/threadpool.h
@@ -59,6 +59,10 @@ namespace LL
          */
         virtual void run();
 
+    protected:
+        // LLWindowWin32Thread should set this flag to true
+        bool mExplicitShutdown { false };
+
     private:
         void run(const std::string& name);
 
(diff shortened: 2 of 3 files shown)


indra/llcommon/threadpool.cpp
```diff
@@ -39,6 +39,11 @@ void LL::ThreadPool::start()
                 run(tname);
             });
     }
+
+    // Special workflow for LLWindowWin32Thread - it's close() should be called explicitly
+    if (mExplicitShutdown)
+        return;
```
akleshchev, line 45: What issues close() call in such case?

LLGuru: > What issues close() call in such case?
As iit was said in the comment above, the destructor of the class ThreadPool calls this method in any case:
```
LL::ThreadPool::~ThreadPool()
{
    close();
}
```

akleshchev (approved):

nat-goodspeed (commented):
I understand that you're deferring the `LLWindowWin32Thread::close()` call to its destructor, and that it's explicitly destroyed by `~LLWindowWin32()`. But I don't know where the `LLWindow` subclass is destroyed?

LLGuru:
> I understand that you're deferring the `LLWindowWin32Thread::close()` call to its destructor, and that it's explicitly destroyed by `~LLWindowWin32()`. But I don't know where the `LLWindow` subclass is destroyed?

The class `LLWindow` destructor is being explicitly called from `LLWindowManager::destroyWindow()`
```
BOOL LLWindowManager::destroyWindow(LLWindow* window)
{
	if (sWindowList.find(window) == sWindowList.end())
	{
		LL_ERRS() << "LLWindowManager::destroyWindow() : Window pointer not valid, this window doesn't exist!" 
			<< LL_ENDL;
		return FALSE;
	}

	window->close();

	sWindowList.erase(window);

	delete window;

	return TRUE;
}
```

nat-goodspeed (approved):
Thank you. I thought we'd need a new call at shutdown time, but your answer addresses my concern.

PR #113, opened 2023-03-07
SL-19128: Fix adding material to face sometimes not respecting protections

cosmic-linden:


indra/newview/llselectmgr.cpp
```diff
@@ -1938,6 +1938,12 @@ void LLSelectMgr::selectionSetGLTFMaterial(const LLUUID& mat_id)
             LLUUID asset_id = mMatId;
             if (mItem)
             {
+                // If success, the material may be copied into the object's inventory
```
akleshchev, line 1941: Looks got to me, but couple questions:
1. Instead of doing this check here, shouldn't it be better to extend allowOperationBy check below (1957) to cover "no transfer"?
2 selectionSetImage handles a case for "te == -1", my assumtion that this never happens and thus isn't covered here? (and actually shouldn't be covered in selectionSetImage)

akleshchev (approved):

indra/newview/llselectmgr.cpp
cosmic-linden: 1. Now that I look closer, I see a potential for cleanup here. I think `handleDropMaterialProtections` handles the important edge cases, so probably best to rely exclusively on that. I will push an update if the code works after cleanup.
2. You're probably right that the -1 case is never handled

indra/newview/llselectmgr.cpp
```diff
@@ -1954,15 +1937,8 @@ void LLSelectMgr::selectionSetGLTFMaterial(const LLUUID& mat_id)
         }
     };
 
-    if (item && !item->getPermissions().allowOperationBy(PERM_COPY, gAgent.getID()))
-    {
-        getSelection()->applyNoCopyPbrMaterialToTEs(item);
```
akleshchev, line 1953: applyNoCopyPbrMaterialToTEs should no longer be in use, so whole function can be removed.

Also dropTextureOneFace handles protextions for textures, so similar operation can be done for applyNoCopyTextureToTEs

indra/newview/llselectmgr.cpp
cosmic-linden: Looking closer at this, I think that this cleanup is not possible. Only thing that seems doable is removing the -1 case. Other cases may still be needed. Probably not worth messing with too much at this testing stage.

indra/newview/llselectmgr.cpp
cosmic-linden: I reverted the cleanup that removed this code. There were various cases that I did not account for which could cause issues with permissions.

PR #115, opened 2023-03-09
SL-19330: Fix texture animation in wrong direction when setting material on another side

cosmic-linden:


indra/newview/llvovolume.cpp
```diff
@@ -670,6 +670,21 @@ void LLVOVolume::animateTextures()
...
+                    {
+                        off_s = 0.0f;
+                        off_t = 0.0f;
+                    }
+                    if (!(result & LLViewerTextureAnim::SCALE))
+                    {
+                        scale_s = 1.0f;
+                        scale_t = 1.0f;
+                    }
+
```
cosmic-linden, line 687: This is just a copy-paste of some similar logic for non-PBR materials up above

RunitaiLinden (approved):

PR #144, opened 2023-03-31
SL-18839: Add basic windows build

bennettgoble:
Add basic builds for win32 and win64 using Github actions. This is a foundation for additional work, and should be able to run in parallel with our existing TeamCity configurations. See an example run [here](https://github.com/secondlife/viewer/actions/runs/4579252214).

Screenshot 2023-03-31 150055

## Limitations

- Windows only
- ReleaseOS configuration
- Channel is not baked into resulting installable
- Not signed
- Installers are zipped
- Build artifacts last 90 days (by default)

At a minimum, this changeset allows us to verify PRs/commits to shared branches such as `contribute` do not break the build. It also allows builds to be manually triggered, which may be useful for producing regular/periodic downloads for projects.

## Future work

There's a lot of work remaining here. This PR is a first step that gives us some initial useful value.

### macOS

We need to get some dedicated macOS hosts set up so that darwin builds don't take an eternity --at least until Github releases more[ powerful macOS managed runners](https://github.blog/2023-03-01-github-actions-introducing-faster-github-hosted-x64-macos-runners/).

### Proprietary libs

We also need to move all proprietary autobuild dependencies from our internal artifact server which requires VPN access to Github release artifacts so that we can use Github credentials to download 3p deps like fmod.

### Releases

Ideally the Second Life viewer release process will incorporate Github Releases in the future. These could be used to help publish changenotes and permanently host project viewer releases, including project viewer nightlies. Publishing files through releases would mean that we can provide installation artifacts that are not wrapped in a zip archive, and can have no shelf life.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,93 @@
+name: Build
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches: [main, contribute]
+    tags: ["*"]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        runner: [windows-large]
+        configuration: [ReleaseOS]
+        addrsize: [64]
+        include:
+          - runner: windows-large
+            configuration: ReleaseOS
+            addrsize: 32
+    runs-on: ${{ matrix.runner }}
+    env:
+      AUTOBUILD_CONFIGURATION: ${{ matrix.configuration }}
+      AUTOBUILD_ADDRSIZE: ${{ matrix.addrsize }}
+      AUTOBUILD_INSTALLABLE_CACHE: ${{ github.workspace }}/.autobuild-installables
+      AUTOBUILD_VARIABLES_FILE: ${{ github.workspace }}/.build-variables/variables
+      AUTOBUILD_VSVER: "170" # vs2k22
+      LOGFAIL: debug # Show details when tests fail
+      GIT_REF: ${{ github.head_ref || github.ref }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      - name: Checkout build variables
+        uses: actions/checkout@v3
+        with:
+          repository: secondlife/build-variables
+          ref: viewer
+          path: .build-variables
+
+      - name: Install autobuild and python dependencies
+        run: pip3 install autobuild llbase
+
+      - name: Cache autobuild packages
+        uses: actions/cache@v3
+        id: cache-installables
+        with:
+          path: .autobuild-installables
+          key: ${{ runner.os }}-${{ matrix.addrsize }}-${{ matrix.configuration }}-${{ hashFiles('autobuild.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.addrsize }}-${{ matrix.configuration }}-
+            ${{ runner.os }}-${{ matrix.addrsize }}-
+
+      - name: Install windows dependencies
+        if: runner.os == 'Windows'
+        run: choco install nsis-unicode
+
+      - name: Build
+        id: build
+        shell: bash
+        env:
+          RUNNER_OS: ${{ runner.os }}
+        run: |
+          # On windows we need to point the build to the correct python
+          # as neither CMake's FindPython nor our custom Python.cmake module
+          # will resolve the correct interpreter location.
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            export PYTHON="$(cygpath -m "$(which python)")"
+            echo "Python location: $PYTHON"
+          fi
+          
+          autobuild configure -- -DVIEWER_CHANNEL="Second Life Test ${GIT_REF##*/}"
+          autobuild  build --no-configure
+
+          # Find artifacts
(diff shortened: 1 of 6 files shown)


indra/newview/installers/windows/installer_template.nsi
```diff
@@ -26,7 +26,6 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Compiler flags
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-Unicode true
```
brad-linden, line 29: whats the story here?  is this needed?  is it temporary until we have a workaround for some problem?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,89 @@
+name: Build
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches: ["contribute"]
```
bennettgoble, line 7: Action trigger criteria is very intentional here: builds are not triggered for every push unless the change is made to an important shared branch such as `contribute` (the default development branch.) It is possible to specify branch patterns here. For example, if we wanted to trigger a build for every DRTSIM commit: `branches: ["contribute", "DRTSIM-*"]`.

There is potential here to aid in a movement towards a git ops workflow, where project viewers, releases, release notes are all driven through git operations.

brad-linden (approved):
> I presume we will want to switch back to using unicode nsis properly when we can?

never mind this will be fixed by upgrading nsis installer elsewhere

indra/newview/installers/windows/installer_template.nsi
bennettgoble: The `Unicode` parameter has been dropped from newer NSIS versions as `Unicode true` is default behavior. We have a couple options here:

1. Update TeamCity agents' NSIS installer
2. Perform some file manipulation on installer_template.nsi either in TC or Github to patch up the file for the respective NSIS installer version
3. Provide separate installer_template files...

My preference would be #1.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,89 @@
...
+    env:
+      AUTOBUILD_CONFIGURATION: ${{ matrix.configuration }}
+      AUTOBUILD_ADDRSIZE: ${{ matrix.addrsize }}
+      AUTOBUILD_INSTALLABLE_CACHE: ${{ github.workspace }}/.autobuild-installables
+      AUTOBUILD_VARIABLES_FILE: ${{ github.workspace }}/.build-variables/variables
+      AUTOBUILD_VSVER: "170" # vs2k22
+      LOGFAIL: debug # Show details when tests fail
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
toastercup, line 30: Similar to as we do elsewhere, should we do:
```suggestion
        uses: actions/checkout@v3
        with:
          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
Otherwise, for builds that are the result of a PR event, the `ref` will point to the merge commit created by GitHub, which could be undesirable for our builds.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,89 @@
...
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches: ["contribute"]
+
+jobs:
+  build:
+    strategy:
+      matrix:
```
toastercup, line 12: :tada:

.github/workflows/build.yaml
```diff
@@ -0,0 +1,89 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: secondlife/build-variables
+          ref: viewer
+          path: .build-variables
+
+      - name: Install autobuild and python dependencies
+        run: pip3 install autobuild llbase
+
+      - name: Cache autobuild packages
```
toastercup, line 42: :1st_place_medal: Warms my heart to see dependency caching, and a nicely-thought-out cache key/restore keys!

toastercup (changes requested):
This is exciting! :exclamation: Just one piece of feedback that may or may not be relevant/necessary..

.github/workflows/build.yaml
```diff
@@ -0,0 +1,92 @@
...
+        run: |
+          # On windows we need to point the build to the correct python
+          # as neither CMake's FindPython nor our custom Python.cmake module
+          # will resolve the correct interpreter location.
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            export PYTHON="$(cygpath -m "$(which python)")"
+            echo "Python location: $PYTHON"
+          fi
+
+          autobuild configure
```
brad-linden, line 73: Had another potential idea:
```suggestion
          autobuild configure -- -DVIEWER_CHANNEL="Second Life Nightly ${GITHUB_REF##*/}"
```

bennettgoble: Good idea; I've adapted it to be more general: "Second Life Test ${GITHUB_REF_NAME}"

toastercup (approved):
Splendid work - thanks for addressing the feedback!

PR #149, opened 2023-04-04
SL-19042: Replace FMOD with VLC for parcel audio

LLGuru:


marchcat-pe (approved):

callumlinden (approved):

akleshchev (approved):

indra/newview/app_settings/settings.xml
```diff
@@ -7077,6 +7077,17 @@
...
+    <key>UseMediaPluginsForStreamingAudio</key>
+    <map>
+      <key>Comment</key>
+      <string>Use media plugins (VLC) for streaming audio.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>Boolean</string>
+      <key>Value</key>
+      <integer>0</integer>
```
akleshchev, line 7089: We probably should make VLC the dafault one when it's present, otherwise what's the point? But ability to toggle back to fmod is still usefull in case we run into issues or for testing.

LLGuru: > Or alternative option - give it proper testing first and enable it by default afterward.

That was the plan

akleshchev: I suspect that by the time D582 reaches testing we will forget this change exists, and it is likely to get regression tested with the option still off and passed as is. I suggest requesting a pre-merge testing next time.

P.S. Updated test plan to include UseMediaPluginsForStreamingAudio since otherwise QA wouldn't even know it's there.

marchcat-pe: We can still request testing for this specific feature now, and once that's done, enable it by default in 582.

PR #152, opened 2023-04-04
SL-19541: Add basic coding standards check

bennettgoble:
Run Linden coding standard git-hooks in a basic Github actions workflow.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,8 @@
 repos:
-  - repo: https://bitbucket.org/lindenlab/git-hooks.git 
-    rev: v1.0.0-beta2
+  - repo: https://github.com/secondlife/git-hooks.git 
+    rev: v1.0.0
     hooks:
       - id: opensource-license
-      - id: jira-issue 
       - id: llsd
       - id: no-trigraphs
       - id: copyright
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,17 +1,16 @@
 repos:
-  - repo: https://bitbucket.org/lindenlab/git-hooks.git 
-    rev: v1.0.0-beta2
+  - repo: https://github.com/secondlife/git-hooks.git 
+    rev: v1.0.0
     hooks:
       - id: opensource-license
-      - id: jira-issue
```
bennettgoble, line 6: I've explicitly removed the jira-issue git hook because it would be a huge barrier to incorporating opensource contributions.

nat-goodspeed: @vir-linden , do we still want to require that check for Linden commits?

@bennettgoble , could we make that conditional for local Linden commits if desired?

bennettgoble: That may be possible through some additional internal-developer-only step. Nothing prevents developers from adding their own git hooks in addition to open source friendly ones.

I'd like to suggest we merge this PR as is and follow up with additional tooling or guidance, if desired, for internal devs.

nat-goodspeed (approved):

PR #156, opened 2023-04-05
Andreyk/sl 19522

akleshchev:
Fix viewer not writing missing/dupplicate keys but assigning 'num' values to file like they are still there.


marchcat-pe:
DRTVWR-477? looks more suitable for 580 or 582 (577 only if urgent)

marchcat-pe (approved):
LGTM
🚢it.

PR #159, opened 2023-04-07
merge main into contribute

marchcat-pe:


bennettgoble (approved):
@marchcat-pe @vir-linden approved.

Are these merge PRs useful or should we just allow updates to `contribute` to be pushed by maintainers? I think this exercise was only mandatory because `contribute` has branch protections turned on which prohibit direct pushes. IMHO, I think opening a PR for every merge from main would be unorthodox and unneeded, and that specific protection could be removed.

marchcat-pe:
@bennettgoble right, it's just for merging `main` into the locked `contribution` since it's locked.

bennettgoble:
> @bennettgoble right, it's just for merging `main` into the `contribution` branch since it's locked.

OK, I've removed that branch protection rule. In the future merge commits can just be pushed to `contribute` unless there is a specific reason for a PR. Thanks, @marchcat-pe

marchcat-pe:
thank you!

indra/newview/pipeline.cpp
```diff
@@ -1775,20 +1775,17 @@ void LLPipeline::unlinkDrawable(LLDrawable *drawable)
...
-    }
+	for (light_set_t::iterator iter = gPipeline.mNearbyLights.begin();
+		 iter != gPipeline.mNearbyLights.end(); iter++)
+	{
+        const LLViewerObject *vobj = iter->drawable->getVObj();
+        if (vobj && vobj->getAvatar()
+            && vobj->isAttachment() && vobj->getAvatar() == muted_avatar)
+		{
+			gPipeline.mLights.erase(iter->drawable);
+			gPipeline.mNearbyLights.erase(iter);
```
nat-goodspeed, line 1786: This isn't valid! `std::set::erase(iterator)` invalidates the passed iterator, and even though it doesn't invalidate any other iterators in the container, the `iter++` at line 1779 tries to increment the invalidated iterator.

Despite its awkwardness, the original loop logic is more correct.

nat-goodspeed (commented):
There's at least one C++ error. Please review.

bennettgoble:
@nat-goodspeed this changeset is a merge commit with code that has already been merged into main and shipped. How should these review comments be addressed? New Jiras?

marchcat-pe:
I've created a task for that: SL-19578

PR #170, opened 2023-04-14
SL-19591 Prompt user to save current settings as Graphics Preset before enabling AutoFPS

marchcat-pe:


indra/newview/llfloaterperformance.cpp
```diff
@@ -672,4 +678,22 @@ void LLFloaterPerformance::updateAutotuneCtrls(bool autotune_enabled)
     getChild<LLTextBox>("wip_desc")->setVisible(autotune_enabled && !auto_tune_locked);
     getChild<LLTextBox>("display_desc")->setVisible(LLPerfStats::tunables.vsyncEnabled);
 }
+
+void LLFloaterPerformance::enableAutotuneWarning()
+{
+    if (!gSavedPerAccountSettings.getBOOL("HadEnabledAutoFPS") && LLPerfStats::tunables.userAutoTuneEnabled)
```
akleshchev, line 684: This probably shouldn't be in draw() and if you are putting it into draw at least make getBool into a static.

PR #177, opened 2023-04-18
SL-19605 handling of batched material overrides in LargeGenericMessage

brad-linden:


brad-linden:
see related server change here: https://github.com/secondlife/server/pull/244

to avoid problems the viewer change should go out first

PR #179, opened 2023-04-18
brad/mac nightly builds

brad-linden:
Experiment with turning on MacOS builds for github based nightly builds for DRTVWR-559


bennettgoble:
@brad-linden You may want to adopt the new large macOS runners so we can have faster builds.

bennettgoble (changes requested):
Suggestion: [Large macOS runners](https://github.blog/changelog/2023-04-24-github-actions-faster-macos-runners-are-now-available-in-open-public-beta/)

bennettgoble (approved):

PR #191, opened 2023-04-24
Fix SL-19675 crash due to thread unsafe LLRefCount usage possibly related to SL-19648

brad-linden:


cosmic-linden (approved):
Looks good. I recommend documenting expected behavior for fromJSON

PR #197, opened 2023-04-28
SL-19669 Integrate the contributed translations (Japanese) by logue

marchcat-pe:


Ansariel:
Is there a reason literally every file got reformatted so the only chance to "fix" the expected thousands of merge conflicts for TPVs is by simply ignoring all of these changes?

PR #199, opened 2023-04-30
BUG-233797/233798 - fix blackout when u/w fog_density < 0 

beqjanus:
raising a negative number to a non-integral power results in a non-real result (which is NaN for our purposes)
Two methods were tested, number 2 is being used as it avoids colour inversion:
1) Force the fog_mod to be integral. The effect is unlikely to be nice, but it is better than blackness.
In this method a few of the combinations are "usable" but the water colour is effectively inverted (blue becomes yellow)
this seems to be unlikely to be a desirable use case for the majority.
2) Force density to be an arbitrary non-negative (i.e. 1) when underwater and modifier is not an integer (1 was arbitrarily chosen as it gives at least some notion of fog in the transition)
This is more restrictive, effectively forcing a density under certain conditions, but allowing the range of #1 and avoiding blackness in other cases
at the cost of overriding the fog density. 

While this is a fix for 233797, it also addresses 233798, which results from trying to linearly transition from an impossible starting position to a new setting, meaning that once things are broken you cannot "unbreak" without some form of reset (reload shader, restart viewer etc.)


beqjanus:

indra/llinventory/llsettingswater.cpp
```diff
@@ -290,6 +290,19 @@ F32 LLSettingsWater::getModifiedWaterFogDensity(bool underwater) const
...
+        // In this method a few of the combinations are "usable" but the water colour is effectively inverted (blue becomes yellow)
+        // this seems to be unlikely to be a desirable use case for the majority.
+        // 2) Force density to be an arbitrary non-negative (i.e. 1) when underwater and modifier is not an integer (1 was aribtrarily chosen as it gives at least some notion of fog in the transition)
+        // This is more restrictive, effectively forcing a density under certain conditions, but allowing the range of #1 and avoiding blackness in other cases
+        // at the cost of overriding the fog density. 
+        if(fog_density < 0.0f && underwater_fog_mod != (F32)llround(underwater_fog_mod) )
+        {
+            fog_density = 1.0f;
+        }
         fog_density = pow(fog_density, underwater_fog_mod);
```
brad-linden, line 306: negative values of fog_density seem extremely non-physical.  when they do work they make water look kind of emissive.  would it be better to simply block them at the `LLSettingsWater::validationList()` stage?

akleshchev (approved):

akleshchev:
Thank you.

PR #203, opened 2023-05-03
DRTVWR-559: Replace debugLoggingEnabled() function with LL_DEBUGS().

nat-goodspeed:
The trouble with debugLoggingEnabled() is that it locked mutexes and searched maps every time that call was reached. LL_DEBUGS() has the same functionality (albeit with idiosyncratic syntax) but performs expensive lookups only once per session, caching the result in a local static variable.

### DIFF
--- a/indra/llcommon/llerror.h
+++ b/indra/llcommon/llerror.h
@@ -467,7 +467,29 @@ typedef LLError::NoClassInfo _LL_CLASS_TO_LOG;
 		LLError::CallSite& _site(_sites[which]);                        \
 		lllog_test_()
 
-// Check at run-time whether logging is enabled, without generating output
+/*
+// Check at run-time whether logging is enabled, without generating output.
+Resist the temptation to add a function like this because it incurs the
+expense of locking and map-searching every time control reaches it.
 bool debugLoggingEnabled(const std::string& tag);
+
+Instead of:
+
+if debugLoggingEnabled("SomeTag")
+{
+    // ... presumably expensive operation ...
+    LL_DEBUGS("SomeTag") << ... << LL_ENDL;
+}
+
+Use this:
+
+LL_DEBUGS("SomeTag");
+// ... presumably expensive operation ...
+LL_CONT << ...;
+LL_ENDL;
+
+LL_DEBUGS("SomeTag") performs the locking and map-searching ONCE, then caches
+the result in a static variable.
+*/ 
 
 #endif // LL_LLERROR_H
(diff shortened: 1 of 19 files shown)


indra/llwindow/llwindowwin32.cpp
```diff
@@ -2308,8 +2308,6 @@ LRESULT CALLBACK LLWindowWin32::mainWindowProc(HWND h_wnd, UINT u_msg, WPARAM w_
 
     LLWindowWin32* window_imp = (LLWindowWin32*)GetWindowLongPtr(h_wnd, GWLP_USERDATA);
 
-    bool debug_window_proc = false; // gDebugWindowProc || debugLoggingEnabled("Window");
```
nat-goodspeed, line 2311: `debug_window_proc` had no effect because the output it supposedly wrapped was solely controlled by `LL_INFOS("Window")`. The only effect of `if (debug_window_proc)` was to skip the cached `LL_INFOS()` test if `(! debug_window_proc)`.

indra/llcommon/llerror.h
```diff
@@ -467,7 +467,29 @@ typedef LLError::NoClassInfo _LL_CLASS_TO_LOG;
...
+
+LL_DEBUGS("SomeTag");
+// ... presumably expensive operation ...
+LL_CONT << ...;
+LL_ENDL;
+
+LL_DEBUGS("SomeTag") performs the locking and map-searching ONCE, then caches
+the result in a static variable.
+*/ 
+
```
RunitaiLinden, line 494: Nice!

indra/llwindow/llwindowwin32.cpp
RunitaiLinden: Oop, sorry for the cruft here.  I put those in to trace events while the window got multithreaded, but we don't need them any more.  Would probably be better to just delete all the code that was compile time if (false)'d out.  Should improve readability and save some microseconds.

RunitaiLinden (approved):

indra/newview/llviewerobject.h
```diff
@@ -28,6 +28,7 @@
 #define LL_LLVIEWEROBJECT_H
 
 #include <map>
+#include <unordered_map>
```
brad-linden, line 31: this intentional?

brad-linden (approved):

indra/newview/llviewerobject.h
nat-goodspeed: The Mac build failed without it, claiming there's no such animal as `std::unordered_map`.

brad-linden: yeah understandable.  I'm curious how this ever worked without it

nat-goodspeed: I won't even deplore our habits as coders, because the C++ language doesn't incentivize checking for the right `#include` every time we reference a new function or class. If our new code compiles, that's good enough...

So (a) many of our .cpp files rely on indirect header-file inclusion, subject to unexpected breakage if the set of `#includes` changes either in the .cpp file or in one of the headers it drags in; and (b) we doubtless have many `#includes` that are no longer needed.

Presumably some IDEs can tag every symbol with the header file declaring it, recognizing whether the current .cpp file `#include`s that same header (possibly redundantly), and can identify directly-included headers none of whose declarations are referenced ... but until such tooling becomes universal, I don't expect us all to suddenly become more meticulous about header usage.

PR #204, opened 2023-05-03
SL-19647: Eliminate LLSDArray entirely.

nat-goodspeed:


simon-linden (approved):
Have used these changes in the puppetry branch and it seems good.   It's definitely nice to clean up the old calls

PR #205, opened 2023-05-04
sl-19676 - Stats on updates and scene loading

simon-linden:
This is work that makes it easier to grab stats on updates coming from the servers and then analyzing how they taper off.   See https://lindenlab.atlassian.net/wiki/spaces/SLT/pages/2872377345/Grabbing+Viewer+Update+Stats


callumlinden:
Looks good - my only concern with changes along these lines is that they introduce small performance hits - I didn't see anything that stood out but figured it was worth mentioning..

callumlinden (approved):

PR #207, opened 2023-05-05
Fix exceptions during shutdown causing early program termination

RyeMutt:
This fixes a few uncaught exceptions during shutdown in the work queue and image threadpool


nat-goodspeed:
Hey Rye, thanks for surfacing this! Prompted by your fix, I think I'd like to make `WorkQueue::post()` return `bool` instead, which should simplify things for many callers. I'll file another PR and ask you to check it out.

PR #208, opened 2023-05-08
SL-19690: Follow up on Rye Mutt's fix for shutdown crashes.

nat-goodspeed:
Rather than continuing to propagate `try`/ `catch (Closed)` (aka `LLThreadSafeQueueInterrupt`) constructs through the code base, make `WorkQueueBase::post()` return `bool` indicating success (i.e. `! isClosed()`). This obviates `postIfOpen()`, which no one was using anyway.

In effect, `postIfOpen()` is renamed `post()`, bypassing the exception when `isClosed()`.

Review existing `try` / `catch` blocks of that sort, changing to test for `post()` returning false.

### DIFF
--- a/indra/llaudio/llaudiodecodemgr.cpp
+++ b/indra/llaudio/llaudiodecodemgr.cpp
@@ -607,40 +607,37 @@ void LLAudioDecodeMgr::Impl::startMoreDecodes()
 
         // Kick off a decode
         mDecodes[decode_id] = LLPointer<LLVorbisDecodeState>(NULL);
-        try
-        {
-            main_queue->postTo(
-                general_queue,
-                [decode_id]() // Work done on general queue
+        bool posted = main_queue->postTo(
+            general_queue,
+            [decode_id]() // Work done on general queue
+            {
+                LLPointer<LLVorbisDecodeState> decode_state = beginDecodingAndWritingAudio(decode_id);
+
+                if (!decode_state)
                 {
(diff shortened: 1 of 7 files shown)


indra/llaudio/llaudiodecodemgr.cpp
```diff
@@ -607,40 +607,37 @@ void LLAudioDecodeMgr::Impl::startMoreDecodes()
...
-                });
-        }
-        catch (const LLThreadSafeQueueInterrupt&)
+                }
+
+                // Disk write of decoded audio is now in progress off-thread
+                return decode_state;
+            },
+            [decode_id, this](LLPointer<LLVorbisDecodeState> decode_state) // Callback to main thread
+            mutable {
```
akleshchev, line 626: Are you sure audio will still exists when vorbis decode finishes and 'this' gets hit?
I feel like this shouldn't be this&mutable and instead check if things are still alive, like:

if (LLAudioDecodeMgr::isInstantExists() && !LLApp::isExiting()) LLAudioDecodeMgr::getInstance()->enqueueFinishAudio(decode_id, decode_state)

nat-goodspeed: Not my code, but isn't that what the `gAudiop` test is all about? At least, that's what the comments state.

akleshchev: You are correct, that should become null long before manager gets destroyed.

RyeMutt:
This looks good to me!

PR #210, opened 2023-05-10
puppetry IK fixes

AndrewMeadows:
This PR fixes a few problems with puppetry inverse kinematics.


indra/llcharacter/CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ if (LL_TESTS)
   include(LLAddBuildTest)
   # UNIT TESTS
   SET(llcharacter_TEST_SOURCE_FILES
-    #llik.cpp
+    llik.cpp
```
AndrewMeadows, line 81: I'm re-enabling the IK unit test projbect although most of the unit tests are temporarily disabled.

indra/llcharacter/llik.cpp
```diff
@@ -49,6 +49,7 @@ namespace
     constexpr const char *UNKNOWN_CONSTRAINT_NAME               ("UNKNOWN_CONSTRAINT");
     constexpr const char *SIMPLE_CONE_CONSTRAINT_NAME           ("SIMPLE_CONE");
     constexpr const char *TWIST_LIMITED_CONE_CONSTRAINT_NAME    ("TWIST_LIMITED_CONE");
+    constexpr const char *SHOULDER_CONSTRAINT_NAME              ("SHOULDER");
```
AndrewMeadows, line 52: I'm adding a `ShoulderConstraint` class.  The shoulder joint was previously using a `LimitedTwistCone` but that was insufficient.

indra/llcharacter/llik.cpp
```diff
@@ -271,35 +275,26 @@ size_t LLIK::Constraint::generateHash() const
...
     const LLQuaternion& local_rot = joint.getLocalRot();
     LLQuaternion adjusted_local_rot = computeAdjustedLocalRot(local_rot);
     if (! LLQuaternion::almost_equal(adjusted_local_rot, local_rot))
     {
+        // Note: we update joint's local-frame rot, but not its world-frame rot
+        // -- that responsibility belongs to external code.
         joint.setLocalRot(adjusted_local_rot);
+        joint.applyLocalRot();
         return true;
     }
```
AndrewMeadows, line 294: `joint.applyLocalRot(`) used to be called explicitly after a constraint had been enforced, so I moved it inside this context.

indra/llcharacter/llik.cpp
```diff
@@ -271,35 +275,26 @@ size_t LLIK::Constraint::generateHash() const
...
-    constexpr F32 MIN_AXIS_LENGTH = 1.0e-5f;
-    LLQuaternion new_local_rot = LLQuaternion::DEFAULT;
-    if (swing_axis.length() > MIN_AXIS_LENGTH)
-    {
-        F32 swing_angle = std::acos(mForward * joint_forward);
-        new_local_rot.setAngleAxis(swing_angle, swing_axis);
-    }
-    return new_local_rot;
-}
-
```
AndrewMeadows, line 302: I'm removing the `minimizeTwist()` methods.  I've learned: it is better to perform twist adjustments inside the `Constraint::enforce()` overrides.  See further below.

indra/llcharacter/llik.cpp
```diff
@@ -483,38 +478,117 @@ LLQuaternion LLIK::TwistLimitedCone::computeAdjustedLocalRot(const LLQuaternion&
...
+}
 
-    LLVector3 joint_forward = mForward * joint_local_rot;
-    LLVector3 swing_axis = mForward % joint_forward;
-    constexpr F32 MIN_SWING_AXIS_LENGTH = 1.0e-3f;
-    if (swing_axis.length() > MIN_SWING_AXIS_LENGTH)
+LLIK::ShoulderConstraint::ShoulderConstraint(LLSD &parameters):
+    LLIK::Constraint(LLIK::Constraint::SHOULDER_CONSTRAINT, parameters)
+{
+}
```
AndrewMeadows, line 500: Shoulder constraint doesn't have any private data members yet.  This may change later but for the now all it does is try to keep the elbow bend pointed down.  This is done in the `enforce()` override below, and the implementation of its `computeAdjustedLocalRot()` method is a no-op.

indra/llcharacter/llik.cpp
```diff
@@ -525,7 +599,7 @@ LLIK::ElbowConstraint::ElbowConstraint(const LLVector3& forward_axis, const LLVe
     mForward.normalize();
     mPivotAxis = mForward % (pivot_axis % mForward);
     mPivotAxis.normalize();
-    mLeft = mPivotAxis % mForward;
+    mPivotXForward = mPivotAxis % mForward;
```
AndrewMeadows, line 602: The name `ElbowConstraint::mLeft` was unclear so I changed it to something more meaningful.

indra/llcharacter/llik.cpp
```diff
@@ -577,6 +651,121 @@ size_t LLIK::ElbowConstraint::generateHash() const
     return seed;
 }
```
AndrewMeadows, line 653: The `ElbowConstraint::enforce()` override is a big contributor to higher stability and accuracy of the arm IK algorithm.  Basically what it does is: rather than change the direction of the lower-arm in order to enforce the constraint at the elbow it assumes its direction is correct and instead spins the upper- and lower-arm (shoulder, elbow) joints to line up the elbow's pivot axis in each's local-frame, all within the twist limits of the elbow's forearm.  Besides providing faster convergence of the IK iterations this also provides effective "back-pressure" on the shoulder joint when the lower arm can't quite reach its orientation target, which would allow lower joints to move in an attempt to help reach the goal... although we are not yet taking full advantage of that since the shoulder doesn't have any twist constraints ATM so most back-pressure doesn't propagate lower than the shoulder.

indra/llcharacter/llik.h
```diff
@@ -61,29 +61,33 @@ class Solver;
...
+
+constexpr Flag IK_FLAG_LOCAL_ROT = 1 << 1; // IK has adjusted local_rot
+constexpr Flag IK_FLAG_ACTIVE = 1 << 5;
+constexpr Flag IK_FLAG_LOCAL_ROT_LOCKED = 1 << 7; // local_rot is locked during IK
+
+constexpr Flag MASK_POS = CONFIG_FLAG_TARGET_POS | CONFIG_FLAG_LOCAL_POS;
+constexpr Flag MASK_ROT = CONFIG_FLAG_TARGET_ROT | CONFIG_FLAG_LOCAL_ROT;
+constexpr Flag MASK_TRANSFORM = MASK_POS | MASK_ROT;
+constexpr Flag MASK_LOCAL = CONFIG_FLAG_LOCAL_POS | CONFIG_FLAG_LOCAL_ROT | CONFIG_FLAG_DISABLE_CONSTRAINT;
+constexpr Flag MASK_TARGET = CONFIG_FLAG_TARGET_POS | CONFIG_FLAG_TARGET_ROT;
```
AndrewMeadows, line 84: I ran out of bits in the `U8` flags, so I upped it to `U16` and employed a `using` declaration to simplify future changes around that.

indra/newview/character/avatar_constraint.llsd
```diff
@@ -11,30 +11,30 @@
...
     'mCollarLeft':
         {
-            'cone_angle':r9,
+            'cone_angle':r3,
             'forward_axis':
                 [r-0.021,r0.085,r0.165],
-            'max_twist':r18,
-            'min_twist':r-18,
+            'max_twist':r3,
+            'min_twist':r-3,
```
AndrewMeadows, line 18: I'm tightening the collar constraints to keep them from breaking out for some difficult target positions.

indra/newview/character/avatar_constraint.llsd
```diff
@@ -11,30 +11,30 @@
...
                 [r0,r1,r0],
             'max_bend':r0,
-            'max_twist':r135,
+            'max_twist':r0.5,
             'min_bend':r-157.5,
-            'min_twist':r-45,
+            'min_twist':r-0.5,
             'pivot_axis':
                 [r0,r0,r1],
             'type':'ELBOW'
```
AndrewMeadows, line 40: The SL avatar mesh is not weighted correctly to the forearm bone and the result is poor mesh animation when the forearm is twisted.  This is why I'm tightening the twist limits of the elbow joints to a very narrow window.  This means ll "twist" of the forearm needs to happen at the wrist... which also has poor mesh blend results but it looks better than when it happens at the forearm.

indra/newview/character/avatar_constraint.llsd
```diff
@@ -43,12 +43,12 @@
...
-            'min_twist':r-45,
+            'max_bend':r157.5,
+            'max_twist':r0.5,
+            'min_bend':r0.0,
+            'min_twist':r-0.5,
             'pivot_axis':
-                [r-0,r-0,r-1],
+                [r0,r0,r1],
             'type':'ELBOW'
         },
```
AndrewMeadows, line 53: I'm flipping the direction of the right elbow's pivot axis, and swizzling the bend limits accordingly.  This doesn't actually change the result, but it makes more sense to me when I try to visualize the constraint's geometry.

indra/newview/llpuppetmotion.cpp
```diff
@@ -212,11 +202,14 @@ LLMotion::LLMotionInitStatus LLPuppetMotion::onInitialize(LLCharacter *character
...
-        //// effectively preventing the spine from moving.
+        //// TEMPORARY HACK: whitelist of sub-roots.
+        //// This HACK forces joints to form the base of their chain
+        //// effectively preventing them from being moved during the IK solution.
+        //// The plan is to disable this when we have a more robust full-avatar IK solution.
         //ids.clear();
         //ids.insert(CHEST_ID);
+        //ids.insert(COLLAR_LEFT_ID);
+        //ids.insert(COLLAR_RIGHT_ID);
         //mIKSolver.setSubRootIds(ids);
```
AndrewMeadows, line 213: While testing changes and trying new things I briefly tried whitelisting the collar bones out of the IK chains.  In the end I decided to keep them enabled, but I'm leaving this experimental code around in case it is needed later.

indra/llcharacter/llik.cpp
```diff
@@ -3092,55 +3089,92 @@ void LLIK::Solver::executeFabrikPass()
...
+    for (chain_map_t::const_reverse_iterator itr = mChainMap.rbegin(); itr != mChainMap.rend(); ++itr)
     {
-        const joint_list_t& chain = data_pair.second;
+        S16 joint_id = enforceConstraintsInward(itr->second);
+        if (joint_id > -1)
+        {
+            // as we proceed we remember the smallest joint_id that was updated on that chain
+            adjusted_joint_ids.push_back(joint_id);
+        }
+    }
```
AndrewMeadows, line 3115: It is inefficient to recompute world-frame transforms that haven't changed, so we spend some CPU cycles above to locate only the joints that have changed (above) and then recompute their world-frame transforms in correct order (below).

indra/llcharacter/llik.cpp
```diff
@@ -3314,7 +3335,8 @@ void LLIKConstraintFactory::initSingleton()
     {
         std::ifstream file(constraint_file);
 
-        if (!LLSDSerialize::deserialize(constraint_data, file, LLSDSerialize::SIZE_UNLIMITED))
+        constexpr llssize MAX_EXPECTED_LINE_LENGTH = 256;
+        if (!LLSDSerialize::deserialize(constraint_data, file, MAX_EXPECTED_LINE_LENGTH))
```
AndrewMeadows, line 3339: Yeah so... `LLSDSerialize::SIZE_UNLIMITED` doesn't actually work.  It wasn't clear to me if it was failing because of how we were actually using it when we tried to parse each line of the file, or if it should never EVER be used, but the result was that the constraints config file would fail to load anything.  Consequentially the IK was happening on unconstrained chains and this, combined with some other bugs, resulted in the horrifying spinning IK arms.  I'm introducing this `MAX_EXPECTED_LINE_LENGTH` value as a quick way to solve the problem without falling down a rabbit hole in the **LLSD** parsing code.

PR #211, opened 2023-05-11
sl-19676 - more loading stats and 360 Interest List mode work

simon-linden:
Some refinement with 360 mode Interest List and viewer stats:

- 360 modes is applied to all regions the viewer can see
- if the agent is in 360 interest list mode, it is applied to all new regions they see or visit
- removed obsolete 360CaptureUseInterestListCap Debug setting - all regions support it now.
- shuffled code to use the IL cap into LLViewerRegion
- object kill messages are now counted in the update stats
- object update failures no longer track message size
- mObjectCacheMissResponses from stats removed, it was useless

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -16964,17 +16964,6 @@
       <key>Value</key>
       <integer>0</integer>
     </map>
-  <key>360CaptureUseInterestListCap</key>
-  <map>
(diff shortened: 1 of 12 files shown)


callumprentice (approved):
Looks good.  Only suggestion I have is that we don't call it 360 mode since there will be a variety of other uses - "Promiscuous Mode" perhaps?  "Deluge Mode"? :)

PR #217, opened 2023-05-17
SL-19575 Create emoji gallery (fix bug with drawing emojis in chat history)

LLGuru:
The core change is a new function `LLEmojiDictionary::isEmoji()` and using it in `LLTextBase::insertStringNoUndo()`


indra/llcommon/llstring.cpp
```diff
@@ -545,29 +545,29 @@ std::string utf8str_truncate(const std::string& utf8str, const S32 max_len)
 
 std::string utf8str_symbol_truncate(const std::string& utf8str, const S32 symbol_len)
 {
-    if (0 == symbol_len)
```
akleshchev, line 548: Our coding standard prefers spaces over tabs and specifically warns not to do unnesesary convertions.
https://wiki.secondlife.com/wiki/Coding_standard

indra/newview/llfloateremojipicker.cpp
```diff
@@ -139,7 +155,7 @@ void LLFloaterEmojiPicker::fillEmojis()
 	mEmojis->clearRows();
 
 	const LLEmojiDictionary::emoji2descr_map_t& emoji2Descr = LLEmojiDictionary::instance().getEmoji2Descr();
-	for (const LLEmojiDictionary::emoji2descr_item_t& it : emoji2Descr)
```
akleshchev, line 142: Why is it not a reference?

indra/llui/lltextbase.cpp
```diff
@@ -2349,10 +2353,9 @@ void LLTextBase::appendLineBreakSegment(const LLStyle::Params& style_params)
 
 void LLTextBase::appendImageSegment(const LLStyle::Params& style_params)
 {
-	if(getPlainText())
```
akleshchev, line 2352: Outr coding standard requests "Use braces for clarity, even when not strictly required", removing braces here goes against that.

indra/newview/llfloateremojipicker.cpp
LLGuru: Because using a reference causes a compilation error on macOS

indra/newview/llchathistory.cpp
```diff
@@ -1227,26 +1225,17 @@ void LLChatHistory::appendMessage(const LLChat& chat, const LLSD &args, const LL
 		mUnreadChatSources.insert(chat.mFromName);
 		mMoreChatPanel->setVisible(TRUE);
 		std::string chatters;
-		for (unread_chat_source_t::iterator it = mUnreadChatSources.begin();
-			it != mUnreadChatSources.end();)
+		for (std::string source : mUnreadChatSources)
```
akleshchev, line 1228: mac Os should have an issue with this one being a referenece (it has no issues doing the same for mNameList), but if does, I suggest reverting to 'iterator' variant.

akleshchev (approved):

PR #220, opened 2023-05-17
Merge branch 'DRTVWR-566-ais' into DRTVWR-567

marchcat-pe:


akleshchev:
1. This handling should already be present in D580
2. It isn't AIS3, but AIS2 failure handling
I think this isn't needed.

marchcat-pe:
So there already is a better implementation. Awesome, thank you!

PR #222, opened 2023-05-19
SL-19635 Implement DeepL tranlation support

akleshchev:
Target is D580 since it's where azure changes are (commit depends onto those and has some light azure changes). Should be safe and we should have enough time for a test, but can cherry pick azure to some other branch and apply DeepL there.


marchcat-pe (approved):
D580 should be good for this.
Thank you!

maxim-productengine (approved):

PR #225, opened 2023-05-23
SL-19744: Remove LLErrorThread and LLAppViewer::handleViewerCrash()

nat-goodspeed:
LLErrorThread is, ironically, not thread-safe. Recently it has been a source of shutdown hangs.

LLErrorThread's sole responsibility is to call handleViewerCrash() when LLApp::isError(). The surprising part is that I can find no code that calls either LLApp::setError() or setStatus(APP_STATUS_ERROR). In other words, I believe that no one ever calls LLAppViewer::handleViewerCrash().

Remove all that unused code. LLErrorThread is supposed to be part of the general LLApp machinery, but to actually be robust, it would have to be rewritten.

I'd like to request a QA pass over this branch to verify that it's no worse -- either while running or providing a crash report -- than the release viewer.

Unless some problem turns up, please pull this into an upcoming maint viewer. I marked it "Draft" because of course it shouldn't be merged directly into main.


marchcat-pe (approved):

marchcat-pe:
Will pull this into DRTVWR-580 Maint S.
Thank you, @nat-goodspeed !

PR #226, opened 2023-05-23
Fix mashup of CXX_FLAGS for RelWithDebInfo and Release configurations for DRTVWR-559

brad-linden:


brad-linden:
fixed mac build.  not sure why `add_compile_definitions()` doesn't seem to be working properly for the Xcode generator for this

PR #229, opened 2023-05-28
INTL-490 Japanese Translation Update

marchcat-pe:
provided by Logue Takacs


Ansariel:
No way this should be pulled in this shape and form - unless you want all TPVs to simply discard the changes for almost 500 files because of the pointless reformatting and whitespace changes which will cause a million merge conflicts!

marchcat-pe:
@Ansariel, thank you for bringing this problem up. The updated version should look better.

marchcat-pe:
The change is still huge though.

Ansariel:
Briefly skimming over the updated version, it seems to be much better and bearable.

logue:
I cut the branch and remade the Japanese translation.
Added auto-tuned translations added in 580, and fixed remaining translation inconsistencies.
https://github.com/logue/viewer/tree/DRTVWR-580-maint-T

Example: Translation of 'world' and 'map' probrem.

Words such as world map (世界地図) and minimap(ミニマップ) were shaken after translation.
So, I unified the map (地図) translations into ’マップ', and concomitantly changed world ('世界') to 'ワールド'.

marchcat-pe:
Thank you @logue!
I'll integrate these changes too.

logue:
When should I do additional translations?
It seems that the XML coding rules were different.

PR #230, opened 2023-05-31
SL-18251 Expand extents to include the Control Avatar bounds

LLGuru:


indra/newview/llvovolume.cpp
```diff
@@ -5110,6 +5110,58 @@ LLControlAVBridge::LLControlAVBridge(LLDrawable* drawablep, LLViewerRegion* regi
...
+		LLVector3 controlAvatarMax(controlAvatar->mDrawable->getSpatialExtents()[1].getF32ptr());
+
+		// Make coordinates relative to the root position
+		controlAvatarMin -= mDrawable->getPosition();
+		controlAvatarMax -= mDrawable->getPosition();
+
+		// Rotate coordinates backward to the root rotation
+		controlAvatarMin = controlAvatarMin * ~mDrawable->getRotation();
+		controlAvatarMax = controlAvatarMax * ~mDrawable->getRotation();
+
```
RunitaiLinden, line 5141: When rotating an AABB, you can't rely on just the min/max corners to contain the original AABB in the new coordinate frame (for example, rotate an axis aligned cube min/max 45 degrees on Z and the min/max points will both be on the XY plane -- your new box will have zero volume).  You have to construct an OBB in the new coordinate frame and generate an AABB from that.

RunitaiLinden (approved):

PR #232, opened 2023-06-05
Fixes for compatibility with new xcode 15 beta

brad-linden:
for any viewer branch after DRTVWR-577

### DIFF
--- a/indra/cmake/00-Common.cmake
+++ b/indra/cmake/00-Common.cmake
@@ -171,6 +171,10 @@ if (DARWIN)
 ## Really?? On developer machines too?
 ##set(ENABLE_SIGNING TRUE)
 ##set(SIGNING_IDENTITY "Developer ID Application: Linden Research, Inc.")
+
+  # required for clang-15/xcode-15 since our boost package still uses deprecated std::unary_function/binary_function
+  # see https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#C++-Standard-Library
+  add_compile_definitions(_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION)
 endif (DARWIN)
 
 if (LINUX OR DARWIN)
--- a/indra/newview/llviewerthrottle.cpp
+++ b/indra/newview/llviewerthrottle.cpp
@@ -84,7 +84,7 @@ LLViewerThrottleGroup::LLViewerThrottleGroup()
 }
 
 
-LLViewerThrottleGroup::LLViewerThrottleGroup(const F32 settings[])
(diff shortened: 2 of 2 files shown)


indra/cmake/00-Common.cmake
```diff
@@ -171,6 +171,10 @@ if (DARWIN)
 ## Really?? On developer machines too?
 ##set(ENABLE_SIGNING TRUE)
 ##set(SIGNING_IDENTITY "Developer ID Application: Linden Research, Inc.")
+
+  # required for clang-15/xcode-15 since our boost package still uses deprecated std::unary_function/binary_function
+  # see https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#C++-Standard-Library
+  add_compile_definitions(_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION)
```
nat-goodspeed, line 177: For the GitHub Actions viewer we've updated Boost to 1.81. But I take your point that this fix will help viewer branches _before_ we merge the `actions` branch. I guess we should put this in and then, if my hunch is correct that a newer Boost doesn't need it, remove it again in the `actions` branch.

indra/newview/llviewerthrottle.cpp
```diff
@@ -84,7 +84,7 @@ LLViewerThrottleGroup::LLViewerThrottleGroup()
 }
 
 
-LLViewerThrottleGroup::LLViewerThrottleGroup(const F32 settings[])
+LLViewerThrottleGroup::LLViewerThrottleGroup(const F32 settings[TC_EOF])
```
nat-goodspeed, line 87: Does C++ no longer accept an arbitrary-length array argument?

nat-goodspeed (approved):

indra/cmake/00-Common.cmake
brad-linden: yeah, good point.  I can try against the actions branch too.  I'll be honest at the time I wasn't really thinking about what the right solution was, I just wanted to get past it to see how many c++ incompatibilities were lurking.  I suspect it won't matter too much until this fall when xcode 15 gets out of beta. hopefully actions branch will be out the door by then

indra/newview/llviewerthrottle.cpp
brad-linden: c++ supports it, but in the header we've declared this method as taking a fixed length array, and it is being more strict in checking that the declaration and definition types match.

indra/cmake/00-Common.cmake
brad-linden: confirmed that this is unneeded with the boost version that is in the actions branch

PR #234, opened 2023-06-06
SL-19808 - Fix texture transforms changing topography of PBR normal textures

cosmic-linden:


indra/newview/app_settings/shaders/class1/deferred/textureUtilV.glsl
```diff
@@ -76,3 +76,47 @@ vec2 texture_transform(vec2 vertex_texcoord, vec4[2] khr_gltf_transform, mat4 sl
...
+
+// Take the rotation only from both transforms and apply to the tangent. This
+// accounts for the change of the topology of the normal texture when a texture
+// rotation is applied to it.
+// *HACK: Assume the imported GLTF model did not have both normal texture
+// transforms and tangent vertices. The use of this function is inconsistent
+// with the GLTF sample viewer when that is the case. See getNormalInfo in
+// https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Viewer/47a191931461a6f2e14de48d6da0f0eb6ec2d147/source/Renderer/shaders/material_info.glsl
+// We may want to account for this case during GLTF model import.
+// -Cosmic,2023-06-06
```
brad-linden, line 88: what does the GLTF spec say for this combination?  is it possible that this is a bug in the GLTF sample viewer?

brad-linden (approved):
can't say I fully understand the tangent transformation math, but looks good enough to me

indra/newview/app_settings/shaders/class1/deferred/textureUtilV.glsl
cosmic-linden: For SL texture animations, that's our own internal thing and we can take our own creative liberties there.

For texture transforms, the KHR_texture_transform spec doesn't say directly, but I think the most relevant bit in the GLTF spec is that tangents should be calculated using MikkTSpace, which we're already doing.

The GLTF sample viewer is probably the de-facto standard here.

PR #235, opened 2023-06-07
SL-19286 Avatar is upside down when viewed from below

LLGuru:
Avatar is upside down when viewed from below after left-clicking the mouse on the avatar and dragging down

### DIFF
--- a/indra/newview/llagent.cpp
+++ b/indra/newview/llagent.cpp
@@ -1371,26 +1371,21 @@ LLVector3 LLAgent::getReferenceUpVector()
 void LLAgent::pitch(F32 angle)
 {
 	// don't let user pitch if pointed almost all the way down or up
-	mFrameAgent.pitch(clampPitchToLimits(angle));
-}
 
-
-// Radians, positive is forward into ground
-//-----------------------------------------------------------------------------
-// clampPitchToLimits()
-//-----------------------------------------------------------------------------
-F32 LLAgent::clampPitchToLimits(F32 angle)
-{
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

indra/newview/llagent.cpp
```diff
@@ -1371,26 +1371,17 @@ LLVector3 LLAgent::getReferenceUpVector()
...
-{
 	// A dot B = mag(A) * mag(B) * cos(angle between A and B)
 	// so... cos(angle between A and B) = A dot B / mag(A) / mag(B)
 	//                                  = A dot B for unit vectors
 
 	LLVector3 skyward = getReferenceUpVector();
 
 	const F32 look_down_limit = 179.f * DEG_TO_RAD;;
-	const F32 look_up_limit   =   1.f * DEG_TO_RAD;
+	const F32 look_up_limit   =  10.f * DEG_TO_RAD;
```
akleshchev, line 1382: So it's going to reserve 10 degrees intead of 1? Please leave a comment in code that explains why this value was choosen.

LLGuru: Do you see any comment about 179 degrees? :-)
This value was chosen empirically, the issue is observed on lower values

akleshchev: I mean a comment like "10 degrees dead zone was choosen to avoid issues with vamera flipping", lets not follow bad practices.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/69593

PR #236, opened 2023-06-07
Attempt to port some build.yaml improvements from DRTVWR-559 over to actions branch

brad-linden:


.github/workflows/build.yaml
```diff
@@ -35,9 +40,18 @@ jobs:
...
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
```
brad-linden, line 54: I think this is what we want for making sure the python install is correctly setup.

> Using the setup-python action is the recommended way of using Python with GitHub Actions because it ensures consistent behavior across different runners and different versions of Python... For more information, see the [setup-python action](https://github.com/marketplace/actions/setup-python).

https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#specifying-a-python-version

callumlinden (approved):

PR #240, opened 2023-06-09
SL-18977 Borders of inventory view in inventory floater seem to be cut a bit

LLGuru:


marchcat-pe (approved):

LLGuru:
image

indra/llui/llscrollcontainer.cpp
```diff
@@ -126,6 +126,7 @@ LLScrollContainer::LLScrollContainer(const LLScrollContainer::Params& p)
 	
 	LLRect horizontal_scroll_rect = mInnerRect;
 	horizontal_scroll_rect.mTop = horizontal_scroll_rect.mBottom + scrollbar_size;
+	horizontal_scroll_rect.mLeft += 3;
```
akleshchev, line 129: '3' in this case is a 'magic number', please use a const variable, like HORIZONTAL_SCROLL_LEFT_PAD.

akleshchev (approved):

indra/llui/llscrollcontainer.cpp
akleshchev: On screen it looks like horizontal scroll overlaps vertical scroll now, are you shure you don' need to reduce with by the same 3 pixels?

PR #242, opened 2023-06-10
SL-18837 set build id based on github workflow run id

brad-linden:


nat-goodspeed (approved):
I like it. But I guess LLVersionInfo needs to be updated also.

Empirically, `$GIT_REF` is (or at least contains) the branch, e.g. "Second Life Test actions". If for a tag build it expands to the tag name, that would let us use a tag naming convention to set VIEWER_CHANNEL according to our existing practice.

indra/newview/llappviewer.cpp
```diff
@@ -3552,7 +3552,7 @@ void LLAppViewer::writeSystemInfo()
 	gDebugInfo["ClientInfo"]["MajorVersion"] = LLVersionInfo::instance().getMajor();
 	gDebugInfo["ClientInfo"]["MinorVersion"] = LLVersionInfo::instance().getMinor();
 	gDebugInfo["ClientInfo"]["PatchVersion"] = LLVersionInfo::instance().getPatch();
-	gDebugInfo["ClientInfo"]["BuildVersion"] = LLVersionInfo::instance().getBuild();
+	gDebugInfo["ClientInfo"]["BuildVersion"] = STRINGIZE(LLVersionInfo::instance().getBuild());
```
nat-goodspeed, line 3555: Please use lowercase `stringize()` instead of the `STRINGIZE()` macro for new usage.

`STRINGIZE()` predates (type-safe) support for variadic functions. It accepts a partial `operator<<()` expression, which works but is eyebrow-raising for future maintainers. Besides, it upsets Dr. Stroustrup.

`stringize()` is a proper variadic function.

For a single argument like this, you could alternatively use `std::to_string()`.

indra/llcommon/llprocessor.cpp
```diff
@@ -746,7 +746,7 @@ class LLProcessorInfoDarwinImpl : public LLProcessorInfoImpl
 		__cpuid(0x1, eax, ebx, ecx, edx);
 		if(feature_infos[0] != (S32)edx)
 		{
-			LL_ERRS() << "machdep.cpu.feature_bits doesn't match expected cpuid result!" << LL_ENDL;
+			LL_WARNS() << "machdep.cpu.feature_bits doesn't match expected cpuid result!" << LL_ENDL;
```
nat-goodspeed, line 749: I admit I'm curious about the circumstances under which this fails, and why it's in this particular PR?

nat-goodspeed:
This Windows build failed in a way reminiscent of the problem that caused me all the flailing with `py.exe`: `Access is denied` errors trying to run `C:/hostedtoolcache/windows/Python/3.11.4/x64/python.exe`. But you cut to the chase somehow, undoing the `py.exe` changes and getting clean builds anyway. What did you do then, and why isn't it working now?

.github/workflows/build.yaml
```diff
@@ -92,8 +93,8 @@ jobs:
             echo "Python location: $PYTHON"
           fi
 
-          autobuild configure -- -DVIEWER_CHANNEL="Second Life Test ${GIT_REF##*/}"
-          autobuild  build --no-configure
+          autobuild configure --id "${{ github.run_id }}" -- -DVIEWER_CHANNEL="Second Life Test ${GIT_REF##*/}"
```
nat-goodspeed, line 96: Hmm, I'm no longer sure `github.run_id` is the right variable. Consider https://github.com/secondlife/viewer/actions/runs/5205525429. What I think we want is version 6.6.13.5205525429 so we could just paste that last component into the .../viewer/actions/runs/ URL. Instead the viewer versions are 231482258 (64-bit viewers) or 231582302 (32-bit viewer).

indra/newview/llappviewer.cpp
brad-linden: good point, thanks

indra/llcommon/llprocessor.cpp
brad-linden: I've been running with this workaround locally for a while now when building on Apple Silicon and running unit tests under rosetta2.  I hoped to get it into the actions branch in advance of github making Apple Silicon M1 runner hosts available shortly so we can use them.  It is unrelated to the other changes here, but will be needed once we have M1 based build runners.

However, I don't know what the right fix is here.  I can't yet find the docs for apple's machdep.cpu.feature_bits contents, but it is unsurprising to me that this will have a new unexpected value when running under rosetta2.

brad-linden:
> This Windows build failed in a way reminiscent of the problem that caused me all the flailing with `py.exe`: `Access is denied` errors trying to run `C:/hostedtoolcache/windows/Python/3.11.4/x64/python.exe`. But you cut to the chase somehow, undoing the `py.exe` changes and getting clean builds anyway. What did you do then, and why isn't it working now?

yes.  previously the fix came from using the setup_python action as documented by github's docs.  I'm concerned that that fix is no longer working...

.github/workflows/build.yaml
brad-linden: Yeah you're right that's not what I expected from reading the docs.  I'll keep investigating this

brad-linden: correction.  github.run_id is properly matching with that URL component as I read it.  for example: https://github.com/secondlife/viewer/actions/runs/5384282553 

even though those builds are currently failing due to the other issues described in this PR, they do properly set the build to be 6.6.13.5384282553 for all 3 builds.

brad-linden:
I have removed all the prior commits trying to work around python build failures.  They are resolved after the latest merge with the main [actions](https://github.com/secondlife/viewer/tree/actions) branch

indra/llcommon/llrand.cpp
```diff
@@ -82,15 +82,15 @@ static LLSeedRand sRandomSeeder;
...
 #endif
 }
 inline F32 ll_internal_random_float()
 {
 #if LL_WINDOWS
-	return (F32)rand() / (F32)RAND_MAX; 
+	return (F32)rand() / (F32)(RAND_MAX+1);
 #else
 	return (F32)drand48();
 #endif
```
nat-goodspeed, line 96: Couldn't the whole body of `ll_internal_random_float()` be:
```
inline F32 ll_internal_random_float()
{
    return F32(ll_internal_random_double());
}
```

.github/workflows/build.yaml
nat-goodspeed: I do see that in your most recent branch build, https://github.com/secondlife/viewer/actions/runs/5559609923 in fact produces viewer version 6.6.13.5559609923, which is what I was hoping for. Not sure what went wrong with the build I first cited above.

nat-goodspeed (approved):

.github/workflows/build.yaml
```diff
@@ -13,24 +13,21 @@ jobs:
...
-        addrsize: [64]
         python-version: ["3.11"]
         include:
-          - runner: windows-large
-            configuration: Release
-            addrsize: 32
-            python-version: "3.11"
           - runner: macos-12-xl
             developer_dir: "/Applications/Xcode_14.0.1.app/Contents/Developer"
+            python-version: "3.11"
```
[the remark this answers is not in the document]
brad-linden: This was a bit superstitious of me.  I had seen builds before where a matrix variable was not properly being inherited as I expected it in one of these include branches.  I'll read up on the docs to be sure, and verify that this is unneeded and remove it

.github/workflows/build.yaml
brad-linden: it looks like just a minor confusion.  double checking now, the change 3ffa57473d4aec6a32807873513f7d880ec4bc67 that it was built from preceded this branch by a couple commits

indra/llcommon/llrand.cpp
brad-linden: oh boy do I have some simplifications in mind for all this.  I should probably write a jira

PR #247, opened 2023-06-15
SL-18977 Borders of inventory view in inventory floater seem to be cut a bit

LLGuru:


marchcat-pe (changes requested):
Looks like the patch includes multiple unnecessary formatting changes.

PR #250, opened 2023-06-21
SL-19692 Z Offset is applied twice in Mouselook

LLGuru:


marchcat-pe (approved):

indra/newview/llagentcamera.cpp
```diff
@@ -1764,25 +1764,32 @@ LLVector3d LLAgentCamera::calcCameraPositionTargetGlobal(BOOL *hit_limit)
...
+
 		head_offset.clearVec();
+		F32 fixup;
+		if (gAgentAvatarp->hasPelvisFixup(fixup))
+		{
+			head_offset[VZ] -= fixup;
+		}
+		if (gAgentAvatarp->isSitting())
+		{
+			head_offset.mdV[VZ] += 0.1;
```
akleshchev, line 1776: This is confusing, why does it need to be adjusted and why by 0.1 specifically?

LLGuru: This is an old expression
I only removed duplicated magic numbers

PR #256, opened 2023-06-25
Fixing a copy-paste oversight in lang_zh.nsi

PanteraPolnocy:


PanteraPolnocy:
Yes, it's only that.

akleshchev:
Ok, thank you, since it appears to already be fixed in both DRTVWR-582 and DRTVWR-559 I'm closing this.

PR #160, opened 2023-04-10
Cherry-pick basic build into contribute

bennettgoble:
Cherry-pick #144 commits into `contribute`.

PR #259, opened 2023-06-28
Merge ReleaseOS build into Maint U

nat-goodspeed:
PR #160 was merged to contribute-frozen just after the previous merge to Maint U (April 10). Signal wants this PR to get out to the community so that pushes to their own forks will build ReleaseOS to surface any unintentional build breakage.


bennettgoble:
Excellent, thanks, @nat-goodspeed. @brad-linden did some work in 559 to add working macOS support as well in #179

.github/workflows/build.yaml
```diff
@@ -0,0 +1,93 @@
+name: Build
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches: [main, contribute]
```
bennettgoble, line 7: Can likely omit `contribute` as it was deleted/moved.

bennettgoble (approved):

indra/newview/installers/windows/installer_template.nsi
```diff
@@ -1,902 +1,902 @@
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
akleshchev, line 1: Was this commited by a mistake? I don't see any changes in this files sans mac's line endings to window's instalator file.

nat-goodspeed: You're right: when I checked "Hide whitespace," GitHub says there are only whitespace differences.

Does NSIS actually care any more? I realize our git-hooks coding policy wants Windows line endings on any file with a relative path including "/windows/", but does it matter to the program that reads that file?

akleshchev: Likely doesn't care, but keeping it this way can be usefull just to prevent 'line ending batteles' due to people opening the thing in different editors.

Reverted this specific file to unlock builds and to prevent merge conflicts (I think we did some changes in other branches to this file)

nat-goodspeed: Thank you.

PR #261, opened 2023-06-30
SL-19816 Home location on world map for invalid locations

LLGuru:
setHomePosRegion() is called below, from setStartPositionSuccess() - if the operation was successful

### DIFF
--- a/indra/newview/llagent.cpp
+++ b/indra/newview/llagent.cpp
@@ -2534,12 +2534,6 @@ void LLAgent::setStartPosition( U32 location_id )
     if (!requestPostCapability("HomeLocation", body, 
             boost::bind(&LLAgent::setStartPositionSuccess, this, _1)))
         LL_WARNS() << "Unable to post to HomeLocation capability." << LL_ENDL;
-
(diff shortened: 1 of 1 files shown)


indra/newview/llagent.cpp
```diff
@@ -2534,12 +2534,6 @@ void LLAgent::setStartPosition( U32 location_id )
     if (!requestPostCapability("HomeLocation", body, 
             boost::bind(&LLAgent::setStartPositionSuccess, this, _1)))
         LL_WARNS() << "Unable to post to HomeLocation capability." << LL_ENDL;
-
-    const U32 HOME_INDEX = 1;
-    if( HOME_INDEX == location_id )
-    {
-        setHomePosRegion( mRegionp->getHandle(), getPositionAgent() );
-    }
```
akleshchev, line 2542: I feel like this should have been moved to setStartPositionSuccess, does assigning a home location works without this?

akleshchev: Nevermind, I see that it's present there.

PR #265, opened 2023-07-06
SL-19951 Define a list of icon groups with their names and icons

LLGuru:


indra/newview/skins/default/xui/en/emoji_groups.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" ?>
```
akleshchev, line 1: I feel like this is going to cause localization issues. this differes significantly from our normal 'name+label' structure and likely won't be picked up.

LLGuru: In case of such issues the xml file and the kiading algorithm could be transformed with no affect to other UI code

Ansariel: This is totally not going to work for localization.

BTW: The checkmark for LLMenuItemCheckGL is still rendered as Emoji and looks ugly!

akleshchev (approved):

marchcat-pe (approved):

LLGuru:
image
image

indra/newview/skins/default/xui/en/emoji_groups.xml
LLGuru: > This is totally not going to work for localization.

There was an intermediate solution without localization files
Now it works for all languages except Turkish

PR #267, opened 2023-07-07
SL-19951 Collect used icons in a special group 'Recently used'

LLGuru:
Two special sections added: "Recently used" and "Frequently used"
image

### DIFF
--- a/indra/llcommon/llstring.h
+++ b/indra/llcommon/llstring.h
@@ -357,6 +357,7 @@ public:
 	static void	replaceNonstandardASCII( string_type& string, T replacement );
 	static void	replaceChar( string_type& string, T target, T replacement );
 	static void replaceString( string_type& string, string_type target, string_type replacement );
+	static string_type capitalize(const string_type& str);
 	static void capitalize(string_type& str);
 	
 	static BOOL	containsNonprintable(const string_type& string);
@@ -1594,6 +1595,15 @@ void LLStringUtilBase<T>::replaceTabsWithSpaces( string_type& str, size_type spa
 		}
 	}
 	str = out_str;
+}
+
+//static
+template<class T>
+std::basic_string<T> LLStringUtilBase<T>::capitalize(const string_type& str)
+{
+	string_type result(str);
+	capitalize(result);
+	return result;
 }
 
 //static
--- a/indra/newview/llfloateremojipicker.cpp
+++ b/indra/newview/llfloateremojipicker.cpp
@@ -27,6 +27,7 @@
 
 #include "llfloateremojipicker.h"
 
+#include "llappviewer.h"
 #include "llbutton.h"
 #include "llcombobox.h"
 #include "llemojidictionary.h"
@@ -37,11 +38,27 @@
 #include "llscrollingpanellist.h"
 #include "llscrolllistctrl.h"
 #include "llscrolllistitem.h"
+#include "llsdserialize.h"
 #include "lltextbox.h" 
 #include "llviewerchat.h" 
 
-size_t LLFloaterEmojiPicker::sSelectedGroupIndex;
-std::string LLFloaterEmojiPicker::sSearchPattern;
+namespace {
+// The following variables and constants are used for storing the floater state
+// between different lifecycles of the floater and different sissions of the viewer
+
+// Floater state related variables
+static U32 sSelectedGroupIndex = 0;
+static std::string sSearchPattern;
+static std::list<llwchar> sRecentlyUsed;
+static std::list<std::pair<llwchar, U32>> sFrequentlyUsed;
+
+// State file related values
+static std::string sStateFileName;
+static const std::string sKeySelectedGroupIndex("SelectedGroupIndex");
+static const std::string sKeySearchPattern("SearchPattern");
+static const std::string sKeyRecentlyUsed("RecentlyUsed");
+static const std::string sKeyFrequentlyUsed("FrequentlyUsed");
+}
 
 class LLEmojiScrollListItem : public LLScrollListItem
 {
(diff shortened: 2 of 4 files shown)


marchcat-pe (approved):

indra/llcommon/llstring.h
```diff
@@ -1596,6 +1597,15 @@ void LLStringUtilBase<T>::replaceTabsWithSpaces( string_type& str, size_type spa
 	str = out_str;
 }
 
+//static
+template<class T>
+std::basic_string<T> LLStringUtilBase<T>::capitalize(const string_type& str)
```
akleshchev, line 1602: we already have basic 'case' operatiions, why not use LLStringUtil::toUpper?

indra/newview/llfloateremojipicker.cpp
```diff
@@ -631,9 +674,186 @@ BOOL LLFloaterEmojiPicker::handleKeyHere(KEY key, MASK mask)
...
+        for (auto& it : sFrequentlyUsed)
+        {
+            it.second = std::max((U32)0, it.second - delta);
+        }
+    }
+}
+
+void LLFloaterEmojiPicker::saveState()
+{
+    if (sStateFileName.empty())
```
akleshchev, line 808: I suggest checking mSecondInstance. Generally we treat files from under second instance as 'read only' to prevent issues.

indra/llcommon/llstring.h
LLGuru: toUpper makes "abc def" => "ABC DEF"
capitalize makes "abc def" => "Abc Def"
we have a procedure capitalize modifying it's argument in-place
but sometimes it is more usable to have a function returning the result but not modifying the argument
I added a function capitalize which is just a wrapper around the procedure capitalize (a syntax sugar)

indra/newview/llfloateremojipicker.cpp
LLGuru: mSecondInstance relates to another opera

the floater is created each time it appears on the screen
but static data members preserve their values even after the floater is destroyed
that's why it is enough to load the saved state only once, on the 1st creating of the floater
the state is saved each time the floater disappears, but the state stays loaded in the memory
so no needs to load it again

akleshchev: >on the 1st creating of the floater the state is saved each time the floater disappears
Second instance does not save (or is not supposed to) library cache, asset or texture cache due to potential overlap with main instance. Probably shouldn't even save chrome cache. And I probably should do the same for keybindings...

Example: if both instances of viewer were logged into the same region that restarted/logged out both agents, both viewers will try to save to file simultaneously, which might cause a mess since you didn't add write protections.

I think you should either check for second instance here; instead of LL_PATH_USER_SETTINGS use LL_PATH_PER_SL_ACCOUNT; or lock files before writing.
I think first option is preferable, third one not practical (due to needing OS specific lock methods)

indra/llcommon/llstring.h
akleshchev: Please add a comment before the function to explain what it does.

indra/newview/llfloateremojipicker.cpp
```diff
@@ -37,11 +37,20 @@
 #include "llscrollingpanellist.h"
 #include "llscrolllistctrl.h"
 #include "llscrolllistitem.h"
+#include "llsdserialize.h"
 #include "lltextbox.h" 
 #include "llviewerchat.h" 
 
-size_t LLFloaterEmojiPicker::sSelectedGroupIndex;
-std::string LLFloaterEmojiPicker::sSearchPattern;
+static U32 sSelectedGroupIndex;
```
akleshchev, line 44: From coding policy:

> Static class member variables start with an 's' followed by an uppercase character

This isn't a memeber variable, I think it can be defined as a filescoped global variable and should be marked as 'g', gSelectedGroupIndex.

Suggestion: it has no reason to be filscoped, just move declaration into class's header and mark as private.

indra/newview/llfloateremojipicker.cpp
```diff
@@ -631,9 +674,186 @@ BOOL LLFloaterEmojiPicker::handleKeyHere(KEY key, MASK mask)
...
+
+    LLSD state;
+    LLSDSerialize::fromXML(state, file);
+    if (state.isUndefined())
+    {
+        LL_WARNS() << "Emoji floater state file is missing or ill-formed: " << sStateFileName << LL_ENDL;
+        return;
+    }
+
+    sSelectedGroupIndex = state[sKeySelectedGroupIndex].asInteger();
```
akleshchev, line 747: Looks like if file loading failes, value will be random? This one likely needs to be initialized with some default value.

indra/llcommon/llstring.h
LLGuru: As you can see, I didn't add any new functionality, the function capitalize() was introduced earlier with no comments
I just overloaded this function for the argument of type a constant reference to the string
image

Moreover, all functions in this class have no comments because their names are understandable intuitive
It would look strange if I add a comment to only one function in this class

indra/newview/llfloateremojipicker.cpp
LLGuru: IMHO no needs to declare these variables in the class definition
No any other class or module should know about these variables

Global variables are variables accessible from any file or module
These variables aren't global, they are niot accessible from outside
Thus the prefix "g" would look strange or confusive in their names

indra/newview/llfloateremojipicker.cpp
LLGuru: The prefix "s" is used for file-scoped variables as well
image

LLGuru: > Looks like if file loading failes, value will be random? This one likely needs to be initialized with some default value.

Usually static variables are being initialized by zero
Anyway, the note makes sense; fixed

Thanks

indra/newview/llfloateremojipicker.cpp
LLGuru: The prefix "s" is used for file-scoped variables as well
image

akleshchev: >The prefix "s" is used for file-scoped variables as well
Check our coding policy document (sadly doesn't cover namespaces)

personally I have no issues with it remaining an 's', since it is a static, just pointing out that it is not according to coding policy.

>No any other class or module should know about these variables
If they will get marked as private, none would

akleshchev (approved):

PR #272, opened 2023-07-12
SL-19831 Check marks in the menu are less visible

LLGuru:
image


LLGuru:
The problem code block was introduced by Kitty Barnett <develop@catznip.com> Nov 7, 2019
https://github.com/secondlife/viewer/commit/5a6ddb2ea666e895890d3cb690cce5101cf12652
"Fallback fonts can have first crack at adding an unknown character + set Twemoji as the viewer's fallback for all emoji blocks"

marchcat-pe (approved):

PR #279, opened 2023-07-13
SL-20009 - race condition - calling cards could be created without name

roxanneskelly:
This happens when a calling card is created before the name is in the name cache.


indra/newview/llfriendcard.cpp
```diff
@@ -534,20 +534,7 @@ void LLFriendCardsManager::syncFriendsFolder()
 	// Create own calling card if it was not found in Friends/All folder
 	if (!collector.isAgentCallingCardFound())
 	{
-		LLAvatarName av_name;
-		LLAvatarNameCache::get( gAgentID, &av_name );
```
roxanneskelly, line 538: Name cache may not have retrieved the name yet, which results in an empty av_name.

akleshchev (approved):

PR #280, opened 2023-07-13
Major improvements to LLLeap functionality

nat-goodspeed:
This PR greatly extends the functionality of LLEventAPI, LLDispatchListener and LLEventDispatcher. It also improves their error messages.

LLDispatchListener (and therefore LLEventAPI, and therefore LLLeap plugin requests) can now handle batched requests. This can be engaged either by an LLSD map (which makes calls in unspecified order and continues even if some produce errors) or an LLSD array (which makes calls in strict order and stops on first error).

LLDispatchListener (and therefore LLEventAPI) can now invoke non-void functions. A return type convertible to LLSD is converted and sent in a reply message.

All the event reply logic has moved to LLDispatchListener: direct calls to LLEventDispatcher no longer look for a "reply" key.

LLEventDispatcher can now dispatch to non-void functions, as above, directly returning LLSD.

A function registered as accepting an array can now be called via LLSD event map, if the map contains a new argskey key whose value is an appropriate LLSD array.

An LLEventDispatcher (subclass) instance can now be streamed to ostream to self-describe. This is used in error messages.

The LLEventDispatcher implementation has changed from boost::fusion to LL::apply(). This should support functions with more parameters, while greatly increasing the readability of the header file.

LLLeap now catches any exception thrown while honoring a child-process request. An LLLeap plugin can no longer crash the viewer with a bad request.

LLLeapListener::getAPI() and getAPIs() now enumerate registered LazyEventAPIs as well, even if they're not yet instantiated.

Add to the viewer's bag of tricks:

    LL::LazyEventAPI        // like LLEventAPI, but instantiated on demand
    // to pass a specific arg count to a variadic function
    LL::apply_n<COUNT>(function, vector)
    LL::apply_n<COUNT>(function, LLSD)
    LL::apply(function, LLSD)
    LL::invoke()            // like std::invoke()
    LL::bind_front()        // like std::bind_front()
    LL::function_arity<SIG> // supports boost::function and std::function
    LL::always_return<T>()  // e.g. to always get LLSD from an arbitrary signature
    LL::make_always_return()
    LL::get_ref()           // given either pointer or reference, return reference
    LL::get_ptr()           // given either pointer or reference, return pointer
    LLCaptureListener<T>    // LLStoreListener that bundles the target variable
    LLSDParam<LLSD> specialization with templated conversions to other types

In support of LazyEventAPI, LLEventPumps adds methods:

    [un]registerTypeFactory()
    [un]registerPumpFactory()

Add tests for apply(), LazyEventAPI and new LLEventDispatcher and LLDispatchListener functionality.


nat-goodspeed:
@AndrewMeadows , this is the retargeted/rebased/simplified PR to replace https://github.com/secondlife/viewer/pull/35 (and also supplants https://github.com/secondlife/viewer/pull/54). I've been trying to get these improvements into the viewer for months now!

marchcat-pe (approved):

akleshchev (approved):

auralinden (approved):
Looks good.  I pulled the changes into the puppetry branch and tested them against our current leap scripts, verified works as expected there.

PR #284, opened 2023-07-17
Aura/drtvwr 558 merge

auralinden:


indra/llcharacter/lljoint.cpp
```diff
@@ -1050,5 +1037,24 @@ void LLJoint::setSkinOffset( const LLVector3& offset )
...
+		if (joint->isAnimatable())
+		{
+			main_axis = joint->getPosition();
+			main_axis.normVec();
+			// only care about first animatable child
+			break;
+		}
+	}
+}
+
```
AndrewMeadows, line 1058: `clampRotation()` is bad: it doesn't do what it claims, and it is also not used anywhere.  This is probably cruft I had removed when looking over **lljoint.cpp**.  I don't remember doing it, but it is what I would do now.

auralinden: Re-removed

AndrewMeadows (commented):
It seems this already merged.  There appears to be a bit of work that should be restored: in particular Rider's conversion from raw to smart pointers in llmotioncontroller.cpp.

PR #288, opened 2023-07-18
Marchcat/d582 xcode15

marchcat-pe:


indra/cmake/00-Common.cmake
```diff
@@ -171,6 +171,10 @@ if (DARWIN)
 ## Really?? On developer machines too?
 ##set(ENABLE_SIGNING TRUE)
 ##set(SIGNING_IDENTITY "Developer ID Application: Linden Research, Inc.")
+
+  # required for clang-15/xcode-15 since our boost package still uses deprecated std::unary_function/binary_function
+  # see https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#C++-Standard-Library
+  add_compile_definitions(_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION)
```
nat-goodspeed, line 177: Hmm please help me remember to pull this when we merge in the actions branch: on that branch we've updated to Boost 1.81, which no longer requires the deprecated *nary_function classes.

PR #289, opened 2023-07-18
Picked the nicest syntax of conflicting merges.

auralinden:


indra/llcharacter/lljoint.cpp
```diff
@@ -67,10 +67,11 @@ void LLVector3OverrideMap::showJointVector3Overrides( std::ostringstream& os ) c
...
-	for (const map_type::value_type& pos_pair : m_map)
+	for (map_type::const_iterator it = m_map.begin();
+		 it != m_map.end(); ++it)
 	{
-		const LLVector3& pos = pos_pair.second;
-		os << " " << "[" << pos_pair.first <<": " << pos << "]" << ((pos_pair==(*max_it)) ? "*" : "");
+		const LLVector3& pos = it->second;
+		os << " " << "[" << it->first <<": " << pos << "]" << ((it==max_it) ? "*" : "");
 	}
 }
```
AndrewMeadows, line 76: Revert the above change.  The previous `for(item : container)` pattern is better than explicit iterators, I think.

This same feedback applies to all the other instances below, modulo specific feedback as necessary.

indra/llcharacter/lljoint.cpp
```diff
@@ -209,8 +210,10 @@ void LLJoint::touch(U32 flags)
 			child_flags |= POSITION_DIRTY;
 		}
 
-		for (LLJoint* joint : mChildren)
+		for (joints_t::iterator iter = mChildren.begin();
+			 iter != mChildren.end(); ++iter)
 		{
+			LLJoint* joint = *iter;
```
AndrewMeadows, line 216: Revert the above change.  `mChildren` really is a container of raw pointers and the old pattern is preferred.

indra/llcharacter/lljoint.cpp
```diff
@@ -261,6 +264,7 @@ LLJoint *LLJoint::findJoint( const std::string &name )
 	for (joints_t::const_iterator iter = mChildren.begin();
 		 iter != mChildren.end(); ++iter)
 	{
+		LLJoint* joint = *iter;
```
AndrewMeadows, line 267: Use `for ( item : container)` pattern here.

AndrewMeadows (changes requested):
No, these are not the changes I was looking for.  I believe the newer and more terse for(item : container) pattern is preferred over the iterator one.

Ansariel:
It would tremendously help simply reading the comments in https://github.com/secondlife/viewer/commit/8abf358d9a21c13a3739115353bb7282474a599c#comments or https://github.com/secondlife/viewer/commit/e9b1a58a9e45d724f05247b7993eaac7a52d0f07#comments where I basically told you exactly what needs to go where...

Ansariel:
Anyway: Had some spare minutes and fixed the merge errors for you: https://github.com/secondlife/viewer/pull/290

indra/llcharacter/lljoint.cpp
```diff
@@ -258,17 +258,16 @@ LLJoint *LLJoint::findJoint( const std::string &name )
 	if (name == getName())
 		return this;
 
-	for (joints_t::const_iterator iter = mChildren.begin();
-		 iter != mChildren.end(); ++iter)
+	for (LLJoint* joint : mChildren)
```
AndrewMeadows, line 261: Prefer the `for (item : contianer)` pattern.  It tends to be more readable and terse.  When `auto` is used for `item`'s **typename** it can even be container agnostic.

PR #295, opened 2023-07-19
Update DejaVu Fonts to 2.37, including LICENSE

larsnaesbye:
Updates the DejaVu Fonts, [adding many glyphs and correcting some errors](https://dejavu-fonts.github.io/NEWS.html). 2.37 seems to be the final version.

I made a brief visual tour of the UI and it looks 'normal', in the sense that no glitches were observed with the updated version of the fonts.


larsnaesbye:
recheck

marchcat-pe:
Check passed.

Thank you!

marchcat-pe (approved):

PR #311, opened 2023-07-27
SL-20053: Fix sometimes unable to unset a PBR material on a prim

cosmic-linden:
Caused by inventory item asset IDs being null. Seems suspicious. But working around it for now seems best.

### DIFF
--- a/indra/newview/lltexturectrl.cpp
+++ b/indra/newview/lltexturectrl.cpp
@@ -693,6 +693,11 @@ void LLFloaterTexturePicker::draw()
 
 const LLUUID& LLFloaterTexturePicker::findItemID(const LLUUID& asset_id, BOOL copyable_only, BOOL ignore_library)
 {
+	if (asset_id.isNull())
+	{
+		return LLUUID::null;
+	}
+
 	LLViewerInventoryCategory::cat_array_t cats;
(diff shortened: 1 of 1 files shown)


brad-linden (approved):
seems like a reasonable defensive check to have regardless

akleshchev:
>Caused by inventory item asset IDs being null. Seems suspicious.
Nothing suspicious there, some freshly created items, including gltf materials, don't have an asset set unless edited.

cosmic-linden:
@akleshchev Thanks for your input. Interesting to know about freshly-created items. That does validate that this is a reasonable fix.

To clarify, I walked through this null UUID issue in a debugger, and saw that items which were *not* freshly created had null UUIDs - in particular, an outfit item from the library (girl next door). That does seems potentially suspicious to me. But good to know about null UUID being the default.

akleshchev:
I wouldn't be surprised if 'girl next door' is the default in everything.

>and saw that items which were not freshly created had null UUIDs - in particular
I will correct my phrasing: null asset id is valid for some item types if content was not edited since creation. It's the acr of editing than saving that assign nuw non-null id.

cosmic-linden:
Yea, that makes sense.

PR #312, opened 2023-07-28
SL-20088 EmojiPicker - replace the image on the activation button with an emoji

LLGuru:
Current look:
image

Proposed look:
image


callumlinden (approved):
Good change :)

PR #317, opened 2023-08-03
SL-20078 Remove Google Translation UI (just fix error processing)

LLGuru:
Before this change (incorrect error message):
image

After this change (correct error message):
image

Using my own API key (no error message):
image

No needs to remove Google Tranclation API, it works well
All we need is correct API key and correct error processing


akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70418

PR #320, opened 2023-08-07
SL-19807 BugSplat Crash: LLDispatchSetEstateExperience::getIDs()

LLGuru:
Check the vector size to avoid of OOB error


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70169

PR #321, opened 2023-08-09
SL-20039 "Save to inventory" should only allow saving identical faces

akleshchev:
Unsure about "func.mIdenticalId && func.mIdenticalOverride;" part. If two faces have identical materials despite different ids, should we asume them identical for all purposes in UI as well? Should I be checking just render material instead of an override? Should viewer decide on case by case basis, like 'save' being enabled if visually identical and picker showing 'multiple' even if visually identical, as long as ids are different?

### DIFF
--- a/indra/newview/llpanelface.cpp
+++ b/indra/newview/llpanelface.cpp
@@ -1829,13 +1829,13 @@ void LLPanelFace::updateUIGLTF(LLViewerObject* objectp, bool& has_pbr_material, 
 
     const bool editable = objectp->permModify() && !objectp->isPermanentEnforced();
     bool has_pbr_capabilities = LLMaterialEditor::capabilitiesAvailable();
+    bool identical_pbr = true;
 
     // pbr material
     LLTextureCtrl* pbr_ctrl = findChild<LLTextureCtrl>("pbr_control");
     if (pbr_ctrl)
     {
         LLUUID pbr_id;
-        bool identical_pbr;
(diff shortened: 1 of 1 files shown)


maxim-productengine (approved):

indra/newview/llpanelface.cpp
```diff
@@ -5098,12 +5102,49 @@ void LLPanelFace::LLSelectedTE::getPbrMaterialId(LLUUID& id, bool& identical, bo
...
+                {
+                    mIdenticalOverride &= (mMaterialOverride == LLGLTFMaterial::sDefault);
+                }
+            }
+            else
+            {
+                mInitialized = true;
+                mPBRId = pbr_id;
+                LLGLTFMaterial* override = object->getTE(te_index)->getGLTFMaterialOverride();
+                if (override)
```
cosmic-linden, line 5129: I think `LLGLTFMaterial::sDefault` is just the default constructor at this point, but as a precaution, we may consider assigning `mMaterialOverride` to  `LLGLTFMaterial::sDefault` here when `override` is null.

cosmic-linden (approved):
LGTM otherwise. I don't see a need to check the render material

indra/newview/llpanelface.cpp
akleshchev: Ok, will do, thank you.

akleshchev: Actually it already is set to sDefault from contructor at string 5091 mMaterialOverride(LLGLTFMaterial::sDefault)

PR #323, opened 2023-08-11
SL-20024: Improve material permissions handling

cosmic-linden:


indra/newview/llmaterialeditor.cpp
```diff
@@ -1792,55 +1819,142 @@ void LLMaterialEditor::loadLive()
...
+    {
+        if (op == PERM_MODIFY && selected_object->isPermanentEnforced())
+        {
+            return false;
+        }
     }
 
-    LLPointer<LLGLTFMaterial> render_material;
-    LLPointer<LLLocalGLTFMaterial> local_material;
+    item_out = selected_object->getInventoryItemByAsset(func.mMaterialId);
```
akleshchev, line 1847: What if object's inventory isn't loaded at this point?

cosmic-linden: I think the buttons should be disabled in that case (see `inventory_pending` in `LLPanelFace::updateUIGLTF`. However, it wouldn't hurt to double-check here.

akleshchev (approved):

indra/llprimitive/llmaterial.h
```diff
@@ -115,8 +115,6 @@ class LLMaterial : public LLRefCount
     void        setDiffuseAlphaMode(U8 alpha_mode);
     U8          getAlphaMaskCutoff() const;
     void        setAlphaMaskCutoff(U8 cutoff);
-    LLUUID      getMaterialID() const;
-    void        setMaterialID(LLUUID const & material_id);
```
brad-linden, line 119: We will also want to ensure these changes get into server side llprimitive project as well to keep them from diverging too much further

brad-linden (approved):

PR #330, opened 2023-08-16
DRTVWR-588: Update to apr_suite 1.7.2

nat-goodspeed:


akleshchev (approved):

nat-goodspeed:
Unfortunately, since the newer apr_suite has acquired a dependency on zlib-ng, it seems we will also need to update boost, googlemock, colladadom, libxml2, minizip-ng, curl, openssl, freetype and libpng. We shouldn't need to rebuild them; we only need to bring over their (consistent) autobuild.xml entries from the actions branch. I'll do that, but it's not my foremost task.

akleshchev (approved):

nat-goodspeed:
After considerable tweaking, I eventually ran into link errors apparently due to compiler version mismatches between GitHub Actions runners and our TeamCity build agents. Abandoning this experiment.

PR #336, opened 2023-08-18
Eliminate some obviated Boost constructs, unused internal timer features

nat-goodspeed:
Boost.Bind used to expose argument placeholders `_1`, `_2` et al. in the global namespace, and we used them that way. That's changed, but instead of qualifying every such reference, we choose to replace them with lambdas.

Boost.Phoenix is another clever library that can cause compile problems with modern compilers. Replace with lambdas.

Whittling away use of APR is always good, with the eventual goal of completely eliminating APR.

The viewer contains a surprising number of mechanisms to request main-loop callbacks. In particular, some are for scheduling future work. Eliminate some convenience functions based on `LLEventTimer`: consider introducing a main-loop `WorkSchedule` instead.

llrand.cpp contained some #if'ed out code that evidently hasn't been used since June 2008. Its only actual effect was to obfuscate the actual code. Delete it.


nat-goodspeed:
This work is still pending, and I still intend to finish it.

larsnaesbye:
Getting rid of Boost will be really great for future development :-)

nat-goodspeed:
There's some very useful stuff in Boost that has not yet made it into the C++ standard, e.g. Boost.Stacktrace and Boost.Process. But I totally agree that when a given Boost feature has been supplanted by a `std` feature, we should use the `std` feature instead.

larsnaesbye:
At the moment I'm loooking at the macOS-specific code, and the Boost filesystem calls in lldir_mac.cpp seem to exist somewhat in C++ 17. Could be cool to get them  swapped too :-)

nat-goodspeed:
Yes, now that we have access to std::filesystem, we can and should cut over to that as we go along.

larsnaesbye:
Would you like me to make a PR for lldir_mac.cpp, [replacing Boost with std](https://www.cppstories.com/2019/05/boost-to-stdfs/) to this branch?

nat-goodspeed:
As you see, this branch needs some love before it could be shipped. I still intend to do that, but...

Meanwhile, I recommend you create a new branch with those changes. It will make progress faster.

AiraYumi:
I think lexical_cast, bind, and placeholders will be replaced with std. I think it's also possible to use a lambda.

nat-goodspeed:
I agree about replacing bind() with lambdas: in general I think they're more readable, and I've replaced a few along the way.

The one exception I've found is when the parameter list is long, or the parameter types are long, and the code only needs (say) two of six. In such cases I confess the implicit parameter list declaration provided by bind() is more succinct.

PR #342, opened 2023-08-23
SL-19299 Viewer crashes after change 'Pick a physics model:' dropdown

LLGuru:
The key changes:
1. File **llmodelpreview.cpp** - Zero pointer `mDefaultPhysicsShapeP` before deleting the correspondent object
2. File **llrefcount.cpp** - Initialize `const S32 gMaxRefCount = LL_REFCOUNT_FREE;`
3. File **llrefcount.h** - Add `assert` for `(mRef != LL_REFCOUNT_FREE) // object is deleted`
4. File **llpointer.h** - Move method `assign()` out of the `ifdef LL_LIBRARY_INCLUDE` block

### DIFF
--- a/indra/newview/llmodelpreview.cpp
+++ b/indra/newview/llmodelpreview.cpp
@@ -270,6 +270,7 @@ void LLModelPreview::rebuildUploadData()
 {
     assert_main_thread();
 
+    mDefaultPhysicsShapeP = NULL;
     mUploadData.clear();
     mTextureSet.clear();
 
--- a/indra/llcommon/llpointer.h
+++ b/indra/llcommon/llpointer.h
@@ -129,16 +129,6 @@ protected:
 	void ref();                             
 	void unref();
 #else
-
-	void assign(const LLPointer<Type>& ptr)
-	{
-		if( mPointer != ptr.mPointer )
-		{
-			unref(); 
-			mPointer = ptr.mPointer;
-			ref();
-		}
-	}
 	void ref()                             
 	{ 
 		if (mPointer)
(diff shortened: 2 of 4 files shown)


marchcat-pe (approved):
It's ok, but formatting changes feel a bit excessive

indra/newview/llmodelpreview.cpp
```diff
@@ -272,6 +272,10 @@ void LLModelPreview::rebuildUploadData()
 
     mUploadData.clear();
     mTextureSet.clear();
+    if (mDefaultPhysicsShapeP && mDefaultPhysicsShapeP->getNumRefs() == LL_REFCOUNT_FREE)
```
akleshchev, line 275: Why does it end in this state?

LLGuru: This was a root cause of the bug
The object was deleted but mDefaultPhysicsShapeP still had it's address and then it was reused
mDefaultPhysicsShapeP  is not a LLPointer and this is right - it shouldn't prevent deleting the object

LLGuru:
> formatting changes feel a bit excessive

Agree, but these giant trailing spaces and mix of spaces and tabs look very ugly
And I guess noone accepts just formatting commit without real reason
That's why I listed the key changes in the description - just to help with this

indra/newview/llmodelpreview.cpp
akleshchev: >mDefaultPhysicsShapeP is not a LLPointer and this is right - it shouldn't prevent deleting the object

It can be "if (mDefaultPhysicsShapeP->getNumRefs() == 1) mDefaultPhysicsShapeP = nullptr;" or in case of a weak pointer, check if it is alive. The problem is you are calling getNumRefs() from a dead object

LLGuru: > The problem is you are calling getNumRefs() from a dead object

Agree. We may null this pointer unconditionally. Fixed. Thanks!

vir-linden (commented):
Why do we want to remove validateRefCount()? It looks like both checks should be valid in both places it was being used.

akleshchev (approved):

RunitaiLinden (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/69608

PR #348, opened 2023-08-28
Fix CMake CMP0148 deprecation warning introduced with CMake 3.27.4

Ansariel:
As of CMake 3.27.4, it will issue a CMP0148 deprecation warning about the way Python is located.

### DIFF
--- a/indra/cmake/Python.cmake
+++ b/indra/cmake/Python.cmake
@@ -40,7 +40,7 @@ elseif (WINDOWS)
     ${regpaths}
     ${pymaybe}
     )
-    include(FindPythonInterp)
+  find_package(Python3 COMPONENTS Interpreter)
 else()
   find_program(python python3)
 


indra/cmake/Python.cmake
```diff
@@ -40,7 +40,7 @@ elseif (WINDOWS)
     ${regpaths}
     ${pymaybe}
     )
-    include(FindPythonInterp)
+  find_package(Python3 COMPONENTS Interpreter)
```
nat-goodspeed, line 43: Hmm! I'm sure this change addresses the deprecation warning, but now I find myself wondering: why did we have this `include(FindPythonInterp)` line at all?

[The CMake FindPythonInterp documentation](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html) says that `FindPythonInterp` sets `PYTHONINTERP_FOUND` and `PYTHON_EXECUTABLE` (along with `PYTHON_VERSION_*`, which we don't seem to reference).

Maybe all we need is to delete line 43 altogether and move lines 47-49 before line 52, turning 52 into `else` ?

akleshchev (approved):

indra/cmake/Python.cmake
akleshchev: This is a contribution and it works so I'm taking it as is, we can improve it later.

akleshchev:
Thank you for your contribution.

PR #350, opened 2023-08-28
SL-20209 EmojiPicker - Show 'Recent' panel instead of the floater on click the button

LLGuru:
image


indra/newview/llpanelemojicomplete.cpp
```diff
@@ -50,6 +51,7 @@ LLPanelEmojiComplete::Params::Params()
 LLPanelEmojiComplete::LLPanelEmojiComplete(const LLPanelEmojiComplete::Params& p)
 	: LLUICtrl(p)
 	, mAutoSize(p.autosize)
+	, mNoScroll(p.noscroll)
```
akleshchev, line 54: if noscroll is optional, it's better to have a default value here, then do an 'if (p.noscroll.isProvided()) mNoScroll = p.noscroll;'
Or noscroll itself can have a default value

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -76,7 +77,7 @@ LLFloaterIMSessionTab::LLFloaterIMSessionTab(const LLSD& session_id)
 	mInputPanels(NULL),
 	mChatLayoutPanelHeight(0)
 {
-    setAutoFocus(FALSE);
+	setAutoFocus(FALSE);
```
akleshchev, line 80: Please make a separate out of formating fixes, it makes commit hard to review. Also we really shouldn't be replacing tabs with spaces, it's against coding policy.

Ansariel:
I don't really know if I should like this change. Yes, you can have the 20 most recently used emojis at hand, but:
- that makes the chat input area really clunky
- if you don't want the recent emojis take up chat window space and you have them collapsed, you now need to do 2 + 1 extra clicks to add an emoji: 1) toggle the panel 2) click on "More" and 3) - after selecting an emoji from the picker - toggle the panel again

There should be a way to combine both the recent emojis panel and toggling the picker directly without cluttering the input panel with 2 full-sized buttons.

indra/newview/llpanelemojicomplete.cpp
LLGuru: It works right similar to autosize
The default value is default for the value type (false for bool)
We should specify the default value for the parameter when it differs from the default value of the type
No needs to specify it explicitly here

Ansariel:
> This idea comes from Steeltoe Linden. If you have any concerns please feel free to ask him directly. Thanks :-)

Well... Of course it does. So I'll have to make it actually usable for our users... :joy:

akleshchev (approved):

akleshchev (approved):

PR #356, opened 2023-08-30
SL-20212 EmojiPicker - Turning the text field grey when it loses focus

LLGuru:
image


akleshchev (approved):

indra/newview/skins/default/xui/en/floater_im_session.xml
```diff
@@ -291,6 +291,7 @@
                          height="20"    
                          is_expandable="true"
                          text_tentative_color="TextFgTentativeColor"
+                         bg_writeable_color="ScriptBackground"
```
akleshchev, line 294: This will work for everything, but shouldn't it be not turning grey only for emoji flaoter?

PR #359, opened 2023-08-30
SL-20226 Object Profile Floater Stays As Inactive Transparency When Active

LLGuru:


akleshchev (approved):
Fix looks good in scope of sidepanel container, but DRTVWR-567 probably is a better place for this as it havily affect object profiles.

indra/newview/llfloatersidepanelcontainer.cpp
```diff
@@ -99,6 +99,10 @@ LLPanel* LLFloaterSidePanelContainer::openChildPanel(const std::string& panel_na
 	{
 	openFloater();
 	}
+	else if (!hasFocus())
+	{
+		setFocus(TRUE);
```
akleshchev, line 104: This probably should do a bringtofront, not just setfocus

LLGuru: It works as expected with  this change

akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70573

PR #364, opened 2023-09-02
SL-20236 BugSplat Crash in LLVertexBuffer::validateRange(708)

LLGuru:
The method LLVector4a::normalize3fast() shouldn't be applied to a zero vector, the result is NaN


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70584

PR #367, opened 2023-09-04
Fix emoji category filter not working correctly

Ansariel:
Fix emoji category filter not working correctly. This stems from the category list using all lower-case names while the (localized) description uses mixed case. (You need to select a different language than English to notice the filter not working, e.g. German)


akleshchev (approved):

akleshchev:
Thank you for the fix!

PR #372, opened 2023-09-05
DRTVWR-587: Simplify Python.cmake by omitting find_package(Python3).

nat-goodspeed:
Elsewhere in CMake land, we reference PYTHONINTERP_FOUND and PYTHON_EXECUTABLE, both of which are explicitly set by Python.cmake. We don't seem to need the find_package(Python3 COMPONENTS Interpreter) call. Given that we take some pains to be careful about which Windows Python interpreter we find, this eliminates a wildcard.

### DIFF
--- a/indra/cmake/Python.cmake
+++ b/indra/cmake/Python.cmake
@@ -40,18 +40,15 @@ elseif (WINDOWS)
     ${regpaths}
     ${pymaybe}
     )
-  find_package(Python3 COMPONENTS Interpreter)
 else()
   find_program(python python3)
-
-  if (python)
-    set(PYTHONINTERP_FOUND ON)
-  endif (python)
 endif (DEFINED ENV{PYTHON})
 
-if (NOT python)
+if (python)
+  set(PYTHONINTERP_FOUND ON)
+else()
   message(FATAL_ERROR "No Python interpreter found")
-endif (NOT python)
+endif (python)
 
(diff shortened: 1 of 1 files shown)


marchcat-pe (approved):

akleshchev (approved):

Ansariel:
Simplify the whole file:
```
# -*- cmake -*-

set(PYTHONINTERP_FOUND)

if (DEFINED ENV{PYTHON})
  # Allow python executable to be explicitly set
  set(python "$ENV{PYTHON}")
  set(PYTHONINTERP_FOUND ON)
elseif (WINDOWS)
  if (DEFINED ENV{VIRTUAL_ENV})
    set(Python3_FIND_VIRTUALENV "ONLY")
  endif()
  find_package(Python3 COMPONENTS Interpreter)
  set(python ${Python3_EXECUTABLE})
else()
  find_program(python python3)

  if (python)
    set(PYTHONINTERP_FOUND ON)
  endif (python)
endif (DEFINED ENV{PYTHON})

if (NOT python)
  message(FATAL_ERROR "No Python interpreter found")
endif (NOT python)

set(PYTHON_EXECUTABLE "${python}" CACHE FILEPATH "Python interpreter for builds")
mark_as_advanced(PYTHON_EXECUTABLE)
```

marchcat-pe (approved):

nat-goodspeed:
I simplified per Ansariel's suggestion (thanks Ansariel!). Please look?

akleshchev (approved):

PR #381, opened 2023-09-20
SL-20287 The 'search_history.txt' file in the user profile is always empty

LLGuru:
This file in not used anymore (as well as LLSearchHistory class)
Deleting it instead of saving the empty history solves the issue


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70636

PR #383, opened 2023-09-21
SL-15039 Prune unused preferences

akleshchev:
Task: Based on generals statistics prune settings (sadly only covers permanent settings) that have no modification counts, use constants where applicable or clean it up. 

This covers settings that start from A-D (mostly) and further changes will be posted separately since it's already a mess and the more I add the harder it will be to review.


LLGuru (approved):

marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/65021

PR #386, opened 2023-09-22
SL-20188 Size of friend list affects time to open Preferences floater

akleshchev:
Don't check every file when it's enough to know if there are any.


maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70534

PR #387, opened 2023-09-25
SL-15039 Prune unused preferences #4

akleshchev:
This feels like a waste of time due to little to no gain and potential conflicts down the line so this is probably the final change. Might clean up UI related settings later, like various paddings, those often overcomplicate things and should either be in related xmls or hardcoded.


akleshchev:
@ChorazinAllen Can you specify which ones? I can restore those. But most of the settings are unused entirely.

maxim-productengine (approved):

marchcat-pe (approved):

marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/65021

PR #392, opened 2023-09-26
SL-20292 Recent emojis panel: reduce spaces between controls

LLGuru:


akleshchev (approved):

Ansariel:
Would also be nice if the category filter in the picker work **again** for other languages.

marchcat-pe (approved):

akleshchev:
@Ansariel , please report that as a bug in jira so that our QA will be able to track and test it.

PR #398, opened 2023-09-28
SL-20361 Avoid of ASSERT (false) in check_rigged_group()

LLGuru:
This assert is often being raised on developer's builds when menu "Develop | Render Metadata | Octree" is checked
Usually this happens when animated objects or control avatars get into the field of view, and this leads to process exit

### DIFF
--- a/indra/newview/llspatialpartition.cpp
+++ b/indra/newview/llspatialpartition.cpp
@@ -1616,6 +1616,7 @@ void pushVertsColorCoded(LLSpatialGroup* group, U32 mask)
 //  - a linked rigged drawable face has the wrong draw order index
 bool check_rigged_group(LLDrawable* drawable)
 {
+#if 0
     if (drawable->isState(LLDrawable::RIGGED))
     {
         LLSpatialGroup* group = drawable->getSpatialGroup();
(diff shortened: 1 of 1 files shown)


indra/newview/llspatialpartition.cpp
```diff
@@ -1715,13 +1708,11 @@ void renderOctree(LLSpatialGroup* group)
 			for (LLSpatialGroup::element_iter i = group->getDataBegin(); i != group->getDataEnd(); ++i)
 			{
 				LLDrawable* drawable = (LLDrawable*)(*i)->getDrawable();
-				if(!drawable || drawable->getNumFaces() == 0)
+				if (!drawable || !drawable->getNumFaces() || !check_rigged_group(drawable))
 				{
```
RunitaiLinden, line 1712: don't call check_rigged_group outside of llassert -- it's very slow, so we don't want to call it outside of dev builds

RunitaiLinden (changes requested):

RunitaiLinden (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70718

PR #399, opened 2023-09-28
SL-20244 On-screen animesh characters that start pelvis offset animations disappear when root goes off-screen

LLGuru:


indra/newview/llspatialpartition.h
```diff
@@ -205,6 +205,7 @@ class LLDrawInfo : public LLRefCount
 LL_ALIGN_PREFIX(64)
 class LLSpatialGroup : public LLOcclusionCullingGroup
 {
+	using super = LLOcclusionCullingGroup;
```
Ansariel, line 208: What is this kind of Java-style keyword thing?

LLGuru: This is a common way we widely use to fill the hole in C++ standard :-)
image

indra/newview/llspatialpartition.cpp
```diff
@@ -847,6 +847,43 @@ void LLSpatialGroup::handleChildAddition(const OctreeNode* parent, OctreeNode* c
 	assert_states_valid(this);
 }
 
+//virtual
+void LLSpatialGroup::rebound()
```
akleshchev, line 851: Are you sure this is the right way? I think our common solution for such cases is to keep parent visible if any of it's children are visible, may be that check needs to be fixed to account avatar's extents properly? How often does rebound happen?

indra/newview/llspatialpartition.cpp
```diff
@@ -847,6 +847,43 @@ void LLSpatialGroup::handleChildAddition(const OctreeNode* parent, OctreeNode* c
...
+            bridge->mDrawable &&
+            bridge->mDrawable->getVObj() &&
+            bridge->mDrawable->getVObj()->isRoot())
+        {
+            LLControlAvatar* controlAvatar = bridge->mDrawable->getVObj()->getControlAvatar();
+            if (controlAvatar &&
+                controlAvatar->mDrawable &&
+                controlAvatar->mControlAVBridge)
+            {
+                llassert(controlAvatar->mControlAVBridge->mOctree);
```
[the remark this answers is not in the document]
akleshchev: It probably can't be null, this part is/was crashy in movePartition() so it got extra safety checks (crashy likely due to invalid control avatar since we recently found an issue there)

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70593

PR #409, opened 2023-10-03
AIS/Thumbnails post merge fixes

akleshchev:


marchcat-pe (approved):
Thank you!

marchcat-pe:
Should we add this to W and X as well?

akleshchev:
>Should we add this to W and X as well?
Probably not, W has too high of a crashrate at the moment, it's unlikely to release first and X is only in planing.

marchcat-pe:
Ok, let's just pick it up from here then.

akleshchev:
Added 'apr crash' crashfix to the list since fix seems to be working well in maint-X and is it is our top crasher.

maxim-productengine (approved):

PR #421, opened 2023-10-06
SL-19311 Grey textures after teleport

akleshchev:
There can be no asset capability during teleport and it can take some time to arrive, texture shouldn't just fail to fetch if capability arrives after content.

I don't like how it works, but based on how WAIT_HTTP_RESOURCE2 just returns false without doing anything else, I assume this is the best avaliable way to wait without shutting down fetching and cache.


maxim-productengine (approved):

nat-goodspeed (commented):
I'm sorry, I'm not familiar with texture-fetching logic.

PR #425, opened 2023-10-07
BUG-234493 Add "Until key release" to the gesture "Wait"

FelixWolf:
**This PR is if the suggestion is accepted.**
Implements change from https://jira.secondlife.com/browse/BUG-234493

Code sanity checks implemented:
- [x] If triggered twice (Such as by holding down a key due to key repeat), don't trigger it again _if_ we are waiting for a key release.
- [x] If we waited for a key to be released, completely ignore any following key waits and fall back to either timed or animation delay. (If none, skip step entirely)
- [x] If triggered by chat command, completely ignore key wait and fall back to either timed delay or animation delay. (If none, skip step entirely)

I don't know of any other sanity checks that might need to be implemented, I think I got them all.

### DIFF
--- a/indra/llcharacter/llmultigesture.cpp
+++ b/indra/llcharacter/llmultigesture.cpp
@@ -68,6 +68,8 @@ void LLMultiGesture::reset()
 	mCurrentStep = 0;
 	mWaitTimer.reset();
 	mWaitingTimer = FALSE;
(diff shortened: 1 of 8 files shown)


marchcat-pe:
The change is integrated into DRTVWR-591 (Maint X).
Please see future release notes for it (may as well appear as SL-20442).

PR #434, opened 2023-10-11
SL-20378 Add menu bar entry and keyboard shortcut to open Notifications floater

LLGuru:


AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70736

PR #435, opened 2023-10-11
SL-20403 FIXED People floater showing arrows on default size

maxim-productengine:
The issue was caused by https://github.com/secondlife/viewer/commit/2a035c080dec2aa16b9628aac7b120e4f15135c2#diff-018dd621e811a33ee38efeabfc2e0fa246aca4a0bd7b169f84f021efac558865 change. So let's just use the new offset if it's needed.

### DIFF
--- a/indra/newview/skins/default/xui/en/panel_main_inventory.xml
+++ b/indra/newview/skins/default/xui/en/panel_main_inventory.xml
@@ -217,6 +217,7 @@
        tab_height="30"
        tab_position="top"
        tab_min_width="100"
+       use_tab_offset="true"
        top="0">
         <inventory_panel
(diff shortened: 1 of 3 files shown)


indra/newview/skins/default/xui/en/panel_main_inventory.xml
```diff
@@ -217,6 +217,7 @@
        tab_height="30"
        tab_position="top"
        tab_min_width="100"
+       use_tab_offset="true"
```
akleshchev, line 220: arguably tabs are the ones that should be fixed to not require an additional offset.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70765

PR #436, opened 2023-10-11
SL-20419 Receiving new objects or items, etc causes inventory to switch from recent to my inventory tab

LLGuru:


akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70782

PR #441, opened 2023-10-12
SL-20453 Fix dupplicate items in menu

akleshchev:
Cleanup. Caused by a merge of main(D567) into D559


Ansariel:
Not all of those appear to be duplicate - at least not in menu_inventory.xml

akleshchev:
QA reported seeing dupplicates of "Use as default for", "New Folder" and most creation options were removed in D567.

I checked what should be present, then just moved menu_inventory.xml from release into pbr and readded "PBR uploads".

Ansariel:
Ah yes. That weird UI design. Something for me to ignore when merging. :D

marchcat-pe (approved):

akleshchev:
>Ah yes. That weird UI design. Something for me to ignore when merging. :D
Looks like it will be added back in SL-20457

PR #443, opened 2023-10-12
SL-20286 Duplication of autocomplete results occurs when entering some search queries in the navigation bar

LLGuru:


akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70635

PR #446, opened 2023-10-13
SL-20464 Don't allow enetring non-unicode material name

maxim-productengine:
Don't allow entering non-unicode name same we do for outfits, settings etc.
Also show 'Save material dialog' again after telling the user to use non-empty name.

### DIFF
--- a/indra/newview/llmaterialeditor.cpp
+++ b/indra/newview/llmaterialeditor.cpp
@@ -1571,7 +1571,11 @@ void LLMaterialEditor::onSaveAsMsgCallback(const LLSD& notification, const LLSD&
         }
         else
         {
-            LLNotificationsUtil::add("InvalidMaterialName");
+            LLNotificationsUtil::add("InvalidMaterialName", LLSD(), LLSD(), [this](const LLSD& notification, const LLSD& response) 
+                {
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

indra/newview/llmaterialeditor.cpp
```diff
@@ -1571,7 +1571,11 @@ void LLMaterialEditor::onSaveAsMsgCallback(const LLSD& notification, const LLSD&
         }
         else
         {
-            LLNotificationsUtil::add("InvalidMaterialName");
+            LLNotificationsUtil::add("InvalidMaterialName", LLSD(), LLSD(), [this](const LLSD& notification, const LLSD& response) 
+                {
+                    LLNotificationsUtil::add("SaveMaterialAs", LLSD().with("DESC", mMaterialName), LLSD(),
+                        boost::bind(&LLMaterialEditor::onSaveAsMsgCallback, this, _1, _2));
+                });
```
akleshchev, line 1578: Isn't text box the one that is supposed to limit input?

maxim-productengine: Currently we can't do that for notifications. But yeah, it would be good to allow use this setting just in notifications.xml.

PR #450, opened 2023-10-16
DRAFT: Send game controller input to viewer for LSL access

AndrewMeadows:
This PR is a proposal for sending game controller input to the SL viewer using **Simple DirectMedia Layer** 2.0 [SDL2](https://www.libsdl.org/) for controller detection.  Corresponding server work to support this feature is complete and will be up on the test grid soon (second half of October 2023).

The new LSL event will have the following format:

`game_control ( key id, list buttons_edge_down, list buttons_edge_up, list buttons_down, list axes)`

The `id` will be that of the source agent.

The elements of the `buttons_*` lists will be  guaranteed to be integers and those of the `axes` list will be floats in the range **[-1, 1]**.

Feedback is welcome.

Added **DO NOT MERGE** to title because I'm pointing the diff at the **actions** branch which I expect to merge into **main** before this work does.


indra/llwindow/llkeyboardsdl.cpp
```diff
@@ -24,7 +24,7 @@
  * $/LicenseInfo$
  */
 
-#if LL_SDL
+#if LL_LINUX
```
AndrewMeadows, line 27: We used to use SDL for the linux client, however support for linux was allowed to bitrot sometime after 2013.  Now that we're using SDL on the other platforms too this llkeyboardsdl.cpp code should probably be renamed **llkeyboardlinux.cpp**.

indra/newview/app_settings/settings.xml
```diff
@@ -3739,6 +3739,28 @@
...
+    <map>
+      <key>Comment</key>
+      <string>Send 'unhandled' keystrokes as GameInput to server</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>Boolean</string>
+      <key>Value</key>
+      <integer>0</integer>
+    </map>
```
AndrewMeadows, line 3763: The initial UI for controlling this feature is very rudimentary to start.  Two checkboxes: one to enable the stream of data, and another to consider the keyboard to be a game controller with lots of buttons.

indra/llwindow/llkeyboard.cpp
```diff
@@ -254,6 +257,10 @@ BOOL LLKeyboard::handleTranslatedKeyDown(KEY translated_key, U32 translated_mask
 	mKeyDown[translated_key] = TRUE;
 	mCurTranslatedKey = (KEY)translated_key;
 	handled = mCallbacks->handleTranslatedKeyDown(translated_key, translated_mask, repeated);
+    if (!handled && LLGameControl::getIncludeKeyboardButtons())
+    {
+        LLGameControl::onButton(translated_key, true);
+    }
```
AndrewMeadows, line 263: Indentation is bad here, but there are plans to eventually convert all of the viewer code over to four-spaces instead of tabs.  Maybe that work will get done before this is shipped, so I'll leave this alone for now.

The more interesting aspect of these specific changes: we treat all "unhandled" keystrokes as game controller input (if enabled).  This will allow LSL scripts on attachments and seats to have access to keyboard control no permissions necessary (e.g. leave the feature disabled if you don't want to broadcast).

nat-goodspeed:
Hey Levi, based on your big DO NOT MERGE, I marked this PR as "draft." This means it can't be merged until marked "ready." Hope that's okay.

indra/llwindow/llkeyboard.cpp
```diff
@@ -254,6 +259,11 @@ BOOL LLKeyboard::handleTranslatedKeyDown(KEY translated_key, U32 translated_mask
 	mKeyDown[translated_key] = TRUE;
 	mCurTranslatedKey = (KEY)translated_key;
 	handled = mCallbacks->handleTranslatedKeyDown(translated_key, translated_mask, repeated);
+    // TODO: capture unhandled keys and translate any mapped buttons to game controller
```
akleshchev, line 262: Unless controller keys are identical to keybord keys or should go throught mTranslateKeyMap, this is a wrong spot for this.

indra/newview/llappviewer.cpp
```diff
@@ -1375,6 +1378,51 @@ bool LLAppViewer::frame()
...
+            msg->addU8Fast(_PREHASH_Index, i);
+            msg->addS16Fast(_PREHASH_Value, state.mAxes[i]);
+        }
+    }
+
+    const std::vector<U8>& buttons = state.mPressedButtons;
+    if (!buttons.empty())
+    {
+        msg->nextBlockFast(_PREHASH_ButtonData);
+        msg->addBinaryDataFast(_PREHASH_Data, (void*)(buttons.data()), (S32)(buttons.size()));
```
akleshchev, line 1418: I think this is a bit too inflecible.
1. Not all functions are server sided, should be possible to map buttons to viewer functions, not just sent to server and let sript sort it out.
2. What if user has some kind of flight controller with minimal amount of buttons? Normally keyboard picks the slack and sends extra functions.
3. What about users that don't have controllers at all? 3p viewers will start abusing and emulating this to get access to these functions and get extended control from scripts.

I think joystick should be it's own thing and we need an extention of llTakeControls (mControlsTakenCount) that doesn't care about joysticks specifically, just allows more inputs with things like axises included.
So that instead of doing strictly 'joystick->server'. it whould be something like 'any input ->map to desired value->server'

indra/newview/llappviewer.cpp
```diff
@@ -1375,6 +1378,51 @@ bool LLAppViewer::frame()
...
+    }
+
+    LLGameControl::clearInput();
```
akleshchev, line 1421: so if I understand it corretly it woll be something like leyboard | joystick -> LLGameControl -> server?

What if user wants some local function on the joystick, like toggling on voice?

indra/llwindow/llkeyboard.cpp
```diff
@@ -272,6 +282,11 @@ BOOL LLKeyboard::handleTranslatedKeyUp(KEY translated_key, U32 translated_mask)
 		// sequence W<return> in chat to move agents forward. JC
 		mKeyUp[translated_key] = TRUE;
 		handled = mCallbacks->handleTranslatedKeyUp(translated_key, translated_mask);
+        // TODO: capture unhandled keys and translate any mapped buttons to game controller
```
akleshchev, line 285: if you want to translate keyboard buttons (our mouse buttons) to game control buttons according to user's settings, it should be LLViewerInput.

AndrewMeadows: Thanks for the input.  Yes, we will add the ability to map keys to game controller input and when we get it working it will be moved to a more correct code context.  This is a work in progress (WIP).  We might add mouse mappings although I'm less certain how that would work right now so am punting on it.

indra/newview/llappviewer.cpp
AndrewMeadows: That is a good question.  I will think about it more when we get closer to something that is shippable, however I don't think mapping game controller input to viewer-only events should be included in a "minimal viable feature" requirement.  If most game controllers are detected, can be remapped on a per-device basis, and the data can reach the server... then we'll probably ship it in that working state and then add bells and whistles in separate PRs.

indra/llwindow/llkeyboard.cpp
```diff
@@ -273,12 +276,49 @@ BOOL LLKeyboard::handleTranslatedKeyUp(KEY translated_key, U32 translated_mask)
...
+	{
+		handled = handleTranslatedKeyUp(translated_key, translated_mask);
+	}
+    if (!handled)
+    {
+        LLGameControl::onKeyUp(translated_key, translated_mask);
+    }
+
+	return handled;
+}
```
AndrewMeadows, line 320: These have been moved from `LLKeyboard`--derived classes into the base class.   After a little cleanup the logic pattern was identical acrossl platforms.

indra/llwindow/llkeyboardmacosx.cpp
```diff
@@ -304,14 +271,4 @@ void LLKeyboardMacOSX::scanKeyboard()
...
-BOOL LLKeyboardMacOSX::translateNumpadKey( const U16 os_key, KEY *translated_key )
-{
-	return translateKey(os_key, translated_key);
-}
-
-U16	LLKeyboardMacOSX::inverseTranslateNumpadKey(const KEY translated_key)
-{
-	return inverseTranslateKey(translated_key);
-}
-
```
AndrewMeadows, line 316: This was some very old cruft... `LLKeyboardMacOSX::translateNumpadKey()` was just a wrapper for `translateKeys()`.  Meanwhile, `LLKeyboardMacOSX::inverseTranslateNumpadKey()` was never called anywhere.  Similar cruftitude existed in other platform implementations.

indra/newview/llappviewer.cpp
akleshchev: >If most game controllers are detected, can be remapped on a per-device basis, and the data can reach the server...

That feel like doing things in wrong order. Viewer already has basic controller/joystick support, why not hook that into server messages and separately hook controllers into existing UI? Ability to hook axises and more butons to scripts will be very welcome by users even without having better joystick support. And having two separate joystick implementations is going to cause confusion.

indra/llwindow/llkeyboard.cpp
akleshchev: >We might add mouse mappings although I'm less certain how that would work right now so am punting on it.

At the moment viewer treats mouse keys the same way as keyboard. There is a bunch of trigerable functions, we map keyboard or mouse keys to those functions based on user settings (be those purely local like turning on voice, or server handled 'click' for LLTakeControl) to be trigered on press and release. Ideally any joystick specific keys should work the same way and just be hooked into LLViewerInput with LLViewerInput triggering assigned action.

PR #452, opened 2023-10-17
SL-20163 Rework LLAvatarPropertiesProcessor::sendRequest()

LLGuru:
Age could be hidden now:
image ---> image

Also the gap between comboboxes is closed:
image ---> image

### DIFF
--- a/indra/newview/llavatarpropertiesprocessor.cpp
+++ b/indra/newview/llavatarpropertiesprocessor.cpp
@@ -114,32 +114,32 @@ void LLAvatarPropertiesProcessor::sendRequest(const LLUUID& avatar_id, EAvatarPr
 		return;
 	}
 
-    std::string cap;
-
-    switch (type)
+    // Try to send HTTP request if cap_url is available
+    if (type == APT_PROPERTIES || type == APT_PICKS || type == APT_GROUPS || type == APT_NOTES)
     {
-    case APT_PROPERTIES:
-        // indicate we're going to make a request
+        std::string cap_url(gAgent.getRegionCapability("AgentProfile"));
+        if (!cap_url.empty())
+        {
+            initAgentProfileCapRequest(avatar_id, cap_url, type);
+            return;
+        }
+
+        // Don't sent UDP request for APT_PROPERTIES
+        if (type == APT_PROPERTIES)
+        {
+            LL_WARNS() << "No cap_url for APT_PROPERTIES, request is not sent" << LL_ENDL;
+            return;
+        }
+    }
+
+    // Send UDP request
+    if (type == APT_PROPERTIES_LEGACY)
+    {
         sendAvatarPropertiesRequestMessage(avatar_id);
-        // can use getRegionCapability("AgentProfile"), but it is heavy
-        // initAgentProfileCapRequest(avatar_id, cap);
-        break;
-    case APT_PICKS:
-    case APT_GROUPS:
-    case APT_NOTES:
-        if (cap.empty())
-        {
-            // indicate we're going to make a request
-            sendGenericRequest(avatar_id, type, method);
-        }
-        else
-        {
-            initAgentProfileCapRequest(avatar_id, cap);
-        }
-        break;
-    default:
+    }
+    else
+    {
         sendGenericRequest(avatar_id, type, method);
-        break;
     }
 }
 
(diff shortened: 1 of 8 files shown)


marchcat-pe (approved):

indra/newview/llavatarpropertiesprocessor.cpp
```diff
@@ -116,32 +114,23 @@ void LLAvatarPropertiesProcessor::sendRequest(const LLUUID& avatar_id, EAvatarPr
...
-    case APT_PICKS:
-    case APT_GROUPS:
-    case APT_NOTES:
-        if (cap.empty())
+        std::string cap_url(gAgent.getRegionCapability("AgentProfile"));
+        if (!cap_url.empty())
         {
-            // indicate we're going to make a request
-            sendGenericRequest(avatar_id, type, method);
+            initAgentProfileCapRequest(avatar_id, cap_url, type);
```
akleshchev, line 122: I don't think this is ready for processing bulk requests

akleshchev (changes requested):

indra/newview/llavatarpropertiesprocessor.cpp
```diff
@@ -302,114 +293,106 @@ void LLAvatarPropertiesProcessor::requestAvatarPropertiesCoro(std::string cap_ur
             << (!status ? " (no HTTP status)" : !result.has("id") ? " (no result.id)" :
                 std::string(" (result.id=") + result["id"].asUUID().asString() + ")")
             << LL_ENDL;
-        LLAvatarPropertiesProcessor* self = getInstance();
-        self->removePendingRequest(agent_id, APT_PROPERTIES);
-        self->removePendingRequest(agent_id, APT_PICKS);
-        self->removePendingRequest(agent_id, APT_GROUPS);
-        self->removePendingRequest(agent_id, APT_NOTES);
```
akleshchev, line 309: And what si going to 'free' viewer to rerequest if this request fails?

indra/newview/llavatarpropertiesprocessor.cpp
```diff
@@ -276,8 +262,13 @@ bool LLAvatarPropertiesProcessor::hasPaymentInfoOnFile(const LLAvatarData* avata
 }
 
 // static
-void LLAvatarPropertiesProcessor::requestAvatarPropertiesCoro(std::string cap_url, LLUUID agent_id)
+void LLAvatarPropertiesProcessor::requestAvatarPropertiesCoro(std::string cap_url, LLUUID agent_id, EAvatarProcessorType type)
 {
+    LLAvatarPropertiesProcessor* self = getInstance();
+
+    // Response received, no longer pending
+    self->removePendingRequest(agent_id, type);
```
akleshchev, line 270: This is wrong, you are going to free viewer to request again, potentially senconds before request arrives so viewer can rerequest same data multipkle times while it's getting to viewer, at minimumm this needs to be after getAndSuspend, when we know if request failed or not and that it actually happened.

indra/newview/llavatarpropertiesprocessor.cpp
```diff
@@ -302,114 +293,106 @@ void LLAvatarPropertiesProcessor::requestAvatarPropertiesCoro(std::string cap_ur
...
+        if (result.has("charter_member")) // won't be present if "caption" is set
+        {
+            avatar_data.caption_index = result["charter_member"].asInteger();
+        }
+        else if (result.has("caption"))
+        {
+            avatar_data.caption_text = result["caption"].asString();
+        }
+
+        self->notifyObservers(agent_id, &avatar_data, type);
```
akleshchev, line 345: This is a potential crash in the making, self might no longer exists

akleshchev: P.S. Arguably this also needs a 'gDisconected' check

indra/newview/llavatarpropertiesprocessor.cpp
```diff
@@ -167,19 +159,16 @@ void LLAvatarPropertiesProcessor::sendAvatarPropertiesRequestMessage(const LLUUI
...
 }
 
-void LLAvatarPropertiesProcessor::initAgentProfileCapRequest(const LLUUID& avatar_id, const std::string& cap_url)
+void LLAvatarPropertiesProcessor::initAgentProfileCapRequest(const LLUUID& avatar_id, const std::string& cap_url, EAvatarProcessorType type)
 {
-    addPendingRequest(avatar_id, APT_PROPERTIES);
-    addPendingRequest(avatar_id, APT_PICKS);
-    addPendingRequest(avatar_id, APT_GROUPS);
-    addPendingRequest(avatar_id, APT_NOTES);
+    addPendingRequest(avatar_id, type);
```
akleshchev, line 164: This is somewhat ineffective, if something sends an apt_properties request then user opens picks for example, it will trigger initAgentProfileCapRequest again (unlikely, but possible), viewer will be requesting and getting same data twice. Instead of should mark all the data as pending so that if user opens a mini profile then picks, both will get updated via same message instead of requiring two.

akleshchev: By this point most APT_* need to be merged into a single one with old system disabled where possible.

PR #460, opened 2023-10-19
SL-20497 FIXED Minimum distance is not updated correctly

maxim-productengine:


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70866

PR #461, opened 2023-10-19
SL-20504 Chat history crashes on shutdown

akleshchev:


AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70874

PR #462, opened 2023-10-19
SL-20163 Rework add/remove observers in LLAvatarPropertiesProcessor

LLGuru:


akleshchev (approved):

indra/newview/llavatarpropertiesprocessor.cpp
```diff
@@ -52,24 +52,23 @@ LLAvatarPropertiesProcessor::~LLAvatarPropertiesProcessor()
...
 {
+	if (!observer)
+		return;
+
 	// Check if that observer is already in mObservers for that avatar_id
-	observer_multimap_t::iterator it;
+	using pair = std::pair<LLUUID, LLAvatarPropertiesObserver*>;
+	observer_multimap_t::iterator begin = mObservers.begin();
+	observer_multimap_t::iterator end = mObservers.end();
+	observer_multimap_t::iterator it = std::find_if(begin, end, [&](const pair& p)
```
marchcat-pe, line 62: might be just `std::find_if(mObservers.begin(), mObservers.end(), [&](const pair& p)`

marchcat-pe (approved):

indra/newview/llavatarpropertiesprocessor.cpp
LLGuru: In such case mObservers.end() will be called twice

marchcat-pe: i mean, there is no need in temporary variables "begin" and "end", since they are used only once

PR #466, opened 2023-10-20
SL-13610 List HID available devices in joystick selection

akleshchev:
Purpose is to list more than one device and not to list things like touchbar.

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -1384,9 +1384,9 @@
             <key>archive</key>
             <map>
               <key>hash</key>
-              <string>a487fff84208a45844602c4a1f68c974</string>
+              <string>69c3e7c8cbf47d6f840011d1fa34cd8b</string>
               <key>url</key>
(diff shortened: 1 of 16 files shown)


AndrewMeadows:
I expect to get my hands on a Mac laptop soon and when I do I will checkout this branch and try it out.  If this PR doesn't get much attention from the main viewer team your could consider targeting the **leviathan/game-control** branch instead since that work would benefit from improved joystick detection on the Mac.  We'll be adding a new competing game controller detection system using **SDL2** which will be independent from **ndof**  (see PR-450): this to allow general game controller input to LSL.  Eventually we'll probably coalesce the non-space-mouse controllers to **SDL2** but that will have to be sorted out later.

akleshchev:
> I expect to get my hands on a Mac laptop soon and when I do I will checkout this branch and try it out.
Thank you, I think PE QA has needed mac compatible controllers as well. Since it works with any HID devices I canfirmed that part fully, the only parts I'm not confident about are filtering (and couldn't check activation), but conditions were copied from ndof so they should work and if not, will ask PE QA to turn on debug and get me data on needed flags.

maxim-productengine (approved):

PR #467, opened 2023-10-23
Fill agent_id fields in LLAvatarData correctly when using capability …

Ansariel:
…for avatar properties request

LLAvatarData.agent_id is historically set with the ID of the agent issuing the avatar properties request, which is gAgentID. Legacy code might also be checking if agent_id is the same as gAgentID in the response handler to verify the response is meant for us. Make sure in case of the new cap we set agent_id to the correct ID.


marchcat-pe (approved):

marchcat-pe:
Thank you, @Ansariel !

This fix  is was integrated as SL-20524 into DRTVWR-591 (Maint X).

PR #469, opened 2023-10-24
SL-20517: When attempting to find orphans, if an object claims to be …

Rider-Linden:
…its own parent the viewer logs a warning and then continues the loop without incrementing the iterator.  This increments the iterator so that loop processing can continue and the viewer does not get stuck on the bad object.


marchcat-pe (approved):

akleshchev (approved):

AndrewMeadows (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70887

PR #472, opened 2023-10-25
SL-20401 FIXED Popping out conversation floater second time moves location

maxim-productengine:


akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70763

PR #477, opened 2023-10-26
Andreyk/d587 build fix

akleshchev:
Test to see if this fixes the issue.


marchcat-pe (approved):
Let's see how it builds.

PR #478, opened 2023-10-26
SL-20536: Fix build error C1128 on Windows RelWithDebInfo

cosmic-linden:


indra/newview/CMakeLists.txt
```diff
@@ -1699,6 +1699,7 @@ if (WINDOWS)
         LINK_FLAGS_DEBUG "/NODEFAULTLIB:\"LIBCMT;LIBCMTD;MSVCRT\" /INCREMENTAL:NO /LARGEADDRESSAWARE"
         LINK_FLAGS_RELEASE "/FORCE:MULTIPLE /MAP\"secondlife-bin.MAP\" /OPT:REF /LARGEADDRESSAWARE"
         )
+    target_compile_options(${VIEWER_BINARY_NAME} PUBLIC /bigobj)
```
brad-linden, line 1702: I don't think it matters since we're not building a library target with downstream dependencies, but I would consider making this PRIVATE

brad-linden (approved):

indra/newview/CMakeLists.txt
cosmic-linden: Makes sense. I'll change that.

PR #480, opened 2023-10-27
SL-20496 FIXED Auto-adjust causes incorrect UI value of Max number of non-impostors

maxim-productengine:
"IndirectMaxNonImpostors" is used just for UI purpose both in Preferences and Advanced Preferences floaters.
"RenderAvatarMaxNonImpostors" should be adjusted instead, which will also trigger update of UI controls ("IndirectMaxNonImpostors").


marchcat-pe (approved):

akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70865

PR #482, opened 2023-10-30
Revert "DRTVWR-587: Simplify Python.cmake"

marchcat-pe:
Reverts secondlife/viewer#479


marchcat-pe:
This breaks GHA builds, so have to revert for now.

PR #488, opened 2023-11-02
SL-20553: Save material button in build floater now depends on agent inventory rather than object inventory

cosmic-linden:


indra/newview/llmaterialeditor.cpp
```diff
@@ -1838,19 +1880,45 @@ bool can_use_objects_material(LLSelectedTEGetMatData& func, const std::vector<Pe
         }
     }
 
-    item_out = selected_object->getInventoryItemByAsset(func.mMaterialId);
-
-    LLPermissions item_permissions;
-    if (item_out)
+    // Look for the item to base permissions off of
+    item_out = nullptr;
+    if (func.mMaterialId != LLGLTFMaterialList::BLANK_MATERIAL_ASSET_ID)
```
akleshchev, line 1885: if assigned id is BLANK_MATERIAL_ASSET_ID, means viewer needs to look for a material with null uuid.

akleshchev (approved):

PR #489, opened 2023-11-02
Fix for SL-19968 objects missing timing bug due to stall during login

brad-linden:
Ensure inventory skeleton loading doesn't block the message system from processing packets.

This addresses the stall during login in 2 ways: some minor optimizations in LLSD inventory skeleton loading to avoid a lot of unnecessary temporary copy allocations, and periodically calling back to `pump_idle_startup_network()` to ensure that UDP messages can still be processed as we are going.

### DIFF
--- a/indra/llinventory/llinventory.cpp
+++ b/indra/llinventory/llinventory.cpp
@@ -904,6 +904,8 @@ bool LLInventoryItem::fromLLSD(const LLSD& sd, bool is_new)
 		mInventoryType = LLInventoryType::IT_NONE;
 		mAssetUUID.setNull();
 	}
+
+#if 0  // old implementation.  makes a LOT of temporary copies and LLSD::safe(impl) calls
 	std::string w;
 
 	w = INV_ITEM_ID_LABEL;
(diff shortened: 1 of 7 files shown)


cosmic-linden (approved):

cosmic-linden (approved):

indra/llinventory/llinventory.cpp
```diff
@@ -1050,6 +1052,159 @@ bool LLInventoryItem::fromLLSD(const LLSD& sd, bool is_new)
...
+        if (i->first == INV_ASSET_TYPE_LABEL)
+        {
+            LLSD const &label = i->second;
+            if (label.isString())
+            {
+                mType = LLAssetType::lookup(label.asString().c_str());
+            }
+            else if (label.isInteger())
+            {
+                S8 type = (U8) label.asInteger();
```
Ansariel, line 1157: Why not cast to S8 directly?

indra/llinventory/llinventory.cpp
```diff
@@ -1050,6 +1052,159 @@ bool LLInventoryItem::fromLLSD(const LLSD& sd, bool is_new)
...
+        if (i->first == INV_INVENTORY_TYPE_LABEL)
+        {
+            LLSD const &label = i->second;
+            if (label.isString())
+            {
+                mInventoryType = LLInventoryType::lookup(label.asString().c_str());
+            }
+            else if (label.isInteger())
+            {
+                S8 type        = (U8) label.asInteger();
```
Ansariel, line 1171: Why not cast to S8 directly?

indra/llinventory/llinventory.cpp
brad-linden: Hmm.  unfortunately, I don't know why it was originally implemented this way.  I copied this code directly from the above block that is now in the `#if 0` and tried to disturb it as little as possible while restructuring the LLSD traversal code for this optimization.  I did not dig in to the business logic of these pieces and attempted to preserve them as-is.

PR #503, opened 2023-11-09
SL-20576 FIXED Viewer crashes when clicking scroll bar after switching views

maxim-productengine:


akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70946

PR #505, opened 2023-11-09
SL-20553: Save material button in build floater now depends on agent inventory rather than object inventory

cosmic-linden:


indra/newview/llmaterialeditor.cpp
```diff
@@ -1312,16 +1313,22 @@ class LLObjectsMaterialItemCallback : public LLInventoryCallback
...
-            updates["name"] = mNewName;
+            if (changed_name)
+            {
+                updates["name"] = mNewName;
+            }
+            if (changed_permissions)
+            {
+                updates["permissions"] = ll_create_sd_from_permissions(mPermissions);
+            }
             update_inventory_item(inv_item_id, updates, NULL);
```
akleshchev, line 1332: Is this still needed if we are renaming after upload as well?

akleshchev (approved):

indra/newview/llmaterialeditor.cpp
cosmic-linden: It's not strictly needed. I mainly included it because there is a delay until the asset is uploaded.

PR #511, opened 2023-11-14
SL-18343: Simple interim GLTF material preview - base color only

cosmic-linden:


indra/newview/llfetchedgltfmaterial.cpp
```diff
@@ -175,3 +177,55 @@ void LLFetchedGLTFMaterial::materialComplete()
...
+        return nullptr;
+    }
+
+    auto fetch_texture_for_ui = [](LLPointer<LLViewerFetchedTexture>& img, const LLUUID& id)
+    {
+        if (id.notNull())
+        {
+            if (LLAvatarAppearanceDefines::LLAvatarAppearanceDictionary::isBakedImageId(id))
+            {
+                LLViewerObject* obj = LLSelectMgr::getInstance()->getSelection()->getFirstObject();
```
akleshchev, line 194: This looks wrong, relying onto selection manager when we are previewing materials will bite us in the rear later.

Baked images are limited in number and are shared between all objects. Shouln't checking LLAvatarAppearanceDictionary::isBakedImageId then assetIdToBakedTextureIndex be enough? If it's a magic id simply assume that this is an object (especially since this particular material can be applied to multiple objects and parameters like 'animated' can change) and baked image should be displayed as baked, we do the same in texture picker.

RunitaiLinden (approved):

akleshchev:
>Copying over this review because it was wiped during the branch deletion:
No, I removed this comment because I noticed that texture picker did the same. Using LLSelectMgr is still likely to be causing issues somewhere without at least subscribing to changes in selection, but if picker does the same it can be fixed later together.

cosmic-linden:
👍

PR #516, opened 2023-11-15
SL-20575 When opening menu cache results and spread load over frames

akleshchev:
1. Instead of checking content recursively each time cache result
2. Spread load over frames

Made for D591 due to other commits (cheaper cof id and cheaper isItemRomovable) already being there.


marchcat-pe (approved):
👍 Excellent work!

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70945

PR #517, opened 2023-11-16
SL-20563 Add 'No Post' option to Snapshot floater

LLGuru:
The "No post-processing" check-box is added and synchronized with the "Build | Options | No Post" menu item
The "Build | Options | No Post" menu item label is changed to "No Post-processing"
image

### DIFF
--- a/indra/newview/llfloatersnapshot.cpp
+++ b/indra/newview/llfloatersnapshot.cpp
@@ -461,13 +461,24 @@ void LLFloaterSnapshotBase::ImplBase::onClickAutoSnap(LLUICtrl *ctrl, void* data
 {
 	LLCheckBoxCtrl *check = (LLCheckBoxCtrl *)ctrl;
 	gSavedSettings.setBOOL( "AutoSnapshot", check->get() );
-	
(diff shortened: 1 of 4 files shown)


marchcat-pe (approved):
👍

indra/newview/llfloatersnapshot.cpp
```diff
@@ -461,15 +461,26 @@ void LLFloaterSnapshotBase::ImplBase::onClickAutoSnap(LLUICtrl *ctrl, void* data
...
 		view->impl->checkAutoSnapshot(view->getPreviewView());
 		view->impl->updateControls(view);
 	}
 }
 
+// static
+void LLFloaterSnapshotBase::ImplBase::onClickNoPost(LLUICtrl *ctrl, void* data)
+{
+    BOOL no_post = ((LLCheckBoxCtrl*)ctrl)->get();
+    gSavedSettings.setBOOL("RenderDisablePostProcessing", no_post);
```
akleshchev, line 477: I think this option should disable itself when snapshot floater closes if set by snapshot floater (until it opens next time?).

PR #519, opened 2023-11-17
SL-20607 Version folder was not created when copying folders to marketplace

akleshchev:
'changed' items were not properly processed and the way to track that was unreliable (one item could have been added other could have been renamed, yet both would have decreased count) so I switched fully to callbacks.

It looks like there was no reason to create cb each time since we do not pass a destroy callback


marchcat-pe (approved):
🚢👍

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70980

PR #521, opened 2023-11-20
Only publish installers to releases

bennettgoble:
Cut down the amount of release artifacts to just installers.

## Details

We are currently publishing a four different kinds of release artifacts:

- Distribution archives (Installers)
- Debug symbols
- autobuild-package.xml(s)
- viewer_version.txt(s)

Before
*Before*

In addition to taking up a lot of storage space (which we pay for) this process goes against the general practice for releases: they should only contain files  general users care about. Debug symbols and other developer files are available as build artifacts should they be needed.

To resolve this, let's only download and publish installers on as part of public releases. This simplifies the release process, as eliminates the need to use the flatten_files utility from viewer-build-utils and it reduces the number of files to download during publication.

After
*After*

### Notes

- Reduces `release` job time from 3+ minutes to ~30 seconds
- Reduces release assets by ~60% (883 ➝ 366MB)
- Eliminates need to use separate release-artifacts action from [viewer-build-util](https://github.com/secondlife/viewer-build-util)


marchcat-pe (commented):
This should be great!
The release job used to take up to 20 minutes for some of my builds before.

nat-goodspeed:
Aside from the installers themselves, the only release assets of nontrivial size are the symbol files. We've published those as separate assets because from time to time we used to have to download them and hand-post them to BugSplat. Now that (re)posting to BugSplat is a separate job that can be rerun independently of the viewer build itself, I'm okay with omitting symbol files from the release page.

We post metadata on the GH release page because we're no longer using codeticket. The GH release page is currently our one source of information about a given viewer build. We could reinstate codeticket and trim down the release page; that would simplify some of the scripting. But if our goal is to eliminate codeticket, then this is the place.

.github/workflows/build.yaml
```diff
@@ -330,29 +330,29 @@ jobs:
     steps:
       - uses: actions/download-artifact@v3
         with:
-          path: artifacts
+          name: Windows-installer
```
nat-goodspeed, line 333: A useful enhancement for `actions/download-artifact` would be to accept a list of artifact names. Part of my rationale for using its "download everything" feature was to avoid requiring changes in two places if we wanted to change the set of posted assets: I didn't want to hard-code the set of assets in the release job.

If we could download a specific list to a specific subdirectory, then we could continue passing `action-gh-release` a `files:` argument specifying everything in that subdirectory.

I understand the speedup benefit, just explaining the rationale for the previous implementation.

bennettgoble: If you look at how actions like action-gh-release are used in the wild it is uncommon to hard-code input parameters in _another_ action. I would encourage us to KISS here. Rather than providing useful indirection, placing artifact specifics in a separate repository increases complexity and is a bit confusing.

bennettgoble: To provide more details: the build.yaml workflow steps define the artifacts that are being uploaded (**Windows-metadata**, et al.) so, imo, it's a bit odd to then hard-code expectations about the artifacts in another repository. I think we want to consolidate this information so that is 1) easy to understand and 2) easy to update.

bennettgoble:
@nat-goodspeed I've updated this PR so that autobuild-package and viewer_version files are included. An example release can be found here: https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%239b50c50

bennettgoble:
@nat-goodspeed @marchcat-pe if this PR looks good which branch should it be merged into?

marchcat-pe:
DRTVWR-587-maint-V, if we want to release it sooner

nat-goodspeed (approved):

PR #522, opened 2023-11-20
SL-20563 Add 'No Post' option to Snapshot floater

LLGuru:


akleshchev (changes requested):
1. Option should only affect snapshot floater, should not affect the world after floater closed (ideally should only affect screenshot without affecting full screen)
2. Should not require build floater to work
3. Quick snapshots (see in advanced->shortcuts) should correctly work with this option

akleshchev (approved):

marchcat-pe (approved):

akleshchev (approved):

PR #524, opened 2023-11-20
Fix failure to save the normalized translation data during collada upload

RyeMutt:
This fixes a case where translation would no longer be applied during rigged mesh upload resulting in broken mesh preview and the actual upload.

image


marchcat-pe:
Created SL-20627 to track testing.

My local build seems to work fine.
@RyeMutt, are there any specific models that should be used to test this fix?

marchcat-pe (approved):

RyeMutt:
Unfortunately the artist didn't share the model with me. 

This originally regressed in https://github.com/secondlife/viewer/commit/75de4d32768bb1359611dc7cd963c9a12f94423d#diff-91281943e063cbcc46b79504435a7eb2e501334b46f9f8b87f36adcad0fd11f1 from what appears to be a bit of overzealous cleanup

marchcat-pe:
Ok, thank you!
This should be sufficient.

PR #525, opened 2023-11-20
Ignore .vscode

bennettgoble:
Minor chore: add the frequently autocreated .vscode directory to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ web/locale.*
 web/secondlife.com.*
 
 .env
+.vscode


nat-goodspeed (approved):

bennettgoble:
Thanks for the review, @nat-goodspeed. Which branch should this be merged into?

nat-goodspeed:
Discussing with Vir; I think it will go directly to main. But merging to main implies "push notification" to active projects to refresh from main. Best case is that when we promote a viewer, merging its branch to main is a fast forward.

PR #526, opened 2023-11-20
SL-20564 reset camera when teleporting Home, so avatar points to the expected direction

maxim-productengine:


indra/newview/llagent.cpp
```diff
@@ -4344,6 +4344,10 @@ void LLAgent::teleportRequest(
 // Landmark ID = LLUUID::null means teleport home
 void LLAgent::teleportViaLandmark(const LLUUID& landmark_asset_id)
 {
+    if (landmark_asset_id == LLUUID())
```
akleshchev, line 4347: isNull() check is a better way.

akleshchev (approved):

indra/newview/llagent.cpp
akleshchev: Also can't user keep moving camera despite teleport going on?

PR #529, opened 2023-11-21
SL-20327 FIXED friend location on the map is not shown after right click>Map

maxim-productengine:


akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70681

PR #534, opened 2023-11-23
fix of XML Header / UTF-8 BOM / EOF / name attribute duplicate 

YuzuruJewell:
Suggestions for fixing problems in xui/en XML files
To help easily output multilingual files using the program, I suggest modifying the XML files of en folder.
I fixed the following four issues.
-The beginning of the XML file is not below
 <?xml version="1.0" encoding="utf-8" standalone="yes"?>
 I fixed it because without it the XML parser can't read it.
-UTF-8 BOM is at the beginning of the file
 I unified them without BOM, because it is mixed with those without BOM and with BOM, 
-The end of the XML file is not LF
 It was unified because it was mixed with LF and without LF.
-There are nodes with the same name attribute within the same node.
 Because it could not be identified the node in Japanese translation, if I found the duplicated name attoribute, I add the character of numbers starting from 2 to the value of the name attribute.
For example, if [name="xxx"] is duplicated, the second [name ="xxx"] is modified to ["name="xxx2"] and the third [name ="xxx"] is modified to [name="xxx3"]


YuzuruJewell:

YuzuruJewell:
recheck

marchcat-pe:
@YuzuruJewell, thank you very much for the contribution and congratulations on your first PR!
Your efforts are greatly appreciated.

marchcat-pe:
Created SL-20640 for further tracking.

marchcat-pe:
recheck

marchcat-pe:
CLA Assistant bot issue should be resolved with the next commit.
Squash/merging this one.

akleshchev:
Related issue:
https://github.com/secondlife/jira-archive-internal/issues/71017

PR #536, opened 2023-11-23
SL-20442 Add an 'Until shortcut key released' option to gestures so we can do properly user-mappable keys

LLGuru:


indra/newview/llgesturemgr.cpp
```diff
@@ -56,6 +56,8 @@
 #include "llappearancemgr.h"
 #include "llgesturelistener.h"
 
+#pragma optimize("", off)
```
akleshchev, line 59: Why is this here?

akleshchev (changes requested):
Please either remove 'pragma optimize' or reduce it to scope of the code that needs it along with the explanation of why.

indra/newview/llgesturemgr.cpp
marchcat-pe: We should avoid using compiler-specific code, especially in such general areas as this.

marchcat-pe (approved):

indra/newview/llgesturemgr.cpp
LLGuru: Removed (there was a temporary hint for debug purposes)

akleshchev (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70809

PR #537, opened 2023-11-24
SL-17076 MacOS crash on clearing ChicletNotificationChannel

akleshchev:
Crash appears to happens inside mItems.clear() of the LLNotificationChannelBase, but there is no apparent reson for it and stack jumped some steps, so I'm doing cleanup more explicitly to see if it's indeed there and not a corruption somewhere earlier.

Viewer runs window in a separate thread, it is possible notification was added in a way that corrupted the list, thus adding thread safety as well.


marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/67213

PR #550, opened 2023-11-30
Fix LLGLTFMaterial hashing

sldevel:
This PR fixes the non-working material hashing for LLGLTFMaterial instances.

There are several issues in the current code, stemming to the fact that the hashing is performed on the block of the member variables:

1.- There are padding bytes between member variables, even after rearranging them to avoid most of the padding; in particular, the std::array's size is not a multiple of 4 bytes (64 bits), and most compilers will pad them to the next 4-byte aligment as a result. Note that C++ standards do not impose the zeroing of padding bytes on construction of a class instance, with only a couple exceptions (such as explicit zero-initialization). Those bytes MUST therefore be zeroed by us on construction. 2.- The TextureTransform strutcure getPacked() method did not touch some of the packed bytes, and as a result could *potentially* cause an issue for hashing when applied to a transform of another material instance. 3.- With the recent addition of the local textures tracking map, the said map cannot be hashed as a block of memory (map pairs will typically be allocated on the heap or on the stack, not in the memory block used by member variables).

This PR solves all these issues and offers proper hashing of LLGLTFMaterial instances.

### DIFF
--- a/indra/llprimitive/llgltfmaterial.cpp
+++ b/indra/llprimitive/llgltfmaterial.cpp
@@ -47,16 +47,49 @@ const char* const LLGLTFMaterial::GLTF_FILE_EXTENSION_TRANSFORM_ROTATION = "rota
 // special UUID that indicates a null UUID in override data
 const LLUUID LLGLTFMaterial::GLTF_OVERRIDE_NULL_UUID = LLUUID("ffffffff-ffff-ffff-ffff-ffffffffffff");
 
+LLGLTFMaterial::LLGLTFMaterial()
+{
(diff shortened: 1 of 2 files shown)


sldevel:
recheck

marchcat-pe:
Thank you for your contribution! It's greatly appreciated.
We'll review and integrate it into the appropriate viewer branch.

RunitaiLinden (approved):
Approved.

Looking at the hacks piling up, I'm thinking it might be time to just implement a hash function that enumerates and hashes the members one at a time, but that's a task for another day.

marchcat-pe (approved):

PR #552, opened 2023-11-30
SL-19801 Inserting emoji characters from system emoji picker does not work on macOS

LLGuru:


marchcat-pe (approved):
I have no idea why Apple did it that way, but if it works...

maxim-productengine (approved):

akleshchev (approved):

PR #556, opened 2023-11-30
SL-20611 Make haze effect local lights.  Incidental fix for EventPump coroutine sometimes doing unsafe work mid-render.

RunitaiLinden:


cosmic-linden (approved):

sldevel:
I'm quite a bit skeptical about the posting of the commands to the mainloop work queue: work queues are for thread-safe registering of tasks (executed in another thread), not for posting (and postponing) coroutine results (especially as the coroutine in question actually runs in the same, main thread as the main coroutine).

Also, you can perfectly poll the mainloop work queue at any given time in the main thread, main coroutine, meaning there is no guarantee the queued event poll message handler will be executed at _the right time_; in my viewer, this in fact leads to a crash because that work queue is also used (polled) for shared GL context images creation, and processing the event polls messages at those times somehow causes illegal LLDrawable deletions (i.e. during sNoDelete = true periods) on TPs... I did not check in LL's viewer code, but this might be the case for it too, should the GL image worker thread be enabled.

Instead, I implemented for the Cool VL Viewer (code will be published tomorrow's release) a queuing of poll event replies in a static vector (local to the lleventpoll.cpp module), with a new LLEventPoll::dispatchMessages() method, called in LLAppViewer::idleNetwork() to trigger the event queue replies processing.

cosmic-linden (approved):

RunitaiLinden:
> I'm quite a bit skeptical about the posting of the event queue replies handling commands to the mainloop work queue: work queues are for thread-safe registering of tasks by another thread, not for posting (and postponing) coroutine results (especially as the coroutine in question actually runs in the same, main thread as the main coroutine).
> 
> Also, you can perfectly poll the mainloop work queue at any given time in the main thread, main coroutine, meaning there is no guarantee the queued event poll message handler will be executed at _the right time_; in my viewer, this in fact leads to a crash because that work queue is also used (polled) for shared GL context images creation, and processing the event polls messages at those times somehow causes illegal LLDrawable deletions (i.e. during sNoDelete = true periods) on TPs... I did not check in LL's viewer code, but this might be the case for it too, should the GL image worker thread be enabled.
> 
> Instead, I implemented for the Cool VL Viewer (code will be published in tomorrow's release) a queuing of poll event replies in a static vector (local to the lleventpoll.cpp module), with a new LLEventPoll::dispatchMessages() static method, called in LLAppViewer::idleNetwork() to trigger the event queue replies processing.

The assertion on LLDrawable deletion was what provoked posting the handler to the mainloop WorkQueue, as the handlers were being invoked between the frustum cull and the render and were deleting Drawables.  It should be true that work on the mainloop WorkQueue is only ever executed under gMainloopWork.runFor(MainWorkTimeNanoSec) in LLAppViewer::idle, and that's a safe place to execute any asynchronous event handlers.  As you say, WorkQueues are used for cross-thread communication, but they can also be used to schedule non-time-critical work to be executed at some point in the future.  As the comment says, it's a hack.  I'll follow up with Nat to help find out why the coroutine is being executed at the wrong time and find a better fix that makes sure the handlers don't fire at all while sNoDelete is set (or maybe this will turn out to be  the least-bad fix).

indra/newview/app_settings/shaders/class1/environment/waterFogF.glsl
```diff
@@ -108,3 +110,35 @@ vec4 applyWaterFogViewLinear(vec3 pos, vec4 color)
     return applyWaterFogViewLinearNoClip(pos, color);
 }
 
+// for post deferred shaders, apply sky and water fog in a way that is consistent with
+// the deferred rendering haze post effects
+vec4 applySkyAndWaterFog(vec3 pos, vec3 additive, vec3 atten, vec4 color)
+{
+    bool eye_above_water = dot(vec3(0), waterPlane.xyz) + waterPlane.w > 0.0;
+    bool obj_above_water = dot(pos.xyz, waterPlane.xyz) + waterPlane.w > 0.0;
```
cosmic-linden, line 118: eye_above_water: Dot product of 0 is 0. Could remove this.
More generally, I'm not sure what's going on in `LLSettingsVOWater::applySpecial`, where waterPlane is defined. The naming convention for `ep` and `enorm` seem inconsistent. `ep` is in eye space, but `enorm` is not. So, I wonder what is the reasoning behind `-ep.dot(enorm)`, if `ep` and `enorm` are not in the same coordinate system?

indra/newview/app_settings/shaders/class1/environment/waterFogF.glsl
```diff
@@ -108,3 +110,35 @@ vec4 applyWaterFogViewLinear(vec3 pos, vec4 color)
...
+    {
+        if (obj_above_water)
+        {
+            color.rgb = atmosFragLighting(color.rgb, additive, atten);
+        }
+        else
+        {
+            color.rgb = applyWaterFogViewLinearNoClip(pos, color).rgb;
+        }
+    }
```
cosmic-linden, line 141: Since there are only two outcomes, this branching can be simplified to only check for `obj_above_water`

cosmic-linden (approved):
LGTM. Noticed a few things.

PR #557, opened 2023-12-04
Fix formatting in autobuild.xml (indents in close tags)

LLGuru:
Also convert CRLF -> LF in [.github/release.yaml](https://github.com/secondlife/viewer/pull/557/files#diff-24062042b314111e98256591b30db43bb6fa4b9f1767cce24303a940808346ae)


marchcat-pe:
Otherwise, please set a target to Maint V.

marchcat-pe (approved):

nat-goodspeed (approved):
I approve the `release.yaml` changes.

As noted, the `autobuild.xml` changes are moot.

marchcat-pe:
@nat, since we're already talking about it, `autobuild installables edit ...` alters the whole autobuild.xml file on current `main` forks. Should we just commit this new altered version somewhere?

nat-goodspeed:
> autobuild installables edit ... alters the whole autobuild.xml file on current main forks.

I understand and agree with the above, up through "alters the whole `autobuild.xml` file." I'm not sure what you mean by adding "on current main forks" ?

> Should we just commit this new altered version somewhere?

The whitespace changes to `autobuild.xml` are transitory. I don't mind them being committed somewhere, but we must all understand that the next `autobuild installables edit` command will discard them.

I was wrong earlier about the code that would need to be changed: the bad indenting is actually in the [llsd](https://github.com/secondlife/python-llsd) repo.

akleshchev (approved):

marchcat-pe:
> I'm not sure what you mean by adding "on current main forks" ?
Current format that we have there is slightly different from the regular `autobuild installables edit` output. Packages order, keys order for every package, etc., so changing one package results in a diff with changes to every package. This is not related to the whitespaces shown in the PR.

The increased probability of a merge conflict here is my only concern. However, I can just push it as a part of the next LLCA package update and we'll never see this big diff again.
I'll create a new PR for that when ready, let's continue this discussion there. Hence, this one can be closed.

marchcat-pe:
Also, since we're going to release Maint V soon,

@LLGuru, please retarget this PR to `DRTVWR-588-maint-W`.

PR #559, opened 2023-12-04
SL-20679 fix for "Sort conversations by recent activity" option

maxim-productengine:


AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71060

PR #562, opened 2023-12-06
SL-20140 Setting shape hand size to 36 won't save

LLGuru:


marchcat-pe (approved):

indra/newview/llscrollingpanelparambase.cpp
```diff
@@ -101,12 +104,10 @@ void LLScrollingPanelParamBase::onSliderMoved(LLUICtrl* ctrl, void* userdata)
 
 F32 LLScrollingPanelParamBase::weightToPercent( F32 weight )
 {
-	LLViewerVisualParam* param = mParam;
-	return (weight - param->getMinWeight()) /  (param->getMaxWeight() - param->getMinWeight()) * 100.f;
+	return (weight - mParam->getMinWeight()) * 100.f;
```
akleshchev, line 107: 1. I feel like this needs an llclamp on input.
2. Are you sure this is the right way? Instead of just changing step size to make 61 values on a scale from 0 to 100 you are changing the scale which is likely undesirable and may backfire with either too big or too small values.

LLGuru: Yes, I'm sure
the step should be exact 1 (in fact - 0.01) because it can be entered manually as an integer

marchcat-pe: In this case, the method should probably be renamed to reflect its new behavior.

akleshchev: The options are to either making everything 'float' to fit 61 values on a 100 scale or to make scale be 0 to 60, I suggest reflecting this in commit's message.

akleshchev (approved):

indra/newview/llscrollingpanelparambase.cpp
vir-linden: It looks like we are changing the meaning of percent here. Shouldn't 100% span the full range from min weight to max weight? Note that the ranges can vary by param.

LLGuru: Yes, we change the available range ov values from `[0 .. 100]` to `[0 .. range_size]`
where **range_size** is `(value_max - value_min) * 100`

In fact, user receives the ability to set exact values which will be used actually (but multiplied by 100 and rounded)
For example, the range `[-0.3 .. 0.3]` will be available as interval `[0 .. 60]`

Eacb parameter will automatically use it's own range defined by **value_min** and **value_max**

PR #567, opened 2023-12-07
SL-20682 renaming of folders seems to fail (but doesn't really)

LLGuru:


marchcat-pe (dismissed):

indra/newview/llinventorymodel.cpp
```diff
@@ -3904,7 +3904,6 @@ void LLInventoryModel::processBulkUpdateInventory(LLMessageSystem* msg, void**)
 			cblist.push_back(InventoryCallbackInfo(callback_id, LLUUID::null));
 		}
 	}
-	gInventory.accountForUpdate(update);
```
akleshchev, line 3907: This is wrong.
1. 'update' in such case is not needed
2. This is a message by server that covers a lot of cases, not just moving items/folders, it's likely to break something.
3. In viewer-server relationship server is the one we trust, viewer should not pre-increment version if it knows server will respond

akleshchev (changes requested):
This change with this scenario causes a version mismatch with server
1. Close inventory and relog into viewer
2. Turn on menu-> develop ->xui -> Show XUI Names
3. Opem onventory and create a folder,
4. Note that folder now shows debug info and field V:1 says that it is version 1
5. Create a new item in this folder (ex: create a script), now version should V:2
6. Copy this item into same folder.

Expected: version 3 Actual: 2

akleshchev (approved):

PR #568, opened 2023-12-07
BUG-234770/SL-20682 - Workaround for version mismatch preventing rename

beqjanus:
See JIRA for details;
TL;DR reparenting things causes the parent folder version to become de-synchronised from the server version. This is a workaround that fixes that damage retrospectively when detected, however fixing the cause is obviously a better option. This patch also adds tags to the logging so that it is consistently reported. Also added a TODO where version is currently slammed to AIS version without a refresh.

### DIFF
--- a/indra/newview/llaisapi.cpp
+++ b/indra/newview/llaisapi.cpp
@@ -277,7 +277,7 @@ void AISAPI::CopyLibraryCategory(const LLUUID& sourceId, const LLUUID& destId, b
     {
         url += ",depth=0";
     }
-    LL_INFOS() << url << LL_ENDL;
+    LL_INFOS("Inventory") << url << LL_ENDL;
 
     std::string destination = destId.asString();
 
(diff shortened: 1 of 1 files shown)


beqjanus:
I just saw a commit from Guru related to this. That is still I WiP, I think, so I have sent this PR in any case. The workaround is probably still better at handling the version mismatch situation than the current log-it and run-away that we have.

marchcat-pe:
Thank you for this contribution.
We'll see how it goes and use the workaround if there's no way to do a proper fix now.

LLGuru:
No needs for this workaronnd because the bug fix is ready:
https://github.com/secondlife/viewer/pull/567

akleshchev:
It's still a correct way to recover from desyncs, but I'm not sure if it's a good idea to have or not. On one hand it's preferable to recovers if user does encounter issues (with a caveit that it should be disabled on test builds?) on another, version might be obsolete because there is any other change in progress (some sort of 'last updated' tracker?), also it probably shouldn't be refetching if mFetch is set, we literaly just got full version.

akleshchev:
Thank you, but pull request #567 was accepted, closing.

PR #574, opened 2023-12-11
Apparently slvoice is proprietary and thus needs to be handled accordingly. Otherwise *OS builds will fail

Nicky-D:
- Mac and Windows builds artifacts are not public anymore, rather they are hidden in the private repository https://api.github.com/repos/secondlife/3p-slvoice/
- For some reason the Linux build is still public, but obviously super outdated

-> Guard the slvoice binary behind the USE _PROPRIETARY flag, this is already used for FMOD and KDU and should be a good fit here

### DIFF
--- a/indra/cmake/Copy3rdPartyLibs.cmake
+++ b/indra/cmake/Copy3rdPartyLibs.cmake
@@ -262,21 +262,23 @@ endif(WINDOWS)
 # It's unclear whether this is oversight or intentional, but anyway leave the
 # single copy_if_different command rather than using to_staging_dirs.
 
-if( slvoice_src_dir )
-    copy_if_different(
-            ${slvoice_src_dir}
-            "${SHARED_LIB_STAGING_DIR_RELEASE}"
-            out_targets
-            ${slvoice_files}
-    )
-    list(APPEND third_party_targets ${out_targets})
-endif()
+if( USE_SLVOICE )
+  if( slvoice_src_dir )
+      copy_if_different(
+              ${slvoice_src_dir}
+              "${SHARED_LIB_STAGING_DIR_RELEASE}"
+              out_targets
+              ${slvoice_files}
+      )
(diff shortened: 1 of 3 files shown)


Nicky-D:

Ansariel:
I would rather raise the question why the SLVoice 3p lib suddenly is not public anymore!

Nicky-D:
Fair enough. This btw has been in this state since mid May, why it did not strike me as an accident but deliberate when being like this for over half a year.

vir-linden:
We may be making SLVoice public soon - let's wait on this until end of the week and then discuss.

sldevel:
> We may be making SLVoice public soon - let's wait on this until end of the week and then discuss.

Well, if TPV viewers cannot even link against Vivox libraries, I foresee a _slight_ issue (euphemism, of course) for voice usage in SL... 

So let's hope SLVoice is made public again _real_ soon. :-P

Ansariel:
@vir-linden any updates on the SLVoice libs?

nat-goodspeed:
As part of migrating builds to GitHub, we deemed it unnecessary to continue using our internal codeticket service. Our TeamCity builds would direct codeticket to post build artifacts to either a public or a private S3 bucket. That was useful because although our 3p-slvoice repo has always been private, per licensing agreements with Vivox, we could still specify that its built result should be posted to a public S3 bucket.

An artifact built from a private GitHub repo is automatically private, though. You don't get to choose. That's been the obstacle.

Meanwhile Vivox is under new management, and we've been asking them to review the 3p-slvoice repo. In other words, is there in fact any content that they would object to making public? If not, can we just post a public repo? That's our current status. Sorry for the delay, but it's not _all_ Linden's fault. Thanks for your patience...

Nicky-D:
I'm going to close this one, as it's (hopefully) tracked somewhere else

PR #576, opened 2023-12-11
SL-20714 Crash accessing mControlAVBridge

akleshchev:
Looks like control avatar was recreated after cleanup then object was deleted


maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71103

PR #577, opened 2023-12-11
SL-20713 Crash at isAvatar

akleshchev:
Likely object was NULL

P.S. but there is another case of isAvatar crash in stateSort()


marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71102

PR #578, opened 2023-12-12
Fix for semi-transparent HUDs rendering opaque

sldevel:
This commit fixes a bug introduced with commit 6472b75bcd70470fe5775d1cf6eb70a75b3d76e5 where the fullbrightF.glsl shader fails to set color.a to final_alpha for HUDs.


cosmic-linden (approved):

cosmic-linden:
Good catch.

PR #581, opened 2023-12-12
SL-20626 BugSplat Crash #1364225: LLRender::flush(1626)

LLGuru:


indra/llrender/llrender.cpp
```diff
@@ -1619,7 +1619,7 @@ void LLRender::flush()
 
 		mCount = 0;
 
-        if (mBuffer)
+        if (mBuffer && LLGLSLShader::sCurBoundShaderPtr)
```
akleshchev, line 1622: Wouldn't this just cause a crash on else? `LL_ERRS() << "A flush call from outside main rendering thread"`

akleshchev (changes requested):
Either not to crash or error should provide a sensible reason

indra/llrender/llrender.cpp
LLGuru: Agree, we can just use zero mask in this case

akleshchev (approved):

indra/llrender/llrender.cpp
```diff
@@ -1623,7 +1623,7 @@ void LLRender::flush()
         {
 
             HBXXH64 hash;
-            U32 attribute_mask = LLGLSLShader::sCurBoundShaderPtr->mAttributeMask;
+            U32 attribute_mask = LLGLSLShader::sCurBoundShaderPtr ? LLGLSLShader::sCurBoundShaderPtr->mAttributeMask : 0;
```
cosmic-linden, line 1626: `LLGLSLShader::sCurBoundShaderPtr` should never be null. If it is null, the viewer would crash, if not here then possibly later. I also see at least one place that assumes `attribute_mask` is never zero. See `vb->setPositionData` in this function for example. I think addressing the shader being null is probably a more robust direction.

In `LLGLSLShader::bind`, I see:

```
if (sCurBoundShader != mProgramObject)  // Don't re-bind current shader
```

I could see this potentially failing if `sCurBoundShader` and `mProgramObject` are both 0. which could happen if `glCreateProgram` fails in `LLGLSLShader::createShader`. This should be logged, so it should be possible to detect if that. is the case.

cosmic-linden:
By the way @akleshchev @LLGuru The attribute mask is used by the shaders to indicate what vertex attributes (position, normals, etc) are used by the shader, and what vertex attributes a vertex buffer can provide to the shader.

akleshchev:
> By the way @akleshchev @LLGuru The attribute mask is used by the shaders to indicate what vertex attributes (position, normals, etc) are used by the shader, and what vertex attributes a vertex buffer can provide to the shader.
Thank you for the explanation, seems like this review needs more work.

indra/llrender/llrender.cpp
LLGuru: Yes, `LLGLSLShader::sCurBoundShaderPtr` should never be NULL

Anyway, we use sometimes the checks of kind `llassert(LLGLSLShader::sCurBoundShaderPtr != NULL);`
or of kind `if (LLGLSLShader::sCurBoundShaderPtr)` - this means that this pointer could and should be checked

I do not see any other crash reason on the line 1626

Using zero `attribute_mask` shouldn't lead to a crash

akleshchev: Checked logs and seems to be hardware specifc, users with specific hardware reproduce it contantly, but no one else.
Logs are full of problems like `ERROR: Undefined function call: "calcAtmospherics(vf3;)".  Function not found.` and `CreateDXGIFactory1 failed: 0x80004002` so I think you are fixing it in a wrong place. Viewer should have detected a problem before this then quit while advising user to update driver.

LLGuru: Yes, we have to fix the creating of a shader by checking the result, on the one hand
And on the other hand, we have to add the check for the absent shader in this function
If viewer won't crash and the user will see the black screen (for example), it would be better behavior than the crash

akleshchev: > If viewer won't crash and the user will see the black screen (for example), it would be better behavior than the crash

It's more user friendly to immediately crash, ideally with a warning 'hey there are graphical problems', than to show a black screen for an undefinite amount of time (because users would try to wait it out). Futhermore crashing informs us that there is a problem that needs fixing, while a 'black screen' does not despite also being a problem.

indra/llrender/llrender.cpp
```diff
@@ -1580,7 +1580,12 @@ void LLRender::flush()
 	if (mCount > 0)
 	{
         LL_PROFILE_ZONE_SCOPED_CATEGORY_PIPELINE;
-        llassert(LLGLSLShader::sCurBoundShaderPtr != nullptr);
+        if (!LLGLSLShader::sCurBoundShaderPtr)
+        {
+            // The shader was not created yet? Report and shatdown the application
+            LL_ERRS() << "There are graphical problems (no shader), try updating driver" << LL_ENDL;
+        }
+
```
cosmic-linden, line 1588: `LLRender::flush` is called in a lot of different places. It would make more sense to check for failed shader compilation/linking.

PR #586, opened 2023-12-13
SL-20729 Crash in getObject

akleshchev:
In "new_folder_from_selected" at gInventory.getObject(*ids.begin()); (see string 3200 and 3224).
According to bugsplat argument id was a null pointer so I assume there was an issue geting id out of the list. List likely was empty.


marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71119

PR #592, opened 2023-12-14
SL-3508 Crash in LLKDUDecodeState::processTileDecode

LLGuru:


indra/llimage/llimage.cpp
```diff
@@ -2130,6 +2130,10 @@ U8* LLImageFormatted::reallocateData(S32 size)
 // virtual
 void LLImageFormatted::deleteData()
 {
+    if (mDecoding)
+    {
+        LL_ERRS() << "LLImageFormatted::deleteData() is called during decoding" << LL_ENDL;
```
marchcat-pe, line 2135: Not sure if we should just crash here. Maybe warning + just drop decoding?

marchcat-pe (approved):
It looks safer now. Let's test!

indra/llkdu/llimagej2ckdu.cpp
```diff
@@ -578,16 +579,18 @@ bool LLImageJ2CKDU::decodeImpl(LLImageJ2C &base, LLImageRaw &raw_image, F32 deco
 															mCodeStreamp.get()));
 				}
 				// Do the actual processing
-				F32 remaining_time = decode_time - decode_timer.getElapsedTimeF32();
+				F32 remaining_time = limit_time ? decode_time - decode_timer.getElapsedTimeF32() : 0;
```
marchcat-pe, line 582: Use `0.0f` here to fix mac build error.

indra/llimage/llimage.cpp
akleshchev: Decoding happens in a thread we have no direct access to (pass it to decode?), but doing so would require mutex locing data, waiting for thread to actually stop, so it makes sense to sync data in another way.

But I have conflicted feelings about what should be done here. We can just mutex lock this data thus ensure it's never an issue. And it shouldn't affect much since we mostly use formated images for fetching and for uploading. The question is, is it really better and how wide this mutex lock should be (should it cover just deletion and decoding? Any access to formated image's data?).

akleshchev: From a threading perspective this isn't the best check, variable need to be set before !getData(), otherwise there is a chance deleteData() will happen in between !getData() and decodeImpl and it such case it would still be unreliable (potentially can be fixed by making variable an atomic?) which bring back question of mutexes...

akleshchev (approved):

indra/llimage/llimage.cpp
LLGuru: Calling `deleteData()` during decoding leads to various crashes
See https://app.bugsplat.com/v2/crashes?c0=functionName&f0=CONTAINS&v0=processTileDecode&database=SecondLife_Viewer_2018

This change will move the crash to the root cause, we'll see the real reason for the crash
I expect we won't receive crashes from `LLKDUDecodeState::processTileDecode` anymore
When (and if) we receive crashes in `LLImageFormatted::reallocateData` we'll fix them too

indra/llkdu/llimagej2ckdu.cpp
marchcat-pe: Well, this was not enough :(
Probably casting to F32 or some refactoring will help.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/72761

PR #594, opened 2023-12-15
SL-20743 Use LLMutex in LLImageBase for internal data thread-safety

LLGuru:
Also some enchancements for https://github.com/secondlife/viewer/pull/593

### DIFF
--- a/indra/llimage/llimage.h
+++ b/indra/llimage/llimage.h
@@ -116,7 +116,11 @@ class LLImageBase
 {
 protected:
 	virtual ~LLImageBase();
-	
+
+	virtual void deleteData();
+	virtual U8* allocateData(S32 size = -1);
+	virtual U8* reallocateData(S32 size = -1);
+
 public:
 	LLImageBase();
 
@@ -125,10 +129,6 @@ public:
 		TYPE_NORMAL = 0,
 		TYPE_AVATAR_BAKE = 1,
 	};
-
-	virtual void deleteData();
-	virtual U8* allocateData(S32 size = -1);
-	virtual U8* reallocateData(S32 size = -1);
 
 	virtual void dump();
 	virtual void sanityCheck();
@@ -171,9 +171,26 @@ private:
 
 	S8 mComponents;
 
-	bool mBadBufferAllocation ;
-	bool mAllowOverSize ;
+	bool mBadBufferAllocation;
+	bool mAllowOverSize;
+
+private:
+    mutable LLSharedMutex mDataMutex;
+
+public:
+    template<bool SHARED>
+    class DataLock : LLSharedMutexLockTemplate<SHARED>
+    {
+    public:
+        DataLock(const LLImageBase* image)
+        : LLSharedMutexLockTemplate<SHARED>(image ? &image->mDataMutex : nullptr)
+        {
+        }
+    };
 };
+
+using LLImageDataLock = LLImageBase::DataLock<false>;
+using LLImageDataSharedLock = LLImageBase::DataLock<true>;
 
 // Raw representation of an image (used for textures, and other uncompressed formats
 class LLImageRaw : public LLImageBase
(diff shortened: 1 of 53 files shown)


indra/llimage/llimage.h
```diff
@@ -356,7 +356,9 @@ class LLImageFormatted : public LLImageBase
 	S8 mDecoded;  // unused, but changing LLImage layout requires recompiling static Mac/Linux libs. 2009-01-30 JC
 	S8 mDiscardLevel;	// Current resolution level worked on. 0 = full res, 1 = half res, 2 = quarter res, etc...
 	S8 mLevels;			// Number of resolution levels in that image. Min is 1. 0 means unknown.
-	
+
+	std::mutex mMutex;
```
marchcat-pe, line 360: Good, but LLMutex might be better since we've already have the wrapper.

indra/newview/llinventorygallerymenu.cpp
```diff
@@ -70,7 +70,7 @@ void modify_outfit(BOOL append, const LLUUID& cat_id, LLInventoryModel* model)
     if (items.size() > max_items())
     {
         LLSD args;
-        args["AMOUNT"] = llformat("%d", static_cast<U32>(max_items));
+        args["AMOUNT"] = llformat("%u", max_items());
```
marchcat-pe, line 73: 👍

indra/llkdu/llimagej2ckdu.cpp
```diff
@@ -579,7 +579,7 @@ bool LLImageJ2CKDU::decodeImpl(LLImageJ2C &base, LLImageRaw &raw_image, F32 deco
 															mCodeStreamp.get()));
 				}
 				// Do the actual processing
-				F32 remaining_time = limit_time ? static_cast<F32>(decode_time - decode_timer.getElapsedTimeF32()) : 0.0f;
+				F32 remaining_time = limit_time ? decode_time - decode_timer.getElapsedTimeF32()() : 0.0f;
```
marchcat-pe, line 582:
```suggestion
				F32 remaining_time = 0.0f;
				if (limit_time)
					remaining_time = decode_time - decode_timer.getElapsedTimeF32();
```

indra/llimage/llimage.h
akleshchev: Agree, please use LLMutex, and probably a good idea to switch LL_ERRS() to LL_WARNS(), since if section is mutex locked,  mDecoding beeing set is a sign of decoding thread terminating in a wrong state, not running.

indra/llcommon/llunittype.h
```diff
@@ -154,6 +154,11 @@ struct LLUnit
         return value();
     }
 
+    LL_FORCE_INLINE storage_t operator () () const
```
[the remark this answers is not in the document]
LLGuru: It's more useful and much more readablem then `static_cast<F32>`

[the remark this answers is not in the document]
marchcat-pe: But omitting the ternary will possibly improve readability even more.

akleshchev (commented):
If you are covering data access with mutex, please cover everything, not just the part from a specific callstack, ex: updateData() calls mImpl->getMetadata(*this); after chacking getData(), meaning there is a chance data will vanish in between.

indra/llimage/llimage.h
akleshchev: LLMutex is an extension of an std mutex, unless specific advantages are needed, I'm not sure if there is a point.

indra/llcommon/llunittype.h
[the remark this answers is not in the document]
akleshchev: Agree with marchcat. Something like
```
F32SecondsImplicit a =5,f;
F32 b = a;
```
Workd just fine. We treat 'a' in this case as any other F32, calling an a() would be confusing.

indra/llkdu/llimagej2ckdu.cpp
LLGuru: Using 3 lines instead of 1 doesn't help to see the whole algorithm
We'd better hide these simple details to allow user see the picture

indra/llimage/llimage.h
LLGuru: I believe this ERRS should never happen after this change

And if it happens for some reason, this WARNS could lead to a crash in other thread,
and we won't see the root cause again (WARNS doesn't show the callstack)

ERRS is actually what we need in this case

indra/llkdu/llimagej2ckdu.cpp
marchcat-pe: It's not about counting the lines. 
Ternaries are fine, but if we need to add more operators in order to use them, it's better to avoid that.

indra/llcommon/llunittype.h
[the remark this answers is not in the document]
LLGuru: My goal was to simplify all this stuff and fit it in a single line,
hide the container-related details and focus on the algorithm itself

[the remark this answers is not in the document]
LLGuru: If you guys are so adamantly against this `operator ()` then I propose the following variant:

`F32 remaining_time = limit_time ? decode_time - decode_timer.getElapsedTimeF32().value() : 0.0f;`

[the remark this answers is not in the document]
akleshchev: Won't something like `F32 remaining_time = limit_time ? (decode_time - decode_timer.getElapsedTimeF32()) : 0.0f;` work as is?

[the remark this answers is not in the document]
akleshchev: Works for me (value() feels like it should be returning an LLSD), but personaly I would just use (F32)decode_timer.getElapsedTimeF32(), I'm not sure why clang thinks it's ambigious when everythng else is explicitly F32.

akleshchev (changes requested):
1. Fix isn't complete. mDecoding and mutex aren't general enough, it covers strictly LLImageJ2C, but what if viewer interupts png or some other format? It's rare, but those are present (some avatar icons had gifs; local textures aren't limited to j2c).
2. WorkQueueBase::callWork() specifically catches exceptions (because if decoding fails with an exception we can safely rocever and just mark image as missing), but in your case mDecoding will be stuck as true and viewer will crash either way. Either it should be a warning and not an LL_ERRS or should be reset once decoder throws an exception.

LLGuru:
> 1. Fix isn't complete. mDecoding and mutex aren't general enough, it covers strictly LLImageJ2C, but what if viewer interupts png or some other format? It's rare, but those are present (some avatar icons had gifs; local textures aren't limited to j2c).

I guess all listed cases don't relate to the current bug, we're talking only about `LLKDUDecodeState::processTileDecode`

akleshchev:
> I guess all listed cases don't relate to the current bug, we're talking only about LLKDUDecodeState::processTileDecode

If KDU crashes here due to data being deleted, OpenJpeg and for example LLImagePNG::decode does too. It's not processTileDecode that is the issue, but deletion of data, so if data can be missing, than everything it can be missing from should be covered.

LLGuru:
> if data can be missing, than everything it can be missing from should be covered.

Agree, but this change seems to be a much more wider than a simple crash fix, so it requires a separate jira ticket
Could you please create one and assign it to me (e.g., "Use LLMutex in LLImageBase for internal data thread-safety")?

Thanks

marchcat-pe:
@LLGuru Ok, please use this one: SL-20743.

akleshchev:
Not LLImageBase, that would cover raw images as well which are secured in a difrerent way and much more widely used, only LLImageFormatted

indra/newview/lltexturecache.cpp
```diff
@@ -2075,6 +2078,7 @@ bool LLTextureCache::writeToFastCache(LLUUID image_id, S32 id, LLPointer<LLImage
 		{
             // Make a duplicate to keep the original raw image untouched.
             raw = raw->duplicate();
+            lock.update(raw);
```
akleshchev, line 2081: What's the point? It's literaly a single use copy.

indra/llkdu/llimagej2ckdu.cpp
akleshchev: Agree with marchcat, and personaly I find longer Ternaries slower to read than clear if+else. They are convinient for some cases and easy to read when very short, but if they require either more code to make work, too long or use other operations inside it's probably better to avoif them.

akleshchev (approved):

RunitaiLinden:
> I think covering base image is excessive since it also covers Raw images, covering formated images would have been enough to avoid the issue. But covering it is indeed safer, might even fix some issues and probably has no major downsides, asking davep to check.

Out of principle, I don't reject PRs out of performance concerns and just wait until performance concerns present themselves and then fix them, but I do worry that this will negate the benefit of LLImageGLThread entirely since the whole point of LLImageGLThread is that you can use an LLImageGL _while it is being updated in a background thread without blocking the main thread_.  Be sure to profile this with Tracy and check for stalls in the main thread waiting on these new mutexes.  It's better to crash once every 20 hours than it is to have frame stalls every second.

I'm also concerned by the introduction of yet another wrapper around yet another flavor of std::mutex, particularly one that maintains an internal std::map of the threads that have locked it.  We should really just be using std mutexes directly.

LLGuru:
Optimized by performance: std::map is replaced with std::unordered_map, LLMutex is replaced with std::mutex
image

akleshchev:
Locking main thread for 11.64 ms is a lot...

Alternative solution suggestion:
1. In LLTextureFetchWorker::resetFormattedData instead of deleting data from mFormattedImage, null the pointer instead of deleting data. It's an LLPointer, so it should clean itself as needed if nothing else is using it.
2. In WAIT_HTTP_REQ part, after it's done making a copy of image buffer, null the pointer then recreate the image (we are going to reallocate data either way)
3. Similar treatment for cache

Crash happens in decode requested by LLTextureFetchWorker, so it should be related to LLTextureFetchWorker's formated image (but I fail to see how it can get reset while fetcher is in a DECODE_IMAGE_UPDATE state)

LLGuru:
> Locking main thread for 11.64 ms is a lot... What region did you check it at? Was that some insanely decorated region or one of simple PE regions?

It was lagland, as it was suggested by @RunitaiLinden - and he said that this result is OK

LLGuru:
> Alternative solution suggestion:

There is no an ideal reacheable: for any proposed enchancement it makes sense to create a separate ticket
Otherwise, I risk continuing to work on this bug until retirement. :-)

RunitaiLinden (approved):
Still have some concerns about mutex performance in general, but profiles didn't show any red flags, so it's fine to go ahead and get this out there to see if it helps the crash rate and keep an eye on profiles and optimize as needed.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71134

PR #595, opened 2023-12-15
SL-20737 Crash clearing texture callbacks

akleshchev:
Shutdown crash


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71128

PR #601, opened 2023-12-19
SL-20744 Update libs to support macos minimum version 10.13

marchcat:


marchcat:
Minimum macos version is 11.7 now, closing.

PR #604, opened 2023-12-25
Fix recenty-activity based sorting of the various chat participant lists

RyeMutt:
Noticed this was broken when I was doing a cleanup pass in Alchemy.


Ansariel:
Isn't this part of one of the maint branches already? Pretty sure I have seen something similar.

marchcat-pe:
Right, it's in DRTVWR-588-maint-W already.

PR #605, opened 2024-01-02
SL-20759 Ctrl+O should close Avatar floater if Wearing tab is opened

maxim-productengine:


akleshchev (approved):

indra/newview/llviewermenu.cpp
```diff
@@ -6651,6 +6652,13 @@ void handle_edit_outfit()
...
 void handle_now_wearing()
 {
+    LLSidepanelAppearance *panel_appearance = dynamic_cast<LLSidepanelAppearance *>(LLFloaterSidePanelContainer::getPanel("appearance"));
+    if (panel_appearance && panel_appearance->isInVisibleChain() && panel_appearance->isCOFPanelVisible())
+    {
+        LLFloaterReg::findInstance("appearance")->closeFloater();
+        return;
+    }
+
     LLFloaterSidePanelContainer::showPanel("appearance", LLSD().with("type", "now_wearing"));
```
akleshchev, line 6662: Looks good, but are you sure it should be here an not in "now_wearing" (showOutfitsInventoryPanel or toggleMyOutfitsPanel)?

maxim-productengine: it seems it's better to change behavior only when using the menu/hotkey.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71151

PR #608, opened 2024-01-03
SL-17896 Don't crash silently if files are missing or out of memory

akleshchev:
Under debug LL_ERRS will show a message as well, but release won't show anything and will quit silently so show a notification when applicable.


marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
https://github.com/secondlife/viewer-private/issues/38

PR #609, opened 2024-01-04
SL-3824 Remove asserts from LLImage

LLGuru:


indra/llimage/llimagebmp.cpp
```diff
@@ -321,11 +321,7 @@ bool LLImageBMP::updateData()
 	if( 0 != mColorPaletteColors )
 	{
 		mColorPalette = new(std::nothrow) U8[color_palette_size];
-		if (!mColorPalette)
-		{
-			LL_ERRS() << "Out of memory in LLImageBMP::updateData()" << LL_ENDL;
-			return false;
-		}
+		llassert_message_return_false(mColorPalette, "Out of memory in LLImageBMP::updateData()")
```
akleshchev, line 324: Disagree, unless color_palette_size is unreasonable viewer should show a notification (done in a separate commit) and quit then report to bugsplat. If viewer can't allocate this tiny amount of data it's going to crash either way next time it allocates a buffer for something.

indra/llimage/llimagedxt.cpp
```diff
@@ -207,10 +207,7 @@ bool LLImageDXT::updateData()
 		height = header->maxheight;
 	}
 
-	if (data_size < mHeaderSize)
-	{
-		LL_ERRS() << "LLImageDXT: not enough data" << LL_ENDL;
-	}
+	llassert_message_return_false(data_size >= mHeaderSize, "LLImageDXT: not enough data")
```
akleshchev, line 210: Disagree. If this actually happens, memory is likely corrupted, viewer should quit, not keep going.

indra/llimage/llimagetga.cpp
```diff
@@ -570,7 +571,7 @@ bool LLImageTGA::decodeColorMap( LLImageRaw* raw_image, bool rle, bool flipped )
 		case 2:	pixel_decoder = &LLImageTGA::decodeColorMapPixel15; break;
 		case 3:	pixel_decoder = &LLImageTGA::decodeColorMapPixel24; break;
 		case 4:	pixel_decoder = &LLImageTGA::decodeColorMapPixel32; break;
-		default: llassert(0); return false;
+		default: LL_WARNS() << "mColorMapBytesPerEntry not in [1..4]" << LL_ENDL; return false;
```
akleshchev, line 574: This has potential to be spammy thus causing slowdowns, I think it should either be a DEBUG or _ONCE.

indra/llcommon/llerror.h
```diff
@@ -90,6 +90,42 @@ const int LL_ERR_NOERR = 0;
...
+// The macro llassert_returning is used in the following macros:
+// - llassert_message_return_value
+// - llassert_message_return
+// - llassert_return_value
+// - llassert_return
+#define llassert_returning(condition, separator, message, flag, value) \
+{                                                                      \
+    if (LL_UNLIKELY(!(condition)))                                     \
+    {                                                                  \
+        LL_WARNS()                                                     \
```
akleshchev, line 107: basic llassert() is develop build only (llassert_always work everywhere) I think this should be kept the same way due to viewer handling massive amount of images. If something happens viewer will freeze due to excessive logging.

indra/llimage/llimage.cpp
```diff
@@ -2405,7 +2405,8 @@ void LLImageBase::generateMip(const U8* indata, U8* mipdata, S32 width, S32 heig
 				*(U8*)data = (U8)(((U32)(indata[0]) + indata[1] + indata[in_width] + indata[in_width+1])>>2);
 				break;
 			  default:
-				LL_ERRS() << "generateMmip called with bad num channels" << LL_ENDL;
+				LL_WARNS() << "generateMmip called with bad num channels (" << nchannels << ")" << LL_ENDL;
```
akleshchev, line 2408: Just shouldn't happen. If it did either developer did a mistake developer needs to know about now instead of from logs sometime next year or viewer is likely suffering from memory corruption and should quit and report this issue to bugsplat, not report a warning.

indra/llimage/llimagetga.cpp
```diff
@@ -1051,7 +1052,7 @@ bool LLImageTGA::decodeTruecolorRle8( LLImageRaw* raw_image )
 // Processing happens during the decode for speed.
 bool LLImageTGA::decodeAndProcess( LLImageRaw* raw_image, F32 domain, F32 weight )
 {
-	llassert_always(raw_image);
+	llassert_return_false(raw_image);
```
akleshchev, line 1055: I think this should stay either llassert_always or better be 'if (!raw_image) LL_ERRS', because if this happened, viewer encountered issues with decoder that are bigger than a single image failure.

indra/llimage/llimagebmp.cpp
```diff
@@ -334,7 +330,7 @@ bool LLImageBMP::updateData()
 
 bool LLImageBMP::decode(LLImageRaw* raw_image, F32 decode_time)
 {
-	llassert_always(raw_image);
+	llassert_return_false(raw_image);
```
akleshchev, line 333: This is a sign of problems with decoder, viewer should crash and report to bugsplat.

indra/llimage/llimagebmp.cpp
```diff
@@ -542,7 +538,7 @@ bool LLImageBMP::decodeTruecolor24( U8* dst, const U8* src )
 
 bool LLImageBMP::encode(const LLImageRaw* raw_image, F32 encode_time)
 {
-	llassert_always(raw_image);
+	llassert_return_false(raw_image);
```
akleshchev, line 541: Again, should report to bugsplat.

indra/llimage/llimagebmp.cpp
LLGuru: Should we check whether the requested memory size is really tiny? :-)

marchcat-pe: No, this type of crashes is common when free RAM is actually insufficient. Even if you can squeeze one more image into it, it will fail on the next one.

indra/llimage/llimagedxt.cpp
```diff
@@ -207,10 +207,7 @@ bool LLImageDXT::updateData()
 		height = header->maxheight;
 	}
 
-	if (data_size < mHeaderSize)
-	{
-		LL_ERRS() << "LLImageDXT: not enough data" << LL_ENDL;
-	}
+	llassert_always_msg(data_size >= mHeaderSize, "LLImageDXT: not enough data");
```
akleshchev, line 210: I thing this is a sign of memory correption, thus should error globaly, but otherwise it should set setLastError("Invalid input"); or something like that.

indra/llcommon/llerror.h
cosmic-linden: I agree. Also, a big use of asserts is to make it easier for developers to spot errors in code, because it triggers a crash breakpoint in the debugger. I think preserving that behavior in a debug build is very useful. This macro is technically not an assert because it doesn't do that.

indra/llimage/llimage.cpp
```diff
@@ -657,7 +655,7 @@ void LLImageBase::sanityCheck()
 		|| mComponents > (S8)MAX_IMAGE_COMPONENTS
 		)
 	{
-		LL_ERRS() << "Failed LLImageBase::sanityCheck "
+		LL_WARNS() << "Failed LLImageBase::sanityCheck "
 			   << "width " << mWidth
 			   << "height " << mHeight
 			   << "datasize " << mDataSize
```
cosmic-linden, line 661: This function isn't used for anything and was probably left as a debugging tool. `LL_ERRS` seems fine here.

cosmic-linden:
I think a lot of these image processing functions are pretty straightforward in how they work. If we're misusing them in a way that will cause a crash later on, better to crash as soon as possible. Otherwise... I think it depends on what exactly we're trying to fix...

cosmic-linden:
@marchcat-pe do you recall why this ticket was created?

marchcat-pe:
It was a follow-up for a series of the crash fixes. The idea, as I recall, was to allow image loading to fail with some logging rather than crashing the viewer forcefully when the conditions weren't met.

akleshchev:
>The idea, as I recall, was to allow image loading to fail with some logging rather than crashing the viewer forcefully when the conditions weren't met.

I'm fairly ceratain we fixed those crashes and handled failed decodings we could handle. And llassert doesn't crash viewer user side.

indra/llcommon/llerror.h
```diff
@@ -90,6 +90,42 @@ const int LL_ERR_NOERR = 0;
...
+// The following 2 macros are used below in llassert_returning
+#define llvalue_flag0(value)
+#define llvalue_flag1(value) value
+
+// The macro llassert_returning is used in the following macros:
+// - llassert_message_return_value
+// - llassert_message_return
+// - llassert_return_value
+// - llassert_return
+#define llassert_returning(condition, separator, message, flag, value) \
```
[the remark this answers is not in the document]
marchcat-pe: Agree, this may be confusing.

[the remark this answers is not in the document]
LLGuru: There is a difference between "return code" and "exit code"
If the macro promises to return a code, it should return this code from the function
(not to exit AKA terminate the process)

As for me, the name of this macro is readable enough
Moreover, user can hover the mouse cursor and see the definition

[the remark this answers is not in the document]
marchcat-pe: Well, if something can confuse someone, it surely will.

indra/llimage/llimage.cpp
```diff
@@ -946,7 +944,7 @@ bool LLImageRaw::setSubImage(U32 x_pos, U32 y_pos, U32 width, U32 height,
 
 void LLImageRaw::clear(U8 r, U8 g, U8 b, U8 a)
 {
-	llassert( getComponents() <= 4 );
+	llassert_return( getComponents() <= 4 );
```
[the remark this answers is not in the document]
akleshchev: Sorry, got confused, it was alright in this place.

indra/llcommon/llerror.h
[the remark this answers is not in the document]
cosmic-linden: Since this is intended to replace plain asserts, `llassert_returning` should invoke `llassert(condition)`.

indra/llimage/llimagedxt.cpp
```diff
@@ -443,23 +438,23 @@ bool LLImageDXT::convertToDXR()
...
+	S32 total_bytes = getDataSize();
+	llassert_return_false(total_bytes > 0);
 	U8* newdata = (U8*)ll_aligned_malloc_16(total_bytes);
-	if (!newdata)
-	{
-		LL_ERRS() << "Out of memory in LLImageDXT::convertToDXR()" << LL_ENDL;
-		return false;
-	}
-	llassert(total_bytes > 0);
+	llassert_message_return_false(newdata, "Out of memory in LLImageDXT::convertToDXR()")
```
cosmic-linden, line 446: @akleshchev / @marchcat-pe make good points about how to handle OOM - this is another example here

PR #614, opened 2024-01-07
SL-20743 Use LLMutex in LLImageBase for internal data thread-safety (update)

LLGuru:


marchcat-pe (approved):

akleshchev (approved):
Can you use std::shared_mutex instead? Like: read operations getting shared locks and write operations getting exlusive locks.

LLGuru:
> Can you use std::shared_mutex instead? Like: read operations getting shared locks and write operations getting exlusive locks.

Yes, I can, and I do: `mSharedMutex` is of type `std::shared_mutex`
image

akleshchev:
My consern is that maintaining a list of threads is an unneeded overhead, a thread local 'islocked' bool or  something similar should have been enough.

akleshchev:
Could you have used a thread_local integer to store how many times it was locked instead of maintaining an actual list?

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71134

PR #616, opened 2024-01-08
Replace BOOST_FOREACH with standard C++ range-based for-loops

Ansariel:
This replaces BOOST_FOREACH with the C++ standard range-based for-loops and eliminates the include of the boost/foreach.hpp header. Unfortunately, there are two occurrances of BOOST_REVERSE_FOREACH in lllayoutstack.cpp. These should be easy to replace once the C++ standard is switched to C++20, which allows the use of std::ranges::view::reverse adaptor.

### DIFF
--- a/indra/llcommon/llsys.cpp
+++ b/indra/llcommon/llsys.cpp
@@ -49,7 +49,6 @@
 #include "llsdutil.h"
 #include <boost/bind.hpp>
 #include <boost/circular_buffer.hpp>
-#include <boost/foreach.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/range.hpp>
 #include <boost/utility/enable_if.hpp>
@@ -905,9 +904,9 @@ void LLMemoryInfo::stream(std::ostream& s) const
 
 	// Max key length
 	size_t key_width(0);
-	BOOST_FOREACH(const MapEntry& pair, inMap(mStatsMap))
+	for (const auto& [key, value] : inMap(mStatsMap))
 	{
-		size_t len(pair.first.length());
+		size_t len(key.length());
 		if (len > key_width)
 		{
 			key_width = len;
@@ -915,10 +914,9 @@ void LLMemoryInfo::stream(std::ostream& s) const
 	}
 
 	// Now stream stats
-	BOOST_FOREACH(const MapEntry& pair, inMap(mStatsMap))
+	for (const auto& [key, value] : inMap(mStatsMap))
 	{
-		s << pfx << std::setw(narrow(key_width+1)) << (pair.first + ':') << ' ';
-		LLSD value(pair.second);
+		s << pfx << std::setw(narrow(key_width+1)) << (key + ':') << ' ';
 		if (value.isInteger())
 			s << std::setw(12) << value.asInteger();
 		else if (value.isReal())
--- a/indra/llui/lllayoutstack.cpp
+++ b/indra/llui/lllayoutstack.cpp
@@ -395,8 +395,7 @@ void LLLayoutStack::updateLayout()
 							: getRect().getHeight();
 
 	// first, assign minimum dimensions
-	LLLayoutPanel* panelp = NULL;
-	BOOST_FOREACH(panelp, mPanels)
+	for (LLLayoutPanel* panelp : mPanels)
 	{
 		if (panelp->mAutoResize)
 		{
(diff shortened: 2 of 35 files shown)


indra/llui/lllayoutstack.cpp
```diff
@@ -409,12 +408,15 @@ void LLLayoutStack::updateLayout()
 	llassert(total_visible_fraction < 1.05f);
 
 	// don't need spacing after last panel
-	space_to_distribute += panelp ? ll_round((F32)mPanelSpacing * panelp->getVisibleAmount()) : 0;
+	{
+		LLLayoutPanel* panelp = mPanels.empty() ? nullptr : mPanels.back();
+		space_to_distribute += panelp ? ll_round((F32)mPanelSpacing * panelp->getVisibleAmount()) : 0;
+	}
```
nat-goodspeed, line 414: Simpler to express this as:

    if (! mPanels.empty())
    {
        space_to_distribute += ll_round(F32(mPanelSpacing) * mPanels.back()->getVisibleAmount());
    }

indra/llui/lllayoutstack.cpp
```diff
@@ -834,7 +836,7 @@ void LLLayoutStack::updatePanelRect( LLLayoutPanel* resized_panel, const LLRect&
 	LLLayoutPanel* other_resize_panel = NULL;
 	LLLayoutPanel* following_panel = NULL;
 
-	BOOST_REVERSE_FOREACH(LLLayoutPanel* panelp, mPanels)
+	BOOST_REVERSE_FOREACH(LLLayoutPanel* panelp, mPanels) // Should replace this when C++20 reverse view adaptor becomes available...
```
nat-goodspeed, line 839: It's tempting to respond to your comment by coding in terms of `rbegin()` and `rend()`, but I see your point: once we have the range library, it will be easier to replace `BOOST_REVERSE_FOREACH()` than the equivalent reverse-iterator logic.

indra/llui/lllayoutstack.cpp
```diff
@@ -974,6 +976,7 @@ void LLLayoutStack::updatePanelRect( LLLayoutPanel* resized_panel, const LLRect&
 												MIN_FRACTIONAL_SIZE,
 												MAX_FRACTIONAL_SIZE);
 			}
+			break;
```
nat-goodspeed, line 979: oh, nice catch.

indra/newview/llpanelgroupbulkban.cpp
```diff
@@ -163,27 +161,26 @@ void LLPanelGroupBulkBan::submit()
 	// remove already banned users and yourself from request.
 	std::vector<LLAvatarName> banned_avatar_names;
 	std::vector<LLAvatarName> out_of_limit_names;
-	bool banning_self = FALSE;
-	std::vector<LLUUID>::iterator conflict = std::find(banned_agent_list.begin(), banned_agent_list.end(), gAgent.getID());
+	bool banning_self{ false };
+	std::vector<LLUUID>::iterator conflict = std::find(banned_agent_list.begin(), banned_agent_list.end(), gAgentID);
```
nat-goodspeed, line 165: Interesting - is there ever a case in which `gAgentID` does not exactly track `gAgent`?

indra/llcommon/llsys.cpp
```diff
@@ -905,20 +904,19 @@ void LLMemoryInfo::stream(std::ostream& s) const
 
 	// Max key length
 	size_t key_width(0);
-	BOOST_FOREACH(const MapEntry& pair, inMap(mStatsMap))
+	for (const auto& [key, value] : inMap(mStatsMap))
```
nat-goodspeed, line 907: Nice! This and elsewhere.

nat-goodspeed (approved):
Thank you for this PR!

indra/llui/lllayoutstack.cpp
Ansariel: What are the plans for switching to C++20? I think there are some issues that would need to be fixed when I tried to use C++20.

indra/llui/lllayoutstack.cpp
Ansariel: Visual Studio found it 😄

indra/newview/llpanelgroupbulkban.cpp
Ansariel: No. I just replaced it because I stumbled over it and it saves a function call.

indra/llui/lllayoutstack.cpp
AiraYumi: Sorry.
Why is this process done in reverse?
Looking at the processing details, wouldn't it work even if it's not a reverse loop?

AiraYumi: > It's tempting to respond to your comment by coding in terms of `rbegin()` and `rend()`, but I see your point: once we have the range library, it will be easier to replace `BOOST_REVERSE_FOREACH()` than the equivalent reverse-iterator logic.

Sorry.
Why is this process done in reverse?
Looking at the processing details, wouldn't it work even if it's not a reverse loop?

marchcat-pe (approved):
Thank you @Ansariel!

indra/llui/lllayoutstack.cpp
nat-goodspeed: > What are the plans for switching to C++20?

I want to work on that, but transitioning away from Jira requires a whole lot of attention just now.

nat-goodspeed: > Why is this process done in reverse?
Looking at the processing details, wouldn't it work even if it's not a reverse loop?

I haven't studied the semantics in detail, but it seems to me that the assignment to `following_panel`, and its assignment to `other_resize_panel`, would be different with a forwards traversal.

PR #623, opened 2024-01-09
Build fix for Visual Studio update (std::vector<PermissionBit>)

LLGuru:


marchcat-pe (approved):
The previous code didn't fail the builds in Visual Studio either, but this one should be better anyway.

PR #624, opened 2024-01-09
SL-19555 Fix picker failing to highlight default material in inventory

akleshchev:
1. Fixed inventory search to look for a correct item
2. Updated constuctor to include inventory pick type from the start (needed for initial search to pick it up)

### DIFF
--- a/indra/newview/lltexturectrl.cpp
+++ b/indra/newview/lltexturectrl.cpp
@@ -151,7 +151,8 @@ LLFloaterTexturePicker::LLFloaterTexturePicker(
 	PermissionMask immediate_filter_perm_mask,
 	PermissionMask dnd_filter_perm_mask,
 	BOOL can_apply_immediately,
-	LLUIImagePtr fallback_image)
+	LLUIImagePtr fallback_image,
+    EPickInventoryType pick_type)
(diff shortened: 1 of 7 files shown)


indra/newview/lltexturectrl.cpp
```diff
@@ -786,27 +785,43 @@ void LLFloaterTexturePicker::draw()
...
-							cats,
-							items,
-							LLInventoryModel::INCLUDE_TRASH,
-							asset_id_matches);
+
+    if (loockup_id.isNull())
+    {
+        // looking for a material with a null id, null id is shared by a lot
+        // of objects as a default value, so have to filter by type as well
+        LLAssetIDAndTypeMatches matches(loockup_id, LLAssetType::AT_MATERIAL);
```
akleshchev, line 808: I'm not doing LLAssetIDAndTypeMatches matches(loockup_id, mInventoryPickType == PICK_MATERIAL 
? AT_MATERIAL : AT_TEXTURE);, because I'm not sure if a situation where we get an AT_TEXTURE_TGA or the like that is impossible so keeping texture type-independent.

cosmic-linden (approved):

PR #625, opened 2024-01-10
SL-20752 Mouselook no longer allows pitch upwards to full 90 degrees

LLGuru:


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71144

PR #626, opened 2024-01-10
SL-20244 Maint A - On-screen animesh characters that start pelvis offset animations disappear when root goes off-screen

LLGuru:


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/70593

PR #627, opened 2024-01-10
SL-20798 FIXED Crash at LLIMMgr::addMessage

maxim-productengine:


marchcat-pe (approved):

indra/newview/llimview.cpp
```diff
@@ -3160,7 +3160,7 @@ void LLIMMgr::addMessage(
 			}
 
             // Fetch group chat history, enabled by default.
-            if (gSavedPerAccountSettings.getBOOL("FetchGroupChatHistory"))
+            if (gSavedPerAccountSettings.getBOOL("FetchGroupChatHistory") && gAgent.getRegion())
```
akleshchev, line 3163: There is a getRegionCapability that does this check inside, also there should be a !chat_url.empty() check in case region has no caps or no this specific cap.

maxim-productengine: Thanks.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71194

PR #629, opened 2024-01-11
Rename OS X to macOS, mostly in comments

larsnaesbye:
Support is for 10.13+ now, and it's been called macOS since 10.12. References in code to specific older versions are unchanged.


larsnaesbye:
Perhaps it should go into [DRTVWR-489](https://github.com/secondlife/viewer/tree/DRTVWR-489) instead?

marchcat-pe (approved):
Thank you very much!

marchcat-pe:
Let's put this into Maint A.

PR #633, opened 2024-01-11
[macOS] remove obsolete check for system version

larsnaesbye:
operatingSystemVersion and NSOperatingSystemVersion exist since OS X 10.10, and we require macOS 10.13+ now


marchcat-pe (approved):
Thank you!

PR #634, opened 2024-01-11
[macOS] Locale problems prior to OS X 10.4 no longer apply

larsnaesbye:
localeconv() returns the right values in later versions, and we only support 10.13+


marchcat-pe (approved):
Thank you!

PR #635, opened 2024-01-11
SL-20629 Fix Probe ambiance being displayed as 0 when it isn't

akleshchev:
1. Slider displays 2 digits, so 0.001 value was effectively invisible. After discussion just increased the value
2. Hide notification if value is actually 0
3. Fix Ambience of '1' when opening floater for the first time (sAutoAdjustProbeAmbiance change).

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -10720,7 +10720,7 @@
     <key>Type</key>
     <string>F32</string>
     <key>Value</key>
-    <real>0.001</real>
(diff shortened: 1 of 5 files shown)


marchcat-pe (approved):

maxim-productengine (approved):

vladyslavasproductengine:
Passed QA.
Verified on the Second Life Release 7.1.3.7617509125 (64bit) on Win10/OSX in the scope of
IQA 86.

PR #636, opened 2024-01-12
SL-20799 FIXED Viewer crashes when trying to save snapshot on disc without empty space

maxim-productengine:


marchcat-pe (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/71195

PR #637, opened 2024-01-13
replaces parts of boost to C++ standard.

AiraYumi:
Some of the functions used in boost are provided by the C++ standard.
This PR replaces that functionality.

1. replace boost::fibers::promise to std::promise
2. replace boost::fibers::future to std::future
3. replace boost::fibers::mutex to std::mutex
4. replace boost::fibers::cv_status to std::cv_status
5. replace boost::fibers::condition_variable* to std::condition_variable*
6. replace boost::fibers::future_status to std::future_status
7. replace boost::fibers::timed_mutex to std::timed_mutex

### DIFF
--- a/indra/llcommon/llcond.h
+++ b/indra/llcommon/llcond.h
@@ -16,7 +16,6 @@
 
 #include "llunits.h"
 #include "llcoros.h"
-#include LLCOROS_MUTEX_HEADER
 #include "mutex.h"
 #include <chrono>
 
--- a/indra/llcommon/llcoros.h
+++ b/indra/llcommon/llcoros.h
@@ -41,10 +41,6 @@
 #include <exception>
 #include <queue>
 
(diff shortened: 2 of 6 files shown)


marchcat-pe (approved):
Thank you!

marchcat-pe:
Sadly, I have to roll this back for now, since our mac builds don't support `std::condition_variable_any` yet.

nat-goodspeed (commented):
No, please DO NOT make this change. You will slow down the viewer.

`boost::fibers::promise` (et al.) are aware of the Boost.Fibers "coroutines" we use within the main thread. When the running coroutine blocks on `boost::fibers::future`, it only blocks _that coroutine:_ control is passed to another ready coroutine.

`std::promise` (et al.) are _not_ aware of viewer coroutines. Blocking on `std::future` blocks the entire main thread and every pending coroutine.

PR #640, opened 2024-01-15
SL-20795 Part of previously typed emojis disappear in the 'Save settings as a preset...' option of the 'Preferences' floater

LLGuru:


indra/llui/llcombobox.cpp
```diff
@@ -548,7 +548,7 @@ void LLComboBox::createLineEditor(const LLComboBox::Params& p)
 		LLLineEditor::Params params = p.combo_editor;
 		params.rect(text_entry_rect);
 		params.default_text(LLStringUtil::null);
-		params.max_length.bytes(mMaxChars);
+		params.max_length.chars(mMaxChars);
```
akleshchev, line 551: This is likely risky and will cause unintended bugs, some viewer messages have strict byte limit server side, so if we set a longer string and put in some unicode it will just be cut. Please, don't do this unless you want to check every use case individually.

akleshchev (changes requested):

indra/llui/llcombobox.cpp
marchcat-pe: Right, using the maximum size in bytes is correct here, but I'm still curious about why this change was made.

akleshchev: Graphical presets have a very short name length limit. Arguably those should be extended to at least ~50 chars (shouldn't be simulator related so it should be fine to do).

marchcat-pe (approved):

akleshchev (approved):

indra/llrender/llfontfreetype.cpp
```diff
@@ -665,7 +665,15 @@ void LLFontFreetype::renderGlyph(EFontGlyphType bitmap_type, U32 glyph_index) co
 		load_flags |= FT_LOAD_COLOR;
 	}
 
-	llassert_always(! FT_Load_Glyph(mFTFace, glyph_index, load_flags) );
+	FT_Error error = FT_Load_Glyph(mFTFace, glyph_index, load_flags);
+	if (FT_Err_Ok != error)
+	{
```
akleshchev, line 670: I suggest logging error here

PR #649, opened 2024-01-17
Put 2k textures behind a feature flag

RunitaiLinden:
https://github.com/secondlife/viewer-issues/issues/21

### DIFF
--- a/indra/newview/llmaterialeditor.cpp
+++ b/indra/newview/llmaterialeditor.cpp
@@ -1872,7 +1872,7 @@ static void pack_textures(
     if (normal_img)
     {
         // create a losslessly compressed version of the normal map
-        normal_j2c = LLViewerTextureList::convertToUploadFile(normal_img, 1024, false, true);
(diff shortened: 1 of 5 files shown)


indra/newview/llmaterialeditor.cpp
```diff
@@ -1872,7 +1872,7 @@ static void pack_textures(
     if (normal_img)
     {
         // create a losslessly compressed version of the normal map
-        normal_j2c = LLViewerTextureList::convertToUploadFile(normal_img, 1024, false, true);
+        normal_j2c = LLViewerTextureList::convertToUploadFile(normal_img, 2048, false, true);
```
cosmic-linden, line 1875: Can probably just use `MAX_IMAGE_SIZE_DEFAULT` here.

cosmic-linden (approved):
I notice addTextureStats is sometimes called with hardcoded 1024*1024 which sounds related. If so, not a blocker. LGTM.

PR #656, opened 2024-01-19
 SL-20606: PBR swatches in build floater should show entire material

cosmic-linden:


cosmic-linden:
Closed in favor of rebase on release/materials_featurette

PR #663, opened 2024-01-23
Create test plan for optimizeAwayAlpha

RunitaiLinden:


Dan-Linden (commented):
How does the tester verify the face does not render in the alpha pass?

RunitaiLinden:
> How does the tester verify the face does not render in the alpha pass?

Added text clarifying that face should stay visible after unchecking Advanced->Render Types->Alpha

Dan-Linden (approved):

PR #664, opened 2024-01-24
SL-20416 Fix Crash Report 1409376 (update)

LLGuru:


akleshchev (approved):

indra/llui/llfloater.cpp
```diff
@@ -506,6 +506,7 @@ void LLFloater::enableResizeCtrls(bool enable, bool width, bool height)
 
 void LLFloater::destroy()
 {
+	gFloaterView->onDestroyFloater(this);
```
akleshchev, line 509: Might need to check getParent() == gFloaterView first

PR #691, opened 2024-01-29
Remove unused buildscripts files

bennettgoble:
BuildParams, build.sh and its support functions are no longer used.


bennettgoble:
Whoops, evidently this is still used. Gross.

PR #697, opened 2024-01-29
Fix for a potential crash in LLReflectionMapManager::registerSpatialGroup()

sldevel:
The spatial partition could potentially be NULL and shall therefore been tested for this case.

Similar fix to 08cf926d3b6eb28e0b9751ba62b1ce01230a150b


akleshchev (approved):

akleshchev:
Just in case, do you have a full callstack? I haven't found registerSpatialGroup in our crash tracking

marchcat:
Thank you for this contribution!

PR #702, opened 2024-01-30
build.yaml: Build shared branches

bennettgoble:
We're currently building every single commit pushed to Github. This is racking up $20k in build charges a month and is generally excessive.

This changeset alters build triggers so that builds automatically run if they are committed to a **shared branch**:

- `release/*` - A release stabilization branch
- `project/*` - A project viewer branch
- `main/*` - The default/stable branch

These branches are also being set up to require pull requests, which are also automatically built.

...need to build another commit? Developers can trigger one using a manual workflow run.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,8 +4,8 @@ on:
   workflow_dispatch:
   pull_request:
   push:
-    branches: ["*"]
-    tags: ["*"]
+    branches: ["main", "release/*", "project/*"]
+    tags: ["Second_Life_*"]
 
 jobs:
   build:


.github/workflows/build.yaml
```diff
@@ -4,8 +4,8 @@ on:
   workflow_dispatch:
   pull_request:
   push:
-    branches: ["*"]
-    tags: ["*"]
+    branches: ["main", "release/*", "project/*"]
+    tags: ["v*"]
```
nat-goodspeed, line 8: I'm not so sure about this tag filter. Our `release` job runs [when the tag starts with `"Second_Life_"`](https://github.com/secondlife/viewer/blob/main/.github/workflows/build.yaml#L329). I think this change would keep such tag builds from running at all, preventing us from getting viewer release builds.

I don't mind the `branches` restriction.

nat-goodspeed (changes requested):

.github/workflows/build.yaml
bennettgoble: Alright, I've opened the tag trigger back up to `*`.

nat-goodspeed: Thanks. I don't think we'll get a lot of spurious tag builds anyway. I suppose we could say

    tags: ["Second_Life_*"]

if we want to suppress tag builds for the (e.g.) "7.1.2-release" tags.

nat-goodspeed (approved):

.github/workflows/build.yaml
bennettgoble: Done

nat-goodspeed (approved):

PR #708, opened 2024-01-31
SL-19986 Ability to drag Favorites to the Overflow menu

LLGuru:


scripts/messages/message_template.msg.sha1
```diff
@@ -1 +1 @@
-d7915d67467e59287857630bd89bf9529d065198
+d7915d67467e59287857630bd89bf9529d065199
```
cosmic-linden, line 1: Why did the message template change?

indra/newview/llpanellandmarks.cpp
```diff
@@ -1047,6 +1047,85 @@ bool LLLandmarksPanel::handleDragAndDropToTrash(BOOL drop, EDragAndDropType carg
...
+            item->setComplete(TRUE);
+            item->updateServer(FALSE);
+
+            gInventory.updateItem(item);
+            gInventory.notifyObservers();
+        }
+
+        LLView::getWindow()->setCursor(UI_CURSOR_ARROW);
+    }
+};
```
akleshchev, line 1070: This looks incomplete, it's a landmark being copied into favorites' overflow, why isn't it calling setSortIndex to make sure position gets updated and saved?

akleshchev: I also don't see any code for a position indicator when hovering over the overflow menu (like it was done for the main favorites panel)

akleshchev:
Since jira is now gone, please link development issues in the menu to the right.

akleshchev (changes requested):
The whole point of the request was an ability to reorder favorites, what's the point if you aren't preserving new order acros sessions?

scripts/messages/message_template.msg.sha1
akleshchev: I think this one gets pulled automatically from server in some cases by TemplateCheck.cmake.

LLGuru: It was changed automatically during the compilation. Removed from PR, thanks

PR #719, opened 2024-02-01
GH-141872 Ability to drag Favorites to the Landmarks panel

LLGuru:
Landmak can be copied by a DragAndDrop operation if only it was already fetched from the server

### DIFF
--- a/indra/newview/lltooldraganddrop.cpp
+++ b/indra/newview/lltooldraganddrop.cpp
@@ -47,6 +47,7 @@
 #include "llinventorybridge.h"
 #include "llinventorydefines.h"
 #include "llinventoryfunctions.h"
+#include "llinventorymodelbackgroundfetch.h"
 #include "llpreviewnotecard.h"
 #include "llrootview.h"
 #include "llselectmgr.h"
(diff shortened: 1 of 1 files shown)


akleshchev (changes requested):
This looks wrong
1. Either it should be in LLFavoritesBarCtrl or it shouldn't be landmark specific
2. Item should be fetched when clicked LLFavoritesBarCtrl::onButtonMouseDown like everything else does it (when interacted, selected e t c, click counts), instead of being specific to drag and drop
3. source == SOURCE_LIBRARY check makes no sense, it is very unlikely that users will ever use library landmarks and your fix will basically not work at for normal inventory

LLGuru:
> Either it should be in LLFavoritesBarCtrl 
The issue is reproduced even when dragging from Inventory floater

LLGuru:
> 1. it shouldn't be landmark specific

> 3. source == SOURCE_LIBRARY check makes no sense

Fixed

akleshchev (approved):

indra/newview/lltooldraganddrop.cpp
```diff
@@ -320,6 +321,16 @@ void LLToolDragAndDrop::beginDrag(EDragAndDropType type,
 		LL_WARNS() << "Attempted to start drag without a cargo type" << LL_ENDL;
 		return;
 	}
+
+    if (type != DAD_CATEGORY)
+    {
+        LLViewerInventoryItem* item = dynamic_cast<LLViewerInventoryItem*>(gInventory.getObject(cargo_id));
```
akleshchev, line 327: the cast is not needed, just call getItem directly

marchcat-pe (approved):

indra/newview/lltooldraganddrop.cpp
LLGuru: The getItem returns LLInventoryItem*, may I be sure that this is exactly LLViewerInventoryItem* ?

akleshchev: 1. check what LLInventoryModel::getObject(uuid) does, it calls getCategory first, then getItem, but since you are making sure this isn't DAD_CATEGORY first half of getObejct is redundant, so it's better to just call getItem directly.
2. LLInventoryObject->LLInventoryItem->LLViewerInventoryItem in this case, you can use either of the three, since base class's LLInventoryObject::getUUID() will be used.

PR #725, opened 2024-02-02
Issue #26 Support texture drop onto PBR materials

akleshchev:
1. Dropped textures should override base color
2. Only allow copy&transfer permitted textures for overrides
3. 'Alt' should replace PBR with a texture

### DIFF
--- a/indra/newview/lltooldraganddrop.cpp
+++ b/indra/newview/lltooldraganddrop.cpp
@@ -1071,10 +1071,64 @@ BOOL LLToolDragAndDrop::handleDropMaterialProtections(LLViewerObject* hit_obj,
 	return TRUE;
 }
 
+void set_texture_to_material(LLViewerObject* hit_obj,
+                             S32 hit_face,
+                             const LLUUID& asset_id,
+                             LLGLTFMaterial::TextureInfo drop_channel)
+{
+    LLTextureEntry* te = hit_obj->getTE(hit_face);
+    if (te)
+    {
+        LLPointer<LLGLTFMaterial> material = te->getGLTFMaterialOverride();
+
+        // make a copy to not invalidate existing
+        // material for multiple objects
+        if (material.isNull())
+        {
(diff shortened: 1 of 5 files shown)


indra/newview/lltooldraganddrop.h
```diff
@@ -249,17 +249,20 @@ class LLToolDragAndDrop : public LLTool, public LLSingleton<LLToolDragAndDrop>
                             ESource source,
                             const LLUUID& src_id,
                             bool all_faces,
+                            bool replace_pbr,
```
cosmic-linden, line 252: Looks like it should be `remove_pbr` for consistency

indra/newview/lltooldraganddrop.cpp
```diff
@@ -1071,10 +1071,64 @@ BOOL LLToolDragAndDrop::handleDropMaterialProtections(LLViewerObject* hit_obj,
...
+                break;
+
+            case LLGLTFMaterial::GLTF_TEXTURE_INFO_NORMAL:
+                {
+                    material->setNormalId(asset_id);
+                }
+                break;
+        }
+        LLGLTFMaterialList::queueModify(hit_obj, hit_face, material);
+    }
```
cosmic-linden, line 1124: `LLGLTFMaterial::setTextureId` could work well here.

cosmic-linden (approved):
No blockers. Just minor style suggestions.

indra/newview/lltooldraganddrop.h
akleshchev: Agree, missed this, but not worth a separate commit.

indra/newview/lltooldraganddrop.cpp
akleshchev: Agree, will remake it if there will be any more tasks, but since it's already merged I don't think it's worth a seaprate commit.

PR #726, opened 2024-02-05
#68175 [SL-17986] Ability to drag Favorites to the Overflow menu

LLGuru:


indra/llui/lluictrl.cpp
```diff
@@ -413,13 +416,26 @@ BOOL LLUICtrl::handleDoubleClick(S32 x, S32 y, MASK mask)
 	return handled;
 }
 
+// virtual
+BOOL LLUICtrl::handleDragAndDrop(S32 x, S32 y, MASK mask, BOOL drop, EDragAndDropType cargo_type,
+    void* cargo_data, EAcceptance* accept, std::string& tooltip_msg)
+{
+    if (mDragAndDropSignal)
```
akleshchev, line 423: This feel out of order, with this implementation if a tab handles drag an drop this way any child (like inventory in said tab) would not get the drop signal. I think it should be like

res = LLView::handleDragAndDrop(...);
if (!res && mDragAndDropSignal)
{
   return (*mDragAndDropSignal)(...)
}

akleshchev (changes requested):

indra/newview/llfavoritesbar.cpp
```diff
@@ -285,22 +286,42 @@ class LLFavoriteLandmarkMenuItem : public LLMenuItemCallGL
...
+    BOOL handleHover(S32 x, S32 y, MASK mask) override
+    {
+        mIsHovering = true;
+        LLToggleableMenu::handleHover(x, y, mask);
+        mIsHovering = false;
+        return TRUE;
+    }
+
+    // virtual
+    void setVisible(BOOL visible) override
```
akleshchev, line 307: feels a bit unsafe to me, might be a good idea to add LLToggleableMenu::setVisible(FALSE) in destructor

akleshchev, line 307: Please consider adding a mIsHovering = false; setVisible(false); on destruction just in case someone subsribes t visibility changes.

indra/llui/lluictrl.cpp
LLGuru: I believe this order allows the object to stop the inherited processing if needed.
In case when the control wants to continue inherited processing it returns false.

If we call inherited processing first, the control can't stop this processing.
So, I still mean the proposed order to be right:
specific processing goes first and controls the execution of default processing.

akleshchev: By default it is normal practice to let children handle drag an drop, because they are in the front and that's what LLView::handleDragAndDrop does.

Children are in the front they must have first dibs by default.

akleshchev (approved):

PR #739, opened 2024-02-06
SL-19585 Replace FMOD with OpenAL

akleshchev:
- Fixed distorted wind by switching it to F32
- Replaced fmod with OpenAl

According to request tried replacing fmod with openal, most of the media gets handled by VLC so only wav, wind and button sounds get handled by the engine. I don't think there are any disadvantages from switching, but QA will have to make sure 3d sounds generated by objects work properly.


maxim-productengine (approved):

Ansariel:
We should rather ditch the old broken OpenAL crap which was last updated in 2009! FMOD supports things like out output device selection and stream title extraction - OpenAL supports... nothing

akleshchev:
> We should rather ditch the old broken OpenAL crap which was last updated in 2009! FMOD supports things like out output device selection and stream title extraction - OpenAL supports... nothing

Viewer is using OpenAL Soft with couple extensions not OpenAL and as far as I can tell everything viewer uses from fmod this implementation is also capable of.

Ansariel:
Still a whole lotta junk!

Only because the LL viewer doesn't have output device selection or doesn't allow display of stream titles currently playing on the parcel audio stream, it doesn't mean other viewers are using it.

Anyway: If this gets merged, I'll revert it instantly for Firestorm.

akleshchev:
>parcel audio stream
That plays through VLC at the momkent, if I remeber correctly. It was switched due to fmod having a number of unresolved issues with streams and complaints. fmod is a 3d sound engine, not a media player so fmod developers focused on the engine part.

>If this gets merged, I'll revert it instantly for Firestorm.
Feel free to do that. It's just that for an organisation like LL fmod has couple additional disadvantages: can't distribute it as a package, must pay for it, potential legal complications, so the task is essentialy to see if openal has any practical disadvantages over openal. Worst case scenario, openal gets a fancy update for ReleaseOS builds.

Ansariel:
> > parcel audio stream
> 
> That plays through VLC at the moment. If I remeber correctly, it was switched due to fmod having a number of unresolved issues with streams and complaints. fmod is a 3d sound engine, not a media player so fmod developers focused on the engine part.
Guess the streaming audio tags extracted in LLStreamingAudio_FMODSTUDIO get pulled out of thin air then... 🤷

marchcat-pe (approved):

akleshchev:
>Guess the streaming audio tags extracted in LLStreamingAudio_FMODSTUDIO get pulled out of thin air then...

Apparently LLStreamingAudio_MediaPlugins() was implemented but not turned on by default... I saw it working, so I assumed it was the default, but viewer defaulted to that due to fmod being turned off... Nothing I can do about. QA will have to test it as well and make there is no loss of functionality.

akleshchev:
This might still get reverted if QA finds any issues or it turns out other projects have to keep the license.

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/69923

PR #744, opened 2024-02-06
Remove old unused Mac artifacts

larsnaesbye:
The resource descriptor macview.r didn't contribute to menu handling, macview_Prefix.h wasn't used either and two others can be removed. They seem to relate to very old (unsupported) versions.


brad-linden:
I'm curious about this.  I also would like to revisit this at some point and get rid of this cruft.  thanks for looking at this

PR #745, opened 2024-02-06
SL-19730 update menu items

maxim-productengine:
When there are multiple objects selected, 4 labels are:
- Take as combined item
- Take copy as combined item
- Take as separate items
- Take copies as separate items

If there is only one item selected, "Take" and "Take copy" remain as before.

### DIFF
--- a/indra/newview/llviewermenu.cpp
+++ b/indra/newview/llviewermenu.cpp
@@ -5249,9 +5249,19 @@ bool visible_take_object()
 	return !is_selection_buy_not_take() && enable_take();
 }
 
-bool visible_take_objects() 
+bool is_multiple_selection() 
 {
-    return visible_take_object() && (LLSelectMgr::getInstance()->getSelection()->getRootObjectCount() > 1);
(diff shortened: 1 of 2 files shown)


marchcat-pe (approved):

indra/newview/llviewermenu.cpp
```diff
@@ -5249,9 +5249,19 @@ bool visible_take_object()
 	return !is_selection_buy_not_take() && enable_take();
 }
 
-bool visible_take_objects() 
+bool is_multiple_selection() 
 {
-    return visible_take_object() && (LLSelectMgr::getInstance()->getSelection()->getRootObjectCount() > 1);
+    return (LLSelectMgr::getInstance()->getSelection()->getRootObjectCount() > 1);
```
akleshchev, line 5254: I'm not sure this is a correct indicator of a multi selection, it is posiible to select a face on a root object and a face on child of other object. You might want to disable that if 'individual' selection or 'face' selection is on.

akleshchev (approved):

PR #755, opened 2024-02-07
Confusion between light diffuse and light color in HDR auto-adjust mode

sldevel:
While the SUNLIGHT_COLOR uniform is initialized with getSunlightColor() when not
auto-adjusting for HDR, it is overwritten with getSunDiffuse()*auto_adjust_factor
when auto-adjusting, causing very bad lighting discrepancies in the latter mode
(especially when not at midday or midnight).

Let's use getSunlightColor() everywhere, shall we ?...


RunitaiLinden:
Fix looks good, thanks much.  Retargeting to release/materials_featurette as that's the current graphics work branch.

RunitaiLinden (approved):

PR #758, opened 2024-02-07
SL-18721 Shutdown fixes #4

akleshchev:
Missed that 'kickWindowThread' requires handle to still be alive, also it should happen after closing queue


marchcat-pe (approved):

maxim-productengine (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/68974

PR #761, opened 2024-02-08
BOOL to bool, pt. 1

larsnaesbye:
The first in (hopefully) a long series of PRs seeing `BOOL/TRUE/FALSE` (int) replaced by `bool/true/false` to make it easier for the compiler to optimize code. This should impact performance and memory use somewhat. Some 3rd party viewers have already done this replacement.

In addition to this, in the long run it will make it easier to improve the macOS Objective-C ([which defines BOOL also](https://developer.apple.com/documentation/objectivec/bool)) bridging code.

143 occurrences of `BOOL` and their `FALSE` and `TRUE` companions were converted in `llcommon`. This leaves 13729(!) to fix, though this number includes comments.

If it gets accepted, I will make a much larger next PR. But of course this needs to be done in parts.

Note: I have only been able to test on macOS and x86, so I have not been able to test #ifdef'd Windows code.

### DIFF
--- a/indra/llcommon/llnametable.h
+++ b/indra/llcommon/llnametable.h
@@ -55,16 +55,16 @@ public:
 		mNameMap[tablename] = data;
 	}
 
-	BOOL checkName(const std::string& name) const
+	bool checkName(const std::string& name) const
 	{
 		return checkName(name.c_str());
 	}
 
 	// "logically const" even though it modifies the global nametable
-	BOOL checkName(const char *name) const
+	bool checkName(const char *name) const
 	{
 		char *tablename = gStringTable.addString(name);
-		return mNameMap.count(tablename) ? TRUE : FALSE;
+		return mNameMap.count(tablename) ? true : false;
 	}
 
 	DATA resolveName(const std::string& name) const
--- a/indra/llcommon/llpriqueuemap.h
+++ b/indra/llcommon/llpriqueuemap.h
@@ -47,17 +47,17 @@ public:
 	{
 		if (mPriority > b.mPriority)
 		{
-			return TRUE;
+			return true;
 		}
 		if (mPriority < b.mPriority)
 		{
-			return FALSE;
+			return false;
 		}
 		if (mData > b.mData)
 		{
-			return TRUE;
+			return true;
 		}
-		return FALSE;
+		return false;
 	}
 
 	F32 mPriority;
@@ -90,18 +90,18 @@ public:
 		mMap.insert(pqm_pair(LLPQMKey<DATA_TYPE>(priority, data), data));
 	}
 
-	BOOL pop(DATA_TYPE *datap)
+	bool pop(DATA_TYPE *datap)
 	{
 		pqm_iter iter;
 		iter = mMap.begin();
 		if (iter == mMap.end())
 		{
-			return FALSE;
+			return false;
 		}
 		*datap = (*(iter)).second;
 		mMap.erase(iter);
 
-		return TRUE;
+		return true;
 	}
 
 	void reprioritize(const F32 new_priority, DATA_TYPE data)
--- a/indra/llcommon/llsys.cpp
+++ b/indra/llcommon/llsys.cpp
@@ -214,7 +214,7 @@ LLOSInfo::LLOSInfo() :
 		DWORD cbData(sizeof(DWORD));
 		DWORD data(0);
 		HKEY key;
-		BOOL ret_code = RegOpenKeyExW(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion"), 0, KEY_READ, &key);
+        LSTATUS ret_code = RegOpenKeyExW(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion"), 0, KEY_READ, &key);
 		if (ERROR_SUCCESS == ret_code)
 		{
 			ret_code = RegQueryValueExW(key, L"UBR", 0, NULL, reinterpret_cast<LPBYTE>(&data), &cbData);
@@ -571,7 +571,7 @@ bool LLOSInfo::is64Bit()
     return true;
 #elif defined(_WIN32)
     // 32-bit viewer may be run on both 32-bit and 64-bit Windows, need to elaborate
-    BOOL f64 = FALSE;
+    bool f64 = false;
     return IsWow64Process(GetCurrentProcess(), &f64) && f64;
 #else
     return false;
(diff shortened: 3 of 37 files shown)


sldevel:
I can only approve and encourage this (lengthy and boring) BOOL eradication process that I achieved 10 years ago already for the Cool VL Viewer.

The gain in the member variables size can be **massive** (e.g. for inventory size in memory, or for UI memory consumption), on the condition that this process is also followed by a member variables reshuffling for all classes using BOOL, so to get rid of all padding bytes. Once the full optimization process completed, you also can notice measurable speed gains, thanks to a better usage of CPU data caches (less padding bytes = more cached member variables).

larsnaesbye:
Thank you for the kind words, @sldevel !
I of course took this upon myself after [our forum discussion](https://community.secondlife.com/forums/topic/506017-macos-arm64-build-for-running-natively-on-apple-silicon-macs/).

RunitaiLinden:
Definitely a fan of this kind of contribution.  We need to figure out a good branching strategy, as we don't typically merge code to main until it goes through a full release cycle, so PRs going to main have to be retargeted elsewhere.  I know for the time being render work should target release/materials_featurette, but I'm not sure what branch general decrufting should target.  Will consult with the team and let you know.

larsnaesbye:
Well, egotistically speaking, [DRTVWR-600-maint-A](https://github.com/secondlife/viewer/tree/DRTVWR-600-maint-A) has some changes not yet merged to `main` that make it easier to compile for me.

indra/llcommon/StackWalker.h
```diff
@@ -103,7 +103,7 @@ class StackWalker
   StackWalker(DWORD dwProcessId, HANDLE hProcess);
   virtual ~StackWalker();
 
-  typedef BOOL (__stdcall *PReadProcessMemoryRoutine)(
```
brad-linden, line 106: I'm pretty sure these cannot be changed, since they are not our BOOL, but rather the WinAPI BOOL from windows.h so changing these would be an ABI break.

this is another reason to move our code away from using the LL BOOL definition, but we need to be extremely careful to separate these out properly

brad-linden (changes requested):

indra/llcommon/StackWalker.h
larsnaesbye: Good point. I will remove the StackWalker files from the commit.

larsnaesbye: Done :)

indra/llcommon/llnametable.h
```diff
@@ -55,16 +55,16 @@ class LLNameTable
 		mNameMap[tablename] = data;
 	}
 
-	BOOL checkName(const std::string& name) const
+    bool checkName(const std::string& name) const
```
brad-linden, line 58: something happened to the indentation here

indra/llcommon/llpriqueuemap.h
```diff
@@ -90,18 +90,18 @@ class LLPriQueueMap
 		mMap.insert(pqm_pair(LLPQMKey<DATA_TYPE>(priority, data), data));
 	}
 
-	BOOL pop(DATA_TYPE *datap)
+    bool pop(DATA_TYPE *datap)
```
brad-linden, line 93: something happened to the indentation here

indra/llcommon/llsys.cpp
```diff
@@ -214,7 +214,7 @@ LLOSInfo::LLOSInfo() :
 		DWORD cbData(sizeof(DWORD));
 		DWORD data(0);
 		HKEY key;
-		BOOL ret_code = RegOpenKeyExW(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion"), 0, KEY_READ, &key);
+        bool ret_code = RegOpenKeyExW(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion"), 0, KEY_READ, &key);
```
brad-linden, line 217: this should stay as BOOL if it is being compared against WinAPI `ERROR_SUCCESS`

indra/llcommon/lluuid.h
```diff
@@ -76,8 +76,8 @@ class LL_COMMON_API LLUUID
 	//
 	// ACCESSORS
 	//
-	BOOL	isNull() const;			// Faster than comparing to LLUUID::null.
-	BOOL	notNull() const;		// Faster than comparing to LLUUID::null.
+	bool	isNull() const;			// Faster than comparing to LLUUID::null.
+    bool	notNull() const;		// Faster than comparing to LLUUID::null.
```
brad-linden, line 80: something happened to the indentation here

brad-linden (changes requested):
found one last spot that needs to go back to `BOOL` other than that looks good.

optionally, the changes in LLApp that don't refer to function declarations or function pointer typedefs could also be restored if you were so inclined

brad-linden:
also, I conferred with the team and `DRTVWR-600-maint-A` is the proper destination for this PR

larsnaesbye:
> also, I conferred with the team and `DRTVWR-600-maint-A` is the proper destination for this PR

Cool! I'll fix both the wrong conversion and indentation mishaps, and then squash into one commit. 
Perhaps if the latest commits from `main` could get merged into `DRTVWR-600-maint-A`, it'd be easy to just change the base?

indra/llcommon/llsys.cpp
Ansariel: It should not stay BOOL, but get changed to LSTATUS

Ansariel:
Way too much stuff got reverted now. Only the typedefs for the Windows API function and their results are affected (and in some cases using the wrong return type anyway)

indra/llcommon/llnametable.h
larsnaesbye: Yeah, a tab became spaces. Tying here to find the git setting that makes this an actual change, because it seems to ignore whitespace changes....

indra/llcommon/llsys.cpp
larsnaesbye: Changed it to LSTATUS.

larsnaesbye:
> Way too much stuff got reverted now. Only the typedefs for the Windows API function and their results are affected (and in some cases using the wrong return type anyway)

Yeah, I know. I took the 'let me get started and add more approach' and so when I hit those problems I figured I'll take them on in the next patch. It's a learning process for me.

indra/llcommon/llnametable.h
larsnaesbye: Indentation should now be fixed.

Ansariel:
I fixed all the remaining stuff: #769 😄

larsnaesbye:
Nice teamwork! :D

marchcat (approved):

brad-linden (approved):
looks good!

PR #769, opened 2024-02-08
BOOL to bool, pt. 1.1

Ansariel:
Part 1.1 of the BOOL replacement series: 😄 

Replace remaining BOOL with bool in llcommon that can be replaced and were left out/reverted in #761. Also remove dead Windows-specific code.


marchcat (approved):
Thank you!

PR #790, opened 2024-02-11
Fixing the 'lightning' -> 'lighting' typo

PanteraPolnocy:
Unless, of course, we have some dedicated lightning effects - then I will stand corrected. ~


marchcat (approved):
Thanks!

PR #793, opened 2024-02-12
#779 Emoji picker is an unintuitive UX disaster

LLGuru:


marchcat (approved):

akleshchev (approved):
Shouldn't these options check mHoveredIcon like left/right do? And is it correct that both function use '> mFocusedIconCol' consition, I would have expected one of them to use '< mFocusedIconCol'

PR #795, opened 2024-02-12
triage#59 Show 'Friends Only' rendering mode

akleshchev:
Always impostor non-friends in 'friends only' mode.

Changes affect 3 floaters dropbox with options: limit by complexity|always show friends|only show friends. In case of graphical settings setting it to 'friends only' disables 'complexity' and 'impostors' sliders.
image
image
image

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -8817,17 +8817,6 @@
         <key>Value</key>
             <integer>1</integer>
         </map>
-  <key>AlwaysRenderFriends</key>
-    <map>
-      <key>Comment</key>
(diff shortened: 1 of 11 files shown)


maxim-productengine:
Should we also disable tuning of non-impostors count in `StatsRecorder::updateAvatarParams()` (see _tunables.updateNonImposters_) for new mode?

akleshchev:
> Should we also disable tuning of non-impostors count in StatsRecorder::updateAvatarParams()

Will check that, thank you.

maxim-productengine (approved):

marchcat (approved):

PR #798, opened 2024-02-13
Add help() function to Lua "builtins."

nat-goodspeed:
`help()` with no argument lists all our viewer builtins.

`help(function, function, ...)` shows help text for each named function. Each
argument can be either a string or the function in question (e.g. `help(help)`).

To support Lua-related text containing line breaks, make `LLTextEditor::pasteTextWithLinebreaks()`
a public template method. Change the existing
implementation, which specifically accepts `(const LLWString&)`, into its
`LLWString` specialization. The generic template passes `llconvert(arg)` to that
specialization, the one real implementation.

Make `LLFloaterLUADebug` methods call `pasteTextWithLinebreaks()` instead of
`insertText()`, which ignores newline characters.

To allow `help()` to accept an actual function as well as a string name, add a
lookup-by-function-pointer map to `LuaFunction`. (A Lua function does not store
a name.) Make the constructor store an entry in the new lookup map as well as
in the original registry map.

Change `LuaFunction::getRegistry()` and `getRegistered()` to `getState()` and
`getRState()`, respectively. Each returns a `std::pair`, but the first binds
non-`const` references while the second binds `const` references.

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -1669,11 +1669,11 @@
               <key>creds</key>
               <string>github</string>
               <key>hash</key>
-              <string>c326445818c4e2e2177d5efde1dd59e7af0d6c9a</string>
+              <string>59bf3d96f9df4b6981c406abac5c46ae276f9b15</string>
               <key>hash_algorithm</key>
               <string>sha1</string>
               <key>url</key>
-              <string>https://api.github.com/repos/secondlife/3p-luau/releases/assets/150694132</string>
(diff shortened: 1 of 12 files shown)


nat-goodspeed:
Partially addresses secondlife/viewer-issues#16.

maxim-productengine (approved):

nat-goodspeed:
Maxim, I also added `leaphelp()`.

nat-goodspeed:
Fixes https://github.com/secondlife/viewer-issues/issues/16.

maxim-productengine (approved):

PR #802, opened 2024-02-14
Viewer#779 Make RenderMaxVRAMBudget more consistent

akleshchev:


cosmic-linden (approved):
A nice addition, however I will note that there does not currently appear to be a UI for this.

Ansariel:
But this is just a first step to solve all the inconsistencies. Open issues remaining (most likely not a complete list):

1. Why are the VRAM reserves for other processes different depending on what method you use?
2. Why does Windows need a reserve, macOS does not - but only in one method - in the other they both get a reserve?
3. Why does one method take VRAM usage for Vertex Buffers into account when estimating how much the viewer is using, the other one not?
4. Why do different methods exist to calculate VRAM usage exist in the first place?

akleshchev:
> A nice addition, however I will note that there does not currently appear to be a UI for this.

People aparently do use this and among listed issues this was the smallest and easiest to handle so I decided to knock it out first and get some idea of how things work in the process.

indra/newview/app_settings/settings.xml
```diff
@@ -9376,7 +9376,7 @@
   <key>RenderMaxVRAMBudget</key>
   <map>
     <key>Comment</key>
-    <string>Maximum amount of texture memory to budget for (in MB), or 0 for autodetect.  Requires restart.</string>
+    <string>Maximum amount of texture memory to budget for (in MB), or 0 for autodetect. OS Windows Specific.</string>
```
Ansariel, line 9379: This is not correct. For the actual calculation where it really matters, it is the same for all OS. It is only OS specific for information purposes (lag meter floater and texture view console), where fake values are shown since that calculation is different.

Ansariel:
From my perspective, there are only 2 ways to handle this:

1. Get rid of getAvailableVRAMMegabytes entirely and base the calculation entirely on estimations in LLViewerTexture as it is currently done for calculation the discard bias, since getAvailableVRAMMegabytes is currently nothing but show.
2. Use an OS specific implementation to determine the available memory. However, on Windows updateVRAMUsage (the method that actually calculates the available memory) is only called once every 3 seconds - which is a long time. And for macOS there isn't even a way to get that number, so it would purely based on estimations still.

Another way would be to use vendor-specific OpenGL extensions (the nvidia works, AMD might work (again) or not, Intel has nothing), so that would be a hit and miss as well.

akleshchev:
> since getAvailableVRAMMegabytes is currently nothing but show

That's diagnostic info (and wroten by davep) potentially more accurate at least for windows and I feel like it's a data point that's best left in UI, If anything UI should be updated to show something like 'estimated memory/OS reported memory'. For UI once in 3 seconds update should be fine.

Ansariel:
> > since getAvailableVRAMMegabytes is currently nothing but show
> 
> That's diagnostic info (and wroten by davep) potentially more accurate at least for windows and I feel like it's a data point that's best left in UI, If anything UI should be updated to show something like 'estimated memory/OS reported memory'. For UI once in 3 seconds update should be fine.

It makes 0 sense if it calculated differently. If I want to know how much VRAM is still available on the GPU, I don't need any reserves in that calculation. And if they are in there, they reflect nothing the viewer is actually doing when it comes to allocating memory for textures. It's plain simple nonsense in it's current form and shape.

cosmic-linden:
> People aparently do use this (likely from advanced->debug settings) and among listed issues this was the smallest and easiest to handle so I decided to knock it out first and get some idea of how things work in the process.

Thank you for clarifying intent. I do think it's reasonable to merge these changes incrementally, as there are a lot of different parts to this.

@akleshchev It sounds like you are in the vicinity of [[SL-17005] "Bias" value is forcing textures to downres](https://github.com/secondlife/jira-archive-internal/issues/67137). QA is currently retesting this, but I wouldn't be surprised if that is still an issue in some cases, given the code issues pointed out.

akleshchev:
>It sounds like you are in the vicinity of https://github.com/secondlife/jira-archive-internal/issues/67137. QA is currently retesting this, but I wouldn't be surprised if that is still an issue in some cases, given the code issues pointed out.

Based on how our gGLManager.mVRAM gets set, I won't be surprised if that's Intel's DVMT causing havok.

akleshchev:
Added a commit to expose that 'bias' memory in texture console (and it reports way too much), but I feel like any further changes will need to wait for davep so this review is done for now.

There are odities like 'LLImageGL::getTextureBytesAllocated() / 1024.0 / 512.0' in LLViewerTexture::updateClass() which make it preferable to ask davep prior to any changes.

Ansariel:
Why is VRAM consumption by VBOs not taken into account in the LLWindowsXXX classes? It is in LLViewerTexture::updateClass

cosmic-linden:
> Based on how our gGLManager.mVRAM gets set, I won't be surprised if that's in part due to Intel's DVMT causing havok, depends on how intel reports avaliable memory.

Interesting. That is worth keeping in mind on Intel machines, although the issue came up for Nvidia graphics as well at the time that ticket was opened.

I recall davep more recently mentioning that there exist some heuristics on Windows that determine whether or not an application is a game, which was found to have implications for performance in the viewer, and could very well have implications for memory allocation as well. I do not recall the state of resolution for that issue.

indra/newview/llviewertexture.cpp
```diff
@@ -99,6 +99,7 @@ U32 LLViewerTexture::sMinLargeImageSize = 65536; //256 * 256.
 U32 LLViewerTexture::sMaxSmallImageSize = MAX_CACHED_RAW_IMAGE_AREA;
 bool LLViewerTexture::sFreezeImageUpdates = false;
 F32 LLViewerTexture::sCurrentTime = 0.0f;
+F32 LLViewerTexture::sFreeVRAMMegobytes = 768.0f;
```
cosmic-linden, line 102: This is a pretty specific number and is worth documenting why it's used.

Ansariel: sFreeVRAMMegabytes instead of sFreeVRAMMegobytes

akleshchev: Declaread 768 as a const MIN_VRAM_BUDGET (value was selected due it being the minimum from updateClass())
Fixed name

indra/newview/llviewertexture.cpp
```diff
@@ -100,6 +100,9 @@ U32 LLViewerTexture::sMaxSmallImageSize = MAX_CACHED_RAW_IMAGE_AREA;
 bool LLViewerTexture::sFreezeImageUpdates = false;
 F32 LLViewerTexture::sCurrentTime = 0.0f;
 
+const F32 MIN_VRAM_BUDGET = 768.f;
```
Ansariel, line 103: Could define it as constexpr

marchcat (approved):

maxim-productengine (approved):

Issue #808, opened 2024-02-15
Do not crash on 'Unable to load basic shader'


LLGuru:
BugSplat collects different crash reports which actually don't relate to a real crashes:
https://github.com/secondlife/triage/issues/55

These fake crash reports can be grouped by **User Description** which is in fact the error message
https://app.bugsplat.com/v2/keycrash?g0=userDescription&stackKeyId=83464&database=SecondLife_Viewer_2018

The most popular error message is _"Unable to load basic shader"_
If the viewer can't load some shader it displays a message box
![image](https://github.com/secondlife/viewer/assets/124201357/be9bff60-4e2b-40f1-b5eb-01dd11aef617)
and then logs the error using `LL_ERRS()` which explicitly leads to a crash

But if the user is already notified, no needs to crash the process and send the crash report
The viewer could shutdown without sending a crash report

## Test Plan
- Imitate the error locally (delete or rename the shader file if possible),
- Start the viewer
- Verify the message box
- Verify soft shutdown with no crash

akleshchev:
>But if the user is already notified, no needs to crash the process and send the crash report
I think the conclusion on this was: let viewer crash and log to bugsplat for the sake of statistics and because it has a potential to be an shader init issue.

PR #809, opened 2024-02-15
#808 Do not crash on 'Unable to load basic shader'

LLGuru:


akleshchev:
1. This is a wrong message to show. We have a dedicated user message for that in strings.xml. MBVideoDrvErr
2. We left this failure crashing intentionaly for posibility to have access to logs and check which gpus fail to work.

LLGuru:
> 1. This is a wrong message to show. We have a dedicated user message for that in strings.xml. MBVideoDrvErr

You mean to use the following call:
`OSMessageBox(mCallbacks->translateString("MBVideoDrvErr"), mCallbacks->translateString("MBError"), OSMB_OK);`

Of course, we could use **MBVideoDrvErr**, but this will hide more impotrant information which is displayed in my variant

akleshchev:
>but this will hide more impotrant information which is displayed in my variant
1. You are showing a user message (which viewer wasn't doing before. It means that at minimum means that it should be localizable, yours aren't
2. You can put that info into logs, and if user does report this issue we will need full logs either way to figure out specifics.

LLGuru:
> 2\. We left this failure crashing intentionaly for posibility to have access to logs and check which gpus fail to work.

All you get from the log is the same file name as is displayed in the message box

Example: https://app.bugsplat.com/v2/crash?database=SecondLife_Viewer_2018&id=1425227
User Description: _Unable to load basic shader deferred/screenSpaceReflUtil.glsl, verify graphics driver installed and current._

SecondLife.log
```
2024-02-15T12:43:52Z WARNING # llrender/llshadermgr.cpp(429) LLShaderMgr::dumpObjectLog : Shader loading from C:\\Program Files\\SecondLifeViewer\\app_settings\\shaders/class1/deferred/screenSpaceReflUtil.glsl
2024-02-15T12:43:52Z WARNING # llrender/llshadermgr.cpp(430) LLShaderMgr::dumpObjectLog : \nFragment shader failed to compile with the following errors:\nERROR: 37:1: error(#247) Function return is not matching type\nERROR: 43:1: error(#247) Function return is not matching type\nERROR: error(#273) 2 compilation errors.  No code generated\n
2024-02-15T12:43:52Z WARNING # llrender/llshadermgr.cpp(406) LLShaderMgr::dumpShaderSource : \n   1: #version 330\n   2: #define FXAA_GLSL_130 1\n   3: #define GBUFFER_FLAG_SKIP_ATMOS   0.0 \n   4: #define GBUFFER_FLAG_HAS_ATMOS    0.34\n   5: #define GBUFFER_FLAG_HAS_PBR      0.67\n   6: #define GET_GBUFFER_FLAG(flag)    (abs(norm.w-flag)< 0.1)\n   7: #define MAX_JOINTS_PER_MESH_OBJECT 110\n   8: #define IS_AMD_CARD 1\n   9: /**\n  10:  * @file class3/deferred/screenSpaceReflUtil.glsl\n  11:  *\n  12:  * $LicenseInfo:firstyear=2007&license=viewerlgpl$\n  13:  * Second Life Viewer Source Code\n  14:  * Copyright (C) 2007, Linden Research, Inc.\n  15:  *\n  16:  * This library is free software; you can redistribute it and/or\n  17:  * modify it under the terms of the GNU Lesser General Public\n  18:  * License as published by the Free Software Foundation;\n  19:  * version 2.1 of the License only.\n  20:  *\n  21:  * This library is distributed in the hope that it will be useful,\n  22:  * but WITHOUT ANY WARRANTY; without even the implied warranty of\n  23:  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU\n  24:  * Lesser General Public License for more details.\n  25:  *\n  26:  * You should have received a copy of the GNU Lesser General Public\n  27:  * License along with this library; if not, write to the Free Software\n  28:  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA\n  29:  *\n  30:  * Linden Research, Inc., 945 Battery Street, San Francisco, CA  94111  USA\n  31:  * $/LicenseInfo$\n  32:  */\n  33: \n  34: // debug stub\n  35: \n  36: float random (vec2 uv) \n  37: {\n  38:     return 0;\n  39: }\n  40: \n  41: float tapScreenSpaceReflection(int totalSamples, vec2 tc, vec3 viewPos, vec3 n, inout vec4 collectedColor, sampler2D source, float glossiness)\n  42: {\n  43:     collectedColor = vec4(0);\n  44:     return 0;\n  45: }\n
2024-02-15T12:43:52Z WARNING #Shader# newview/llviewershadermgr.cpp(675) LLViewerShaderMgr::loadBasicShaders : Failed to load fragment shader deferred/screenSpaceReflUtil.glsl
2024-02-15T12:43:52Z INFO # newview/llappviewer.cpp(3215) LLAppViewer::writeDebugInfo : Writing debug file C:\\Users\\gustavomolina\\AppData\\Roaming\\SecondLife\\logs\\static_debug_info.log
2024-02-15T12:43:52Z ERROR # newview/llviewershadermgr.cpp(445) LLViewerShaderMgr::setShaders : Unable to load basic shader deferred/screenSpaceReflUtil.glsl, verify graphics driver installed and current.
```

static_debug_info.log
```
<llsd>
    <map>
    <key>FatalMessage</key>
        <string>2024-02-15T12:43:52Z ERROR # newview/llviewershadermgr.cpp(445) LLViewerShaderMgr::setShaders : Unable to load basic shader deferred/screenSpaceReflUtil.glsl, verify graphics driver installed and current.</string>
    <key>FoundOtherInstanceAtStartup</key>
        <boolean>0</boolean>
    </map>
</llsd>
```

If we really need these crash reports we'd better use some other call stack to separate these crash reports in a special BugSplat

cosmic-linden:
> Of course, we could use **MBVideoDrvErr**, but this will hide more impotrant information which is displayed in my variant

Then perhaps we should create an enhanced variant of this localization string which replaces the driver shader name into the string. I don't know the convention for doing that off the top of my head, but I believe the minimap hovertext code does this.

I also agree that having the viewer crash makes sense to get that information into BugSplat.

akleshchev:
> Then perhaps we should create an enhanced variant of this localization string which replaces the driver shader name into the string.
It's trivial, just add some kind of [SHADER] into the string, then LLTrans::getString(name, shader_llsd_arg) with set argument in code. Ex: splash_msg

LLGuru:
Hi @cosmic-linden ,
thank you for your opinion

> I also agree that having the viewer crash makes sense to get that information into BugSplat.

I believe that if the root cause of the issue is on the viewer side, then our QA will face and report it (even without BugSplat)
But if the reason is a corrupted local configuration on the user side, then we can't fix this ourselves (even with BugSplat)
In fact, in both cases we don't need to receive such chash reports to BugSplat
Does it make sense? :-)

akleshchev:
> I believe that if the root cause of the issue is on the viewer side, then our QA will face and report it (even without BugSplat)
Our QA has a limited pool of hardware and can't reasonably test everything.

> In fact, in both cases we don't need to receive such crash reports to BugSplat
We actually want to know about both cases.

In first case we can figure out specific hardware (I think it happened  already with this specific crash on a MacOS device). In second case if it happens too often we would want to add additonal logging. But to know that it happens we will want to log it bugsplat side.

PR #815, opened 2024-02-15
Replace more BOOL with bool

Ansariel:
This replaces almost all fake booleans in llmath except in LLVolume and LLXForm which are base classes for various classes in other projects.

Also did some cleanup and remove old unused functions and function declarations without actual definitions.


marchcat (approved):
Thank you!

Issue #811, opened 2024-02-15
 Do not crash on 'Default configuration file settings_files.xml not found'


LLGuru:
BugSplat collects different crash reports which actually don't relate to a real crashes:
https://github.com/secondlife/triage/issues/55

These fake crash reports can be grouped by User Description which is in fact the error message
https://app.bugsplat.com/v2/keycrash?g0=userDescription&stackKeyId=83464&database=SecondLife_Viewer_2018

One of the most popular error messages is _"Default configuration file settings_files.xml not found"_

In fact, this file is not present sometimes, and viewer can't start in this case
The solution could be to make viewer be able to start without this file
As an option, it's contents could be embedded to the binary as a resource

## Test Plan

- Delete or rename file settings_files.xml
- Start viewer
- Verify no crash on startup

LLGuru:
Fixed for Windows in https://github.com/secondlife/viewer/pull/817

PR #817, opened 2024-02-16
[viewer#811] Do not crash on 'Default configuration file settings_files.xml not found'

LLGuru:
In the context of https://github.com/secondlife/viewer/issues/811

For Windows we can pack XML files into the binary and read them from there on run-time instead of reading the file

### DIFF
--- a/indra/llfilesystem/lldir_win32.cpp
+++ b/indra/llfilesystem/lldir_win32.cpp
@@ -390,6 +390,15 @@ bool LLDir_Win32::fileExists(const std::string &filename) const
 	}
 }
 
+// virtual
+bool LLDir_Win32::skinExists(const std::string& subdir, const std::string &skin) const
+{
+    if (subdir == LLDir::XUI && skin == "en")
+        return true;
+
+    return LLDir::skinExists(subdir, skin);
+}
+
 
 /*virtual*/ std::string LLDir_Win32::getLLPluginLauncher()
 {
--- a/indra/llxml/llxmlnode.cpp
+++ b/indra/llxml/llxmlnode.cpp
@@ -651,33 +651,83 @@ bool LLXMLNode::updateNode(
 }
 
 // static
+std::string LLXMLNode::getXML(const std::string& filename)
+{
+    // Try to read from file
+    LLFILE* fp = LLFile::fopen(filename, "rb"); /* Flawfinder: ignore */
+    if (fp)
+    {
+        fseek(fp, 0, SEEK_END);
+        U32 length = ftell(fp);
+        fseek(fp, 0, SEEK_SET);
+
+        LL_DEBUGS("XMLNode") << "reading XML from file: " << filename << LL_ENDL;
+        std::vector<char> buffer(length);
+        size_t nread = fread(buffer.data(), 1, length, fp);
+        fclose(fp);
+
+        return std::string(buffer.data(), nread);
+    }
+
+#ifdef _WIN64
+    // Try to read from resource
+    std::string app_dir = gDirUtilp->getAppRODataDir();
+    // E.g., if finename starts with "C:\\Program Files\\SecondLifeViewer"
+    if (filename.substr(0, app_dir.size()) == app_dir)
+    {
+        // Convert the rest of filename to resource_id
+        // E.g., "app_settings\\settings_files.xml" => L"IDR_APP_SETTINGS_SETTINGS_FILES_XML"
+        std::string resource_id = filename.substr(app_dir.size() + 1);
+        LLStringUtil::replaceChar(resource_id, '\\', '_');
+        LLStringUtil::replaceChar(resource_id, '%', '_');
+        LLStringUtil::replaceChar(resource_id, '.', '_');
+        LLStringUtil::toUpper(resource_id);
+        resource_id = "IDR_" + resource_id;
+        std::wstring resource_idw(resource_id.begin(), resource_id.end());
+
+        HMODULE hmodule = GetModuleHandleW(nullptr);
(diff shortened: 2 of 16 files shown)


akleshchev:
>For macOS I'm still investigating how to do the similar thing

On MacOS those files are a part of a single dmg packagate, normaly user can't tamper with that and intall only an executable.

LLGuru:
> unless you are willing to turn every file into a resource

Souns good :-) I can do this

LLGuru:
> Please add a milestone and link an issue.

Added to https://github.com/secondlife/viewer/issues/811

cosmic-linden:
If I go to `%appdata%\Roaming\SecondLife\user_settings`, and delete `settings.xml` there, the viewer just regenerates the file there, and starts up normally without issue.

Meanwhile, if I go to `C:\Program Files\SecondLifeViewer\app_settings` and attempt to delete `settings.xml` there, Windows asks for administrative permissions first.

It seems reasonable that if a Windows user is messing around in the Program Files folder, then bad things could happen. So, it seems reasonable that the viewer should not attempt to recover from that. But there are reasons why the files would be unreadable for reasons outside the user's control, which @akleshchev makes good suggestions on.

indra/llxml/llxmlnode.cpp
```diff
@@ -653,31 +653,67 @@ bool LLXMLNode::updateNode(
...
+                    size_t size = SizeofResource(nullptr, hresource);
+                    if (LLXMLNode::parseBuffer(text, size, node, defaults_tree))
+                        return true;
+                }
+            }
+        }
+    }
+#endif
+
+	// Try to read from file
```
akleshchev, line 693: This doesn't cover all files, for example removing commands.xml still causes a crash.

I'm not against the approach of moving files into resources, but since it can't cover some files at all (like updater and plugins) I would rather have viewer complain to user about a bad install than trying to fix consequences one file at a time.

marchcat: There is still fallback code below, so it will likely be fine.

LLGuru: > This doesn't cover all files
I do not pretend on this, just want to make the viewer more stable and fast

marchcat (dismissed):

indra/llxml/llxmlnode.cpp
akleshchev: >There is still fallback code below, so it will likely be fine.
I mean that:
1. Instead of failing on a first file with LL_ERRS and a visual warning in D591 it will just fail at next one and not nessesaryly with an LL_ERRS and defientely without a warning. If you are doing this, than move the visual warning to next file that is most likely to crash viewer. 
2. Files were moved into resources yet remain unused. In case of settings.xml it means extra 500kb of weight.

akleshchev: >There is still fallback code below, so it will likely be fine.
Actually this is going to break preview tools... It looks like it copies all files (including floaters) which means viewer will load resources first and live-time modification via previe tool wouldn't work.

marchcat: Oh. That's a problem.

akleshchev: >There is still fallback code below, so it will likely be fine.

Actually this is going to break preview tools... It looks like it copies all files (including floaters) which means viewer will load resources first and live-time modification via previe tool wouldn't work.

akleshchev (changes requested):

indra/newview/CMakeLists.txt
```diff
@@ -1531,6 +1531,47 @@ if (WINDOWS)
...
+             "IDR_${RESOURCE_ID}\tXML\t\"../${XML_FILE}\"\n"
+             )
+
+    endforeach()
+
+    file(APPEND ${CMAKE_CURRENT_BINARY_DIR}/viewerRes.rc
+         "\n"
+         "/////////////////////////////////////////////////////////////////////////////\n"
+         "\n"
+         )
```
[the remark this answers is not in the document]
marchcat: Wait... Does this mean XUI files as well?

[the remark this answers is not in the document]
akleshchev: Yes, that's why preview tool broke.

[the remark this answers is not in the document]
akleshchev: What tools exactly? If preview tools will start working, likely any tool for external modifications would to.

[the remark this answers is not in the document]
marchcat: Translations parser, for example. It compares different revisions using git, so won't work correctly with binary files.

akleshchev (commented):
Issue got fixed, so I'm removing the 'request change', but looks like it's still in progress.

indra/llfilesystem/lldir_win32.cpp
```diff
@@ -390,6 +390,15 @@ bool LLDir_Win32::fileExists(const std::string &filename) const
 	}
 }
 
+// virtual
+bool LLDir_Win32::skinExists(const std::string& subdir, const std::string &skin) const
+{
+    if (subdir == LLDir::XUI && skin == "en")
```
akleshchev, line 396: Why 'en' only if you are making a copy of all skins?

LLGuru: Because we use this check for 2 skins only, 'en' and 'en-us'
For xui we provide 'en' but not 'en-us'
Thus this check is very simple

LLGuru:
Replaced with https://github.com/secondlife/viewer/pull/914

LLGuru:
Replaced with https://github.com/secondlife/viewer/pull/914

PR #821, opened 2024-02-16
BOOL to bool, pt. 4

larsnaesbye:
Another bunch of conversions aiming to eventually remove the custom BOOL definition and enabling more optimizations.


marchcat (approved):
Thank you very much!

PR #825, opened 2024-02-17
Enable indent_with_spaces hook for source files

marchcat:


Ansariel:
Should be the other way: Tabs are way way better than spaces

marchcat:
That's a philosophical question 😸
Nonetheless, our coding standard requires spaces.

Ansariel:
> That's a philosophical question 😸 Nonetheless, our coding standard requires spaces.

Easy fix: Just change it back to tabs as it was originally until somebody decided to change it.

marchcat:
We now have a crazy mix of tabs and spaces, so changing the standard will still require processing all files to make them fit. And in this case, changing from tabs to spaces seems a little simpler

Ansariel:
The vast majority of the code is tabs though...

.pre-commit-config.yaml
```diff
@@ -9,6 +9,8 @@ repos:
       - id: end-of-file
         files: \.(cpp|c|h|py|glsl|cmake|txt)$
         exclude: language.txt
+      - id: indent-with-spaces
+        files: \.(cpp|c|h|py|glsl)$
```
bennettgoble, line 13: `inl`?

marchcat:
Converted back to draft until the code is ready (should remove all the tabs first).

Nicky-D:
@marchcat in case you'd really replace all tabs with spaces, there is another pet peeve of mine.

There is an unholy amount of trailing white spaces in the code base. If you to through the task and create a huge white space diff anyway, ca we get those trailing whitespaces dealt with too?

marchcat:
@Nicky-D yes, that diff will include trimming those whitespaces too.

marchcat:
This was a part of #1373.

PR #831, opened 2024-02-18
Replace BOOL with bool in llwindow and dependent classes

Ansariel:
This replaces all BOOL with bool in llwindow except for Windows API calls and Objective-C (@larsnaesbye or maybe @marchcat  maybe you should look into those since I don't know anything about it and I can't tell if the changes are compiling on macOS right now).

This also resulted in a big chunk of dependent classes that inherited classes from llwindow and are overriding their methods.

Also fixed the broken unit tests that were failing after recent changes.


marchcat:
@Ansariel, thank you for the PR!

GH builds don't work outside our repos, so I'll try to build your code on my mac. Will post an update when ready.

marchcat (approved):
There were couple minor build issues on mac; I'll fix them after merging this PR.

PR #833, opened 2024-02-18
BOOL to bool, pt. 5

larsnaesbye:
Fewer conversions this time, mostly in llmessage.

### DIFF
--- a/indra/llmath/llvolume.cpp
+++ b/indra/llmath/llvolume.cpp
@@ -95,7 +95,7 @@ const F32 SKEW_MAX	=  0.95f;
 const F32 SCULPT_MIN_AREA = 0.002f;
 const S32 SCULPT_MIN_AREA_DETAIL = 1;
 
-bool gDebugGL = FALSE; // See settings.xml "RenderDebugGL"
+bool gDebugGL = false; // See settings.xml "RenderDebugGL"
 
 bool check_same_clock_dir( const LLVector3& pt1, const LLVector3& pt2, const LLVector3& pt3, const LLVector3& norm)
 {
--- a/indra/llmath/llvolume.cpp
+++ b/indra/llmath/llvolume.cpp
@@ -95,7 +95,7 @@ const F32 SKEW_MAX	=  0.95f;
 const F32 SCULPT_MIN_AREA = 0.002f;
 const S32 SCULPT_MIN_AREA_DETAIL = 1;
 
-BOOL gDebugGL = FALSE; // See settings.xml "RenderDebugGL"
+bool gDebugGL = FALSE; // See settings.xml "RenderDebugGL"
 
 bool check_same_clock_dir( const LLVector3& pt1, const LLVector3& pt2, const LLVector3& pt3, const LLVector3& norm)
 {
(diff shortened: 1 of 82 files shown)


indra/llmath/llvolume.cpp
```diff
@@ -95,7 +95,7 @@ const F32 SKEW_MAX	=  0.95f;
 const F32 SCULPT_MIN_AREA = 0.002f;
 const S32 SCULPT_MIN_AREA_DETAIL = 1;
 
-BOOL gDebugGL = FALSE; // See settings.xml "RenderDebugGL"
+bool gDebugGL = FALSE; // See settings.xml "RenderDebugGL"
```
Ansariel, line 98: Oh no!  A "FALSE"! 😄

larsnaesbye: TY! Fixed :)

indra/llmessage/lltemplatemessagereader.cpp
```diff
@@ -286,9 +286,9 @@ void LLTemplateMessageReader::getU8(const char *block, const char *var,
 void LLTemplateMessageReader::getBOOL(const char *block, const char *var, 
 										  bool &b, S32 blocknum )
 {
-	U8 value;
-	getData(block, var, &value, sizeof(U8), blocknum);
-	b = (BOOL) value;
+	bool value;
+	getData(block, var, &value, sizeof(bool), blocknum);
```
Ansariel, line 290: Does this actually work? I'm concerned the simulator is most likely sending U8 for bool and that it might break if you now just read less data than expected. I would probably keep reading and U8 and then use a static_cast to convert it to bool.

Ansariel:
Question: Are you building with tests? If not, you should probably do so as I expect there being tests that potentially break because from the method signature changes.

indra/llmessage/lltemplatemessagereader.cpp
larsnaesbye: Reverted so we cast again.

larsnaesbye:
> Question: Are you building with tests? If not, you should probably do so as I expect there being tests that potentially break because from the method signature changes.

I would suspect tests do run when calling `autobuild build -c RelWithDebInfoOS` ?

Ansariel:
> > Question: Are you building with tests? If not, you should probably do so as I expect there being tests that potentially break because from the method signature changes.
> 
> I would suspect tests do run when calling `autobuild build -c RelWithDebInfoOS` ?

I don't think they include tests. If you add -DLL_TESTS:BOOL=TRUE it will definitely build and execute the tests, too.

larsnaesbye:
`autobuild build -c RelWithDebInfoOS -- -DLL_TESTS:BOOL=On` seems to run the tests  - and yes some of them fail now. Will fix :)

larsnaesbye:
@Ansariel one test seems to still fail after the latest commit: 
```
Unit test group_started name=permission
[permission, 11] fail: 'setBaseBits():failed '
```

And I'm not sure why....

Ansariel:
> @Ansariel one test seems to still fail after the latest commit:
> 
> ```
> Unit test group_started name=permission
> [permission, 11] fail: 'setBaseBits():failed '
> ```
> 
> And I'm not sure why....

Difficult to say without actually seeing it during the build. I know one of the GLTF tests is failing (@marchcat do you have an idea why?) and something else was failing. I might just have to check after the PR got merged.

larsnaesbye:
It seems the test fails even before this patch is applied. So it's not related to the changes I introduced here....

marchcat:
Thanks!

I usually build with tests disabled too (since GH builds run them anyway), but will specifically check them before and after this change.

marchcat:
The mentioned GLTF test fails even before all these bool changes (for example, it failed for me at https://github.com/secondlife/viewer/commit/2d9e00eff5ca1e526351e77800dc3abdf4ff9b68).

marchcat (approved):

PR #835, opened 2024-02-19
MS Visual Studio build fix for '[[fallthrough]]'

LLGuru:


marchcat (changes requested):
I think we shouldn't use any compilers that don't support c++17.
Something is wrong with your local env.

akleshchev:
Agreee with marchcat, you are likely using wrong branch of build variables.

Ansariel:
The fallthrough attribute is part of C++17 standard which is enabled for all builds. This is a problem in your build environment, not a problem in the code.

PR #836, opened 2024-02-19
Converted remaining BOOLs into bool in llinventory and llmessage

Ansariel:
Converted remaining BOOLs into bool in llinventory and llmessage.

Also changed LLPreeditor::standouts_t to std::deque since it got turned into std::vector<bool> previously, which is a implementation dependent specialization where some STL container functions will perform unexpectedly depending on the implementation.

@marchcat I'm a bit confused about llsdmessagebuilder.cpp line 357: Was casting to BOOL* even correct before? A bool in the message template is stored in a U8 while BOOL is an S32, so the size should be different!?

### DIFF
--- a/indra/llcharacter/llkeyframemotion.cpp
+++ b/indra/llcharacter/llkeyframemotion.cpp
@@ -1339,12 +1339,14 @@ bool LLKeyframeMotion::deserialize(LLDataPacker& dp, const LLUUID& asset_id, boo
 		return false;
 	}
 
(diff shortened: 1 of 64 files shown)


marchcat:
Thank you @Ansariel!

Not sure if it was correct, but there was a further logical check anyway, so it should have been fine.
The updated version looks much better.

marchcat (approved):

marchcat:
Oops, wrong button. Merging now.

PR #837, opened 2024-02-19
[Linux]gcc is broken for in place template instantiation.

beqjanus:
gcc is broken and cannot compile the latest llmutex.h without the following mods. 

See https://stackoverflow.com/questions/49707184/explicit-specialization-in-non-namespace-scope-does-not-compile-in-gcc
for details.


marchcat (approved):
Thank you!

PR #842, opened 2024-02-20
viewer#840 Xcode buildfix

marchcat:


indra/llrender/llfontfreetype.cpp
```diff
@@ -423,7 +423,7 @@ bool LLFontFreetype::hasGlyph(llwchar wch) const
 LLFontGlyphInfo* LLFontFreetype::addGlyph(llwchar wch) const
 {
 	if (mFTFace == NULL)
-		return false;
+		return NULL;
```
nat-goodspeed, line 426: While we're getting away from old `#define` symbols, we could start propagating `nullptr` too...

marchcat: I have decided to leave it as NULL since there are other appearances of it in the file.
But you are right, it's another `#define` to be replaced.

PR #844, opened 2024-02-20
viewer#819 Avoid reading the same XML file multiple times

LLGuru:
Cache parsed XML files for reuse

### DIFF
--- a/indra/llui/llfloater.cpp
+++ b/indra/llui/llfloater.cpp
@@ -3419,12 +3419,16 @@ bool LLFloater::isVisible(const LLFloater* floater)
     return floater && floater->getVisible();
 }
 
-bool LLFloater::buildFromFile(const std::string& filename)
+bool LLFloater::buildFromFile(const std::string& filename, bool cacheable)
 {
     LL_PROFILE_ZONE_SCOPED;
+
+    llassert_msg(!cacheable || !mSingleInstance || !mReuseInstance,
+        "No needs to cache XML for floater with mSingleInstance AND mReuseInstance flags set");
+
 	LLXMLNodePtr root;
 
-	if (!LLUICtrlFactory::getLayeredXMLNode(filename, root))
+	if (!LLUICtrlFactory::getLayeredXMLNode(filename, root, LLDir::CURRENT_SKIN, cacheable))
 	{
 		LL_WARNS() << "Couldn't find (or parse) floater from: " << filename << LL_ENDL;
 		return false;
--- a/indra/llxml/llxmlnode.cpp
+++ b/indra/llxml/llxmlnode.cpp
@@ -650,34 +650,56 @@ bool LLXMLNode::updateNode(
 	return true;
 }
 
-// static
-bool LLXMLNode::parseFile(const std::string& filename, LLXMLNodePtr& node, LLXMLNode* defaults_tree)
+static std::map<std::string, LLXMLNodePtr> sXMLCache;
+static LLSharedMutex sXMLCacheMutex;
+
+static void saveToCache(const std::string& filename, LLXMLNodePtr& node)
 {
-	// Read file
-	LL_DEBUGS("XMLNode") << "parsing XML file: " << filename << LL_ENDL;
-	LLFILE* fp = LLFile::fopen(filename, "rb");		/* Flawfinder: ignore */
-	if (fp == NULL)
-	{
-		node = NULL ;
-		return false;
-	}
-	fseek(fp, 0, SEEK_END);
-	U32 length = ftell(fp);
-	fseek(fp, 0, SEEK_SET);
+    LLExclusiveMutexLock lock(&sXMLCacheMutex);
+    sXMLCache.emplace(filename, node.notNull() ? node->deepCopy() : nullptr);
+}
 
-	U8* buffer = new U8[length+1];
-	size_t nread = fread(buffer, 1, length, fp);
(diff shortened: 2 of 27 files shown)


indra/llxml/llxmlnode.cpp
```diff
@@ -650,29 +650,52 @@ bool LLXMLNode::updateNode(
...
+        LLSharedMutexLock lock(&sXMLCacheMutex);
+        auto it = sXMLCache.find(file_id);
+        if (it != sXMLCache.end())
+        {
+            node = it->second.notNull() ? it->second->deepCopy() : nullptr;
+            return node.notNull();
+        }
+    }
+
+    LL_DEBUGS("XMLNode") << "parsing XML file: " << filename << LL_ENDL;
```
akleshchev, line 680: I have a feeling this is going to break XUI preview tool since it relies on re-reading the same file.

akleshchev:
Agree with Cosmic.

By this point xml files being parsed live and being changeable is a feature, not a bug, if some things are called too often from large xml files I suggest caching them individually like notifications do with mTemplates or LLtrans does with sStringTemplates.

akleshchev:
P.S. probably not going to be usable in this case since affected files are panels, but in case of floaters we have mReuseInstance

indra/llxml/llxmlnode.cpp
```diff
@@ -650,34 +650,106 @@ bool LLXMLNode::updateNode(
 	return true;
 }
 
+static std::map<std::string, LLXMLNodePtr> sXMLCache;
+static LLSharedMutex sXMLCacheMutex;
```
nat-goodspeed, line 654: fwiw you could consider `LockStatic`, although I see that lockstatic.h lacks a usage example in its comments.

Also `LockStatic` should be extended to support the shared/exclusive lock distinction you use below.

akleshchev:
>I do like some of these changes, notably the signature of LLXMLNode::parseBuffer().

I believe that leaked from a different pull request
https://github.com/secondlife/viewer/pull/817

>Is it considered a feature that a user could edit the .xml files in the installed viewer directory and see the changes?

XUI preview tools works based on that, at minimum changiing xmls 'live' should stay functional on local builds.

indra/llxml/llxmlnode.cpp
LLGuru: Fixed

LLGuru:
> Is it considered a feature that a user could edit the .xml files in the installed viewer directory and see the changes?

In the recent upload the cache is applied only for XML files readen from the resource.
XML files readen from the file sysem aren't cached and aren't checked in the cache.
This allows users edit XML files directly in the repository where they are placed.

I suppose that the main cost in not a reading but a parsing of the XML content.
This cache could shorten the way required for creating `LLXMLNode` instance.
Unfortunately we can't use the same `LLXMLNode` instance many times
because some code corrupts them after reading, so I use `deepCopy()`

Also I suppose that 99.99% of users don't edit XML files and don't have a repository with them.
Those users will use XML content from the resource, and this content will be cached
Other 0.001% of users (developers) work with repository, and their workflow stays the same

akleshchev:
>In the recent upload the cache is applied only for XML files readen from the resource.

But now it will be Windows only which partially defeats the purpose and would still cache files that are intended to be 'read once' (most single instance floaters since they get initialized once) if those are in the resources. The issue Cosmic pointed out remains unresolved.

LLGuru:
> By this point xml files being parsed live and being changeable is a feature, not a bug

Developers working with XML files directly from the repository can continue working the same way.
External files aren't being cached, and they have the max priority when reading.

But user having XML content in the resource will parse each file not more than once per session.
All subsequental reading will be redirected to the cache, and we won't spend the time on parsing.

LLGuru:
> I would recommend making this explicit by adding a `cache` flag to these functions

Good idea, @cosmic-linden , I've already implemented it, thanks :-)

akleshchev (approved):

indra/llui/llfloater.cpp
```diff
@@ -3419,12 +3419,12 @@ bool LLFloater::isVisible(const LLFloater* floater)
     return floater && floater->getVisible();
 }
 
-bool LLFloater::buildFromFile(const std::string& filename)
+bool LLFloater::buildFromFile(const std::string& filename, bool cacheable)
```
akleshchev, line 3422: Might be good idea to add a check from single instance floaters and either warn or override value with false when both 'cacheable' and single instance. Because both at the same time make no sense.

LLGuru: Done

indra/llxml/llxmlnode.cpp
```diff
@@ -669,52 +708,70 @@ bool LLXMLNode::parseFile(const std::string& filename, LLXMLNodePtr& node, LLXML
         LL_DEBUGS("XMLNode") << "parsing XML file: " << filename << LL_ENDL;
         if (parseBuffer(buffer.data(), nread, node, defaults_tree))
         {
+            saveToCache(file_id, node);
```
cosmic-linden, line 711: Missing `cacheable` check

cosmic-linden (changes requested):

indra/llxml/llxmlnode.cpp
LLGuru: Fixed, thanks

cosmic-linden (approved):

PR #848, opened 2024-02-20
BOOL to bool

Ansariel:
Converts the remaining parts of llxml (LLControlGroup) and llui from BOOL to bool


larsnaesbye:
Cool! I have worked on all those pesky `postBuild()` too, but since we're not coordinating that's what happens :-D

marchcat (approved):
Wow, this one is huge.

Thank you!

PR #849, opened 2024-02-21
secondlife/viewer-issues#72: Fix material preview making things disappear. General cleanup.

cosmic-linden:


doc/testplans/material_preview.md
```diff
@@ -0,0 +1,81 @@
+# Material Preview
```
Geenz, line 1: I would probably ask someone like @Dan-Linden or @AtlasLinden to review the test plan to make sure it parses well.

Geenz (approved):
Generally looks good - though I think we need to revisit LLRenderTarget and the ref swapping at some point.  Not a blocker for merge though.

cosmic-linden:
> Generally looks good - though I think we need to revisit LLRenderTarget and the ref swapping at some point. Not a blocker for merge though.

Agreed that the ref swapping is a bit odd. I wanted to avoid messing too much with the guts of the render pipeline for this more specialized thing for now.

Geenz:
@cosmic-linden Fair.  I do want to get a pass at render targets specifically after featurettes - we'll figure that out after release though.

PR #851, opened 2024-02-21
Convert remaining BOOLs in llxml and introduce std::string_view

Ansariel:
For some reason, there were some stray FALSE/TRUE left in llxml. Used the opportunity to also introduce std::string_view to the viewer and make use of it for the viewer settings.


marchcat (approved):
Thank you!

PR #855, opened 2024-02-21
Convert remaining BOOL to bool

Ansariel:
This converts all remaining BOOL into bool in newview plus some sneaky ones that have been hiding in tests.

Changes beyond simple conversion:

- Changed some fake booleans represented as S8 in LLViewerFetchedTexture
- Fix an unsafe operation in LLVOCachePartition::setCullHistory
- LLSkyTex::createGLImage parameters were incorrect and an index was reflected by BOOL

Exempted from the changes were calls to the Windows API in llappviewerwin32.cpp, lldirpicker.cpp, llfloaterjoystick.cpp, llfilepicker.cpp, llviewerjoystick.cpp. Also exempted were gbooleans (TRUE/FALSE) defined in glib for Linux.

There is also a remaining BOOL in the interface for the physicsextensions lib which cannot be changed as of now.


indra/newview/llviewertexture.h
```diff
@@ -468,21 +468,21 @@ class LLViewerFetchedTexture : public LLViewerTexture
 	S8  mDesiredDiscardLevel;			// The discard level we'd LIKE to have - if we have it and there's space	
 	S8  mMinDesiredDiscardLevel;	// The minimum discard level we'd like to have
 
-	S8  mNeedsAux;					// We need to decode the auxiliary channels
-	S8  mHasAux;                    // We have aux channels
-	S8  mDecodingAux;				// Are we decoding high components
-	S8  mIsRawImageValid;
-	S8  mHasFetcher;				// We've made a fecth request
-	S8  mIsFetching;				// Fetch request is active
+	bool mNeedsAux;					// We need to decode the auxiliary channels
```
Ansariel, line 471: These were used as booleans, but were S8 - maybe to save memory?

indra/newview/llvocache.cpp
```diff
@@ -1039,10 +1039,10 @@ S32 LLVOCachePartition::cull(LLCamera &camera, bool do_occlusion)
 }
 #endif // LL_TEST
 
-void LLVOCachePartition::setCullHistory(BOOL has_new_object)
+void LLVOCachePartition::setCullHistory(bool has_new_object)
 {
 	mCullHistory <<= 1;
-	mCullHistory |= has_new_object;
+	mCullHistory |= static_cast<U32>(has_new_object);
```
Ansariel, line 1045: Would be unsafe operation otherwise because mCullHistory is an U32.

marchcat (approved):
Thank you, @Ansariel!
The change is huge, let's see how it builds.

PR #861, opened 2024-02-21
[DO NOT MERGE] Remove BOOL/TRUE/FALSE definitions

Ansariel:
This is more of a way to start a discussion about either removing the definitions for the fake booleans or finding a way to prevent them being sprinkled into the codebase especially by merging releases that are in queue before this maintenance release.

The easiest way to prevent this from happening would be to remove the definitions as suggested in this PR. However, one case of BOOL is used in LLPathingLib being part of the llphysicsextensionsstub 3p library - the viewer builds successfully after manually injecting the BOOL definition into the header it is used. So it would have to be updated and rebuilt. But this also might mean there is the risk of the actual Havok lib being affected as well, which would then mean to also provide an updated version to TPVs.

If the definitions should remain for now, what can be done to prevent further usage of them?

@marchcat @larsnaesbye @nat-goodspeed what are your thoughts/ideas on this?

### DIFF
--- a/indra/llcommon/llpointer.h
+++ b/indra/llcommon/llpointer.h
@@ -83,7 +83,6 @@ public:
 	const Type&	operator*() const				{ return *mPointer; }
 	Type&	operator*()							{ return *mPointer; }
 
-	operator BOOL()  const						{ return (mPointer != NULL); }
 	operator bool()  const						{ return (mPointer != NULL); }
 	bool operator!() const						{ return (mPointer == NULL); }
 	bool isNull() const							{ return (mPointer == NULL); }
@@ -203,7 +202,6 @@ public:
 	const Type*	operator->() const				{ return mPointer; }
 	const Type&	operator*() const				{ return *mPointer; }
 
-	operator BOOL()  const						{ return (mPointer != NULL); }
 	operator bool()  const						{ return (mPointer != NULL); }
 	bool operator!() const						{ return (mPointer == NULL); }
 	bool isNull() const							{ return (mPointer == NULL); }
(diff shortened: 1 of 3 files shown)


larsnaesbye:
Could we move them to a new temporary .h file and include that file only in those places where they are needed for now?

larsnaesbye:
I suppose this is something we should ask @nat-goodspeed about? Can we get a fix for BOOL and friends in a new build of llphysicsextensions_source ?

Ansariel:
At least I applied the changes from this PR to our merge branch in Firestorm, so I will find out if somebody sneaks those pesky fake booleans in again. 😉

marchcat:
> Can we get a fix for BOOL and friends in a new build of llphysicsextensions_source ?

Sounds good for me.

nat-goodspeed:
Getting a new build of the llphysicsextensions family sounds like a good idea, but -- the repo is in a state that makes it really unclear exactly how to build each variant. A few months ago I made a push to replace, in the viewer's autobuild.xml, the last few Windows and Mac package builds still left over from TeamCity with GitHub builds. llphysicsextension is about the only one remaining because of the ambiguity.

larsnaesbye:
OK, alternatively: move the defs to LLPathingLib for now?

marchcat:
This one can be merged after https://github.com/secondlife/viewer/pull/874
Just checked and it builds fine (on Visual studio, at least). Now it's time to figure out how to build the -tpv package.

Ansariel:
We need an updated llphysicsextensionstub lib - that's something different than source (which only LL uses when building the viewer) and llphysicsextensions_tpv which TPVs use as part of the sub-license. If you build the OS version, llphysicsextensions_stub is used.

marchcat:
Right, working on that now.

PR #872, opened 2024-02-22
Replaced an overlooked BOOL in lllegacyatmospherics.h

Ansariel:
This BOOL was sneaky and only showed up in a Linux build


marchcat (approved):
Thanks!

PR #874, opened 2024-02-22
viewer#871 Update llphysicsextensions_source to version 1.0.c178d39

marchcat:


autobuild.xml
```diff
@@ -1451,42 +1459,50 @@
...
               <key>hash_algorithm</key>
               <string>sha1</string>
               <key>url</key>
-              <string>https://api.github.com/repos/secondlife/llphysicsextensions_source/releases/assets/117009336</string>
+              <string>https://api.github.com/repos/secondlife/llphysicsextensions_source/releases/assets/153023528</string>
             </map>
             <key>name</key>
             <string>linux64</string>
           </map>
+          <key>windows</key>
```
Ansariel, line 1471: We can omit 32bit Windows.

And it's shuffling the nodes again! AAAAAAAAAhhhhhhh

marchcat:
Should be better now :)

autobuild.xml
akleshchev: yeah, that's very annoying.

Ansariel:
Do you have looked into this again? It turns out that apparently not only llphysicsextensionstub contains that header, but the TPV Havok library, too. So we would need an updated version on those as well.

marchcat:
> TPV Havok library

Ok, sure. Let's rebuild it as well.

Ansariel:
> > TPV Havok library
> 
> Ok, sure. Let's rebuild it as well.

Just to make sure: It needs an update of

- llphysicsextensions_source
- llphysicsextensions_tpv
- llphysicsextensions_stub

Currently in this PR is only the update for source, which does not help TPVs since we don't have access to the source. We need an updated stub and tpv version.

Ansariel:
2 months later, how's the status on the Havok lib updates?

larsnaesbye:
I'd like to know as well.

Ansariel:
@marchcat Since you asked in #1317: We cannot merge #861 before we finally get the updated Havok libs.

marchcat:
I see, thanks..

We still have some issues with building those packages, will update when ready.

PR #878, opened 2024-02-22
remove another unused old macOS file

larsnaesbye:
This file was partner to the removed classic macOS resource file.


marchcat (approved):
Thank you!

PR #879, opened 2024-02-23
Lua listen_events(), await_event() => get_event_pumps(), get_event_next().

nat-goodspeed:
Don't set up a Lua callback to receive incoming events, a la listen_events().
Don't listen on an arbitrary event pump, a la await_event().

Instead, the new get_event_pumps() entry point simply delivers the reply pump
and command pump names (as listen_events() did) without storing a Lua
callback.

Make LuaListener capture incoming events on the reply pump in a queue. This
avoids the problem of multiple events arriving too quickly for the Lua script
to retrieve. If the queue gets too big, discard the excess instead of blocking
the caller of post().

Then the new get_event_next() entry point retrieves the next (pump, data) pair
from the queue, blocking the Lua script until a suitable event arrives. This
is closer to the use of stdin for a LEAP plugin. It also addresses the
question: what should the Lua script's C++ coroutine do while waiting for an
incoming reply pump event?

Recast llluamanager_test.cpp for this new, more straightforward API.

Move LLLeap's and LuaListener's reply LLEventPump into LLLeapListener, which
they both use. This simplifies LLLeapListener's API, which was a little
convoluted: the caller supplied a connect callback to allow LLLeapListener to
connect some listener to the caller's reply pump. Now, instead, the caller
simply passes a bool(pumpname, data) callback to receive events incoming on
LLLeapListener's own reply pump.

Fix a latent bug in LLLeapListener: if a plugin called listen() more than once
with the same listener name, the new connection would not have been saved.

While at it, replace some older Boost features in LLLeapListener and LLLeap.

### DIFF
--- a/indra/cmake/run_build_test.py
+++ b/indra/cmake/run_build_test.py
@@ -122,19 +122,17 @@ def main(command, arguments=[], libpath=[], vars={}):
     # Make sure we see all relevant output *before* child-process output.
     sys.stdout.flush()
     try:
-        return subprocess.call(command_list)
-    except OSError as err:
+        return subprocess.run(command_list).returncode
+    except FileNotFoundError as err:
         # If the caller is trying to execute a test program that doesn't
         # exist, we want to produce a reasonable error message rather than a
         # traceback. This happens when the build is halted by errors, but
         # CMake tries to proceed with testing anyway <eyeroll/>. However, do
         # NOT attempt to handle any error but "doesn't exist."
-        if err.errno != errno.ENOENT:
-            raise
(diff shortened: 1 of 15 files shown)


maxim-productengine (approved):

nat-goodspeed:
Thank you, Maxim. Before I realized you had already approved, I added a few cleanup commits to lua_function.{h,cpp}, lualistener.h, lleventcoro_test.cpp, debug.h and print.h. Unless you want me to retract or improve these, once the current build succeeds, I'll merge.

maxim-productengine:
Additional changes looks good too, ship it.

PR #883, opened 2024-02-23
Fix for deferred avatar eyes shader failing to link on Intel GPU

brad-linden:
vary_position was being read by the diffuseF.glsl shader for mirror clipping, but had not been written by the vertex shader


Geenz (approved):
Looks good to me.

PR #891, opened 2024-02-26
First mirrors optimization pass

Geenz:
Adds a ton of stuff for conservative updates on mirrors, including per face conservative updates and the like that scale based upon face visibility.  I think the face layout I have may be wrong - but overall all of this combined boosts the framerate on my systems to be well within spec (50% impact on framerate).


Geenz:
Also noticed - this brings us slightly more up to date with main.  I must have accidentally merged up with main in the last couple of weeks - my bad!

indra/newview/pipeline.cpp
```diff
@@ -1068,6 +1075,9 @@ void LLPipeline::refreshCachedSettings()
         LLViewerShaderMgr::instance()->clearShaderCache();
         LLViewerShaderMgr::instance()->setShaders();
     }
+    RenderHeroProbeUpdateRate = gSavedSettings.getS32("RenderHeroProbeUpdateRate");
+    RenderHeroProbeConservativeUpdateMultiplier = gSavedSettings.getS32("RenderHeroProbeConservativeUpdateMultiplier");
+
```
RunitaiLinden, line 1080: Prefer to not use connectRefreshCachedSettings and statics, but use static LLCachedControl.

RunitaiLinden (approved):
Approved with comment... not a blocker, but probably refrain from using connectRefreshCachedSettings and instead use static LLCachedControl.

brad-linden (approved):

indra/newview/pipeline.cpp
```diff
@@ -786,9 +790,12 @@ bool LLPipeline::allocateScreenBuffer(U32 resX, U32 resY, U32 samples)
...
         allocateScreenBuffer(res, res, samples);
 
-		res = mHeroProbeManager.mProbeResolution; // We also scale the hero probe RT to the probe res since we don't super sample it.
-        mRT = &mHeroProbeRT;
-        allocateScreenBuffer(res, res, samples);
+		if (RenderMirrors)
+        {
+            res = mHeroProbeManager.mProbeResolution;  // We also scale the hero probe RT to the probe res since we don't super sample it.
+            mRT = &mHeroProbeRT;
+            allocateScreenBuffer(res, res, samples);
```
cosmic-linden, line 797: Not a blocker but I notice that the hero probe target isn't released when RenderMirrors is set to false.

cosmic-linden (approved):

PR #895, opened 2024-02-26
Generate release notes from tag `release` to the built tag.

nat-goodspeed:
Explicitly specify both `target_commitish` and `previous_tag` inputs to the GitHub release-notes generator. Unless they're both explicitly passed, action-gh-release guesses wrong about the range of commits to consider for release notes generation.

This change uses the new action-gh-release feature `previous_tag` (secondlife-3p/action-gh-release#2). Once that PR is merged, we can revert to referencing `secondlife-3p/action-gh-release@v1`.

This behavior also depends on maintaining a new viewer tag `release` to track the most recent release, updating it every time we promote a new default download viewer.

Also add cross-links between the build page and the corresponding release page.

relnotes:
This text should be copied to the release page.

It should include everything after the `relnotes:` line.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -382,7 +382,7 @@ jobs:
             ${{ needs.build.outputs.relnotes }}
           prerelease: true
           generate_release_notes: true
-          target_commitish: ${{ github.ref }}
+          target_commitish: ${{ github.ref_name }}
           append_body: true
           fail_on_unmatched_files: true
           files: |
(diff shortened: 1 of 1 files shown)


nat-goodspeed:
Note that [the build page](https://github.com/secondlife/viewer/actions/runs/8055448524) has a link to [the release page](https://github.com/secondlife/viewer/releases/tag/Second_Life_Test%2388ebb92), and vice-versa. (Scroll down on the release page.)

This PR did not yet exist at the time of that tag build, which is why the `relnotes:` text wasn't found or copied.

marchcat (approved):

nat-goodspeed:
I'll be interested to see the next tag build.

marchcat:
I'll likely build one tonight.

PR #898, opened 2024-02-27
BUG-234595/Jira-public-Issue#11494 Mac texture loading/unloading

beqjanus:
This removes the "special case" for Mac which was resulting in values a few orders of magnitude smaller than on windows.

My guess is that the MacOS case was written/tested on HiDpi which allows the values to be close enough to Windows/Linux that the effect is not noticed (hence it only affects some Mac users). However, the fact that there was ever a "special" handler makes me wonder if by removing it I am reverting a previous issue that was not clear from the commit history. 

TL;DR This works, tested by Intel and ARM Mac users who had seen the issue. No side effects have been identified so far.


marchcat (approved):
Thank you!

AtlasLinden:
Archived internal issue: https://github.com/secondlife/jira-archive-internal/issues/70923

PR #903, opened 2024-02-28
triage#86 Use Emoji font in LineEditor by default

LLGuru:
We can replace font SansSerif with Emoji for the whole UI
The difference between these fonts is only glyphs for emojis

Using SansSerifSmall in LineEditor:
image

Using EmojiSmall in LineEditor:
image

### DIFF
--- a/indra/llui/llui.cpp
+++ b/indra/llui/llui.cpp
@@ -592,11 +592,13 @@ namespace LLInitParam
 	
 	void ParamValue<const LLFontGL*>::updateBlockFromValue(bool make_block_authoritative)
 	{
-		if (getValue())
+		// It is important to use getRawValue() instead of getValue() here
+		// because getValue() calls validateBlock() and updateValueFromBlock()
+		if (getRawValue())
 		{
-			name.set(LLFontGL::nameFromFont(getValue()), make_block_authoritative);
-			size.set(LLFontGL::sizeFromFont(getValue()), make_block_authoritative);
-			style.set(LLFontGL::getStringFromStyle(getValue()->getFontDesc().getStyle()), make_block_authoritative);
+			name.set(LLFontGL::nameFromFont(getRawValue()), make_block_authoritative);
+			size.set(LLFontGL::sizeFromFont(getRawValue()), make_block_authoritative);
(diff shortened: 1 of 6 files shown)


marchcat (approved):
This looks good. But please wait for the merges and then we'll decide where to put the change.

indra/llui/llui.cpp
```diff
@@ -592,11 +592,13 @@ namespace LLInitParam
 	
 	void ParamValue<const LLFontGL*>::updateBlockFromValue(bool make_block_authoritative)
 	{
-		if (getValue())
+		// It is important to use getRawValue() instead of getValue() here
+		// because getValue() calls validateBlock() and updateValueFromBlock()
+		if (getRawValue())
```
akleshchev, line 597: This is clearly wrong. if you have to avoid validation you are likely missing something.

akleshchev (changes requested):

indra/llui/llui.cpp
akleshchev: This breaks TOS font size:
image

akleshchev: > updateBlockFromValue() serves to copy Value to Block

My understanding is that block's values take priority over value and it is correct unless programmer specificaly chose to override value from xml.

LLGuru: I don't think that some of these terms could have a priority
And when we use these two methods we shouldn't think of priority
When we explicitly call `updateBlockFromValue()` we expect that Value should be copied to Block and not vice versa

akleshchev: Note how all other updateBlockFromValue are implemented - you are making an outliner. Besides this is what broke TOS and font test floaters.

marchcat: Does this affect other floaters?

akleshchev: May be some of preview floaters, trackbal, may be some others, haven't checked.
In my opinion 'font test' floater sudenly not behaving as it should is enough of a reason to request a change.

akleshchev: It does affect a lot more, check the W, N, E, S leters:
image

marchcat: Right. "Font test" is named that way for a reason.

PR #905, opened 2024-02-29
 Refactor `require()` to make it easier to reason about Lua stack usage.

nat-goodspeed:
Push throwing Lua errors down into `LLRequireResolver::findModule()` and
`findModuleImpl()` so their callers don't have to handle the error case. That
eliminates `finishrequire()`.

`require()` itself now only retrieves (and pops) the passed module name and
calls `LLRequireResolver::resolveRequire()` to do the actual work.

`resolveRequire()` is now `void`. It only instantiates `LLRequireResolver` and calls
its `findModule()` method.

`findModule()` is now also `void`. It's guaranteed to either push the loaded Lua
module or throw a Lua error. In particular, when `findModuleImpl()` cannot find
the specified module, `findModule()` throws an error. That replaces
`ModuleStatus::NotFound`.

Since `std::filesystem::path::append()` aka `operator/()` detects when its right
operand is absolute and, in that case, discards the left operand, we no longer
need `resolveAndStoreDefaultPaths()`: we can just invoke that operation inline.

When `findModule()` pushes `_MODULES` on the Lua stack, it uses `LuaRemover` (below)
to ensure that `_MODULES` is removed again no matter how `findModules()` exits.

`findModuleImpl()` now accepts the candidate pathname as its argument. That
eliminates `mAbsolutePath`.

`findModuleImpl()` now returns only `bool`: `true` means the module was found and
loaded and pushed on the Lua stack, `false` means not found and nothing was
pushed; no return means an error was reported.

Push running a newly found module's source file down into `findModuleImpl()`.
That eliminates the distinction between `Cached` and `FileRead`, which obviates
`ModuleStatus`: a `true` return means either "previously cached" or "we read it,
compiled it, loaded it and ran it." That also eliminates the need to store the
module's textual content in `mSourceCode`.

Similarly, once loading the module succeeds, `findModuleImpl()` caches it in
`_MODULES` right away. That eliminates `ResolvedRequire` since we need not pass
the full pathname of the found module (or its contents) back up through the
call chain.

Move `require()` code that runs the new module into private `runModule()`method,
called by `findModuleImpl()` in the not-cached case. `runModule()` is the only
remaining method that can push either a string error message or the desired
module, because of its funny stack manipulations. That means the check for a
string error message on the stack top can move down to `findModuleImpl()`.

Add `LuaRemover` class to ensure that on exit from some particular C++ block,
the specified Lua stack entry will definitely be removed. This is different
from `LuaPopper` in that it engages `lua_remove()` rather than `lua_pop()`.

Also ditch obsolete `await_event()` Lua entry point.

### DIFF
--- a/indra/newview/llluamanager.cpp
+++ b/indra/newview/llluamanager.cpp
@@ -195,30 +195,6 @@ lua_function(get_event_next,
     return 2;
 }
 
-lua_function(await_event,
-             "await_event(pumpname [, timeout [, value to return if timeout (default nil)]]):\n"
-             "pause the running Lua chunk until the next event on the named LLEventPump")
-{
-    auto pumpname{ lua_tostdstring(L, 1) };
-    LLSD result;
-    if (lua_gettop(L) > 1)
-    {
-        auto timeout{ lua_tonumber(L, 2) };
-        // with no 3rd argument, should be LLSD()
-        auto dftval{ lua_tollsd(L, 3) };
-        lua_settop(L, 0);
-        result = llcoro::suspendUntilEventOnWithTimeout(pumpname, timeout, dftval);
-    }
-    else
-    {
-        // no timeout
-        lua_pop(L, 1);
-        result = llcoro::suspendUntilEventOn(pumpname);
-    }
-    lua_pushllsd(L, result);
-    return 1;
-}
-
 void LLLUAmanager::runScriptFile(const std::string& filename, script_finished_fn cb)
 {
     // A script_finished_fn is used to initialize the LuaState.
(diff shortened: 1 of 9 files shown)


indra/newview/llluamanager.cpp
```diff
@@ -383,31 +381,59 @@ LLRequireResolver::LLRequireResolver(lua_State *L, const std::string& path) : mP
...
 }
 
-[[nodiscard]] LLRequireResolver::ResolvedRequire LLRequireResolver::resolveRequire(lua_State *L, std::string path)
+/**
+ * Remove a particular stack index on exit from enclosing scope.
+ * If you pass a negative index (meaning relative to the current stack top),
+ * converts to an absolute index. The point of LuaRemover is to remove the
+ * entry at the specified index regardless of subsequent pushes to the stack.
+ */
+class LuaRemover
```
maxim-productengine, line 390: Ok, I like this addittion for clearing the stack. _MODULES is removed much cleaner now.

PR #914, opened 2024-03-01
viewer#811 win: Move XML files to the resource

LLGuru:
In replacement of https://github.com/secondlife/viewer/pull/817

### DIFF
--- a/indra/newview/viewer_manifest.py
+++ b/indra/newview/viewer_manifest.py
@@ -59,6 +59,10 @@ class ViewerManifest(LLManifest):
         # files during the build (see copy_w_viewer_manifest
         # and copy_l_viewer_manifest targets)
         return 'package' in self.args['actions']
+
+    # This differs for Windows, it delivers XML files as resources
+    def is_packaging_xml_files(self):
+        return True;
     
     def construct(self):
         super(ViewerManifest, self).construct()
@@ -67,10 +71,12 @@ class ViewerManifest(LLManifest):
 
         if self.is_packaging_viewer():
             with self.prefix(src_dst="app_settings"):
-                self.exclude("logcontrol.xml")
-                self.exclude("logcontrol-dev.xml")
+                # XML files are delivered as resources on Windows
+                if self.is_packaging_xml_files():
+                    self.exclude("logcontrol.xml")
+                    self.exclude("logcontrol-dev.xml")
+                    self.path("*.xml")
                 self.path("*.ini")
-                self.path("*.xml")
 
                 # include the entire shaders directory recursively
                 self.path("shaders")
@@ -136,8 +142,10 @@ class ViewerManifest(LLManifest):
 
 
             with self.prefix(src_dst="character"):
+                # XML files are delivered as resources on Windows
+                if self.is_packaging_xml_files():
+                    self.path("*.xml")
                 self.path("*.llm")
-                self.path("*.xml")
                 self.path("*.tga")
 
             # Include our fonts
@@ -155,9 +163,12 @@ class ViewerManifest(LLManifest):
                             self.path("*.j2c")
                             self.path("*.png")
                             self.path("textures.xml")
-                    self.path("*/xui/*/*.xml")
-                    self.path("*/xui/*/widgets/*.xml")
-                    self.path("*/*.xml")
+
+                    # XML files are delivered as resources on Windows
+                    if self.is_packaging_xml_files():
+                        self.path("*/xui/*/*.xml")
+                        self.path("*/xui/*/widgets/*.xml")
+                        self.path("*/*.xml")
 
                     # Update: 2017-11-01 CP Now we store app code in the html folder
                     #         Initially the HTML/JS code to render equirectangular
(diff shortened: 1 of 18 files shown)


indra/newview/viewer_manifest.py
```diff
@@ -492,6 +503,10 @@ def test_for_no_msvcrt_manifest_and_copy_action(self, src, dst):
         else:
             print("Doesn't exist:", src)
         
+    # This differs for Windows, it delivers XML files as resources
+    def is_packaging_xml_files(self):
+        return False;
```
Ansariel, line 508: First thing to set to return True after merging into Firestorm...

akleshchev: Can you specify reasons? Benefits are a bit questionable given that everything else stays as separate files, but I don't think there are any disadvantages either as long as preview and localization tools keep working. I feel like moving skin files into resources can be a potential issue since people will have to get repo's files now, but I don't know that for sure, agruably you should be working with viewer's repo if you are translating.

Ansariel: Can you specify how this should fix issues with the installer or the users anti virus tool borking installations, given that DLLs might also not be installed properly and preventing the viewer from starting. Or SLPLugin.exe regularly getting flagged as malicious and deleted/moved to a quarantine folder?

This just seems to prevent from users customizing the viewer, e.g. dropping in custom skins they created, or simply update translations which will be done by LL either once in a blue moon, might not fit or are simply wrong. I will for sure undo this.

akleshchev: > Can you specify how this should fix issues with the installer

That was my argument as well. It does make viewer slightly more reliable (ex: against things that would modify text files and warranties that warning strings load), but most of the time it doesn't fix anything and just bumps problem to the next file (.msg in this case) instead of fixing our install process and validation.

> This just seems to prevent from users customizing the viewer

Does it actually happen? This code is supposed (haven't checked it yet) to load local files before resources so it should still be possible to have custom files. To me this change does feel like a disadvantage in terms of customizability, but I don't know if anyone actually modifies those files and won't fail a review on a feeling.

Ansariel: > > This just seems to prevent from users customizing the viewer
> 
> Does it actually happen? This code is supposed (haven't checked it yet) to load local files before resources so it should still be possible to have custom files. To me this change does feel like a disadvantage in terms of customizability, but I don't know if anyone actually modifies those files and won't fail a review on a feeling.

Yes. Or change shortcuts in the menu because they don't like them.  With that change you would have to checkout or download a zipped version of the repository and make sure you grab the files for the correct version seems a bit... cumbersome.

vir-linden: This fix is specific to windows. Is the problem specific to windows?

vir-linden: I'm skeptical of the need to handle this case rather than just warning about it. How hard would it be to just detect the missing file case and tell the user to reinstall?

akleshchev: > How hard would it be to just detect the missing file case and tell the user to reinstall?

That's what I did in DRTVWR-591, [here](https://github.com/secondlife/viewer/commit/51088cde7f5a0bdaf9249bfdd5d31b9b212403ab)

> This fix is specific to windows. Is the problem specific to windows?

Mostly, it's overwhelmingly more frequent on windows with couple cases on MacOS.

nat-goodspeed (commented):
Is there any reason to suspect that missing-file crashes are due to any cause other than the user removing them? Is there some agent at work (anti-virus?) that might remove files from the installed viewer directory without the user's awareness?

Otherwise I'm inclined to say: you messed up your install. Reinstall and sin no more.

indra/newview/viewer_manifest.py
LLGuru: @Ansariel , for those users who want to experiment with XML files without checking out the repository I would like to propose the feature "Export XML files", some button which will unpack all XML contents from resources to external files in proper directories

LLGuru: As you can see, if there is a file in the directory and the correspondent resource, the file has a priority, and it will be used, not the resource
This is done specially for developers working with XML files at run-time

LLGuru:
> this is a desirable change

@nat-goodspeed , looking how often the viewer reads any XML files at run-time, I think that this change could improve the performance as well as data safety, and even extend the HDD/SSD service life :-)

sldevel:
> looking how often the viewer reads any XML files at run-time, I think that this change could improve the performance as well as data safety, and even extend the HDD/SSD service life :-)

Thing is, this is an idiosyncratic (Windows only) implementation, and I personally do not like either the idea that the advanced users would loose direct access to the XUI files to perform live modifications to them (I always do this when I modify a floater or panel: I modify the XML file in an editor and reopen the corresponding floater to see if the result is good, then adjust again, etc, in quick modify/test successions).

If your only aim is to speed up thing, then you could cache the parsed XML file result and reuse that result instead of re-parsing it each time, using a static unordered_map with the HBXX128::digest(full_path_file_name) hash (an LLUUID) as the key, and as the value, a
struct { void* parsed_buffer; time_t last_file_write_date; }
When trying to parse a file, first lookup into your static map if you already parsed that file, and if yes, also check it has not since been written to (via the stored timestamp). In the favorable case, you can simply reuse the parsed data as is, else you need to (re-)parse the file.

PR #915, opened 2024-03-01
Convert remaining TRUE/FALSE to true/false

Ansariel:
@marchcat Apparently there were some more sneaky TRUEs and FALSEs hidden in the code. That should be really all now - hopefully.


marchcat (approved):
Wow, thank you! 
How it worked when we disabled the defines?

PR #918, opened 2024-03-01
Revert the "Update emoji_categories in Polish translation"

PanteraPolnocy:
The reason of reverting this one is the fact, that the grammar errors and typos - not only in Polish language, but in German and all the rest - are all around 3P packages as well, so updating category names in the viewer will break the connection with the 3P. Who had the bright idea of designing it like that and not exposing it for translators to correct? Not to mention, why on earth the EMOJI CODES are translated as well, instead of - for example - tooltip names? Everyone in Poland would look for :potato: and not a :ziemniak:


PanteraPolnocy:
... even Github changes : potato : to an actual :potato:

PanteraPolnocy:
For reference, I found this. https://github.com/secondlife/jira-archive/issues/11824

akleshchev (approved):

marchcat (approved):

akleshchev:
Yeah, bad design, git fails to merge this, will do a separate commit.

marchcat:
@PanteraPolnocy Thank you for noticing this problem, we'll fix this in the upcoming release.

PanteraPolnocy:
@marchcat I cannot take creadit here, as it was @Ansariel who pointed that issue out when I grumbled about the state of this file in our repository; She warned, that 3p packages will be a problem here.

akleshchev:
And here I though that thing was fixed...

PR #920, opened 2024-03-01
GameControl: Flycam mode works

AndrewMeadows:
This PR moves Flycam and Avatar motion support to GameControl input.

The old Flycam feature through the SpaceMouse system has been restricted to only work for SpaceMouse and other 3Dconnexion devices, not general gamepads.  Gamepad support now uses GameControl.

Also, slightly simpler UI for mapping GameControl inputs.


simon-linden (approved):
Let's get it all together

PR #922, opened 2024-03-02
Strongly attenuate the blue hue seen on shiny materials

sldevel:
This a proposal for a workaround to attenuate the "ugly blue hue" badly impacting shiny materials (PBR and legacy alike) when they are not shielded from the sky by a reflection probe (meaning all shinies all over SL for now, since very few places only started the conversion to PBR and reflection probes).

It got the immense advantage of not needing to touch the existing environment settings (which won't suffice anyway to remove the blue hue, short of making the sky render grey).

The "strength" of this workaround can be adjusted via two new debug settings: "RenderSkyReflectionDesaturation" and "RenderSkyReflectionDarkening". They both affect the "blue horizon" and "blue density" colors as passed to shaders when rendering the reflections irradiance maps. "RenderSkyReflectionDesaturation" desaturates the colors (the higher the setting, the more desaturated; setting this to 1.0 or lower disables entirely the workaround), while "RenderSkyReflectionDarkening" darkens them. With the default values (2.0 and 1.5 respectively), I get acceptable/tolerable shinies in sky boxes or on sky platforms not shielded from the sky with a reflection probe, while not impacting too much outdoors mirror-like PBR materials which are reflecting the sky.

Note that this also improves a lot the water bodies color, making them appear less "sky-blue" and therefore more realistic.

### DIFF
--- a/indra/llmath/v3color.h
+++ b/indra/llmath/v3color.h
@@ -77,6 +77,17 @@ public:
 
 	void setHSL(F32 hue, F32 saturation, F32 luminance);
 	void calcHSL(F32* hue, F32* saturation, F32* luminance) const;
+
+    // Desaturates the color by a given factor, also optionnally darkens it. HB
+    inline void desaturate(F32 factor, F32 darkening = 1.f)
+    {
+        if (factor > 1.f)
+        {
+            F32 h, s, l;
+            calcHSL(&h, &s, &l);
+            setHSL(h, s / factor, l / llmax(darkening, 1.f));
+        }
+    }
 	
 	const LLColor3&	setToBlack();					// Clears LLColor3 to (0, 0, 0)
 	const LLColor3&	setToWhite();					// Zero LLColor3 to (0, 0, 0)
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -10744,6 +10744,29 @@
     <key>Value</key>
     <real>0.7</real>
   </map>
+
+  <key>RenderSkyReflectionDarkening</key>
+    <map>
+    <key>Comment</key>
+    <string>Darkening factor for the colors fed to the reflection probes shaders to attenuate the blue hue. 1 or less to disable.</string>
+    <key>Persist</key>
+    <integer>1</integer>
+    <key>Type</key>
+    <string>F32</string>
+    <key>Value</key>
+    <real>1.5</real>
+    </map>
+  <key>RenderSkyReflectionDesaturation</key>
+    <map>
+    <key>Comment</key>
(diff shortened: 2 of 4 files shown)


Jenna-Huntsman:
This patch might be a bit overreaching as it stands right now as it affects all sky settings, regardless of if they are built for PBR (i.e. reflection probe ambiance of not-0), but hooking it into the Auto-Adjust Legacy functions may be better.

sldevel:
> This patch might be a bit overreaching as it stands right now as it affects all sky settings, regardless of if they are built for PBR (i.e. reflection probe ambiance of not-0), but hooking it into the Auto-Adjust Legacy functions may be better. (This would still mean that existing presets don't need to be touched, but new built-for-PBR presets don't get stamped on)

The very purpose of this patch is precisely to render all "new built-for-PBR presets" needless: the "fixes" in these presets do not even work anyway, like I already demonstrated (see the second screenshot in my reply to Dan Linden in [this feedback topic](https://feedback.secondlife.com/bug-reports/p/pbr-blue-mesh)).

We urgently need something that makes PBR viewers render properly all existing SL scenes without a need to fiddle with existing environment settings, and to prevent creator to design "fixed" settings that would only make things worst in the end, once the renderer fully fixed.
I however do not consider this workaround as a definitive fix; this is just what I came up with and makes PBR (barely) acceptable for me in most SL scenes, but I'm no OpenGL renderer and shader expert, by far, and there are definitely things that could be done at that the shader level too (i.e. modify the shaders used for reflections rendering so that they account for the EE settings "scaling" that is required to avoid ugly blue hue (or red, if your sky is red: Mars parcel anyone ?)). Also, this workaround could be exponentially toned down for faces acting like mirrors (faces 100% metallic or close, and 0% rough or close), so that the sky is still perfectly reflected on those, but again this would need to be done at the shader level, using the draw info for each face...

You can see this workaround at play in the Cool VL Viewer v1.32.0.12 I released yesterday, and can test your "new built-for-PBR presets" with it as well (feedback welcome).

RunitaiLinden:
Closed and won't merge.  We need less color fudging, not more.  The color fudging of legacy skies has been backed out in maint, should be going to release soon.  The blue sheen has also been fixed in maint by modifying the new default midday (which somehow got blown out to be unrealistically bright).  The bug reports I'm getting from FS beta are actually trending the other way (not blue enough).  I verified locally that the lighting model is correct compared to substance by importing an HDRI to make skies match on both sides, and I'm putting together a "local HDRI preview" mode for artists to duplicate the lighting from Substance in SL so they can have a WYSIWYG experience without having to stand on their head and recite the alphabet backwards.

Once that's done, it should alleviate the need for a new default midday that approximates an HDRI, and then we can look at whether or not the new default midday is still needed at all.

PR #923, opened 2024-03-03
more BOOL to bool (pt. 7?)

larsnaesbye:
Converting a bunch of viewer functions and constants that have BOOL, and internal use of FALSE and TRUE. Tests run.


Ansariel:
Those files aren't use anymore at all. They should be deleted instead.

larsnaesbye:
Well in that case that's what's sorely needed now. Purging all unused code!

PR #927, opened 2024-03-04
PBR Terrain UI second pass

cosmic-linden:


indra/newview/llfloaterregioninfo.cpp
```diff
@@ -1554,31 +1529,96 @@ bool LLPanelRegionTerrainInfo::refreshFromRegion(LLViewerRegion* region)
...
-				LLUUID tmp_id(compp->getDetailAssetID(i));
-				asset_ctrl->setImageAssetID(tmp_id);
-			}
-		}
+        if (set_texture_swatches)
+        {
+            for(S32 i = 0; i < LLTerrainMaterials::ASSET_COUNT; ++i)
+            {
+                buffer = llformat("texture_detail_%d", i);
+                asset_ctrl = getChild<LLTextureCtrl>(buffer);
```
akleshchev, line 1573: I think it would have been nicer to have local variables for texture controls and use those instead of doing relatively expensive search on each refresh.

akleshchev (approved):

cosmic-linden:
@akleshchev You bring up some good points, in particular `onSimulatorFeaturesReceived`, which I am pretty sure isn't doing anything useful and ought to be cleaned up. Will merge this for now and continue work in a follow-up PR.

PR #929, opened 2024-03-05
Do not automatically close issues

bennettgoble:
Mark issues as stale but do not close them.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -17,7 +17,8 @@ jobs:
         with:
           stale-pr-message: This pull request is stale because it has been open 30 days with no activity. Remove stale label or comment or it will be closed in 7 days
           days-before-stale: 30
-          days-before-close: 7 
+          days-before-close: 7
+          days-before-issue-close: -1
           exempt-pr-labels: blocked,must,should,keep
           stale-pr-label: stale
       - name: Print outputs


Geenz (approved):
Looks good to me.

.github/workflows/stale.yaml
```diff
@@ -17,7 +17,8 @@ jobs:
         with:
           stale-pr-message: This pull request is stale because it has been open 30 days with no activity. Remove stale label or comment or it will be closed in 7 days
```
brad-linden, line 18: I think we should add a `stale-issue-message` as well.  a lot of the confusion here was that it was not at all clear that the bot was operating on this issue

marchcat (approved):
This will be useful, thanks!

brad-linden (approved):
looks good with a suggested improvement

.github/workflows/stale.yaml
bennettgoble: Should there be a `stale-issue-message` even if nothing happens after the label is added?

brad-linden: I don't know, but in this case things happened on the issue (it was referenced in commits and PRs) and it was closed anyways.  so some paper trail seems warranted

brad-linden: I guess that doesn't matter so much if we are no longer closing them however

PR #941, opened 2024-03-07
viewer-private#41 Texture fail to apply to materials in some cases

akleshchev:
Shift-dropping textures can fail if one of 'early' faces has nomod material

### DIFF
--- a/indra/newview/lltooldraganddrop.cpp
+++ b/indra/newview/lltooldraganddrop.cpp
@@ -1077,51 +1077,71 @@ void set_texture_to_material(LLViewerObject* hit_obj,
                              LLGLTFMaterial::TextureInfo drop_channel)
 {
     LLTextureEntry* te = hit_obj->getTE(hit_face);
-    if (te)
+    if (!te)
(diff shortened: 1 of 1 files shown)


marchcat (approved):

maxim-productengine (approved):

indra/newview/lltooldraganddrop.cpp
```diff
@@ -1077,51 +1077,71 @@ void set_texture_to_material(LLViewerObject* hit_obj,
...
-            case LLGLTFMaterial::GLTF_TEXTURE_INFO_BASE_COLOR:
-            default:
-                {
-                    material->setBaseColorId(asset_id);
-                }
-                break;
+    if (hit_obj->isInventoryDirty() && hit_obj->hasInventoryListeners())
+    {
+        hit_obj->requestInventory();
+        return;
```
akleshchev, line 1094: Should this just fall through or cancel? We do the same check when applying limited permission items, so it isn't without precendent.

Issue #681, opened 2024-01-26
Mirrors quality pass


Geenz:

Dan-Linden:
Please add a test plan.

Dan-Linden:
This was a Developer task.
Geenz says this can be closed.

PR #942, opened 2024-03-07
Mirrors quality pass 1

Geenz:
Closes out #681

Quality!  In this case, probe blending and a few tweaks here and there to make mirrors generally "make sense" when a mirror probe is placed.  Supports box and sphere probes, with some tweaked falloff for convenience.  Also better generalizes our existing intersection code for boxes and spheres.


RunitaiLinden (approved):

indra/newview/llreflectionmap.cpp
```diff
@@ -166,7 +166,7 @@ void LLReflectionMap::autoAdjustOrigin()
             
         }
     }
-    else if (mViewerObject)
+    else if (mViewerObject && !mViewerObject->isDead())
```
cosmic-linden, line 169: More `isDead` checks is likely good, and could probably be propagated to other `mViewerObject` null checks in this class, such as `LLReflectionMap::isRelevant`

indra/newview/llreflectionmapmanager.cpp
```diff
@@ -906,6 +906,8 @@ void LLReflectionMapManager::updateUniforms()
         // the box probe
         LLMatrix4 refBox[LL_MAX_REFLECTION_PROBE_COUNT];
 
+        LLMatrix4 heroBox;
```
cosmic-linden, line 909: Just a random thought. iirc LLMatrix4 uses OpenGL matrix conventions, but there is something different about the matrix calculations (floating point precision-related, possibly). I wonder if there is value in a convention of using glm data types when we know the data is about to be passed directly to the shader. I don't have a compelling use case for that at this time, though.

cosmic-linden (approved):
LGTM. Just a few random thoughts which may not be actionable at this time.

PR #943, opened 2024-03-07
Fixed a few std::string Windows complaints and merged main

simon-linden:
Fixed a few std::string usage complaints on Windows and merged main branch


nat-goodspeed (changes requested):
All of these references look like excellent use cases for [std::string::back()](https://en.cppreference.com/w/cpp/string/basic_string/back).

simon-linden:
> All of these references look like excellent use cases for [std::string::back()](https://en.cppreference.com/w/cpp/string/basic_string/back).

Good point, I changed those

PR #950, opened 2024-03-08
Remove gender-based restrictions on shape and physics sliders

JamesFoxpaws:
This patch removes the artificial restriction on the ability to create shapes and physics. There are many cases, regardless of your identity, that having access to all possible shape sliders can be used to fine-tune your appearance. I have been working with this modification for about two years now without issue.

The 'Male' and 'Female' radio buttons have instead been replaced with a single slider named 'Gender' which control the `shape_body` group. The overall affect of this new slider controls the shape and position of your shoulders, as well as secondary effects related to arm length and overall height. Normally you would have this set to 0 for 'Female' and 100 for 'Male', but values in-between have been proven to be immensely useful for designing interesting shapes. For example, I am often around 88% Male on a Jake body and maybe around 8% Male on an Reborn.

A decision will have to be made as to what the formerly Male and Female 'Torso Muscle' sliders should be named as currently you will have both with identical names. 'Torso Overtone' and 'Torso Undertone' may be more appropriate names, but I did not feel comfortable making this decision due to localization considerations. 

Furthermore, creation of a 'New Shape' will default to the female shape values. Something like 'New Shape (M)' and 'New Shape (F)' would be an adequate solution, but is beyond the scope of this PR.

Finally, if a user sets a Gender value to somewhere in-between 0 and 100, it is very likely you will be misgendered by furniture. I highly recommend the experience.

DEMO VIDEO: https://i.imgur.com/TfCdgEo.mp4
(Ignore the shape juddering half-way through, that is simply 'Undo Changes' being clicked and is normal)

closes https://github.com/secondlife/viewer/issues/949


JamesFoxpaws:

marchcat (approved):

PR #952, opened 2024-03-08
viewer#944 Make 'Debug Unicode' feature available for all LLViews

LLGuru:
Main steps:
1. Define functions getFont() and getText() in the class LLView to generalize their interface
2. Implement DebugUnicode feature in the class LLView using functions getName(), getFont() and getText()
3. Add to the tooltip debug output the control name and the font name and size if available

This works on text blocks and on buttons the same way as follows:
image
image
image

After fixing the issue (the 2nd commit):
image
image
image

### DIFF
--- a/indra/llui/llviewmodel.cpp
+++ b/indra/llui/llviewmodel.cpp
@@ -81,7 +81,7 @@ void LLTextViewModel::setValue(const LLSD& value)
 {
 	// approximate LLSD storage usage
 	LLViewModel::setValue(value);
-    mDisplay = utf8str_to_wstring(value.asString());
+    mDisplay = utf8str_to_wstring(mStringValue = value.asString());
 
     // mDisplay and mValue agree
     mUpdateFromDisplay = false;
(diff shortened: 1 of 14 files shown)


indra/llui/llviewmodel.cpp
```diff
@@ -101,23 +101,32 @@ void LLTextViewModel::setDisplay(const LLWString& value)
     mUpdateFromDisplay = true;
 }
 
+#define UPDATE_FROM_DISPLAY_IF_NEEDED                                   \
```
marchcat, line 104: Please don't.

marchcat (changes requested):

indra/llui/llviewmodel.cpp
LLGuru: As you wish! 👌

marchcat: Much better now, thanks!

marchcat (approved):

PR #956, opened 2024-03-09
Allow fading emojis characters in fading text.

sldevel:
In the recent emojis implementation, the text color alpha is ignored (emojis are always rendered with an opaque white color), causing them to fail and fade properly with the rest of the text they are printed with.

This trivial patch fixes this issue.


marchcat:
Thanks for the contribution!

marchcat (approved):

PR #957, opened 2024-03-09
Restore how script dialogs used to render by prioritizing non-color fallback fonts.

sldevel:
With the emojis support, a new font was added, which not only provides emojis but also fancy colorful replacements for UTF-8 characters that used to be supported by our fallback (monochrome) fonts: this causes discrepancies and unwanted/undesired changes in scripted objects menus (e.g. an empty circle or square may render as a black, full one, a heart may render red instead of white), not to mention the larger font size used by the emoji characters...

This patch restores the aspect of such menus/dialogs/UI elements with UTF-8 characters that *are* supported by the usual fallback fonts (fonts which may also vary from one viewer to another, and from one OS to another), so that everything keeps working/rendering as it always did so far, while not impairing the use of new colorful emojis.

Note: this should also fix https://github.com/secondlife/viewer/issues/944

### DIFF
--- a/indra/llrender/llfontfreetype.cpp
+++ b/indra/llrender/llfontfreetype.cpp
@@ -356,9 +356,7 @@ void LLFontFreetype::clearFontStreams()
 
 void LLFontFreetype::addFallbackFont(const LLPointer<LLFontFreetype>& fallback_font, const char_functor_t& functor)
 {
-	// Insert functor fallbacks before generic fallbacks
-	mFallbackFonts.insert((functor) ? std::find_if(mFallbackFonts.begin(), mFallbackFonts.end(), [](const fallback_font_t& fe) { return !fe.second; }) : mFallbackFonts.end(),
-	                      std::make_pair(fallback_font, functor));
+	mFallbackFonts.emplace_back(fallback_font, functor);
 }
 
 F32 LLFontFreetype::getLineHeight() const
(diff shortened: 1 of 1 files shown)


marchcat:
Thank you!

marchcat (approved):

cosmic-linden:
Behavior of graphical vs non-graphical emojis is covered by the Unicode standard here:

https://www.unicode.org/reports/tr51/#Presentation_Style

I don't fully grok Henri's change from my quick glance, but I get the impression that it is arbitrarily selecting between textual and emoji representation based on the contents of the specific font, without checking the presentation selector. If that is the case, then that may not be Unicode compliant, depending on the contents of the fonts. If the viewer cannot distinguish unicode characters by textual representation, that would be a limitation of the viewer's font system, in which case that would need to be fixed to fully address this bug.

I'm not fully sure what's going on here, because wide chars are being used, which has some differences from UTF-8 that I don't fully understand.

That said, even if that is the case, I think this could be an OK fix in the short term if the proper fix is difficult, but a proper fix would be needed to ensure interoperability with other software, such as copy and paste and web services.

The ideal fix for this change, as far as I understand it, would be to default to textual representation of emojis in all SL UIs when no presentation selector is provided. That would be a sensible change given the history of the platform, especially when LSL comes into play, while still allowing for the new and shiny colorful emojis. In order to support this change, colored emojis from the emoji picker should include the emoji presentation selector, if they are not doing so already (`U+FE0F` in Unicode/UTF-8).

Default emoji presentation varies, even in the web browser. Off the top of my head, emoji without an emoji presentation selector can have variable appearance in browser tabs in Chrome vs Firefox, when a zero-width joiner emoji is involved. I think also that some

sldevel:
This patch is just a _restoration_ of how things used to work in SL (and LSL) before the emojis fonts introduction, the latter causing some UTF-8 characters to be rendered differently (in color, instead of in a monochrome font, and with arbitrary colors, at that, which results in bad surprises in LSL dialogs/menus that used to render fine so far).

The idea is to un-break what the emojis viewer broke (see, for example this feedback: https://feedback.secondlife.com/bug-reports/p/utf-8-without-u-fe0e-or-u-fe0f-should-not-display-as-emoji ).

You can test this code in the Cool VL Viewer v1.32.0.13 (which does not have a built-in emojis selector, but can render emojis nonetheless, including in LSL dialogs, but yet keeps rendering the latter dialogs as they used to for UTF-8 characters part of fallback monochrome fonts).

As for perfecting LL's emojis selector, this would be another topic entirely...

cosmic-linden:
@sldevel To be clear, you are not remotely obligated to follow up on this fix. My remarks are meant mainly for the team, and any individuals who are interested in the technical aspects of this.

sldevel:
> @sldevel To be clear, you are not remotely obligated to follow up on this fix. My remarks are meant mainly for the team, and any individuals who are interested in the technical aspects of this.

Well, I just try to avoid any misunderstanding...

Another suggestion for a "proper" fix would be to use another font than the TwemojiSVG.ttf one (which contains colorful UTF-8 special characters in excess of **actual** emojis, causing "collision" with monochrome UTF-8 fonts providing the said special characters: e.g. arrows, etc). If we could find a font that provides emojis **only**, then such deconfliction code would become moot (even if it won't impair the usage of such emojis-only-fonts either).

vir-linden:
Henri, could you give some examples of which characters you are concerned about in this fix (other than the link above to issue that mentioned right-pointing triangle)? Ideally we would like to address this in a way that gives minimal surprises to both current unicode users and new emoji enthusiasts.

sldevel:
> Henri, could you give some examples of which characters you are concerned about in this fix (other than the link above to issue that mentioned right-pointing triangle)? 

Well, I cannot list all characters...

So far I saw issues with: ⚙ ▶ ▲ ♥ ☉ ⚫
Which I use in my own scripts and are ruined by the new emojis font without that patch (see Chorazin's recent comment in https://feedback.secondlife.com/bug-reports/p/emoji-display-should-be-consistent-between-lldialog-text-and-lldialog-buttons )...

But there are many, many scripts in SL using dialogs with UTF-8 special symbols, that are affected as well: you will however have to find out by yourself, is you want to make an exhaustive list (pretty much an impossible task).

> Ideally we would like to address this in a way that gives minimal surprises to both current unicode users and new emoji enthusiasts.

I totally agree. The proposed patch is just here to repair what just got broken, but it will cause (minor) side effects, such as a few "yellow face" emojis being printed via the fallback UTF-8 monochrome fonts...

This patch should be seen _as an emergency fix_, but for full fledged emojis support, you will want a more refined approach such as the two possible other solutions I exposed in my last message above:

1. Find a better emojis-only font (or recompile/rebuild TwemojiSVG.ttf with just the actual emojis built in), which would avoid any unwanted collision.
2. Refine LLStringOps::isEmoji() to return false for all the UTF-8 characters we do not want to see overridden with the TwemojiSVG.ttf fancy colorful glyphs.

As a side note, LLStringOps::isEmoji() could be simplified to use the same code as LLEmojiDictionary::isEmoji() (this would get rid of the ICU library dependency too): it would then be much simpler to code it "manually" to exclude all the UTF-8 characters we want to see rendered via the monochrome fonts, or possibly to make it load (in a static std::vector, at first run) a list of characters to exclude, contained in, for example an app_settings/non_emojis.xml file (easier to modify over time without the need to recompile the viewer)...

sldevel:
Addendum regarding the "minor side effect" induced by this patch (e.g. monochrome emojis for some "yellow faces"): this can be solved by specifying a _font="Emoji"_ tag (instead of _font="SansSerif"_) in the <text_editor ...> and <line_editor ...> XUI controls for the chat history, chat console, chat input line, IM history, IM input line: in this case (and even with this patch), the viewer will still try first to render the characters with the Emoji font glyphs first, and when the glyph is not found (which will be the case for all non-emojis, non-special characters), it will search in the fallback fonts.
It means that the emojis will still printed as such in these specific text/line editors.

On the other hand, https://github.com/secondlife/viewer/commit/326055ba82c22fedde186c6a56bafd4fe87e613a will need to be reverted (and the _font="SansSerif"_ tag will have to be used for all UI elements of the LSL dialogs), so that the patch can work its magic and prioritize the monochrome UTF-8 special characters over the fancy colorful ones, ensuring LSL dialogs render as they used to...

marchcat:
Integrated #982; this one can be closed.

Issue #955, opened 2024-03-09
PR#950 breaks ALL existing shapes


FelixWolf:
### Environment

N/A

### Description

The change made to avatar_lad.xml will break existing shapes, and is incomplete code wise.
Ref: https://github.com/secondlife/viewer/pull/950/commits/028fc448322b2388759703aa466c09c7c0d585ec#diff-6671a74c3f95a8f854245aa102deb48aa2cbcacd695a73df202a2fca66cd6abaL6920

The way avatars work is that once the gender slider is greater than 0.5f, code sees the avatar as male, otherwise female.
Various parameters use this to drive specific blend shapes and morphs depending on the avatar shape, simply removing all instances of `sex="male"` or `sex="female"` without understanding what they do will cause problems.

Specifically the following issues are present:
1. This breaks existing shapes because the parameters for boobs are disabled for males, but aren't set to zero, as a result, male shapes will unintentionally have boobs on viewers using this change. See `LLPolySkeletalDistortion::apply`.
2. There are now two `Muscular_Torso` params (Due to the fact there was one for each sex). Originally these were separated for female and male blend shape.<br/>![image](https://github.com/secondlife/viewer/assets/6495643/29047338-f584-4a58-b1fd-25f2c3273cfb)
3. Setting the gender slider to 0 will not enable the boobs sliders, as the trigger for that is tied to `onCommitSexChange` from the `AvatarSex` control.
4. In addition to above, this may not update various code properly.
5. There is unused code tied to elements that are no longer accessible.

We were testing enabling this in Alchemy the other day(See https://git.alchemyviewer.org/FelixWolf/alchemy-viewer/-/commit/e0c37a4a017cc869670ea53554a7856789138d31) and there were a few hiccups.
Noteably: My shape became corrupt. (I managed to rescue it by reading logs and dumping the asset manually and re-creating it) There is a chance this was coincidence, but this needs to be ruled out before full merge into the viewer. In specific, the shape was no longer in the system because it apparently didn't pass validation, but the inventory asset pointer was updated anyway.

### Reproduction steps

N/A

sldevel:
Yes, years ago, I did test this (old) tweak too, for possible inclusion in the Cool VL Viewer and got the same bad issues, which caused me to reject it entirely.

Please, revert and keep at bay !

marchcat:
Thanks! Reverting..

bennettgoble:
The original PR mentioned the torso muscle parameters issue needed a resolution:

[...]

PR #958, opened 2024-03-09
Revert "Remove gender-based restrictions on shape and physics sliders"

marchcat:
Reverts secondlife/viewer#950

see https://github.com/secondlife/viewer/issues/955 for more details.


Ansariel:
Can we merge this please until there is an actually working version that doesn't break shapes?

akleshchev (approved):

PR #966, opened 2024-03-12
Merge main into DRTVWR-600-maint-A

Ansariel:
Looks like nobody wanted to do the merge, so I did it myself. 😉 
@marchcat


marchcat (approved):
Wow! Thank you very much @Ansariel!

I was in the middle of doing that, but you were faster.

PR #968, opened 2024-03-12
viewer#961 Add a notification appearing when user tries to insert a disallowed character

LLGuru:
Example:
image

### DIFF
--- a/indra/llui/lllineeditor.cpp
+++ b/indra/llui/lllineeditor.cpp
@@ -82,9 +82,9 @@ template class LLLineEditor* LLView::getChild<class LLLineEditor>(
 
 LLLineEditor::Params::Params()
 :	max_length(""),
-    keystroke_callback("keystroke_callback"),
-	prevalidate_callback("prevalidate_callback"),
-	prevalidate_input_callback("prevalidate_input_callback"),
+	keystroke_callback("keystroke_callback"),
+	prevalidator("prevalidator"),
+	input_prevalidator("input_prevalidator"),
 	background_image("background_image"),
 	background_image_disabled("background_image_disabled"),
 	background_image_focused("background_image_focused"),
@@ -112,6 +112,8 @@ LLLineEditor::Params::Params()
 	default_text("default_text")
 {
 	changeDefault(mouse_opaque, true);
+	addSynonym(prevalidator, "prevalidate_callback");
+	addSynonym(input_prevalidator, "prevalidate_input_callback");
 	addSynonym(select_on_focus, "select_all_on_focus_received");
 	addSynonym(border, "border");
 	addSynonym(label, "watermark_text");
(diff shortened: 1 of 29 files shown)


indra/llui/lltexteditor.cpp
```diff
@@ -1136,6 +1140,8 @@ S32 LLTextEditor::addChar(S32 pos, llwchar wc)
 		S32 delta = 0;
 		if (mPrevalidateFunc)
 		{
+			LLUI::getInstance()->reportBadKeystroke();
```
akleshchev, line 1143: This looks like it would pop up on every validating case.
I think such message would be usefull when trying to input utf-8 into ansi-only string, but there is no point showing it for something like validateNonNegativeS32 which obviously only accepts nubers.

indra/newview/skins/default/xui/en/notifications.xml
```diff
@@ -413,6 +413,14 @@ Initialization with the Marketplace failed because of a system or network error.
         yestext="OK"/>
     </notification>
 
+    <notification
+     icon="alertmodal.tga"
+     name="InvalidKeystroke"
+     type="alertmodal">
+        There was an invalid keystroke entered. Please input a valid text.
```
akleshchev, line 420: This message is too generic. If it doesn't provide a list of allowed symbols (declare and get a string from TypeValuesHelper?) it's better to replace it with a failure sound.

akleshchev:
I feel like this behavior needs a fair bit of polish.

indra/llui/lllineeditor.cpp
```diff
@@ -1590,19 +1595,28 @@ BOOL LLLineEditor::handleKeyHere(KEY key, MASK mask )
 				deselect();
 			}
 
-			BOOL need_to_rollback = FALSE;
+			bool need_to_rollback = false;
+			bool prevalidator_failed = false;
 
 			// If read-only, don't allow changes
-			need_to_rollback |= (mReadOnly && (mText.getString() == rollback.getText()));
+			need_to_rollback |= (mReadOnly == TRUE) && (mText.getString() == rollback.getText());
```
marchcat, line 1602: But why?
Please use lowercase `bool` if it's required.

akleshchev:
Cause for the ticket: We have some esoteric character requirements in some of the field. Just ehoing a character back is not helpfull, those fields should in some form indicate those requirements, like:
image

indra/llui/lllineeditor.cpp
LLGuru: > But why?
> Please use lowercase `bool` if it's required.

Just trying to make Visual Assistant happy
image

marchcat: It will get fixed by itself after maint A merge.

LLGuru: Fixed
image

PanteraPolnocy:
Could this possibly work similarly to renaming inventory items now (character is just not entered at all, the "thunk" sound is made)? Instead of producing a notification-popup that steals focus from the text field while typing, forcing the user to use the mouse, Enter, or Spacebar to close it?

akleshchev:
Agree, at minimum this is a wrong notification type. It should have been the one that pops up in a corner and fades out without stealing focus. Also it's probably an overkill to use this for everything, instead of just for 'esoteric' text fields.

LLGuru:
> It should have been the one that pops up in a corner and fades out without stealing focus kind of like a hint

Fixed
image

vir-linden (approved):
Test build behavior looks good.

akleshchev (approved):

marchcat (approved):

PR #980, opened 2024-03-13
Fix github actions dependency deprecations

brad-linden:
update to actions/setup-python@v5 and actions/cache@v4 which use node20 runtime


brad-linden:
I intend to squash this when merging

bennettgoble (approved):

nat-goodspeed (approved):
Thank you for addressing this, Brad.

PR #981, opened 2024-03-13
Remove obsolete LLSINGLETON_C11 and LLSINGLETON_EMPTY_CTOR_C11 macros

Ansariel:
@marchcat LLSINGLETON was changed to include the override keyword, this makes LLSINGLETON_C11 redundant and obsolete.


marchcat (approved):
Awesome! Thank you!

PR #982, opened 2024-03-13
Proposal # 2 to restore how UI/dialogs used to render by prioritizing fallback fonts.

sldevel:
With the emojis support, a new font was added, which not only provides emojis but also fancy colorful replacements for UTF-8 characters that used to be supported by our fallback (monochrome) fonts: this causes discrepancies and unwanted/undesired changes in scripted objects menus (e.g. an empty circle or square may render as a black, full one, a heart may render red instead of white), not to mention the larger font size used by the emoji characters...

This patch restores the aspect of such menus/dialogs/UI elements with UTF-8 characters that *are* supported by the usual fallback fonts (fonts which may also vary from one viewer to another, and from one OS to another), so that everything keeps working/rendering as it always did so far, while not impairing the use of new colorful emojis.

This second proposal ensures that:
- "genuine" emojis (in the 0x1f000-0x1ffff range), will *always* be rendered using the new emojis font (this solves, for example, the monochrome "yellow faces" issue seen with some characters in my first proposal).
- Special UTF-8 characters (in the 0x2000-0x32FF range) which have been used by scripters so far, will render as they used to, using the monochrome fallback fonts (this repairs scripted dialogs menus).
- Remaining special characters, that do not have a corresponding glyph in the monochrome font, but do have one in the emojis font, will use the latter font to render.

It also got the nice side-effect of removing the dependency on the ICU4C library.

Note however that the recent commit https://github.com/secondlife/viewer/commit/326055ba82c22fedde186c6a56bafd4fe87e613a will need to be reverted to allow this patch to actually fix scripted dialogs.

Also, some cleanup might be needed in skins/default/xui/*/emoji_characters.xml to remove from it the special UTF-8 characters that will no longer be rendered with fanciful colors, but instead with the monochrome font glyphs.


marchcat (approved):
Thanks! This one looks more suitable.

marchcat:
Also closed #957.

LLGuru:
Hi @sldevel

Initialy the dialog had looked as follows
image

After my change it had looked as follows:
image

After your revert of my change and your "Proposal #2" it became looking as follows:
image

It seems something had been broken with the text on the panel (aboce the buttons)

Do you have any idea about the root cause?
Could it be the new version of `LLStringOps::isEmoji()` ?

Thanks

AtlasLinden:
Related issue: https://github.com/secondlife/viewer/issues/944

kylelinden:
LLdialog and UI should not contain emojis, proposed solution approved.

PR #990, opened 2024-03-13
viewer#945 Some worn items are missing from Avatar floater's Outfits tab

akleshchev:
Make sure that if worn outfit folder is out of sync, it's going to get fetched asap.

P.S. Likely won't fix the issue (still trying to figure out why a single item might be not visible), but should make things more reliable.

### DIFF
--- a/indra/newview/lloutfitslist.cpp
+++ b/indra/newview/lloutfitslist.cpp
@@ -885,18 +885,10 @@ void LLOutfitListBase::refreshList(const LLUUID& category_id)
         ++items_iter)
     {
         LLViewerInventoryCategory *cat = gInventory.getCategory(*items_iter);
-        if (!cat)
+        if (cat)
         {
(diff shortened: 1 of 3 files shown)


indra/newview/lloutfitslist.cpp
```diff
@@ -885,18 +885,10 @@ void LLOutfitListBase::refreshList(const LLUUID& category_id)
         ++items_iter)
     {
         LLViewerInventoryCategory *cat = gInventory.getCategory(*items_iter);
-        if (!cat)
+        if (cat)
         {
-            LLInventoryObject* obj = gInventory.getObject(*items_iter);
```
akleshchev, line 890: This makes no sense. If getCategory failed, object would be an LLViewerInventoryItem. If said item is a link to a category, casting it to LLViewerInventoryCategory is going to cause issues.

maxim-productengine (approved):

marchcat (approved):

Issue #988, opened 2024-03-13
Ensure that terrain UI is still usable when the language is not English and the PBR terrain feature flag is unset


cosmic-linden:
Some UI keys have changed/added. Ensure that they do not break translations, as this would break the guarantee of the feature flag.

cosmic-linden:
It seems that there is nothing badly broken about translations when the PBR terrain feature is turned off, so I am bumping the priority of this down to "Normal." I've pushed a partial translation fix which prevents text from getting cut off by removing some obviously obsolete text, however I will note that translations are generally the duty of the translation team.

georgebproductengine:
Passed QA.

Verified on the Second Life Release 7.1.8.9103842320 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/239. Some buttons appear cropped in Spanish and French.
<img width="711" alt="Screenshot 2024-05-27 at 11 12 00" src="https://github.com/secondlife/viewer/assets/150668104/cd2209c6-7642-472f-8c70-7f856b7a8bc6">
<img width="674" alt="Screenshot 2024-05-27 at 11 13 12" src="https://github.com/secondlife/viewer/assets/150668104/fed866cb-ba87-49c4-ad4b-509449c7ff3f">

PR #993, opened 2024-03-14
secondlife/viewer#988: Fix terrain composition label getting cut off

cosmic-linden:
Normally I wouldn't mess with non-English translations. However, in the spirit of making sure that the terrain feature flag doesn't break anything when turned off, I checked the translations, and noticed that some text was cut off which wasn't before. I am thus providing this fix as a matter of due diligence.


marchcat (approved):
👍

PR #995, opened 2024-03-14
Revert 326055ba82c22fedde186c6a56bafd4fe87e613a after 2f452d06e6964b0edf26b0b3f6eaa156e3fa2d48

sldevel:
It is necessary to revert commit 326055ba82c22fedde186c6a56bafd4fe87e613a for 2f452d06e6964b0edf26b0b3f6eaa156e3fa2d48 to work its magic and repair script dialogs so that they render as they used to before the emojis fonts introduction.

Note that this revert won't prevent to use the new emojis should scripters want them in their new scripted dialogs: it just ensures existing scripted dialogs using special UTF-8 characters (which are not genuine emojis) will render as they used to, using the monochrome fallback fonts.

### DIFF
--- a/indra/newview/lltoastnotifypanel.cpp
+++ b/indra/newview/lltoastnotifypanel.cpp
@@ -298,8 +298,8 @@ void LLToastNotifyPanel::init( LLRect rect, bool show_images )
     // init font variables
     if (!sFont)
     {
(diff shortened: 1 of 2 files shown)


marchcat (approved):

marchcat:
Thanks!

PR #1003, opened 2024-03-14
SL-18721 Restore release behavior

akleshchev:
Closing window correctly caused a significant amount of logout freezes with no known reproes. Temporarily returning to old behavior were thread was killes without closing window and will reenable in later maints to hopefully get a scenario or at least more data of what is causing the freeze.


marchcat (approved):

AtlasLinden:
Archived issue: https://github.com/secondlife/jira-archive-internal/issues/68974

Issue #1006, opened 2024-03-15
Applying terrain with one or more null material IDs is allowed, not caught by the validator


cosmic-linden:

cosmic-linden:
I'm going to just prevent setting the PBR materials from being set to null, and have them default to the blank material instead. See https://github.com/secondlife/viewer-issues/issues/80 for selecting better default PBR terrain materials to show to the user.

DmitrySProductengine:
Failed QA.
Verified on the Second Life Release 7.1.8.9103842320 (64bit) on Win10/OSX in the scope of [IQA #231](https://github.com/secondlife/iqa/issues/231).

This ticket does not specify the exact test plan for verification.
I checked the test plan part (Featurettes Test Plan - [3. PBR Terrain](https://lindenlab.atlassian.net/wiki/spaces/QA/pages/3054436428/Featurettes+Test+Plan#3.-PBR-terrain)) and everything is running correctly.
In step 7, where the checkbox is enabled, blank PBR materials are displayed.
However, the test plan does not describe the checks or behavior of blank or default PBR materials.

The reason for the failure was the different behavior of the viewer on Win and OSX:
1. OSX - the viewer does not allow the use of blank and default PBR materials. (perhaps this is the expected behavior of this ticket).
7.1.8.9103842320 - OSX - default and blank PBR:
<img width="1680" alt="7 1 8 9103842320 - OSX - default and blank PBR" src="https://github.com/secondlife/viewer/assets/136308794/1b83ca70-2f23-4086-a5ae-221dedafd34b">

2. Win10 - the viewer does not allow the use of blank PBR materials. But it uses the default PBR materials (which look like blank ones).
7.1.8.9103842320 - Win10 - blank PBR

7.1.8.9103842320 - Win10 - default PBR


Also, if possible, please add a test plan to check this ticket.

cosmic-linden:
Sounds like the issue on Mac could be related to https://github.com/secondlife/viewer/issues/989 ? cc: @brad-linden

cosmic-linden:
A test plan has been added as well.

cosmic-linden:
The experience of using the material picker for terrain and setting to blank is not good. I think the right behavior here is that "Default" should be greyed out (texture fallback not allowed), and "Blank" should work and be allowed to be set, resulting in the blank GLTF material being applied. This would be consistent with the meaning of the buttons in the material picker for prims.

[...]

PR #1013, opened 2024-03-18
secondlife/viewer#1006: Prevent setting PBR terrain material swatches to null

cosmic-linden:


akleshchev (approved):

indra/newview/llfloaterregioninfo.cpp
```diff
@@ -1394,7 +1395,7 @@ LLPanelRegionTerrainInfo::LLPanelRegionTerrainInfo()
     }
     for (S32 i = 0; i < LLTerrainMaterials::ASSET_COUNT; ++i)
     {
-        mLastSetMaterials[i] = LLUUID::null;
+        mLastSetMaterials[i] = LLGLTFMaterialList::BLANK_MATERIAL_ASSET_ID;
```
akleshchev, line 1398: We might want to move this value into indra_constants

cosmic-linden: Good point, as that might prevent some race conditions in the future.

Issue #1027, opened 2024-03-20
Transparent PBR terrain has noise on minimap


cosmic-linden:
## Test plan

See, "Minimap" in https://github.com/secondlife/viewer/blob/release/materials_featurette/doc/testplans/pbr_terrain_composition.md

cosmic-linden:
Some day the minimap tiles will be rendered with a shader, but that day is not today.

cosmic-linden:
@irinabproductengine Added test plan to original post in ticket

dmytrohproductengine:
Passed QA.
Verified on the Second Life Release 7.1.8.9103842320 (64bit) on Win10/OSX in the scope of [secondlife/iqa#231](https://github.com/secondlife/iqa/issues/231).

PR #1028, opened 2024-03-21
secondlife/viewer#1027: Fix noise on transparent PBR materials on minimap

cosmic-linden:


indra/newview/llvlcomposition.cpp
```diff
@@ -42,6 +42,8 @@
 #include "llviewercontrol.h"
 
 
+extern LLColor4U MAX_WATER_COLOR;
```
brad-linden, line 45: seems like this declaration belongs in llsurface.h (and a similar extern declaration in llworld.cpp should probably be consolidated there along with it.)

possibly we also want this to be `const` as well?

brad-linden (approved):
looks good with a suggestion for possible organizational cleanup

indra/newview/llvlcomposition.cpp
cosmic-linden: In hindsight, I probably should have checked the coding standard for conventions on global variables: https://wiki.secondlife.com/wiki/Coding_standard

It looks like static class members are the future. My understanding is that the gVariableName convention is superseded by `LLSimpleton`. There are some possible exceptions though, like indra_constants.cpp .

Will keep this in mind for the future.

Side note: Before seeing `MAX_WATER_COLOR` in llworld.cpp, I did not know was possible to use extern in this way.

PR #1029, opened 2024-03-21
Add Develop menu option 'Debug Camera Controls'

LLGuru:
It is useful for investigating camera behavior issues like https://github.com/secondlife/jira-archive-internal/issues/69593

image

### DIFF
--- a/indra/llui/llview.cpp
+++ b/indra/llui/llview.cpp
@@ -60,6 +60,7 @@ static const S32 LINE_HEIGHT = 15;
 S32		LLView::sDepth = 0;
 bool	LLView::sDebugRects = false;
 bool	LLView::sDebugUnicode = false;
+bool	LLView::sDebugCamera = false;
 bool	LLView::sIsRectDirty = false;
 LLRect	LLView::sDirtyRect;
 bool	LLView::sDebugRectsShowNames = true;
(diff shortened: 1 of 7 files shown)


marchcat:
Doesn't this duplicate Develop -> Consoles -> Camera?

LLGuru:
> Doesn't this duplicate Develop -> Consoles -> Camera?

Absolutely not
1. This tool shows info for two cameras: Viewer Camera and Agent Cameera
2. This tool shows the original values stored in the class vriables, and not calculated
3. This tool relates to the CAMERA CONTROLS floater because it shows the result
4. This tool is placed inside the CAMERA CONTROLS floater and looks more readable

marchcat:
Okay, in this case, calling it from the menu should (_probably_) open a camera controls panel if it's not already visible.
Also, merging these two tools into one might be advisable. However, I still have no opinion on this matter.

akleshchev:
> Also, merging these two tools into one might be advisable.
Agree, I would rather have Consoles -> Camera be updated to show more info.

akleshchev (approved):
I don't see issues code wise, but 'feature' wise I suggest asking vir or getting more input about desired behavior elsewhere.

marchcat (approved):

vir-linden:
I'm good moving forward with this design

PR #1032, opened 2024-03-22
triage#112 'No emoji found for' text is not translated to other languages in the 'Choose Emoji' floater

LLGuru:


indra/newview/skins/default/xui/en/strings.xml
```diff
@@ -2699,6 +2699,9 @@ If you continue to receive this message, please contact Second Life support for
 	<string name="IMMainland">mainland</string>
 	<string name="IMTeen">teen</string>
 
+    <!-- floater emoji picker -->
+    <string name="NoEmojiForFilter">No emoji found for '[FILTER]'</string>
```
akleshchev, line 2703: This sting is specific for a signle floater, shouldn't it be declared in emoji loater's xml?

LLGuru: > This sting is specific for a signle floater, it's preferable to declared in floater_emoji_picker.xml.

It seems it is placed in the right place
image

akleshchev: > but I didn't find the way to define a string with parameters in the floater

example:
```
						args["[MINUTES]"] = llformat("%d", (seconds/60));
						std::string buf = parent_floater->getString ("Minutes", args);
```

marchcat: > This Emoji picker already has couple strings declared in own xml file

Agree, it's better to keep them in one place.

LLGuru: > example:

Moved, thanks

marchcat (approved):

akleshchev (approved):

PR #1036, opened 2024-03-22
WebRTC: Move processIceUpdates into a coroutine

roxanneskelly:
processIceUpdates is now handled as a separate coroutine to 'break things up a bit' and reduce the time spent in the state machine coroutine.

Also:
* Added Tracy profiling categories for WebRTC Voice
* Fixed a few crashes.
* Position updates to the server now only happen when position changes.


indra/newview/llvoicewebrtc.cpp
```diff
@@ -2070,123 +2104,92 @@ void LLVoiceWebRTCConnection::OnIceCandidate(const llwebrtc::LLWebRTCIceCandidat
...
+        httpOpts->setWantHeaders(true);
+
+        LLSD result = httpAdapter->postAndSuspend(httpRequest, url, body, httpOpts);
+        LLSD httpResults = result[LLCoreHttpUtil::HttpCoroutineAdapter::HTTP_RESULTS];
+        LLCore::HttpStatus status = LLCoreHttpUtil::HttpCoroutineAdapter::getStatusFromLLSD(httpResults);
+
+        if (!status)
+        {
+            // couldn't trickle the candidates, so restart the session.
+            setVoiceConnectionState(VOICE_STATE_SESSION_RETRY);
```
akleshchev, line 2189: postAndSuspend can take undetermined amount of time and coroutine can persit till end of shutdown. At this point LLVoiceWebRTCConnection might no longer exists.

In case of coroutines you need to check if class is still alive in some way after any 'suspend' (ex: call checkStop(), not sure why suspend doesn't call it internally after a yield)

callumlinden (approved):

callumlinden (approved):

indra/newview/llvoicewebrtc.cpp
akleshchev: Same issue with breakVoiceConnectionCoro

roxanneskelly: thanks, will fix

akleshchev: LLVivoxVoiceClient::sShuttingDown exists for the same reason - vivox's corouitine stubbornly refuses to die.

marchcat (approved):

indra/newview/llvoicewebrtc.cpp
roxanneskelly: Used sShuttingDown in LLWebRTCVoiceClient.  Should be good now.

PR #1038, opened 2024-03-23
Add fiber.lua, which permits calling leap.request() even from Lua's main thread.

nat-goodspeed:
llluamanager_test.cpp verifies that Lua's main thread can call `leap.request()` and receive the viewer's response.

I'm sitting at an airport right now, but if you get a chance, please try cleaning up the one-shot coroutines in your test scripts?


indra/llcommon/llevents.h
```diff
@@ -151,13 +151,32 @@ typedef boost::signals2::signal<bool(const LLSD&), LLStopWhenHandled, float>  LL
...
 /// Result of registering a listener, supports <tt>connected()</tt>,
 /// <tt>disconnect()</tt> and <tt>blocked()</tt>
 typedef boost::signals2::connection LLBoundListener;
 /// Storing an LLBoundListener in LLTempBoundListener will disconnect the
 /// referenced listener when the LLTempBoundListener instance is destroyed.
 typedef boost::signals2::scoped_connection LLTempBoundListener;
 
+/// Accepting (const LLListener&) allows either LLEventListener or LLVoidListener
+/// TODO: but compiler considers the constructor call ambiguous??
+class LLListener
```
maxim-productengine, line 165: We already have LLListener class, which is related to audio support.

maxim-productengine (changes requested):
It seems LLListener class should be renamed, but everything else looks good.

I pulled these changes locally and tested the scripts with the "one line `request`" - it works! I'll update the test scripts, when this PR is merged.

maxim-productengine (approved):

PR #1040, opened 2024-03-25
Add keystroke event support and allow adding text lines to the line editor

maxim-productengine:
The script can be signed for 'keystroke' event now.
Added the command for appending text line (text editor) from the script.
Script updates: 
Removed `util.contains` which is the same as `table.find` in luau; 
call leap.request() from LUA main thread.

### DIFF
--- a/indra/newview/scripts/lua/test_luafloater_demo.lua
+++ b/indra/newview/scripts/lua/test_luafloater_demo.lua
@@ -2,6 +2,7 @@ XML_FILE_PATH = "luafloater_demo.xml"
 
 leap = require 'leap'
 coro = require 'coro'
+fiber = require 'fiber'
 
 --event pump for sending actions to the floater
 COMMAND_PUMP_NAME = ""
@@ -63,4 +64,4 @@ function process_events(waitfor)
   end
 end
 
-coro.launch(process_events, catch_events)
+fiber.launch("catch_events", process_events, catch_events)
--- a/indra/newview/scripts/lua/test_luafloater_demo.lua
+++ b/indra/newview/scripts/lua/test_luafloater_demo.lua
@@ -6,12 +6,7 @@ coro = require 'coro'
 --event pump for sending actions to the floater
 COMMAND_PUMP_NAME = ""
 --table of floater UI events
-event_list={}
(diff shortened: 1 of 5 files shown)


indra/newview/scripts/lua/test_luafloater_demo.lua
```diff
@@ -74,5 +64,3 @@ function process_events(waitfor)
 end
 
 coro.launch(process_events, catch_events)
```
nat-goodspeed, line 66: Use `fiber.launch()` now, and pass a name. `LuaState::expr()` now detects if a script has imported `fiber.lua`, directly or indirectly, and if so it follows up with an implicit `fiber.run()` call that runs launched coroutines to completion. But `fiber.run()` only knows about coroutines launched with `fiber.launch()`.

indra/newview/scripts/lua/test_luafloater_gesture_list.lua
```diff
@@ -71,4 +61,3 @@ function process_events(waitfor)
 end
 
 coro.launch(process_events, catch_events)
```
nat-goodspeed, line 63: same comment about `fiber.launch()`

nat-goodspeed (changes requested):

indra/newview/scripts/lua/test_luafloater_demo.lua
```diff
@@ -2,6 +2,7 @@ XML_FILE_PATH = "luafloater_demo.xml"
 
 leap = require 'leap'
 coro = require 'coro'
```
nat-goodspeed, line 4: At this point you could ditch `require 'coro'` too.

indra/newview/scripts/lua/test_luafloater_gesture_list.lua
```diff
@@ -2,6 +2,7 @@ XML_FILE_PATH = "luafloater_gesture_list.xml"
 
 leap = require 'leap'
 coro = require 'coro'
```
nat-goodspeed, line 4: `require 'coro'` no longer needed.

nat-goodspeed (approved):

PR #1041, opened 2024-03-25
viewer#1016 Incorrect behavior of Physics Shapes rendering

akleshchev:
I'm not entirely sure if this is the right thing to do, but follows server's physics correctly so far. Since I'm lacking test cases and 'thin' models, this will fall to QA to check.

### DIFF
--- a/indra/newview/llphysicsshapebuilderutil.cpp
+++ b/indra/newview/llphysicsshapebuilderutil.cpp
@@ -28,6 +28,26 @@
 
 #include "llphysicsshapebuilderutil.h"
 
+#include "llmeshrepository.h"
+
+bool LLPhysicsVolumeParams::hasDecomposition() const
+ {
+    if (!isMeshSculpt())
+    {
+        return false;
+    }
+
+    LLUUID mesh_id = getSculptID();
+    if (mesh_id.isNull())
+    {
+        return false;
+    }
+
+    LLModel::Decomposition* decomp = gMeshRepo.getDecomposition(mesh_id);
+
+    return decomp != NULL;
+}
+
 /* static */
 void LLPhysicsShapeBuilderUtil::determinePhysicsShape( const LLPhysicsVolumeParams& volume_params, const LLVector3& scale, PhysicsShapeSpecification& specOut)
 {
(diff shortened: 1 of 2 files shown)


marchcat (approved):

sldevel:
See my comment in the commit (with working code). I do not think this fix is valid, but Firestorm and other TPVs (mine, probably Alchemy, etc) got it right for a while already.

akleshchev:
@sldevel can you provide a link to that comment please?
I don't see any comments in pull request's commit (draft comment?), nor in original files.

If code in question is bigger than a single string, it might be better to just make a pull request/contribution.

sldevel:
> @sldevel can you provide a link to that comment please?

https://github.com/secondlife/viewer/pull/1041/files/61c0031d0bf9ed03af847bdfcb0caaa73547b5fa

> If code in question is bigger than a single string, it might be better to just make a pull request/contribution.

Sorry, I do not have time for this currently...

Plus, it's not my code, but Beq janus' from Firestorm.

But I found the corresponding commit by looking at their code and the comment in it citing the https://jira.firestormviewer.org/browse/FIRE-23053 issue, and then via a search for "FIRE-23053" in their git commits short log:
https://github.com/FirestormViewer/phoenix-firestorm/commit/8be317a99e0ebd1ffa0abfeb74bd066aaa123af2

akleshchev:
@sldevel 

> https://github.com/secondlife/viewer/pull/1041/files/61c0031d0bf9ed03af847bdfcb0caaa73547b5fa

Shows zero comments there. You likely made a draft comment, in such case it's only visible to you.

sldevel:
Shows zero comments there. You likely made a draft comment, in such case it's only visible to you.

Bleh... I hate github & Co (Iam a 'patch', 'diff' and 'Nedit' old fart)... Here is a copy:

This test is always false at this point, since we already tested for shouldForceConvex() in the first 'if ()' (shouldForceConvex() is therefore always false when this part of the code is reached). This means PhysicsShapeSpecification::PRIM_CONVEX can never happen... Is this really valid ?

I looked in my own viewer code (which has been fixed a while ago, IIRC with code changes by Beq Janus, that I borrowed from Firestorm), and I have instead:

```
if (volume_params.shouldForceConvex())
{
	specOut.mType = PhysicsShapeSpecification::USER_CONVEX;
}
// Make a simpler convex shape if we can.
else if (volume_params.isConvex() ||	// is convex
		 min_size_counts > 1)			// two or more small dimensions
{
	specOut.mType = PhysicsShapeSpecification::PRIM_CONVEX;
}
else if (volume_params.isSculpt())
{
	if (volume_params.isMeshSculpt())
	{
		// Check if one dimension is smaller than min
		if (!has_decomp &&
			(scale[0] < SHAPE_BUILDER_USER_MESH_CONVEXIFICATION_SIZE ||
			 scale[1] < SHAPE_BUILDER_USER_MESH_CONVEXIFICATION_SIZE ||
			 scale[2] < SHAPE_BUILDER_USER_MESH_CONVEXIFICATION_SIZE))
		{
			specOut.mType = PhysicsShapeSpecification::PRIM_CONVEX;
		}
		else
		{
			specOut.mType = PhysicsShapeSpecification::USER_MESH;
		}
	}
	else
	{
		specOut.mType = PhysicsShapeSpecification::SCULPT;
	}
}
else
{
	// Resort to mesh
	specOut.mType = PhysicsShapeSpecification::PRIM_MESH;
}
```

Where 'has_decomp' is an additional boolean passed to the method, which is true when the mesh got a physics decomposition.

sldevel:
Shows zero comments there. You likely made a draft comment, in such case it's only visible to you.

Bleh... I hate github & Co (I am a 'patch', 'diff' and 'Nedit' old fart)... Here is a copy:

This test is always false at this point, since we already tested for shouldForceConvex() in the first 'if ()' (shouldForceConvex() is therefore always false when this part of the code is reached). This means PhysicsShapeSpecification::PRIM_CONVEX can never happen... Is this really valid ?

I looked in my own viewer code (which has been fixed a while ago, IIRC with code changes by Beq Janus, that I borrowed from Firestorm), and I have instead:

```
if (volume_params.shouldForceConvex())
{
	specOut.mType = PhysicsShapeSpecification::USER_CONVEX;
}
// Make a simpler convex shape if we can.
else if (volume_params.isConvex() ||	// is convex
		 min_size_counts > 1)			// two or more small dimensions
{
	specOut.mType = PhysicsShapeSpecification::PRIM_CONVEX;
}
else if (volume_params.isSculpt())
{
	if (volume_params.isMeshSculpt())
	{
		// Check if one dimension is smaller than min
		if (!has_decomp &&
			(scale[0] < SHAPE_BUILDER_USER_MESH_CONVEXIFICATION_SIZE ||
			 scale[1] < SHAPE_BUILDER_USER_MESH_CONVEXIFICATION_SIZE ||
			 scale[2] < SHAPE_BUILDER_USER_MESH_CONVEXIFICATION_SIZE))
		{
			specOut.mType = PhysicsShapeSpecification::PRIM_CONVEX;
		}
		else
		{
			specOut.mType = PhysicsShapeSpecification::USER_MESH;
		}
	}
	else
	{
		specOut.mType = PhysicsShapeSpecification::SCULPT;
	}
}
else
{
	// Resort to mesh
	specOut.mType = PhysicsShapeSpecification::PRIM_MESH;
}

```
Where 'has_decomp' is an additional boolean passed to the method, which is true when the mesh got a physics decomposition.

akleshchev:
Thank you, that makes sense.

marchcat (approved):

Issue #1050, opened 2024-03-25
Fix mirror clipping for PBR terrain


cosmic-linden:
PBR terrain is appearing in mirrors when it shouldn't. i.e. a hil behind the mirror renders in the mirror.

Geenz suspects this could be due to missing uniforms in the PBR terrain shader relative to the texture terrain shader.

cosmic-linden:
Definitely should be fixed, but not deemed a blocker for shipping the featurettes viewer.

vladyslavasproductengine:
Could you please provide a test plan for verification?

cosmic-linden:
@vladyslavasproductengine I think it would be best for this to be tested as part of the overall testing of the mirrors feature.

I believe the test plan for mirrors is not finished, so I would recommend deferring testing for this ticket in the meantime. cc: @Dan-Linden

Dan-Linden:
Test will be added to https://lindenlab.atlassian.net/wiki/spaces/QA/pages/3054436428/Featurettes+Test+Plan#2.8-Mirrors-test-cases

PR #1058, opened 2024-03-26
secondlife/viewer#1050: Fix mirror clipping for PBR terrain. Also fixes environment settings for PBR terrain.

cosmic-linden:


indra/newview/llviewershadermgr.cpp
```diff
@@ -331,6 +331,7 @@ void LLViewerShaderMgr::finalizeShaderList()
     mShaderList.push_back(&gDeferredPBROpaqueProgram);
     mShaderList.push_back(&gDeferredAvatarProgram);
     mShaderList.push_back(&gDeferredTerrainProgram);
+    mShaderList.push_back(&gDeferredPBRTerrainProgram);
```
cosmic-linden, line 334: This one's interesting. Mirror uniforms are applied in `LLEnvironment::update`, but the PBR terrain shader wasn't in the list of shaders, so `mUniformsDirty` was never set to true. I am sure mirrors weren't the only thing affected by this.

marchcat (approved):

PR #1095, opened 2024-03-31
Remove BuildParams

bennettgoble:
This file is no longer used (?)


marchcat (approved):
I don't see it either.

brad-linden (approved):

nat-goodspeed (approved):
I approve. `BuildParams` was used by [the `sling-buildscripts` `build.sh`](https://github.com/secondlife/sling-buildscripts/blob/master/hg/bin/build.sh), which we used on TeamCity but do not use on GitHub.

PR #1077, opened 2024-03-27
Remove dead googlemock dependency and related setup code

RyeMutt:
This removes dead test-related dependencies and setup code

PR #1054, opened 2024-03-26
Port from JsonCPP to Boost.Json for json parsing and serializing

RyeMutt:
This ports the viewer to Boost.Json to eliminate the now-unneeded dependency on JsonCPP 

This depends on an update to Boost 1.84+

PR #1097, opened 2024-03-31
Upgrade Boost to 1.84.0

bennettgoble:
Bump Boost from 1.81.0 to 1.84.0. This updated package includes the `boost.json` module, which has been added to replace the viewer's old jsoncpp dependency (#1054.)

This PR depends on googlemock being removed in f6590d7e27cf86cc6a4bbf1a7207e6f7ea9b1150 (#1077), which is why it targets `DRTVWR-600-maint-A`

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -104,25 +104,39 @@
             <key>archive</key>
             <map>
               <key>hash</key>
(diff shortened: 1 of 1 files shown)


marchcat (approved):
Thanks!

PR #1100, opened 2024-04-01
[WebRTC] Add UI for AGC, Echo Cancellation, and Noise Reduction; use Linden STUN servers.

roxanneskelly:
Plumb some audio processing settings through llwebrtc so that the viewer proper can set them.
Also, include a first-pass UI for setting those values (English only)
(TODO: Tweak UI and add non-English versions.)

Also, plumb STUN/TURN server settings through from llwebrtc and set them according to grid.
(TODO: pass them up from login)

### DIFF
--- a/indra/newview/llvoicewebrtc.cpp
+++ b/indra/newview/llvoicewebrtc.cpp
@@ -2374,7 +2374,7 @@ void LLVoiceWebRTCConnection::breakVoiceConnectionCoro()
     LLSD body;
     body["logout"]         = TRUE;
     body["viewer_session"] = mViewerSession;
-    body["voice_server_type"] = REPORTED_VOICE_SERVER_TYPE;
+    body["voice_server_type"] = WEBRTC_VOICE_SERVER_TYPE;
 
     LLCoreHttpUtil::HttpCoroutineAdapter::ptr_t httpAdapter(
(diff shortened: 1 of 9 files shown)


marchcat (approved):
> add non-English versions

Alternatively, this can be done later as part of the viewer translation update.

indra/newview/llvoicewebrtc.cpp
```diff
@@ -301,6 +302,13 @@ void LLWebRTCVoiceClient::updateSettings()
     setRenderDevice(outputDevice);
     F32 mic_level = gSavedSettings.getF32("AudioLevelMic");
     setMicGain(mic_level);
+    
+    llwebrtc::LLWebRTCDeviceInterface::AudioConfig config;
+    config.mEchoCancellation = gSavedSettings.getBOOL("VoiceEchoCancellation");
```
akleshchev, line 307: If updateSettings() gets called more often than once per frame, you might want to replace getBOOL with LLCachedControl<bool>

roxanneskelly: Hmm, I'm only seeing updateSettings getting called once when we change an audio setting under the debugger.  I'll make the change anyway, as it's the way of the future.

PR #1107, opened 2024-04-02
enable tracy for linux and build fix

AiraYumi:
Allows tracy to be enabled on Linux.

This may be incorrect because it was modified based on the Firestorm Viewer source code.
That's why it's a draft.

I could not build it without modifying indra/llcommon/llprofiler.h .
I don't know of any other way to fix it.


AiraYumi:
Is it better to target pull requests related to the linux version to maint-b from now on?

marchcat (approved):
Thank you!

marchcat:
> Is it better to target pull requests related to the linux version to maint-b from now on?

Yes, let's keep it there.

PR #1110, opened 2024-04-02
Add `startup.lua` module with `startup.ensure()`, `startup.wait()` functions.

nat-goodspeed:
This lets a calling script verify that it's running at the right point in the
viewer's life cycle. A script that wants to interact with the SL agent
wouldn't work if run from the viewer's command line -- unless it calls
`startup.wait("STATE_STARTED")`, which pauses until login is complete.

Modify `test_luafloater_demo.lua` and `test_luafloater_gesture_list.lua` to find
their respective floater XUI files in the same directory as themselves.
Make them both capture the `reqid` returned by the `"showLuaFloater"`operation,
and filter for events bearing the same `reqid`. This paves the way for a given
script to display more than one floater concurrently.

Make `test_luafloater_demo.lua` (which does not require in-world resources) wait
until `'STATE_LOGIN_WAIT'`, the point at which the viewer has presented the
login screen.

Make `test_luafloater_gesture_list.lua` (which interacts with the agent) wait
until `'STATE_STARTED'`, the point at which the viewer is fully in world.

Either or both can now be launched from the viewer's command line.

### DIFF
--- a/indra/llcommon/llleaplistener.cpp
+++ b/indra/llcommon/llleaplistener.cpp
@@ -70,46 +70,46 @@ LLLeapListener::LLLeapListener(const std::string_view& caller, const Callback& c
 {
     LLSD need_name(LLSDMap("name", LLSD()));
     add("newpump",
-        "Instantiate a new LLEventPump named like [\"name\"] and listen to it.\n"
-        "[\"type\"] == \"LLEventStream\", \"LLEventMailDrop\" et al.\n"
-        "Events sent through new LLEventPump will be decorated with [\"pump\"]=name.\n"
-        "Returns actual name in [\"name\"] (may be different if collision).",
+R"-(Instantiate a new LLEventPump named like ["name"] and listen to it.
+["type"] == "LLEventStream", "LLEventMailDrop" et al.
+Events sent through new LLEventPump will be decorated with ["pump"]=name.
+Returns actual name in ["name"] (may be different if collision).)-",
         &LLLeapListener::newpump,
         need_name);
     LLSD need_source_listener(LLSDMap("source", LLSD())("listener", LLSD()));
     add("listen",
-        "Listen to an existing LLEventPump named [\"source\"], with listener name\n"
-        "[\"listener\"].\n"
-        "If [\"tweak\"] is specified as true, tweak listener name for uniqueness.\n"
-        "By default, send events on [\"source\"] to the plugin, decorated\n"
-        "with [\"pump\"]=[\"source\"].\n"
-        "If [\"dest\"] specified, send undecorated events on [\"source\"] to the\n"
-        "LLEventPump named [\"dest\"].\n"
-        "Returns [\"status\"] boolean indicating whether the connection was made,\n"
-        "plus [\"listener\"] reporting (possibly tweaked) listener name.",
+R"-(Listen to an existing LLEventPump named ["source"], with listener name
+["listener"].
+If ["tweak"] is specified as true, tweak listener name for uniqueness.
+By default, send events on ["source"] to the plugin, decorated
+with ["pump"]=["source"].
+If ["dest"] specified, send undecorated events on ["source"] to the
+LLEventPump named ["dest"].
+Returns ["status"] boolean indicating whether the connection was made,
+plus ["listener"] reporting (possibly tweaked) listener name.)-",
         &LLLeapListener::listen,
         need_source_listener);
     add("stoplistening",
-        "Disconnect a connection previously established by \"listen\".\n"
-        "Pass same [\"source\"] and [\"listener\"] arguments.\n"
-        "Returns [\"status\"] boolean indicating whether such a listener existed.",
+R"-(Disconnect a connection previously established by "listen".
+Pass same ["source"] and ["listener"] arguments.
+Returns ["status"] boolean indicating whether such a listener existed.)-",
         &LLLeapListener::stoplistening,
         need_source_listener);
     add("ping",
-        "No arguments, just a round-trip sanity check.",
+"No arguments, just a round-trip sanity check.",
         &LLLeapListener::ping);
     add("getAPIs",
-        "Enumerate all LLEventAPI instances by name and description.",
+"Enumerate all LLEventAPI instances by name and description.",
         &LLLeapListener::getAPIs);
     add("getAPI",
-        "Get name, description, dispatch key and operations for LLEventAPI [\"api\"].",
+R"-(Get name, description, dispatch key and operations for LLEventAPI ["api"].)-",
         &LLLeapListener::getAPI,
         LLSD().with("api", LLSD()));
     add("getFeatures",
-        "Return an LLSD map of feature strings (deltas from baseline LEAP protocol)",
+"Return an LLSD map of feature strings (deltas from baseline LEAP protocol)",
         static_cast<void (LLLeapListener::*)(const LLSD&) const>(&LLLeapListener::getFeatures));
     add("getFeature",
-        "Return the feature value with key [\"feature\"]",
+R"-(Return the feature value with key ["feature"])-",
         &LLLeapListener::getFeature,
         LLSD().with("feature", LLSD()));
 }
--- a/indra/llcommon/llleaplistener.cpp
+++ b/indra/llcommon/llleaplistener.cpp
@@ -80,11 +80,13 @@ LLLeapListener::LLLeapListener(const std::string_view& caller, const Callback& c
     add("listen",
         "Listen to an existing LLEventPump named [\"source\"], with listener name\n"
         "[\"listener\"].\n"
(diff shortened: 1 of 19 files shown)


indra/llcommon/llleaplistener.cpp
```diff
@@ -80,11 +80,13 @@ LLLeapListener::LLLeapListener(const std::string_view& caller, const Callback& c
...
         "Listen to an existing LLEventPump named [\"source\"], with listener name\n"
         "[\"listener\"].\n"
+        "If [\"tweak\"] is specified as true, tweak listener name for uniqueness.\n"
         "By default, send events on [\"source\"] to the plugin, decorated\n"
         "with [\"pump\"]=[\"source\"].\n"
         "If [\"dest\"] specified, send undecorated events on [\"source\"] to the\n"
         "LLEventPump named [\"dest\"].\n"
-        "Returns [\"status\"] boolean indicating whether the connection was made.",
+        "Returns [\"status\"] boolean indicating whether the connection was made,\n"
+        "plus [\"listener\"] reporting (possibly tweaked) listener name.",
```
Nicky-D, line 89: That's a lot of quoting and finger gymnastics. Raw string literals (https://en.cppreference.com/w/cpp/language/string_literal) could make this string a lot more pleasant to type and read

indra/llcommon/lua_function.cpp
```diff
@@ -68,6 +68,15 @@ int lluau::loadstring(lua_State *L, const std::string &desc, const std::string &
     return luau_load(L, desc.data(), bytecode.get(), bytecodeSize, 0);
 }
 
+std::filesystem::path lluau::source_path(lua_State* L)
+{
+    //Luau lua_Debug and lua_getinfo() are different compared to default Lua:
+    //see https://github.com/luau-lang/luau/blob/80928acb92d1e4b6db16bada6d21b1fb6fa66265/VM/include/lua.h
+    lua_Debug ar;
+    lua_getinfo(L, 1, "s", &ar);
```
Nicky-D, line 76: No error checking or pointer verification (L could be nullptr) makes me sad

indra/llcommon/lua_function.cpp
```diff
@@ -684,6 +705,38 @@ std::pair<LuaFunction::Registry&, LuaFunction::Lookup&> LuaFunction::getState()
     return { registry, lookup };
 }
 
+/*****************************************************************************
+*   source_path()
+*****************************************************************************/
+lua_function(source_path, "return the source path of the running Lua script")
+{
```
Nicky-D, line 712: What's going to happen shall L be nullptr?

Also I assume `lua_function` is a macro hiding the real function signature. I'm not sure if I like that or not. But at least a better comment why it seems void yet returns 1 would be helpful

indra/newview/llluamanager.cpp
```diff
@@ -375,16 +368,16 @@ void LLRequireResolver::findModule()
 
     // not already cached - prep error message just in case
     auto fail{
-        [L=L, path=mPathToResolve]()
+        [L=L, path=mPathToResolve.u8string()]()
```
Nicky-D, line 371: I'd not be surprised if on Windows this would let the viewer fall flat on it's face in mPathToResolve contains a unicode path. Like C:\Prögrämme\ (that's not a real thing).
Something to create a test case for I suggest

indra/newview/scripts/lua/Queue.lua
```diff
@@ -1,6 +1,12 @@
 -- from https://create.roblox.com/docs/luau/queues#implementing-queues,
 -- amended per https://www.lua.org/pil/16.1.html
 
+-- While coding some scripting in Lua
+-- I found that I needed a queua
+-- I thought of linked list
+-- But had to resist
+-- For fear it might be too obscua. 
+
```
Nicky-D, line 9: Nat in almost Haiku land. I approve

indra/newview/scripts/lua/leap.lua
```diff
@@ -40,25 +40,25 @@
 
 local fiber = require('fiber')
 local ErrorQueue = require('ErrorQueue')
--- local debug = require('printf')
-local function debug(...) end 
+local function dbg(...) end
```
Nicky-D, line 43: I know to little about lua  but ... I saw this snipped a few times now, would it not make sense to either define such a function in C++ or some global lua file, to have it everywhere and everywhere the same?

indra/newview/llluamanager.cpp
maxim-productengine: Good point. It seems that using `utf8str_to_utf16str` should be enough to fix it.

indra/newview/llluamanager.cpp
```diff
@@ -358,8 +351,8 @@ class LuaRemover
 // push the loaded module or throw a Lua error
 void LLRequireResolver::findModule()
 {
-    // If mPathToResolve is absolute, this replaces mSourceChunkname.parent_path.
-    auto absolutePath = (std::filesystem::path((mSourceChunkname)).parent_path() / mPathToResolve).u8string();
```
maxim-productengine, line 362: I did a quick test locally (without this changeset), and, as Nicky-D said, unicode symbols in the path cause the issue. But changing this line to
` auto absolutePath = (std::filesystem::path((utf8str_to_utf16str(mSourceChunkname))).parent_path() / mPathToResolve).u8string();` fixes it.

maxim-productengine (approved):

indra/llcommon/lua_function.cpp
nat-goodspeed: You have a good point, but this all agrees with embedded Lua convention: nobody expects ~~the Spanish~~ a `nullptr` `lua_State*`. I could defend this whole source file against `(! L)` (by skipping the whole function body? by aborting the viewer?) but I suspect that if invalid `lua_State*` pointers start flying around, we'd just crash in the Lua VM anyway.

indra/llcommon/lua_function.cpp
nat-goodspeed: Coding a new Lua-accessible C++ entry point involves a certain amount of boilerplate. I'm particularly bothered by frameworks that require you to (a) code a function and then (b) register that function at runtime in some other function far removed, and this is one such. It's too easy to code a new function in all sincerity, and then spend debugging time trying to figure out why your Lua script can't use it.

`lua_function.h` [defines a macro](https://github.com/secondlife/viewer/blob/lua-startup/indra/llcommon/lua_function.h#L161-L171) that declares and instantiates [a `LuaFunction` subclass](https://github.com/secondlife/viewer/blob/lua-startup/indra/llcommon/lua_function.h#L133-L142) and introduces the definition of its `static int call(lua_State* L)` method. Initializing the static instance takes care of registering the particular entry point function.

Every Lua entry point accepts exactly `(lua_State*)` and returns `int`, as described in [the Lua manual](https://www.lua.org/manual/5.1/manual.html#lua_CFunction). Incoming parameters are found on the Lua stack via operations on the `lua_State*`, return values are pushed on the same stack. The returned `int` specifies the number of stack entries to be considered return values.

If the Lua VM calls any such entry point with a null `lua_State*`, things are already badly busted. If anyone else intentionally calls one of these entry points, it's a coding error.

indra/newview/llluamanager.cpp
nat-goodspeed: Okay, except (a) it needs to work on Posix as well as Windows and (b) [we should be actively moving away from `llutf16string`](https://github.com/secondlife/viewer/blob/lua-startup/indra/llcommon/llstring.h#L635-L645). I'll submit an alternative.

indra/newview/scripts/lua/Queue.lua
nat-goodspeed: Classic limerick form!

indra/newview/scripts/lua/leap.lua
nat-goodspeed: I've spent enough time debugging these modules to want a quick way to turn debugging output on or off on a per-module basis. This incantation:
```
local function dbg(...) end
-- local dbg = require 'printf'
```
lets me turn on module-specific debugging output by merely deleting the `--` prefix.

Nicky-D: I'd still argue for a global function, taking as first argument the location it is coming from. (Much like the underused LL_XXX("where") )

Not only will this enrich the final logs, but also logging can then be tweaked on a "where are you from" basis

indra/newview/llluamanager.cpp
nat-goodspeed: > C:\Prögrämme\

Windows Heavy Metal Edition

indra/llcommon/llleaplistener.cpp
nat-goodspeed: https://github.com/secondlife/viewer/pull/1110/commits/3f876a6a1d138c31266afb6c39df7090e304efe3

nat-goodspeed:
test comment

Issue #1116, opened 2024-04-02
HDRI Split screen breaks reflection probes.


RunitaiLinden:
### Description

When HDRI split screen is enabled, reflection probes have seams in them.  Make reflection probe updates always use the HDRI sky when present.  Also render the full EEP sky when split screen is zero.

sergiiiproductengine:
Deferred QA.
Verified on the Second Life Release 7.1.9.10220184061 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/301.

The issue was verified with next steps:
1. Logg into the viewer to the http://maps.secondlife.com/secondlife/Bug%20Island%202/103/148/24 location with a lots of reflection probes.
2. Load HDRI using Develop > Render Tests > HDRI Preview (HDRI was taken from the https://polyhaven.com/hdris).
3. Adjust RenderHDRISplitScreen setting via Advanced > Show Debug Settings from 1.0 to zero value

Observed behaviour: 

- after step 2 HDRI appears on all reflection probes
- after step 3 reflection probes still show HDRI (with any split screen values)

In general reflection probes do not looks broken after spliting, but could you please provide the test plan to verify this issue more clearly and properly?

PR #1119, opened 2024-04-02
#1116 HDRI preview enhancements

RunitaiLinden:
Don't split sky in radiance and irradiance maps
Allow for applying HDRI sky to only irradiance map Allow for showing entire EEP sky (clouds and all) when split is set to zero


Geenz (approved):
👍

PR #1127, opened 2024-04-03
viewer#1117 Use attacment info for declouding logic

akleshchev:
1. Use attachment count from simulator to determine cloud' state, but don't wait forever (wait 120s)
2. Reduce FIRST_APPEARANCE_CLOUD_MAX_DELAY since we no longer need to wait for potential attachments. If we got here, we know that we either waited long enough or that all attachments are already there.


marchcat (approved):

akleshchev:
Despite added 120s wait time avatars now load faster, since that is a 'something we are sure is there haven't loaded' delay, it is the 'wait just in case we don't know something' delay that was causing bulk of the waiting.

PR #1147, opened 2024-04-06
Linux GHA builds

Nicky-D:
Add Ubuntu 22.04 runner and Linux dependencies

Do not even try to touch ReleaseFS for Linux yet (this needs KDU, Havok, FMOD)

@bennettgoble this could work as a first rev. to get automated Linux builds via the standard pipeline. I'll see what else missed after a first run

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,7 +11,7 @@ jobs:
   build:
     strategy:
       matrix:
-        runner: [windows-large, macos-12-xl]
+        runner: [windows-large, macos-12-xl, ubuntu-22.04]
         configuration: [Release, ReleaseOS]
         python-version: ["3.11"]
         include:
@@ -20,6 +20,8 @@ jobs:
         exclude:
           - runner: macos-12-xl
             configuration: ReleaseOS
+          - runner: ubuntu-22.04
+            configuration: Release
     runs-on: ${{ matrix.runner }}
     outputs:
       viewer_channel: ${{ steps.build.outputs.viewer_channel }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
   build:
     strategy:
       matrix:
-        runner: [windows-large, macos-12-xl]
+        runner: [windows-large, macos-12-xl, ubuntu-22.04]
```
bennettgoble, line 14: Like with `windows-large` we have a larger runner available for linux builds, which this build can use by selecting `linux-large` instead of the commodity `ubuntu-22.04`

Screenshot 2024-04-06 085638

Nicky-D: Thank you for the info. I moved that build over to the large runner. I'm not sure if that has has access to AUTOBUILD_GITHUB_TOKEN  (the previous build failed due to that).

Though it is strange anyway that ReleaseOS comes with HAVOK=On when kicked off by GHA (I verified with Windows, there ReleaseOS always builds with Havok)

.github/workflows/build.yaml
```diff
@@ -11,7 +11,8 @@ jobs:
   build:
     strategy:
       matrix:
-        runner: [windows-large, macos-12-xl]
+        runner: [windows-large, macos-12-xl, linux-large]
+        ]
```
bennettgoble, line 15: Oops, little artifact left here.

Nicky-D:
Looks like we are in business:
- The build will run on ubuntu-larg
- Only ReleaseOS will be built
- HAVOK=On will be forcefully changed in build.sh (and crashreporting)

For some reason the Windows/OSX runner do have access to a github token and can download the havok source even during PR builds, the Ubuntu one could not.

For me this is good enough atm, as I cannot do much more regarding builds w/o the secret bits for "Release"

bennettgoble (approved):

marchcat (approved):

PR #1150, opened 2024-04-07
[WebRTC] Explicit decline message for webrtc p2p sessions.

roxanneskelly:
WebRTC P2P uses a different mechanism for managing p2p voice calls than Vivox, as WebRTC uses the existing messaging infrastructure to handle calls, declines, accepts, etc.  Vivox uses a vivox-supplied mechanism for p2p calls.

This required us to add explicit p2p call and p2p decline messaging.

Also, in this PR, are some linux-related code fixes and some renaming of items to be more clear.

### DIFF
--- a/indra/llwebrtc/llwebrtc.cpp
+++ b/indra/llwebrtc/llwebrtc.cpp
@@ -515,16 +515,6 @@ void LLWebRTCImpl::updateDevices()
         char name[webrtc::kAdmMaxDeviceNameSize];
         char guid[webrtc::kAdmMaxGuidSize];
         mTuningDeviceModule->PlayoutDeviceName(index, name, guid);
-
-#if WEBRTC_LINUX
-        // Linux audio implementation (pulse and alsa)
-        // return empty strings for the guid, so 
-        // use the name for the guid
-        if (!strcmp(guid, ""))
-        {
-            strcpy(guid, name);
-        }
-#endif // WEBRTC_LINUX
         mPlayoutDeviceList.emplace_back(name, guid);
     }
 
@@ -543,15 +533,6 @@ void LLWebRTCImpl::updateDevices()
         char name[webrtc::kAdmMaxDeviceNameSize];
         char guid[webrtc::kAdmMaxGuidSize];
         mTuningDeviceModule->RecordingDeviceName(index, name, guid);
-#if WEBRTC_LINUX
-        // Linux audio implementation (pulse and alsa)
-        // return empty strings for the guid, so
-        // use the name for the guid
-        if (!strcmp(guid, ""))
-        {
-            strcpy(guid, name);
-        }
-#endif  // WEBRTC_LINUX
         mRecordingDeviceList.emplace_back(name, guid);
     }
 
--- a/indra/llwebrtc/llwebrtc.cpp
+++ b/indra/llwebrtc/llwebrtc.cpp
@@ -26,7 +26,6 @@
 
 #include "llwebrtc_impl.h"
 #include <algorithm>
-#include <format>
 #include <string.h>
 
 #include "api/audio_codecs/audio_decoder_factory.h"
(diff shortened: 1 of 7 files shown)


indra/llwebrtc/llwebrtc.cpp
```diff
@@ -516,13 +515,23 @@ void LLWebRTCImpl::updateDevices()
         char name[webrtc::kAdmMaxDeviceNameSize];
         char guid[webrtc::kAdmMaxGuidSize];
         mTuningDeviceModule->PlayoutDeviceName(index, name, guid);
+
+#if WEBRTC_LINUX
```
Nicky-D, line 519: I suggest turning this either into a function ala
```
LLWebRTCVoiceDevice createWebRTCVoiceDevice( std::string name, std::string guid )
{
#if LL_LINUX
    if( guid.empty() )
        guid = name;
#endif

    return {name, guid };
```

and then call this in the two places. Or just be lazy and modify the ctor of LLWebRTCVoiceDevice

```
    LLWebRTCVoiceDevice(const std::string &display_name, const std::string &id) :
        mDisplayName(display_name),
        mID(id)
    {
        if( mID.empty() )
            mID = mDisplayName;
    };
```

in any case is my suggestion to get rid of:
- the strcpy which can easily lead to buffer overflows if one of the enums above changes
- the duplicate code of checking for an empty GUID

indra/newview/llimview.cpp
```diff
@@ -2727,7 +2727,7 @@ LLIncomingCallDialog::LLIncomingCallDialog(const LLSD& payload) :
 
 void LLIncomingCallDialog::onLifetimeExpired()
```
Nicky-D, line 2728: This function looks quite bad in the webview. I suspect it has spaces and tabs mixed. Settling on one would be preferred

indra/llwebrtc/llwebrtc_impl.h
```diff
@@ -35,6 +35,7 @@
 #define WEBRTC_POSIX 1
 #elif __linux__
```
Nicky-D, line 36: I suppose the LL-way is to check for LL_LINUX, LL_WINDOWS, LL_DARWIN. This makes grepping easier.

Which at the same time makes me curious why there is a need for WEBRTC_xxx when there is already such a thing as LL_xxx.

And another small nitpick
LL_**WINDOWS** <> LL_**WIN**
LL_**DARWIN** <> WEBRTC_**MAC** (you're my hero btw if you'd manage to get DARWIN changed to OSX or MAC)

indra/llwebrtc/llwebrtc.cpp
roxanneskelly: Done (the second option)

indra/newview/llimview.cpp
roxanneskelly: Should be better now

indra/llwebrtc/llwebrtc_impl.h
roxanneskelly: I'm trying to use the WebRTC way of doing everything (thread model, memory, platform identifiers, etc.) in llwebrtc.  It's the 'interface' that wraps webrtc in a very simple wrapper for the viewer.  Ultimately, the goal is to move it out of the viewer into something downloaded by autobuild (or whatever a TPV dev uses) where I won't have access to viewer-related headers and build-system constants.

indra/llwebrtc/llwebrtc.cpp
Nicky-D: Thank you :)

indra/newview/llimview.cpp
Nicky-D: Yes, looks awesome now

indra/llwebrtc/llwebrtc_impl.h
Nicky-D: Got it, that makes sense.
For grepability  I'd still prefer _WINDOWS and _DARWIN (even though I'd personally prefer _MAC)
But then  the suffix is the same across the sources.

roxanneskelly: The constants listed are used in the webrtc headers so I'm kinda stuck using WEBRTC_*, unfortunately.

marchcat (approved):

Issue #759, opened 2024-02-08
2k texture PBR terrain does not render on the minimap


cosmic-linden:
- [x] Prerequisite: https://github.com/secondlife/viewer-private/issues/151

cosmic-linden:
See `LLVLComposition::generateMinimapTileLand` - in particular, `tex->getRawImageLevel() != ddiscard` evaluates to true

cosmic-linden:
There may be a component to this bug which relates to texture loading.

Minimap generation is also tied to terrain rebuilds. Which may have implications for [this ticket](https://github.com/secondlife/jira-archive-internal/issues/69752). i.e. there potentially is a performance implication to this fix.

Will continue looking into this later. I'm not convinced this is strictly related to 2K PBR terrain. However, that's the case that currently reproduces reliably for me. So, I will assume this is PBR terrain-related for now.

cosmic-linden:
Takeaways for the glorious future:

- Get rid of calls to setBoostLevel as it keeps textures in memory forever
- Don't use raw images for minimap in favor of shader. Reduce generation of raw images where possible.

cosmic-linden:
Re-opening as this issue still reproduces for me.

I noticed something which may be relevant. Unlike their 1K PBR counterparts, my 2K PBR testing terrain has all-black emissive textures which share the same texture ID (due to deduplication). Because of how the minimap code is currently designed, this causes it to perturb the state of `LLViewerTexture` several times. This could be breaking the state machine which manages the resolution of the raw images.

I'm not sure if this is the cause of the bug, but it could be making the bug much more reproducible.

If so, it has implications for the minimap not regenerating when editing the terrain composition.

cosmic-linden:
Will be fixed during the course of https://github.com/secondlife/viewer/issues/1184

sergiiiproductengine:
Passed QA.

Verified on the Second Life Release 7.1.7.8944739879 on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/213.
Verified on Aditi on PMFP 2024-04-13.8669470587 region.

Note: materials taken from the "2k terrain textures" box at secondlife://Aditi/secondlife/Rumpus%20Room%202048/196/54/28 was used for verifying this issue.

PR #1162, opened 2024-04-08
secondlife/viewer#759: Fix minimap loading for 2k PBR materials

cosmic-linden:
## Test plan

Please see "Minimap" in https://github.com/secondlife/viewer/blob/release/materials_featurette/doc/testplans/pbr_terrain_composition.md


RunitaiLinden (approved):

cosmic-linden:
Takeaways for the glorious future:

- Get rid of calls to setBoostLevel  as it keeps textures in memory forever
- Don't use raw images for minimap in favor of shader. Reduce generation of raw images where possible.

PR #1198, opened 2024-04-11
CI: adopt xz compression, actions/*-artifact@v4

bennettgoble:
Move towards packaging artifacts with xz, which offers modestly higher compression ratios and faster decode time, and adopt viewer-build-utils@v2 and GHA's newer upload/download-artfact@v4.

### CI stage time improvements

- Upload executable (Windows Release) `5-7min → 36s`
- Upload symbols (Windows Release) `33s → 5s`
- Similar improvements with macOS

### Dropping windows releaseOS

Tucked into this PRs commits is the deprecation of automatic windows ReleaseOS configuration builds This change was made because the release/maint-b branch contains a new ReleaseOS linux configuration which we can build in its stead. Building linux ReleaseOS instead of windows will save a considerable amount on runners.

### Todo

- [x] Merge https://github.com/secondlife/viewer-build-util/pull/16 and release a new v2 version
- [x] Update this PR, bumping viewer-build-util actions to v2

### Links

- Example build: https://github.com/secondlife/viewer/actions/runs/8653991373
- Example release: https://github.com/secondlife/viewer/releases/tag/Second_Life_Test%230f94ea8

### Questions

Is this the right branch to target? As the changes cut 5 minutes off from mac and windows builds should they be more widely disseminated? Perhaps merged into main?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,13 +12,10 @@ jobs:
     strategy:
       matrix:
         runner: [windows-large, macos-12-xl]
-        configuration: [Release, ReleaseOS]
+        configuration: [Release]
         include:
           - runner: macos-12-xl
             developer_dir: "/Applications/Xcode_14.0.1.app/Contents/Developer"
-        exclude:
-          - runner: macos-12-xl
-            configuration: ReleaseOS
     runs-on: ${{ matrix.runner }}
     outputs:
       viewer_channel: ${{ steps.build.outputs.viewer_channel }}
@@ -100,7 +97,7 @@ jobs:
 
       - name: Determine source branch
         id: which-branch
-        uses: secondlife/viewer-build-util/which-branch@v1
+        uses: secondlife/viewer-build-util/which-branch@v2
         with:
           token: ${{ github.token }}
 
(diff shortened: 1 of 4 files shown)


bennettgoble:
@nat-goodspeed thanks for the initial review. I've updated the PR so that it uses the new v2 viewer-build-utils release and targets main.

nat-goodspeed (approved):
Good.

Since we have so many repeated references to certain action repos, should we use a variable for their specific version/branch? Having to tweak so many different places to be consistent feels wrong.

bennettgoble:
> Since we have so many repeated references to certain action repos, should we use a variable for their specific version/branch? Having to tweak so many different places to be consistent feels wrong.

Maybe, though I've not seen this done before. It may be best to leave actions explicitly versioned, as they can be automatically bumped by dependabot.

There are certainly a few areas where we could combine actions: post-bugsplat-mac + post-bugsplat-windows could become post-bugsplat, and the same for sign-pkg.

nat-goodspeed:
I take your point about dependabot. Would dependabot have picked up on the upload-artifact@v4 and viewer-build-util@v2 changes and filed a PR covering all those references? If so, maybe all we need is to be patient and let dependabot do its thing.

The major reason for breaking out separate Windows and Mac jobs for posting symbols and sign-and-package is that these have historically, and even recently, been less than wonderfully reliable operations.

I could see combining posting symbols to BugSplat because it's the same service, even though we have in fact had to repost (typically) Mac symbols separately from Windows symbols.

But I'm not in favor of combining the sign-and-package jobs because they use different services with different EV certs and have demonstrated distinct failure modes.

PR #1199, opened 2024-04-11
CI: Remove python-version from matrix

bennettgoble:
Drop python version from matrix configuration as it's always 3.11.


brad-linden (approved):
we can re-add later if needed

marchcat (approved):

nat-goodspeed (approved):

PR #1204, opened 2024-04-11
Merge main into maint-A

Ansariel:
@marchcat So we don't have to wait so long again to get it merged. ;)

There was quite a nice merge conflict in lltranslate.cpp, but I think it should be correct like this.


marchcat:
Thank you, @Ansariel!
This was my plan for tonight 😆

marchcat (approved):

PR #1205, opened 2024-04-11
Chore/cmake pretty up

Nicky-D:
This PR adds:
- An option to compile with address sanitation enabled (~2 times slowdown)
- Fixes the option GCC_DISABLE_FATAL_WARNINGS
- Tweaks the mold linker logic a little bit, the most important change is to not make it fail when mold is not found. This seemed to be the desired outcome going by the logic.

@bennettgoble ping for you, re. mold


bennettgoble (approved):
LGTM, thanks!

indra/cmake/Linker.cmake
```diff
@@ -1,7 +1,7 @@
 include_guard(GLOBAL)
 
-if (${LINK_WITH_MOLD})
-  find_program(MOLD_BIN mold REQUIRED)
```
bennettgoble, line 4: Ah, thanks. Don't want to mark it required if there's optional handling.

indra/cmake/00-Common.cmake
```diff
@@ -159,6 +159,7 @@ if (LINUX)
   set(CLANG_WARNINGS
       ${GCC_CLANG_COMPATIBLE_WARNINGS}
       # Put clang specific warning configuration here
+      -Wno-unknown-warning-option
```
bennettgoble, line 162: Hmm, the point of breaking out warnings into separate GCC_WARNINGS / GCC_CLANG_COMPATIBLE_WARNINGS / CLANG_WARNINGS variables was so that this kind of carpet ignore clause could be avoided. Is there an alternative? (Just accept the single warning?)

Nicky-D: That whole ordeal started with snipped in lsdutil.cpp
``` C++

#if LL_WINDOWS
...
#elif LL_LINUX
#	include <netinet/in.h>
#pragma GCC diagnostic ignored "-Wstringop-truncation" // It's actually okay what happens here
#elif LL_DARWIN
...
#endif
```

Admittedly in that moment I took out the sledgehammer to get rid of an error that seemed/is totally unrelated to my changes and just popped up 'out of nowhere`.

With your explanation the rationale and intent behind your change of splitting the flags makes a lot more sense. 

A solution not needing that -Wno-unknown-warning-option would then be:
- Hunt down all #pragma GCC
- Put some lipstick on them by surrounding them with "#ifndef <define_that_clang_sets>"  (__clang__?)

Suggestion:
- We accept ` -Wno-unknown-warning-option` for this PR, as I would prefer to not taint this one by mixing it with a clang/gcc cleanup
- Follow up PR to remove the one above and try to work with __clang__ and the preprocessor

marchcat (approved):
Thanks!

PR #1212, opened 2024-04-12
Make it so that reset skeleton is a networked ViewerEffect

FelixWolf:
This fixes https://github.com/secondlife/jira-archive/issues/4306

This change does the following:
* Create the `LLHUDEffectResetSkeleton(LLHUDEffect)` class
* Create the `LL_HUD_EFFECT_RESET_SKELETON` viewer effect type
* Move handling of `resetSkeleton(bool resetAnimations)` from floater callbacks to the `LLHUDEffectResetSkeleton::update` function
* Change the "Reset Skeleton" floater callbacks to create the `LLHUDEffectResetSkeleton` viewer effect
* Allow networked resetting of skeleton of own avatar
* Allow networked resetting of skeleton of owned animesh
* Allow local resetting of skeleton of any avatar (Existing behavior)
* Allow local resetting of skeleton of any animesh (Existing behavior)

This creates a new viewer effect message with the ID of 17. The following format is used in the `ViewerEffect.Effect[].TypeData` section:
```cpp
struct LLHUDEffectResetSkeleton_data {
    LLUUID TargetObject; // It is "valid" for this to be NULL_KEY, in which it will default to the sender's ID
    uint8_t ResetAnimations;
};
```
The `ResetAnimations` field is prepared in a way that only the first bit is checked, allowing this to be moved to a flags in the future if need be.

Ownership of a effect is determined by the `ViewerEffect.Effect[].AgentID` field, which I am suspecting is verified by the simulator based off the `ViewerEffect.AgentData.AgentID` field. My testings show that you cannot spoof a ViewerEffect's AgentID, but this should be double checked.

### DIFF
--- a/indra/newview/llhudeffectresetskeleton.cpp
+++ b/indra/newview/llhudeffectresetskeleton.cpp
@@ -1,0 +1,207 @@
+/** 
+ * @file llhudeffectresetskeleton.cpp
+ * @brief LLHUDEffectResetSkeleton class implementation
+ *
+ * $LicenseInfo:firstyear=2024&license=viewerlgpl$
+ * Second Life Viewer Source Code
+ * Copyright (C) 2024, Linden Research, Inc.
+ * 
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation;
+ * version 2.1 of the License only.
+ * 
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ * 
+ * Linden Research, Inc., 945 Battery Street, San Francisco, CA  94111  USA
+ * $/LicenseInfo$
+ */
+
+#include "llviewerprecompiledheaders.h"
+
+#include "llhudeffectresetskeleton.h"
+
+#include "llagent.h"
+#include "llviewerobjectlist.h"
+#include "llvoavatar.h"
+#include "message.h"
+
+// packet layout
+const S32 TARGET_OBJECT = 0; // This is to allow for targetting owned animesh
+const S32 RESET_ANIMATIONS = 16; //This can also be a flags if needed
+const S32 PKT_SIZE = 17;
+
+//-----------------------------------------------------------------------------
+// LLHUDEffectResetSkeleton()
+//-----------------------------------------------------------------------------
+LLHUDEffectResetSkeleton::LLHUDEffectResetSkeleton(const U8 type) : 
+	LLHUDEffect(type)
+{
+}
+
+//-----------------------------------------------------------------------------
+// ~LLHUDEffectResetSkeleton()
+//-----------------------------------------------------------------------------
+LLHUDEffectResetSkeleton::~LLHUDEffectResetSkeleton()
+{
+}
+
+//-----------------------------------------------------------------------------
+// packData()
+//-----------------------------------------------------------------------------
+void LLHUDEffectResetSkeleton::packData(LLMessageSystem *mesgsys)
+{
+	// Pack the default data
+	LLHUDEffect::packData(mesgsys);
+
+	// Pack the type-specific data.  Uses a fun packed binary format.  Whee!
+	U8 packed_data[PKT_SIZE];
+	memset(packed_data, 0, PKT_SIZE);
+
+	// pack both target object and position
+	// position interpreted as offset if target object is non-null
+	if (mTargetObject)
+	{
+		htolememcpy(&(packed_data[TARGET_OBJECT]), mTargetObject->mID.mData, MVT_LLUUID, 16);
+	}
+	else
+	{
+		htolememcpy(&(packed_data[TARGET_OBJECT]), LLUUID::null.mData, MVT_LLUUID, 16);
+	}
+	
+	U8 resetAnimations = (U8)mResetAnimations;
+	htolememcpy(&(packed_data[RESET_ANIMATIONS]), &resetAnimations, MVT_U8, 1);
+
+	mesgsys->addBinaryDataFast(_PREHASH_TypeData, packed_data, PKT_SIZE);
+}
+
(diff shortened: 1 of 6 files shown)


FelixWolf:
I'm not entirely happy about repeating the
```cpp
			bool owned = false;
			if(avatar->isAnimatedObject())
			{
				owned = avatar->mOwnerID == gAgent.getID();
			}
			else
			{
				owned = avatar->getID() == gAgent.getID();
			}
```

Perhaps I missed a function to check this properly, but it seems to be the only way to check ownership of both a avatar or animated object depending on type consistently.

marchcat:

Let's put it into Maint C.

indra/newview/llviewermenu.cpp
```diff
@@ -6505,7 +6506,19 @@ class LLAvatarResetSkeleton: public view_listener_t
...
+				owned = avatar->mOwnerID == gAgent.getID();
+			}
+			else
+			{
+				owned = avatar->getID() == gAgent.getID();
+			}
+			LLHUDEffectResetSkeleton* effectp = (LLHUDEffectResetSkeleton*)LLHUDManager::getInstance()->createViewerEffect(LLHUDObject::LL_HUD_EFFECT_RESET_SKELETON, owned);
+			effectp->setSourceObject(gAgentAvatarp);
+			effectp->setTargetObject((LLViewerObject*)avatar);
+			effectp->setResetAnimations(false);
```
akleshchev, line 6521: If the point is to allow resetting selected objects, you should check not just ownership but LLPermissions.

indra/newview/llhudeffectresetskeleton.cpp
```diff
@@ -0,0 +1,207 @@
...
+	
+	LLUUID target_id;
+	htolememcpy(target_id.mData, &(packed_data[TARGET_OBJECT]), MVT_LLUUID, 16);
+
+	if (target_id.isNull())
+	{
+		target_id = source_id;
+	}
+	
+	objp = gObjectList.findObject(target_id);
```
akleshchev, line 127: This needs to check if source has permission to modify target, otherwise some people will remove the 'on send' check and start spamming 'hud effect' at everyone around them

akleshchev (changes requested):

akleshchev:
And someone from server side needs to check if value for LL_HUD_EFFECT_RESET_SKELETON is reserver for something else...

indra/newview/llhudeffectresetskeleton.cpp
```diff
@@ -0,0 +1,207 @@
...
+		owned = mTargetObject->mOwnerID == mSourceObject->getID();
+	}
+	else
+	{
+		owned = mTargetObject->getID() == mSourceObject->getID();
+	}
+
+	if (owned)
+	{
+		if (mTargetObject->isAvatar() || mTargetObject->isAnimatedObject())
```
akleshchev, line 200: This sems wrong. Animated objects have attached control avatars but aren't avatars so you need to mVObjp->getControlAvatar() and then reset skeleton of the control avatar.

indra/newview/llhudeffectresetskeleton.cpp
```diff
@@ -0,0 +1,207 @@
...
+	}
+	
+	bool owned = false;
+	if(mTargetObject->isAnimatedObject())
+	{
+		owned = mTargetObject->mOwnerID == mSourceObject->getID();
+	}
+	else
+	{
+		owned = mTargetObject->getID() == mSourceObject->getID();
```
akleshchev, line 195: This will cover objects and agents, but not animated(control) avatars, if target object is a control avatar, you need to either treat it like an object and compare against 'source' id (unless animated avatars are supposed to use this as well and reset themselves) or get control avatars' control object and get it's permissions.

FelixWolf:
I've realized now that I don't think this should stop animesh animations, as those are controlled simulator side, and would cause a desync.
i will fix this too.

akleshchev:
> I don't think this should stop animesh animations

Agree, thanks for the contribution.

FelixWolf:
So looks like `LLPermissions` isn't in memory for objects, and instead has to be requested, and the only API that I can find that does that is LLSelectMgr, which would only have it if they selected the object.

I'm half tempted to just strip out the ability to reset animesh for the time being, as it can easily just be de-rezzed and rezzed again if need be.

bennettgoble:
> And someone from server side needs to check if value for LL_HUD_EFFECT_RESET_SKELETON is reserver for something else...

It's available and doesn't conflict.

FelixWolf:
After much testing (and not being able to get animesh to reset properly due to a variety of issues, such as mOwnerID not being populated, requiring to poll LLPermissions, animations not controllable via viewer, etc), I've decided to just strip the animesh reset portion for now.
I'm leaving the protocol unchanged, so that when this is revisited in the future, it is forward compatible.
I'll be submitting the change briefly once I confirm the code works.

FelixWolf:
Er, accidentally force pushed out of force of habit. Wanted to add a link to this issue just in case someone else comes by and wants to make it work for animesh.
Probably should have just did a regular push, but doesn't seem to conflict with anything.
My bad.

indra/newview/llviewermenu.cpp
```diff
@@ -6505,7 +6506,17 @@ class LLAvatarResetSkeleton: public view_listener_t
         }
 		if(avatar)
         {
-            avatar->resetSkeleton(false);
+			if(avatar->getID() == gAgent.getID())
```
akleshchev, line 6509: Could have used gAgentID instead of gAgent.getID()

akleshchev (approved):

marchcat (approved):

akleshchev:
```
/Users/runner/work/viewer/viewer/indra/newview/llviewermenu.cpp:6511:159: error: use of undeclared identifier 'owned'
                                LLHUDEffectResetSkeleton* effectp = (LLHUDEffectResetSkeleton*)LLHUDManager::getInstance()->createViewerEffect(LLHUDObject::LL_HUD_EFFECT_RESET_SKELETON, owned);
```

FelixWolf:
Oh, whoops! One sec.

akleshchev:
```
/Users/runner/work/viewer/viewer/indra/newview/llviewermenu.cpp:6518:5: error: use of undeclared identifier 'effectp'
                                effectp->setResetAnimations(false);
                                ^
```

FelixWolf:
Fixed the compile issue, and also went ahead and put the recommendations in as well.

FelixWolf:
Oh 🤦🏻‍♂️
This is what I get for developing on linux. Sorry, one sec.

brad-linden (approved):

AtlasLinden:
Internal jira issue: https://github.com/secondlife/jira-archive-internal/issues/86216

PR #1226, opened 2024-04-13
Chore/pragma gcc cleansweep

Nicky-D:
Follow up to PR 1205

Remove #pragma GCC so we can get rid of  -Wno-unknown-warning-option for clang.

This resulted in  -Wno-stringop-overflow &  -Wno-stringop-truncation also being removed, as compilation was successful after making a nicer LLERROR_CRASH implementation for Linux

Ping for @bennettgoble due to suggestion of this in PR 1025


Nicky-D:
I see I accidentally snuck my CMakePreset.json in here. Wasn't there plan really, but now that it is here can we maybe keep it?

I used it extensively because such a file enables a workflow with clion (can run, debug, and build in there).

bennettgoble:
> I see I accidentally snuck my CMakePreset.json in here. Wasn't there plan really, but now that it is here can we maybe keep it?
> 
> I used it extensively because such a file enables a workflow with clion (can run, debug, and build in there).


Could you submit the CMakePresets.json addition as a separate PR? We could probably entertain the notion.

bennettgoble (changes requested):
Put CMakePresets into a separate PR.

bennettgoble (approved):

indra/llcommon/llerror.h
```diff
@@ -95,6 +95,11 @@ const int LL_ERR_NOERR = 0;
 #define LL_STATIC_ASSERT(func, msg) static_assert(func, msg)
 #define LL_BAD_TEMPLATE_INSTANTIATION(type, msg) static_assert(false, msg)
 #else
+#if LL_LINUX
+// We need to include Singal Linden to make the build moat better (also so we ggot access to raise and SIGSEGV
```
bennettgoble, line 99: I appreciate the reference --let's keep the comments germane, however. Remove my name and I will gladly merge. :)

Nicky-D: Signal Linden has left the building

akleshchev:

PR #1227, opened 2024-04-13
Inject XZ_DEFAULTS=-T0 into the environment. This will speed up archive generation.

Nicky-D:
This will make sure xz archive compression will use all available threads.

@bennettgoble this might be of interest to you, as it can shave off a good amount of time from builds


bennettgoble (approved):

bennettgoble:
Thanks!

Issue #912, opened 2024-02-29
BugSplat Crash #1412267: nvoglv64+0xadcd00


sl-service-account:
Test plan:
* Extract the angel_2016_08_05.dae file from this zip file [angel_2016_08_05.dae.zip](https://github.com/user-attachments/files/19555715/angel_2016_08_05.dae.zip)
* Set graphics quality to High+
* Wear the Bitsy avatar from the Library.
* Right-click your nametag > Take off > Detatch all
* Build menu > Upload > Model and choose the angel_2016_08_05.dae file.

Observed: The Upload Model window will appear for a second or so, then the viewer will crash.

Expected: No crash

----

**BugSplat** [Crash 1412267](https://app.bugsplat.com/v2/crash?id=1412267&database=SecondLife_Viewer_2018)
[Crash 1433959](https://app.bugsplat.com/v2/crash?database=SecondLife_Viewer_2018&id=1433959)

**Application Version:**
7.1.1.7039128750
7.1.3.7790341084

**Callstack:**
<table><tr><td>nvoglv64</td><td></td></tr><tr><td>nvoglv64</td><td></td></tr><tr><td>0x0000029bca6f29b8</td><td></td></tr></table>

akleshchev:
Can consistently repro on High+ when uploading angel.dae

CPU: AMD Ryzen 7 3800X 8-Core Processor              (3892.69 MHz)
Memory: 32687 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4046)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2060 SUPER/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.4617
OpenGL Version: 4.6.0 NVIDIA 546.17

```
Window size: 1759x1102
Font Size Adjustment: 96pt
UI Scaling: 0.95
Draw distance: 64m
Bandwidth: 3000kbit/s
LOD factor: 1.75
Render quality: 5
Advanced Lighting Model: Enabled
Texture memory: 7251MB
Disk cache: Max size 1638.4 MB (100.0% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: FMOD Studio 2.02.13
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Not Connected
[...]

PR #1228, opened 2024-04-14
secondlife/viewer#912 BugSplat Crash 1412267: nvoglv64+0xadcd00

LLGuru:


marchcat (approved):
Nice refactoring.
Not sure if it'll help with the crash though.

akleshchev (approved):
Looks nicer, but probably won't affect video driver crash.

P.S. Please link an issue before closing.

LLGuru:
> Nice refactoring.

Thanks

> Not sure if it'll help with the crash though.

Refactoring really helps to see the logic of the algorithm
And when we see the logic we find points to fix like this:
image

Issue #1072, opened 2024-03-27
BugSplat Crash: Second Life Release!LLPersistentNotificationStorage::LLPersistentNotificationStorage()


sl-service-account:
**BugSplat** [Key Crash 91331](https://app.bugsplat.com/v2/keycrash?stackKeyId=91331&database=SecondLife_Viewer_2018)

**Sample Version:**
7.1.4.8149792635

**Sample Callstack:**
[...]

PR #1229, opened 2024-04-14
secondlife/viewer#1072 Crash: LLPersistentNotificationStorage()

LLGuru:
https://github.com/secondlife/viewer/issues/1072

Crash in the constructor of the class-singleton

We have only 3 crashreports on BugSplat dated March 27, March 26 and March 16:
https://app.bugsplat.com/v2/keycrash?stackKeyId=91331&database=SecondLife_Viewer_2018

The only visible reason is a creation of 2 different singletons based on a single check,
the second singleton existance is not being checked, and theoretically this could be a reason

In any case, it would be more correct to check each simgleton separately

Also this change removes some trainling spaces in the source code around

### DIFF
--- a/indra/newview/llstartup.cpp
+++ b/indra/newview/llstartup.cpp
@@ -436,7 +436,7 @@ bool idle_startup()
 		{
 			LLNotificationsUtil::add(gViewerWindow->getInitAlert());
 		}
-			
+
 		//-------------------------------------------------
 		// Init the SOCKS 5 proxy if the user has configured
 		// one. We need to do this early in case the user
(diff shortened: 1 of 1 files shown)


akleshchev:
Plese link an issue.

akleshchev:
And how does this fix the crash of a constructor?

This fix makes no sense, either both storages are inited simultaneously or both are reset simulataneously since they don't get initialized in any other part of code.

P.S. Please link an issue to this pull request.

Nicky-D:
@akleshchev is Mac compiled with fpo (usually this is the default)?

Could be worth a try to compile without and test how this fares with call stacks.

 "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" should do the trick.

On Mx Mac's there's also Rosetta. Might be another thing that interferes with proper call stacks.

LLGuru:
> @akleshchev is Mac compiled with fpo (usually this is the default)?

Unfortunately we have no steps to reproduce this crash
As I wrote above, we have only 3 reports from this crash

So this change could be treated as a 100% safe refactoring
with some chance to fix the issue

akleshchev:
> with some chance to fix the issue

How does checking LLDoNotDisturbNotificationStorage::instanceExists() helps with the crash in LLPersistentNotificationStorage's constructor?

Please link to the ticket.

> @akleshchev is Mac compiled with fpo (usually this is the default)?

Sorry, not sure what 'fpo' is and google is not helpful for something this short, can you provide a link? As far as I can tell build arguments are minimalistic so everything important should be default.

Callstacks are readable, call hierarhy is correct, but they do not match logs and make zero sense in this particular case. calltack is here: https://github.com/secondlife/viewer/issues/1072/

PR #1230, opened 2024-04-14
Stop vo cache related tp hangs

beqjanus:
The vo cache was extended for PBR to include a new extras cache, which holds the TE metadata for PBR texture channels. Only the `add` functionality was included, and so the cache grows forever. This is exacerbated by a server-side "bug" that sends a lot of empty overrides, which are added to the cache. Our users were reporting significant delays when TPing to a region, typically the worst affected would be the users home region or build region where they spend most time (I saw 15 seconds, but others were seeing far longer to the point of being disconnected). The underlying cause of the TP hangs turned out to be SLEC files in the object cache. The largest I have seen reported is over 1GB. 

This code was released in the most recent PBR beta for FS and users report that the issues have vanished (\o/)

This change adds vocache extras support for: 
* killobject()
* purge and remove of cache itself.
* filtering empty extras updates.
* verifying on import (checking that the localid referenced in the extras cache has a main cache entry to be "extra" to)

The result should be a typical extras cache size of a few MB.

The change also adds a Version number to the cache format. Aside from being good practice this makes it easier to migrate to a more efficient cache later (streaming LLSD as XML just sets my teeth on edge) as the existence of "extras" becomes the norm rather than the exception.

I can't easily compile this locally, so if I've derped the merge, let me know, and I'll take a look.

Beq


marchcat (approved):

marchcat (changes requested):
macos build fails with the following:

```
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:59:45: error: no member named 'VERSION_LABEL' in 'LLGLTFOverrideCacheEntry'
const std::string LLGLTFOverrideCacheEntry::VERSION_LABEL = {"GLTFCacheVer"};
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:60:37: error: no member named 'VERSION' in 'LLGLTFOverrideCacheEntry'
const int LLGLTFOverrideCacheEntry::VERSION = 1;
          ~~~~~~~~~~~~~~~~~~~~~~~~~~^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1561:104: error: use of undeclared identifier 'filename'
        LL_DEBUGS("GLTF", "VOCache") << "Read " << cache_entry_map.size() << " entries from object cache " << filename << ", expected " << num_entries << ", success=" << (success?"True":"False") << LL_ENDL;
                                                                                                              ^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1682:1: error: function definition is not allowed here
{
^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1697:1: error: function definition is not allowed here
{
^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1829:1: error: function definition is not allowed here
{
^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1840:1: error: function definition is not allowed here
{
^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1936:2: error: expected '}'
}
 ^
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1567:1: note: to match this '{'
{
^
8 errors generated.
```

beqjanus:
Odd I'll take a look. It compiles on Mac in FS, I must have placed it differently. CLang is always so fussy.

beqjanus:
Those two should fix it. I missed them in the merge it seems.

beqjanus:
While I think about it, @brad-linden should be added as a reviewer as he is apparently reviewing this whole area.

bennettgoble:
@marchcat, small note: it looks like maint-c needs to merge from main.

marchcat:
@bennettgoble, right, I'm going to update all the maint branches at once.

marchcat (approved):

marchcat (changes requested):
One more build error:

```
/Users/runner/work/viewer/viewer/indra/newview/llvocache.cpp:1600:9: error: variable 'versionNumber' set but not used [-Werror,-Wunused-but-set-variable]
    int versionNumber=0;
        ^
```

indra/newview/llvocache.cpp
```diff
@@ -1602,8 +1602,18 @@ void LLVOCache::readGenericExtrasFromCache(U64 handle, const LLUUID& id, LLVOCac
     {
         std::string versionStr = line.substr(LLGLTFOverrideCacheEntry::VERSION_LABEL.length()+1); // skip the version label and ':'
         versionNumber = std::stol(versionStr);
-        std::getline(in, line); // read the next line for the region UUID check
     }
+    // For future versions we may call a legacy handler here, but realistically we'll just consider this cache out of date.
+    // The important thing is to make sure it gets removed.
+    if(versionNumber != LLGLTFOverrideCacheEntry::VERSION && versionNumber != 0)
+    {
+        LL_WARNS << "Unexpected version number " << versionNumber << " for extras cache for handle " << handle << LL_ENDL;
```
marchcat, line 1610: Should be `LL_WARNS()`.

There also are multiple errors after this point, but I believe they are caused by this macro.

marchcat (approved):
Builds are fine now, thank you!

brad-linden (approved):
This looks great.  I would also like to pull this into `release/materials_featurette`. should it go into both branches or should we just change the target branch of this PR?

marchcat:
@brad-linden, please feel free to retarget. Maint C is the latest one in the queue, so the featurette should be released sooner anyway.

brad-linden:
link to secondlife/jira-archive#10582

PR #1231, opened 2024-04-15
Fix broken shaders due to 72730609f371ae0417a2d117fdc9d805ce8858fb

Ansariel:
72730609f371ae0417a2d117fdc9d805ce8858fb causes shader compiler errors:
0(799) : error C1106: overloaded function declaration "sampleReflectionProbesLegacy" differs only in parameter qualifiers
0(799) : error C1106: overloaded function declaration "sampleReflectionProbesLegacy" differs only in parameter qualifiers
0(799) : error C1106: overloaded function declaration "sampleReflectionProbesLegacy" differs only in parameter qualifiers


RunitaiLinden (approved):
LGTM -- I typically just inout everything because I think of it as pass-by-reference, but this is more correct.

marchcat (approved):
Thanks! Merging.

Issue #1200, opened 2024-04-11
Avatar rotates 180 degrees when viewed from the top and below using "Front View" mode in the "Camera Controls" floater


dmytrohproductengine:
### Environment

Second Life Release 7.1.6.8632452945 (64bit)
Release Notes

You are at 211.9, 108.9, 21.5 in Product Engine located at simhost-04b8dd6c12cd9dd86.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine/212/109/22
(global coordinates 336,852.0, 307,309.0, 21.5)
Second Life Server 2024-03-18.8333615376
Release Notes

CPU: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz (3696.01 MHz)
Memory: 32688 MB
OS Version: Microsoft Windows 10/11 64-bit (Build 22631.3155)
Graphics Card Vendor: ATI Technologies Inc.
Graphics Card: Radeon RX 5500 XT

Windows Graphics Driver Version: 31.0.12027.9001
OpenGL Version: 4.6.0 Core Profile Context 22.20.27.09.230330

Window size: 1920x1009
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 20167MB
Disk cache: Max size 1638.4 MB (19.2% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: FMOD Studio 2.02.20
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.399bd0e

Packets Lost: 0/6,420 (0.0%)
April 11 2024 04:37:07

### Description

See Reproduction steps.

### Reproduction steps

1. Open the Viewer and log in.
2. Click the "Camera controls" toolbar button and choose the "Front View" option.
3. Left-click and hold the mouse button on the avatar.
4. Drag down/up the mouse until it stops (an avatar is viewed from the top/below).
5. Observe the avatar's screen position.

**Actual result:**
Avatar rotates 180 degrees when viewed from the top and below using "Front View" mode in the "Camera Controls" floater.

**Expected result:**
[...]

PR #1240, opened 2024-04-15
viewer#1200 Avatar rotates 360 degrees when viewed from the top and below

LLGuru:
See https://github.com/secondlife/viewer/issues/1200

This change applies `pitch()` to a copy of the `mFrameAgent` and checks the resulting vertical angle
The new value of `mFrameAgent` is used only if the resulting vertical angle doesn't exceed the limits

This method is more precise than any attempts to guess the resulting angle before applying `pitch()`


indra/newview/llagent.cpp
```diff
@@ -1453,48 +1453,70 @@ LLVector3 LLAgent::getReferenceUpVector()
 	return up_vector;
 }
 
-
+#pragma optimize("", off)
```
marchcat, line 1456: ?

LLGuru: > ?

I felt that I would forget to remove it :-)

marchcat (approved):
Let's test it.

indra/newview/llagentcamera.h
```diff
@@ -112,6 +112,7 @@ class LLAgentCamera
 	//--------------------------------------------------------------------
 public:
 	void switchCameraPreset(ECameraPreset preset);
+	ECameraPreset getCameraPreset() const { return mCameraPreset; }
```
akleshchev, line 115: Why is this needed?

indra/newview/llagent.cpp
```diff
@@ -1453,48 +1453,69 @@ LLVector3 LLAgent::getReferenceUpVector()
...
-		LLVector3 agent_focus_pos = getPosAgentFromGlobal(gAgentCamera.calcFocusPositionTargetGlobal());
-		LLVector3 look_dir = agent_focus_pos - viewer_camera_pos;
-		F32 angle_from_skyward = angle_between(look_dir, skyward);
-		if (angle_from_skyward + angle < look_up_limit)
-		{
-			angle = look_up_limit - angle_from_skyward;
-		}
-	}
+    // don't let user pitch if pointed almost all the way down or up
+    if (gAgentCamera.getCameraMode() == CAMERA_MODE_THIRD_PERSON ||
```
akleshchev, line 1474: Why did yu add this cheack? Won't this negatively affect avatar customization?

marchcat: Good point.

indra/newview/llagentcamera.h
LLGuru: See the second file in this commit

indra/newview/llagent.cpp
LLGuru: Because each preset requires different checks

LLGuru: The avatar customization mode won't be affected because it doesn't use vertical angles close to 1 and 179

indra/newview/llagentcamera.h
akleshchev: And what would happen with custom presets?

indra/newview/llagent.cpp
akleshchev: And CAMERA_MODE_FOLLOW? That's a mode where script overrides control over camera and otherwise acts as a CAMERA_MODE_THIRD_PERSON, doesn't appear to affect mouselook

indra/newview/llagentcamera.h
maxim-productengine: I just want to mention that all 3 "default" presets (Front, Rear, Side) can be overwritten. For example, you can set the values of 'Rear' and 'Side' to be the same as 'Front'. Basically, they are almost the same as 'custom' presets, but have their own buttons and default values, stored in the file.

indra/newview/llagent.cpp
akleshchev: I checked and it behaves exactly like CAMERA_MODE_THIRD_PERSON

indra/newview/llagentcamera.h
LLGuru: > So in this case, it looks like those bugs will appear in the very first round of testing. Why wait, then?

Nothing is changed for custom presets
If we don't have bugs with them right now, why should they appear after this change?

indra/newview/llagent.cpp
LLGuru: With CAMERA_MODE_THIRD_PERSON your camera could rotate any directions without any action from your side
So this bug doesn't make sense for CAMERA_MODE_THIRD_PERSON

When the vertical angle takes critical values 1 or 179, camera could rotate, and nobody treats this like a bug

indra/newview/llagentcamera.h
maxim-productengine: I meant something like that:
- Open 'Camer controls'
- Select 'Front view'
- Click 'Save as preset' button
- Choose 'Replace a preset' and select 'Rear'
- (then repeat and select 'Side')
- Click 'Save as preset' button once again
- And save as new preset.

After that the bug will be reproducible on all camera preset views, including REAR, GROUP, CUSTOM.

akleshchev: Test plan for QA now includes testing presets, so I'm tentatively approving this.

akleshchev (approved):

akleshchev:
But may be retarget it to maint C?

marchcat:
> retarget it to maint C?

Sounds good.

indra/newview/llagentcamera.h
LLGuru: If the user explicitly sets incorrect preset settings, the bugs like this are expected, but they won't be treated as bugs

For example, if user choses the color scheme with white color for font and background, the text becomes unreadable, but this is not a buy, yes? :-)

marchcat (approved):

PR #1247, opened 2024-04-16
Call suspend() periodically to avoid viewer freeze

maxim-productengine:


indra/newview/llfloaterluadebug.cpp
```diff
@@ -123,8 +123,10 @@ void LLFloaterLUADebug::completion(int count, const LLSD& result)
     if (count < 0)
     {
         // error: show error message
-        mResultOutput->insertText("*** ");
-        mResultOutput->pasteTextWithLinebreaks(result.asString());
+        LLStyle::Params params;
+        params.readonly_color = LLUIColorTable::instance().getColor("LtRed");
+        mResultOutput->appendText(result.asString(), false, params);
```
nat-goodspeed, line 128: Will this preserve line breaks in the message?

nat-goodspeed (approved):
I like it!

indra/newview/llfloaterluadebug.cpp
maxim-productengine: Yes, error message is posted on the new line.

PR #1254, opened 2024-04-17
Cleanup Install Linux dependencies

AiraYumi:
Organize the packages to be installed when building the Linux version viewer.


marchcat (approved):
Thank you!

PR #1255, opened 2024-04-17
secondlife/viewer#1072 Crash: LLPersistentNotificationStorage()

LLGuru:


indra/llcommon/llsingleton.h
```diff
@@ -584,6 +584,15 @@ class LLSingleton : public LLSingletonBase
...
 
+    // Hasn't this singleton been started to create yet?
+    // Use this to avoid creating singletons twice.
+    static bool isInstanceUninitialised()
+    {
+        // defend any access to sData from racing threads
+        LockStatic lk;
+        return lk->mInitState == UNINITIALIZED;
+    }
+
```
nat-goodspeed, line 595: I'm not enthused about adding this to `LLSingleton`.

akleshchev:
Suggestion: fix logging then add LL_INFOS() into LLPersistentNotificationStorage() or into reset() (in such case it will need to report if that's an init or actual reset).

Log would show if this is first or second login attempt and we can have that data. If crash happens during first login attempt, contructor will be hit regardles and this fix is pointles. If this is a second login attempt logs would reveal what actually happens and confirm that constructor shouldn't be hit.

indra/newview/llstartup.cpp
```diff
@@ -938,17 +938,28 @@ bool idle_startup()
 		LLFile::mkdir(gDirUtilp->getLindenUserDir());
 
 		// As soon as directories are ready initialize notification storages
-		if (!LLPersistentNotificationStorage::instanceExists())
+		if (LLPersistentNotificationStorage::isInstanceUninitialised())
 		{
 			// check existance since this part of code can be reached
 			// twice due to login failures
 			LLPersistentNotificationStorage::initParamSingleton();
-			LLDoNotDisturbNotificationStorage::initParamSingleton();
```
nat-goodspeed, line 946: I'm sorry, this feels to me like flailing.

Can we zoom out a notch? Why are `LLPersistentNotificationStorage` and `LLDoNotDisturbNotificationStorage` derived from `LLParamSingleton` instead of `LLSingleton`, since neither constructor requires any arguments? Why call `initParamSingleton()` here at all, under various combinations of circumstances? Why not just let any subsequent `instance()` or `getInstance()` call initialize these singleton classes?

nat-goodspeed (changes requested):

indra/newview/llstartup.cpp
akleshchev: Originally it was a static class, then a param singleton to have a strict initialization time, so that people won't attempt to access it too early. It probably can be safely converted to a LLSingleton.

PR #1261, opened 2024-04-17
viewer#1260 Fix thumbnail preview being blury and not loading

akleshchev:
1. Switched 'inspect' to thumbnails to minimalize differences
2. Reporting larger area to bump priority
3. Change scaling behavior. Old mechanics worked fine for icons that were scaled down from large images to ~32, but for thumbnails it can result in 256 image scaling down to ~200 before being scaled up to UI's scale (scale factor), causing extra loss of quality.

This might be not a straight fix (somehow preview image was leaking into thumbnail and back?), but by showing thumbnails as thumbnails and enforcing separation it bypasses the issue (still tracking how that happens).

### DIFF
--- a/indra/newview/llinspecttexture.cpp
+++ b/indra/newview/llinspecttexture.cpp
@@ -112,7 +112,6 @@ public:
 
 protected:
 	LLPointer<LLViewerFetchedTexture> m_Image;
-	S32         mImageBoostLevel = LLGLTexture::BOOST_NONE;
 	std::string mLoadingText;
 };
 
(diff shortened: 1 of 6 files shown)


marchcat (approved):
Awesome, thanks!

I'd even put this closer to the release (Maint YZ may be good candidate).

akleshchev:
Back to drawing board, I missed some cases.

akleshchev:
I think I founf the issue. During gltf workers were remade, before that scaled and unsaled textures were fine sharing workers (no point not to), but now when worker exists for a standard image scaled one things that worker creation failed and vice versa, thus one can't make a request.

maxim-productengine (approved):

akleshchev:
Standard and scaled textures couldn't share workers and if one finished
a reuquest, second one failed to start a new one. Fixed.

marchcat (approved):

maxim-productengine (approved):

PR #1262, opened 2024-04-17
Reintroduce `LLCoros::killreq()` to request killing a named coroutine.

nat-goodspeed:
Make `LLCoros` constructor echo "LLApp" status-change events on new "LLCoros"
event pump.

Rename `LLCoros::kill()` to `killreq()` because this operation only registers a
request for the named coroutine to terminate next time it calls `checkStop()`.
Add a new `CoroData` member to record the name of the coroutine requesting
termination. `killreq()` sets that and also posts "killreq" to "LLCoros".

Add an optional final-cleanup callback to `LLCoros::checkStop()`. Make
`checkStop()` check for a pending `killreq()` request as well as viewer
termination. Introduce new `LLCoros::Killed` exception for that case.

Introduce `LLCoros::getStopListener()`, with two overloads, to encapsulate some
of the messy logic to listen (perhaps temporarily) for viewer shutdown. Both
overloads are for use by code at the source end of a queue or promise or other
resource for which coroutines might still be waiting at viewer shutdown time.
One overload is specifically for when the caller knows the name of the one and
only coroutine that will wait on the resource (e.g. because the caller _is_ that
coroutine). That overload honors `killreq()`.

Use `getStopListener()` to simplify the four existing places where we set up
such a listener. Add a fifth: also make `WorkQueue` listen for viewer shutdown
(resolving a TODO comment).

Remove `LLLUAmanager::terminateScript()`, `getTerminationList()` and the static
`sTerminationList`. In the Lua interrupt callback, instead of checking
`sTerminationList`, call `LLCoros::checkStop()`.

Change `LLFloaterLUAScripts` terminate-script logic to call `LLCoros::killreq()`
instead of posting on "LLLua" and calling `LLLUAmanager::terminateScript()`.

Drop `LLApp::setStatus()` posting to "LLLua" `LLEventPump`: the above makes that
moot.

### DIFF
--- a/indra/llcommon/lualistener.cpp
+++ b/indra/llcommon/lualistener.cpp
@@ -41,7 +41,7 @@ LuaListener::LuaListener(lua_State* L):
     // Listen for shutdown events.
     mShutdownConnection(
         LLCoros::getStopListener(
-            LLEventPump::inventName("LuaState"),
+            "LuaState",
             mCoroName,
             [this](const LLSD&)
             {
--- a/indra/llcommon/lualistener.cpp
+++ b/indra/llcommon/lualistener.cpp
@@ -38,21 +38,16 @@ LuaListener::LuaListener(lua_State* L):
         "LuaListener",
         [this](const std::string& pump, const LLSD& data)
         { return queueEvent(pump, data); })),
-    // Listen for shutdown events on the "LLApp" LLEventPump.
+    // Listen for shutdown events.
     mShutdownConnection(
-        LLEventPumps::instance().obtain("LLLua").listen(
+        LLCoros::getStopListener(
             LLEventPump::inventName("LuaState"),
-            [this](const LLSD& status)
+            mCoroName,
+            [this](const LLSD&)
             {
-                auto coro_name = status["coro"].asString();
-                auto statsd = status["status"].asString();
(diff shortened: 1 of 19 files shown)


indra/llcommon/lualistener.cpp
```diff
@@ -38,21 +38,16 @@ LuaListener::LuaListener(lua_State* L):
         "LuaListener",
         [this](const std::string& pump, const LLSD& data)
         { return queueEvent(pump, data); })),
-    // Listen for shutdown events on the "LLApp" LLEventPump.
+    // Listen for shutdown events.
     mShutdownConnection(
-        LLEventPumps::instance().obtain("LLLua").listen(
+        LLCoros::getStopListener(
             LLEventPump::inventName("LuaState"),
```
maxim-productengine, line 44: `inventName(..)` will be called in `LLCoros::getStopListener` anyway.

maxim-productengine (approved):

nat-goodspeed:
Maxim, please see [this new test](https://github.com/secondlife/viewer/pull/1262/files#diff-3235afe16a771071f34813e4f0a1749fecb4673118520e0ceaea4d1750aa0ae6R449-R467).

maxim-productengine (approved):
Looks good!

PR #1264, opened 2024-04-18
Fix for SL-20652 shadows on mac breaking alpha blended objects.

brad-linden:
fix secondlife/jira-archive-internal#71030


brad-linden:
so I think I understand more why it was "working" with the reserved texture units set to a lower number like 9.  It only seems to trigger an actual GL error INVALID_OPERATION if there are multiple usages of a texture unit by samplers of a different type.  If you are "lucky" and they are of the same type, then you can get away with it and it will seem to work but breakage will probably show up elsewhere.

RunitaiLinden (approved):

PR #1265, opened 2024-04-18
Use PipeWire for Linux volume catcher

makidoll:
I've been trying to fix the Linux volume catcher for Firestorm and figured out how terrible Pulseaudio is. I decided to rewrite the volume catcher using the PipeWire API and it works great! The only downside is that users who are still using Pulseaudio won't be able to control their web media volume, however the current implementation doesn't seem to be working anyway.

Chromium still uses Pulseaudio, however I'm certain they will implement PipeWire eventually. The general consensus is that all Linux users should switch to PipeWire as soon as possible anyway. It seems that PipeWire implements the Pulseaudio server better than the official Pulseaudio server too which is kinda funny.

Let me know if you'd like any changes!
And thank you for developing Second Life, and making it open source <3

### DIFF
--- a/indra/media_plugins/cef/linux_volume_catcher_pw.cpp
+++ b/indra/media_plugins/cef/linux_volume_catcher_pw.cpp
@@ -1,0 +1,451 @@
+/** 
+ * @file linux_volume_catcher_pw.cpp
+ * @brief A Linux-specific, PipeWire-specific hack to detect and volume-adjust new audio sources
+ *
+ * @cond
+ * $LicenseInfo:firstyear=2010&license=viewerlgpl$
+ * Second Life Viewer Source Code
+ * Copyright (C) 2010, Linden Research, Inc.
+ * 
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation;
+ * version 2.1 of the License only.
+ * 
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ * 
+ * Linden Research, Inc., 945 Battery Street, San Francisco, CA  94111  USA
+ * $/LicenseInfo$
+ * @endcond
+ */
+
+/*
+  The high-level design is as follows:
+  1) Connect to the PipeWire daemon
+  2) Find all existing and new audio nodes
+  3) Examine PID and parent PID's to see if it belongs to our process
+  4) If so, tell PipeWire to adjust the volume of that node
+  5) Keep a list of all audio nodes and adjust when we setVolume()
+ */
+
+#include "linden_common.h"
+
+#include "volume_catcher.h"
+
+#include <unordered_set>
+#include <mutex>
+
+extern "C" {
+#include <pipewire/pipewire.h>
+#include <spa/pod/builder.h>
+#include <spa/param/props.h>
+
+#include "apr_pools.h"
+#include "apr_dso.h"
+}
+
+#include "media_plugin_base.h"
+
+SymbolGrabber gSymbolGrabber;
+
+#include "linux_volume_catcher_pw_syms.inc"
+
+////////////////////////////////////////////////////
+
+class VolumeCatcherImpl
+{
+public:
+	VolumeCatcherImpl();
+	~VolumeCatcherImpl();
+
+	bool loadsyms(std::string pw_dso_name);
+	void init();
+	void cleanup();
+
+	void pwLock();
+	void pwUnlock();
+
+	void setVolume(F32 volume);
+	// void setPan(F32 pan);
+
+	void handleRegistryEventGlobal(
+		uint32_t id, uint32_t permissions, const char* type,
+		uint32_t version, const struct spa_dict* props
+	);
+
+	class ChildNode
+	{
+	public:
+		bool mActive = false;
+
+		pw_proxy* mProxy = nullptr;
+		spa_hook mNodeListener {};
+		spa_hook mProxyListener {};
+		VolumeCatcherImpl* mImpl = nullptr;
+
+		void updateVolume();
+		void destroy();
+	};
+
+	bool mGotSyms = false;
+
+	F32 mVolume = 1.0f; // max by default
+	// F32 mPan = 0.0f; // center
+
+	pw_thread_loop* mThreadLoop;
+	pw_context* mContext;
+	pw_core* mCore;
+	pw_registry* mRegistry;
+	spa_hook mRegistryListener;
+
+	std::unordered_set<ChildNode*> mChildNodes;
+	std::mutex mChildNodesMutex;
+};
+
+VolumeCatcherImpl::VolumeCatcherImpl()
+{
+	init();
+}
+
+VolumeCatcherImpl::~VolumeCatcherImpl()
+{
+	cleanup();
+}
+
+// static void debugClear()
+// {
+// 	auto file = fopen("/home/maki/git/firestorm-viewer/log.txt", "w");
+// 	fprintf(file, "\n");
+// 	fclose(file);
+// }
+
+// static void debugPrint(const char* format, ...)
+// {
+//     va_list args;
+//     va_start(args, format);
+// 	auto file = fopen("/home/maki/git/firestorm-viewer/log.txt", "a");
+//     vfprintf(file, format, args);
+// 	fclose(file);
+//     va_end(args);
+// }
+
+static void registryEventGlobal(
+	void *data, uint32_t id, uint32_t permissions, const char *type,
+	uint32_t version, const struct spa_dict *props
+)
+{
+	static_cast<VolumeCatcherImpl*>(data)->handleRegistryEventGlobal(
+		id, permissions, type, version, props
+	);
+}
+
+static const struct pw_registry_events REGISTRY_EVENTS = {
+	.version = PW_VERSION_REGISTRY_EVENTS,
+    .global = registryEventGlobal,
+};
+
+bool VolumeCatcherImpl::loadsyms(std::string pw_dso_name)
+{
+	return gSymbolGrabber.grabSymbols({ pw_dso_name });
+}
+
+void VolumeCatcherImpl::init()
+{
+	// debugClear();
+	// debugPrint("init\n");
+	
+	mGotSyms = loadsyms("libpipewire-0.3.so.0");
+	if (!mGotSyms) return;
+
+	llpw_init(NULL, NULL);
+
+	mThreadLoop = llpw_thread_loop_new("SL Plugin Volume Adjuster", NULL);
+	if (!mThreadLoop) return;
+
+	pwLock();
+
+	mContext = llpw_context_new(
+		llpw_thread_loop_get_loop(mThreadLoop), NULL, 0
+	);
+	if (!mContext) return pwUnlock();
+
+	mCore = llpw_context_connect(mContext, NULL, 0);
+	if (!mCore) return pwUnlock();
+
+	mRegistry = pw_core_get_registry(mCore, PW_VERSION_REGISTRY, 0);
+
+	// debugPrint("got registry\n");
+
+	spa_zero(mRegistryListener);
+
+	pw_registry_add_listener(
+		mRegistry, &mRegistryListener, &REGISTRY_EVENTS, this
+	);
+
+	llpw_thread_loop_start(mThreadLoop);
+
+	pwUnlock();
+
+	// debugPrint("started thread loop\n");
+}
+
+void VolumeCatcherImpl::cleanup()
+{
+	mChildNodesMutex.lock();
+	for (auto* childNode : mChildNodes) {
+		childNode->destroy();
+	}
+	mChildNodes.clear();
+	mChildNodesMutex.unlock();
+
+	pwLock();
+	if (mRegistry) llpw_proxy_destroy((struct pw_proxy*)mRegistry);
+	spa_zero(mRegistryListener);
+	if (mCore) llpw_core_disconnect(mCore);
+	if (mContext) llpw_context_destroy(mContext);
+	pwUnlock();
+
+	if (!mThreadLoop) return;
+
+	llpw_thread_loop_stop(mThreadLoop);
+	llpw_thread_loop_destroy(mThreadLoop);
+
+	// debugPrint("cleanup done\n");
+}
+
+void VolumeCatcherImpl::pwLock() {
+	if (!mThreadLoop) return;
+	llpw_thread_loop_lock(mThreadLoop);
+}
+
+void VolumeCatcherImpl::pwUnlock() {
+	if (!mThreadLoop) return;
+	llpw_thread_loop_unlock(mThreadLoop);
+}
+
+// #define INV_LERP(a, b, v) (v - a) / (b - a)
+
+// #include <sys/time.h>
+
+void VolumeCatcherImpl::ChildNode::updateVolume()
+{
+	if (!mActive) return;
+
+	F32 volume = std::clamp(mImpl->mVolume, 0.0f, 1.0f);
+	// F32 pan = std::clamp(mImpl->mPan, -1.0f, 1.0f);
+
+	// debugClear();
+	// struct timeval time;
+	// gettimeofday(&time, NULL);
+	// double t = (double)time.tv_sec + (double)(time.tv_usec / 1000) / 1000;
+	// debugPrint("time is %f\n", t);
+	// F32 pan = std::sin(t * 2.0d);
+	// debugPrint("pan is %f\n", pan);
+
+	// uint32_t channels = 2;
+	// float volumes[channels];
+	// volumes[1] = INV_LERP(-1.0f, 0.0f, pan) * volume; // left
+	// volumes[0] = INV_LERP(1.0f, 0.0f, pan) * volume; // right
+
+	uint32_t channels = 1;
+	float volumes[channels];
+	volumes[0] = volume;
+
+	uint8_t buffer[512];
+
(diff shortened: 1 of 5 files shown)


makidoll:

makidoll:
recheck

AiraYumi:
You probably need to add the libpipewire-0.3-dev installation to your build.yaml.

Nicky-D:
@makidoll the current version, even maybe not ideal, does work. I had tested it before submission.

The one from Firestorm did not, because it was modified and they had decided to remove g_main_context_iteration from pump

Nicky-D:
Hmm, I could not find it so fast. But I believe (might be wrong of course). That webrtc does use Puleaudio too. So I'm not sure how close we're too pipewire and if it makes sense to use it.

makidoll:
Glib mainloop doesn't seem to work for me and using mainloop iterate causes problems that crashes the plugin. It's strange. I could try to take a closer look as to why the Pulse volume catcher is broken. We don't have to use the PipeWire API, but it does seem to work a lot better.

makidoll:
Alright the issue is literally just `libpulse-mainloop-glib.so.0` not getting loaded. I'm on Arch Linux using WirePlumber and the library is definitely in my `/usr/lib` folder. If I try to load `libpulse.so.0` instead, everything works fine. I also added a CMake variable `USE_VOLUME_CATCHER_PW` set to false by default.

yaynstuff:
> Hmm, I could not find it so fast. But I believe (might be wrong of course). That webrtc does use Puleaudio too. So I'm not sure how close we're too pipewire and if it makes sense to use it.

It uses pipewire if its enabled in the gn settings when building webrtc - but Henri argued during an exchange of emails  the case to to completely disable pulseaudio and pipewire in LL's webrtc build for Linux completely, which I believe will make it fall back to ALSA. His argument was compatibility with systems that don't have pulse or pipewire.....

yaynstuff:
> @makidoll the current version, even maybe not ideal, does work. I had tested it before submission.
> 
> The one from Firestorm did not, because it was modified and they had decided to remove g_main_context_iteration from pump

I looked back into this - you did a CMake cleanup a while back, about a year ago, maybe a bit more and it reintroduced a whole bunch of old stuff from LL when Ansa merged it in - I note this in discord from a while back:

`long story short, the cef media plugin for Linux is crashing due to glib assertions being thrown and Linux users are greeted with no splash page, just a big black space and no embedded browser and support for any media that would use that plugin`

```
-#include <pulse/glib-mainloop.h> // There's no special reason why we want the *glib* PA mainloop, but the generic polling implementation seems broken.
+#include <pulse/mainloop.h>
```
From memory the glib mainloop was broken and causing issues - its a long time ago now so I dont remember all the specifics. Zi reworked it and I think thats when it all changed.

makidoll:
This is my first contribution to sl or firestorm. Just seeing a black screen is indeed an issue. Pulse has never been great and PipeWire is the definitive audio system for Linux. I'm certain software will prefer PW over Pulse over ALSA. PipeWire doesn't crash at all and in my experience does Pulseaudio better than the official Pulseaudio software. I personally don't like the glib mainloop but it crashes without it. I could look into why precisely. I really do think it's a good idea that we prefer one implementation over the other as that's what most Linux software does.

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+// static void debugPrint(const char* format, ...)
+// {
+//     va_list args;
+//     va_start(args, format);
+// 	auto file = fopen("/home/maki/git/firestorm-viewer/log.txt", "a");
+//     vfprintf(file, format, args);
+// 	fclose(file);
+//     va_end(args);
+// }
```
Nicky-D, line 137: This should probably go or be reworked. As right now those paths will be not usable to anyone.

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+bool VolumeCatcherImpl::loadsyms(std::string pw_dso_name)
+{
+	return gSymbolGrabber.grabSymbols({ pw_dso_name });
+}
+
+void VolumeCatcherImpl::init()
+{
+	// debugClear();
+	// debugPrint("init\n");
```
Nicky-D, line 162: I am aware this can aid in future debugging. On the other hand I am not a fan of having a lot of commented out code commited (there is more of that later)

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+	
+	mGotSyms = loadsyms("libpipewire-0.3.so.0");
+	if (!mGotSyms) return;
+
+	llpw_init(NULL, NULL);
+
+	mThreadLoop = llpw_thread_loop_new("SL Plugin Volume Adjuster", NULL);
+	if (!mThreadLoop) return;
```
Nicky-D, line 170: I'm not sure about the current LL styleguide, but afaik those returns should be
```
   if( ... )
      return;
``

Best if some Linden can chime in on that if they prefer the other syntax or don't care. (@nat-goodspeed ?)

File: indra/media_plugins/cef/linux_volume_catcher_pw.cpp [hunk 4]
```diff
@@ -0,0 +1,451 @@
...
+	spa_pod_builder builder;
+	spa_pod_builder_init(&builder, buffer, sizeof(buffer));
+
+	spa_pod_frame frame;
+	spa_pod_builder_push_object(&builder, &frame, SPA_TYPE_OBJECT_Props, SPA_PARAM_Props);
+	spa_pod_builder_prop(&builder, SPA_PROP_channelVolumes, 0);
+	spa_pod_builder_array(&builder, sizeof(float), SPA_TYPE_Float, channels, volumes);
+	spa_pod* pod = static_cast<spa_pod*>(spa_pod_builder_pop(&builder, &frame));
+
+	mImpl->pwLock();
```
Nicky-D, line 272: It would be better to wrap those locks into some sort of RAII, just in case something rogue throws

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+	mActive = false;
+
+	mImpl->mChildNodesMutex.lock();
+	mImpl->mChildNodes.erase(this);
+	mImpl->mChildNodesMutex.unlock();
+	
+	spa_hook_remove(&mNodeListener);
+	spa_hook_remove(&mProxyListener);
+
+	mImpl->pwLock();
```
Nicky-D, line 289: see above about RAII

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+	do
+	{
+		if(aPid == myPid) return true;
+		aPid = getParentPid( aPid );
+	} while(aPid > 1);
+
+	return false;
+}
+
```
Nicky-D, line 333: getParentPid and isPluginPid are function introduced in the pa volume catcher and duplicated here. The should be refactored into a common base (probably PluginMediaBase)

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+	// const char* appName = spa_dict_lookup(info->props, PW_KEY_APP_NAME);
+	// debugPrint("got app: %s\n", appName);
+	
+	auto* const childNode = static_cast<VolumeCatcherImpl::ChildNode*>(data);
+	// debugPrint("init volume to: %f\n", childNode->mImpl->mDesiredVolume);
+
+	childNode->updateVolume();
+
+	childNode->mImpl->mChildNodesMutex.lock();
```
Nicky-D, line 350: see above about RAII

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
```diff
@@ -0,0 +1,451 @@
...
+	llpw_proxy_add_listener(proxy, &childNode->mProxyListener, &PROXY_EVENTS, childNode);
+}
+
+void VolumeCatcherImpl::setVolume(F32 volume)
+{
+	// debugPrint("setting all volume to: %f\n", volume);
+
+	mVolume = volume;
+
+	mChildNodesMutex.lock();
```
Nicky-D, line 407: see above about RAII

Nicky-D:
I suppose it will be hard/impossible to find a consensus among user and distributions.
While I personally don't care and have no preference for either, I'm not fond of having some packages depend on solution A and other packages on solution B.

I guess it would be possible to compile the volume catcher against pulse *and* pipewire, then probe the system at runtime and use what's installed. Letting user ovveride the default order via the environment.
Though that is possibly quite a bit of extra work

marchcat (approved):
Wow, this is huge!

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
nat-goodspeed: Please, yeah. When skimming through code, it's too easy to miss the controlled statement of an `if` when it's on the same line.

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
nat-goodspeed: Definitely use RAII for locking and unlocking.

indra/media_plugins/cef/linux_volume_catcher_pw.cpp
akleshchev: Perhaps do something like this?
```
        std::string log_file = /home/maki/git/firestorm-viewer/log.txt
        LL_DEBUGS("VolumeCatched") << "Printing state to file: " << log_file;
	    debugClear(log_file );
	    debugPrint(log_file , "init\n");
        LL_CONT << LL_ENDL;
```

bennettgoble:
We may be able to accept this after revisions so that those looking to build the viewer from source with native Pipewire support. For compatibility reasons, leaving Pulseaudio as default is probably the right choice. Like a couple others here, I don't have a strong opinion: I run Pipewire, but always install pulse and alsa compatibility shims.

bennettgoble:
...and: welcome and thank you for your submission, @makidoll. It's wonderful seeing first time contributors! 😊

makidoll:
> We may be able to accept this after revisions so that those looking to build the viewer from source with native Pipewire support. For compatibility reasons, leaving Pulseaudio as default is probably the right choice. Like a couple others here, I don't have a strong opinion: I run Pipewire, but always install pulse and alsa compatibility shims.

PipeWire is designed to provide backwards compatibility out of the box, so you usually have to uninstall Pulse.

I was thinking of writing code to try to load PipeWire and if it fails, to switch. Or would you rather we keep it disabled for now. I'm worried my friends who don't compile won't get to benefit because the Pulse volume catcher has been very unreliable for them.

AiraYumi:
pipewire is not the default audio until 22.10.
Therefore, users who have installed 22.04 may not have pipewire installed.

Therefore, if you want to use pipewire by default,
You need to copy pipewire to the viewer's installation folder so that users who do not have pipewire installed can use pulseaudio without having to install the pipewire library.

makidoll:
image

I've added an option to enable PipeWire in runtime. It's disabled by default.
Just press "Stop All" on web media after enabling it to reset the plugin.

indra/media_plugins/cef/linux/volume_catcher_linux.cpp
```diff
@@ -0,0 +1,103 @@
...
+		if (PIPEWIRE != nullptr) {
+			delete PIPEWIRE;
+			PIPEWIRE = nullptr;
+		}
+
+		if (PULSEAUDIO == nullptr) {
+			// debugPrint("volume catcher using pulseaudio\n");
+			PULSEAUDIO = new VolumeCatcherPulseAudio();
+		}
+	}
```
Nicky-D, line 69: I don't think you need all of this (and a second pimpl2 pointer.
Just set pimpl to either implementation and it will be used accordingly.

This will also safe you from all the boilerplate of pointer checking.

indra/media_plugins/cef/linux/volume_catcher_linux.h
```diff
@@ -0,0 +1,151 @@
...
+
+#include <pipewire/pipewire.h>
+
+#include "apr_pools.h"
+#include "apr_dso.h"
+}
+
+#include "media_plugin_base.h"
+
+class VolumeCatcherPulseAudio : public virtual VolumeCatcherImpl
```
Nicky-D, line 52: Is virtual inheritance really needed here? I could not see the diamond hierarchy coming up and needing `virtual`

indra/media_plugins/cef/linux/volume_catcher_pipewire.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include "volume_catcher_linux.h"
+
+extern "C" {
+#include <spa/pod/builder.h>
+#include <spa/param/props.h>
+}
+
+SymbolGrabber pwSymbolGrabber;
+#undef LL_SYMBOL_GRABBER
+#define LL_SYMBOL_GRABBER pwSymbolGrabber
```
Nicky-D, line 49: Please don't try to use this with macro magic if it can be avoided. The symbol grabber sadly is already bnad enough.

It should be possible to pass the desired grabber into LL_GRAB_SYMBOL.
Or maybe it is an option to rework the symbol grabber in such a way, only one instance is needed rather than two.
Mostly greabSymbols would need some love here, to not return false for the symbols of teh opposite volume catcher
I think with some rework the symbol grabber can be made to load symbols for either implementation.

It should be possible to pass the desired

indra/media_plugins/cef/linux/volume_catcher_pipewire.cpp
```diff
@@ -0,0 +1,336 @@
...
+	for (auto* childNode : mChildNodes) {
+		childNode->destroy();
+	}
+	mChildNodes.clear();
+	mChildNodesMutex.unlock();
+
+	pwLock();
+	if (mRegistry) llpw_proxy_destroy((struct pw_proxy*)mRegistry);
+	spa_zero(mRegistryListener);
+	if (mCore) llpw_core_disconnect(mCore);
```
Nicky-D, line 146: Here's a few more if that need to be splint into two lines

makidoll:
I've done everything I can for now. I just need some recommendations on the virtual inheritance issue and perhaps figuring out how to log debug messages from the volume catcher if it's necessary. I hope my RAII mutex lock changes and symbol grabber macro magic changes are good.

Nicky-D:
Thank you for those updated @makidoll really appreciated.

Regarding the virtual inheritance. I think you can just drop them. For the DEBUG logging you can use the standard macros. My suggestion would be that I clone you repository and make a PR suggestion against it, so it's not all on your shoulder and you get some input.

How do you feel about that?

makidoll:
Oh, thank you! That's fine by me. I realized outputting to stderr might be more reliable than stdout, I'm not sure.

makidoll:
I couldn't figure out how to set an application name without creating an audio stream or such.

Otherwise, here's a command that shows if the PulseAudio implementation is being used:
`pw-dump | jq '.[] | select(.info.props."application.name"=="SL Plugin Volume Adjuster") | length'`

You can test it without logging in. Just make sure to restart the viewer when toggling, which isn't necessary when switching implementations. Makes me wonder if the PulseAudio implementation is being cleaned up correctly...

AiraYumi:
It seems that it can be compiled with gcc (gcc13) at the moment.

akleshchev:
Build failures:
1. End of line. Unlear which file is shte cause, potentiall either media_plugin_base.cpp or media_plugin_base.h

2. Linux build: volume_catcher_linux.cpp:58:9: error: delete called on 'VolumeCatcherImpl' that is abstract but has non-virtual destructor [-Werror,-Wdelete-abstract-non-virtual-dtor]
        delete pimpl;
        ^

marchcat:
> which file is the cause

indra/media_plugins/base/media_plugin_base.cpp
indra/media_plugins/base/media_plugin_base.h

Nicky-D:
I created a PR against makidolls repo a few days ago, to address the issues. But have not heard back so far. 
Otherwise I can create a new PR with the original changes and build fix if desired?

marchcat:
We can wait a bit longer, Maint B is relatively far from being released.

makidoll:
Hello! I've rebased the pr you sent. I didn't get a notification about it for some strange reason, really didn't meant to wait 9 days. It should be possible for you to directly commit to my branch

akleshchev:
No newline at end of file:
indra/media_plugins/base/media_plugin_base.cpp
indra/media_plugins/base/media_plugin_base.h

Nicky-D:
@makidoll in fact I tried directly pushing, but it did not work. Maybe this would only work for Lindens. Not a bit problem though.

The current error is about missing libsndio, the dependency on libsndio was dropped from OpenAL and SDL a while ago, does pulseaudio need it?

makidoll:
@Nicky-D Pulse doesn't need sndio, it's just for openal. Strange because `libsndio-dev` is being installed in "Install Linux dependencies"

makidoll:
I updated the branch to latest master. Making a build should work now. I ran merge with --no-ff. Not sure if i should fast forward.

AiraYumi:
Can someone please address the windows build error?

I have sent a pull-request to fix the pre-commit error.

akleshchev:
Redefinition of VolumeCatcherImpl, `class VolumeCatcherImpl : public LLSingleton<VolumeCatcherImpl>` is declared in windows_volume_catcher.cpp and is conflicting with your new definition in volume_catcher.h

makidoll:
Sorry for the delay. Was able to rebase the prs. I recently found a bug with this system which may be concerning.

The volume for Chromium gets remembered system wide. I often turn web media volume off and when I launch system installed Chromium, the volume is remembered from what SL set it to. This doesn't apply to Google Chrome or any other forks. I wonder if PipeWire lets you set volume ephemerally.

makidoll:
Figured `SPA_PROP_channelVolumes` is memorized system-wide and `SPA_PROP_softVolumes` is temporary. Made changes to reset Chromium volume back to 100% and apply SL's volume as temporary.

akleshchev (approved):

marchcat (approved):

PR #1269, opened 2024-04-18
Update open_libndofdev and openal to latest versions.

Nicky-D:
- With the updated openal the dependency on sndio can be dropped
- The updated libndofdev is compiled against the correct version of SDL2 and thus NDOF can be enabled again


marchcat (approved):
Thanks!

PR #1273, opened 2024-04-18
#1139 Fix for world going black/white in some scenes (NaNs)

RunitaiLinden:
NaNs were coming from normal encode/decode.  Take advantage of available gbuffer space to send full normal instead of encoding.


Geenz (approved):
This looks fine for the time being given we're not super hurting for G-Buffer space _yet_.  Just keep an eye on it if we start adding clearcoat support which has a second normal for the clearcoat BRDF and we should be good.  That or switch to Forward+ or something later.

PR #1279, opened 2024-04-18
secondlife/viewer#1249 LLFloater class refactoring

LLGuru:
Main points:
- Add LLFloater::mClosing with class LLScopedBooleanToggler and use it in LLFloater::closeFloater()
- Refactoring in LLFloater: use more compact syntax in for and if operators
- Remove extra spaces in LLFloater

### DIFF
--- a/indra/llui/llemojihelper.cpp
+++ b/indra/llui/llemojihelper.cpp
@@ -155,9 +155,12 @@ void LLEmojiHelper::setHostCtrl(LLUICtrl* hostctrl_p)
 		mHostHandle.markDead();
 		mEmojiCommitCb = {};
 
-		if (!mHelperHandle.isDead())
+		if (LLFloater* floater = mHelperHandle.get())
 		{
-			mHelperHandle.get()->closeFloater();
+			if (floater->getVisible() && !floater->isClosing())
+			{
+				mHelperHandle.get()->closeFloater();
+			}
 		}
 
 		if (hostctrl_p)
--- a/indra/llui/llfloater.cpp
+++ b/indra/llui/llfloater.cpp
@@ -244,7 +244,6 @@ static LLWidgetNameRegistry::StaticRegistrar sRegisterFloaterParams(&typeid(LLFl
 
 LLFloater::LLFloater(const LLSD& key, const LLFloater::Params& p)
 :	LLPanel(),	// intentionally do not pass params here, see initFromParams
- 	mDragHandle(NULL),
 	mTitle(p.title),
 	mShortTitle(p.short_title),
 	mSingleInstance(p.single_instance),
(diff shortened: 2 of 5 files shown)


indra/llui/llfloater.cpp
```diff
@@ -602,31 +586,27 @@ LLControlGroup*	LLFloater::getControlGroup()
...
 		{
 			gFocusMgr.setMouseCapture(NULL);
 		}
 	}
 
-	for(handle_set_iter_t dependent_it = mDependents.begin();
-		dependent_it != mDependents.end(); )
+	for (const LLHandle<LLFloater>& handle : mDependents)
 	{
-		LLFloater* floaterp = dependent_it->get();
```
akleshchev, line 623: Personaly I would have preferred old variant, it is cleaner (I suggest checking if this matches out coding policy).

indra/llui/llfloater.cpp
```diff
@@ -2996,16 +2951,13 @@ void LLFloaterView::draw()
 	refresh();
 
 	// hide focused floater if in cycle mode, so that it can be drawn on top
-	LLFloater* focused_floater = getFocusedFloater();
-
-	if (mFocusCycleMode && focused_floater)
+	if (LLFloater* focused_floater = mFocusCycleMode ? getFocusedFloater() : nullptr)
```
akleshchev, line 2954: This looks ugly... a ternary inside if?

akleshchev:
This is 99% cosmetics, I don't see anything worthy of adding into 'hot pottatoe' of a maint-X
Also tickets milestone is maintenance B, I suggest retargeting.

> LLFloater::mClosing
What does this fixes? If previous fix is crashing or something like that, I suggest reverting it and moving everything into Maint-B which is the intended target for this work.

indra/llui/llfloater.cpp
marchcat: Yes, this should work fine and is compact; however, we need to check our coding conventions.

marchcat:
I agree. Let's wait for the Maint X release and then rebase this to a later one.

indra/llui/llemojihelper.cpp
```diff
@@ -155,9 +155,12 @@ void LLEmojiHelper::setHostCtrl(LLUICtrl* hostctrl_p)
 		mHostHandle.markDead();
 		mEmojiCommitCb = {};
 
-		if (!mHelperHandle.isDead())
+		if (LLFloater* floater = mHelperHandle.get())
 		{
-			mHelperHandle.get()->closeFloater();
```
akleshchev, line 160: Implementing isClosing for this feels excessive. What does this solves?

Can't you do something like this?

```
LLFloater* helper = mHelperHandle.get();
mHelperHandle.markDead()
helper->closeFloater();
```

indra/llui/llfloater.cpp
LLGuru: As for me, it's readable enough

But the more important thing is that `getFocusedFloater()` isn't beong called when `mFocusCycleMode` is false

akleshchev: Ternary is harder and more confusing to read by definition, not debugable. Good rule of thumb is to use it for short checks only and to never nest it. You are nesting it inside another if.

> getFocusedFloater() isn't being called when mFocusCycleMode is false
Since getFocusedFloater is const, it changes nothing by definition, compiler should be smart enough to optimize and not call getFocusedFloater() before mFocusCycleMode. And if you wanted to be explicit about that, you could have done that without nesting a ternary.

PR #1297, opened 2024-04-22
Add demo script with idle and notification interactions

maxim-productengine:


nat-goodspeed:
Maxim, I'm not ignoring this PR, I'm playing with some timing related changes myself. I'm not that comfortable with `Flioater:idle()` -- I don't think that timing functionality should be tied to floaters.

Big Picture, across all viewer projects, I'd like to centralize all C++ time-related actions around `LLEventTimer`.

You've also pointed out to me that we should have a way to sleep just the calling Lua coroutine (vs. the whole script), and I plan to implement that too.

nat-goodspeed (changes requested):
My review is [this PR](https://github.com/secondlife/viewer/pull/1811) against this branch.

nat-goodspeed:
I'll let this build complete, then if it's successful I'll approve and merge.

nat-goodspeed (approved):

PR #1317, opened 2024-04-24
Fix BOOL vs bool issues after merge

Ansariel:
...and an additional broken test
@marchcat


marchcat (approved):
Uh, I thought I had fixed them all. Thank you very much, @Ansariel!

BTW, maybe we should merge #861 as well?

Issue #1331, opened 2024-04-25
The animation with null id is attempted to fetch


LLGuru:
### Environment

Second Life Release 7.1.6.8745209917 (64bit)

### Description

Looking into the log after visiting locations with many people we see multiple lines of kind:
> LLKeyframeMotion::onInitialize : Attempted to fetch animation  with null id

### Reproduction steps

- Start the viewer
- Visit any location with many people
- Close the viewer
- Analyze the log

Actual behavior:
There are multiple lines in the log related to an animation with null id

Expected behavior:
There are no lines in the log related to an animation with null id

dmytrohproductengine:
Passed QA.
Verified on the Second Life Release 7.1.7.8993168213 (64bit) on Win10/OSX in the scope of [secondlife/iqa#226](https://github.com/secondlife/iqa/issues/226).

PR #1336, opened 2024-04-25
secondlife/viewer#1331 The animation with null id is attempted to fetch

LLGuru:
For https://github.com/secondlife/viewer/issues/1331

Many motions in many avatars refer "Closed_Mouth" but it isn't known in the viewer library and can't be started
We may silently ignore this emote_name to prevent spamming the log but warn for other unknown motion names

### DIFF
--- a/indra/llcharacter/llkeyframemotion.cpp
+++ b/indra/llcharacter/llkeyframemotion.cpp
@@ -509,7 +509,8 @@ LLMotion::LLMotionInitStatus LLKeyframeMotion::onInitialize(LLCharacter *charact
         }
         else
         {
-            LL_INFOS("Animation") << "Attempted to fetch animation " << mName << " with null id for character " << mCharacter->getID() << LL_ENDL;
+            LL_INFOS("Animation") << "Attempted to fetch animation '" << mName << "' with null id"
+                << " for character " << mCharacter->getID() << LL_ENDL;
         }
 
 		return STATUS_HOLD;
@@ -661,14 +662,13 @@ bool LLKeyframeMotion::setupPose()
 //-----------------------------------------------------------------------------
 bool LLKeyframeMotion::onActivate()
 {
-	// If the keyframe anim has an associated emote, trigger it. 
-	if( mJointMotionList->mEmoteName.length() > 0 )
+	// If the keyframe anim has an associated emote, trigger it.
+	if (mJointMotionList->mEmoteID.notNull())
 	{
-		LLUUID emote_anim_id = gAnimLibrary.stringToAnimState(mJointMotionList->mEmoteName);
 		// don't start emote if already active to avoid recursion
-		if (!mCharacter->isMotionActive(emote_anim_id))
+		if (!mCharacter->isMotionActive(mJointMotionList->mEmoteID))
 		{
-			mCharacter->startMotion( emote_anim_id );
+			mCharacter->startMotion(mJointMotionList->mEmoteID);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


indra/newview/llvoavatar.cpp
```diff
@@ -2137,7 +2137,7 @@ void LLVOAvatar::resetSkeleton(bool reset_animations)
     LL_DEBUGS("Avatar") << avString() << " reset starts" << LL_ENDL;
     if (!isControlAvatar() && !mLastProcessedAppearance)
     {
-        LL_WARNS() << "Can't reset avatar; no appearance message has been received yet." << LL_ENDL;
+        LL_WARNS() << "Can't reset avatar '" << getFullname() << "'; no appearance message has been received yet." << LL_ENDL;
```
akleshchev, line 2140: Plese don't use full name here, use agent id. Reasons:
1. Controls avatars have no proper name, they are objects, name is not unique and of little value.
2. Name can change or be requested by uuid
3. Lets not show avatar name in logs an extra time

indra/llcharacter/llkeyframemotion.cpp
```diff
@@ -1440,14 +1466,14 @@ bool LLKeyframeMotion::deserialize(LLDataPacker& dp, const LLUUID& asset_id, boo
 		if (!dp.unpackString(joint_name, "joint_name"))
 		{
 			LL_WARNS() << "can't read joint name"
-                       << " for animation " << asset_id << LL_ENDL;
+                       << " for animation " << asset() << LL_ENDL;
```
akleshchev, line 1469: Suggestion: when editing infos() and warns(), please add tags

indra/newview/llvoavatar.cpp
```diff
@@ -11292,7 +11293,10 @@ void LLVOAvatar::setOverallAppearanceNormal()
 		return;
 
     LLVector3 pelvis_pos = getJoint("mPelvis")->getPosition();
-	resetSkeleton(false);
+    if (isControlAvatar() || mLastProcessedAppearance)
```
akleshchev, line 11296: This check makes no sense. control avatars won't get here.

indra/llcharacter/llkeyframemotion.cpp
```diff
@@ -1308,16 +1316,34 @@ bool LLKeyframeMotion::deserialize(LLDataPacker& dp, const LLUUID& asset_id, boo
...
+			return false;
+		}
+		// "Closed_Mouth" is a very popular emote name we should ignore
+		if (joint_motion_list->mEmoteName == "Closed_Mouth")
+		{
+			joint_motion_list->mEmoteName.clear();
+		}
+		else
+		{
+			joint_motion_list->mEmoteID = gAnimLibrary.stringToAnimState(joint_motion_list->mEmoteName);
```
akleshchev, line 1339: Motion uploader probably doesn't have ids and this might break it

indra/newview/llvoavatar.cpp
LLGuru: This check makes sense just because without it we receive many stupid warnings from resetSceleton()

akleshchev: There is a isControlAvatar() check at string 11292, checking it here makes no sense.

LLGuru: I just try to preserve the parity in the logics

LLControlAvatar is inherited of LLVOAvatar,
and technically it is possible to call setOverallAppearanceNormal() from this class

It is unsafe practice to make some assumptions of impossibility some conditions
We'd better leave this check here

akleshchev (approved):

marchcat (approved):

maxim-productengine (approved):

PR #1337, opened 2024-04-26
Cached minimap textures still using extra memory/loading still not reliable

cosmic-linden:


marchcat (approved):

indra/llimage/llimage.cpp
```diff
@@ -1874,6 +1874,75 @@ void LLImageRaw::compositeRowScaled4onto3( U8* in, U8* out, S32 in_pixel_len, S3
...
+            dst_pixel[0] = llmin(255, dst_pixel[0] + src_pixel[0]);
+            dst_pixel[1] = llmin(255, dst_pixel[1] + src_pixel[1]);
+            dst_pixel[2] = llmin(255, dst_pixel[2] + src_pixel[2]);
+        }
+	}
+}
+
+void LLImageRaw::addEmissiveScaled(LLImageRaw* src)
+{
+	LL_INFOS() << __FUNCTION__ << LL_ENDL;
```
Ansariel, line 1932: Is this supposed to be here?

cosmic-linden: "Supposed to" is a hard question to answer given the amount of logspam already present in llimage.cpp, but I see no reason to keep this particular one.

PR #1344, opened 2024-04-27
Add timestamp to snapshot file names

### LINKED ISSUE #1343: Snapshots: Timestamp in Filename
Create snapshots with a timestamp (Snapshot_2023-11-20-074526.jpg) filename rather than a serial (Snapshot_002.jpg) name.

# Details

Second Life has long taken screenshots and saved them with a simple number suffix: _XXX.png/jpg. This makes organizing images tedious, as the filename doesn't necessarily match the order/time in which the image was captured.

There are also two formats, with the 360 snapshot feature including a bunch of ancillary information:

1. Standard Snapshot - Snapshot_001.jpg

1. 360 Snapshots - sl360_Linden_Lab_HQ_1_4096x2048_128_20231120_074526.jpg

There could be more consistency here, with the same pattern used for both. Or, if it's simpler: just add a timestamp to standard snapshots.

https://secondlife.canny.io/admin/board/feature-requests/p/snapshots-timestamp-in-filename

bennettgoble:
This changeset adds a timestamp in the format of "YYYY-MM-DD_HMS" to snapshot filenames. This is useful for understanding when a snapshot was taken, chronologically ordering files, and is less confusing than the current method of adding a number to the snapshot name, as it does not result in interleaving of old and new snapshots inside a directory.

Closes #1343

### DIFF
--- a/indra/llcommon/lldate.cpp
+++ b/indra/llcommon/lldate.cpp
@@ -86,6 +86,15 @@ std::string LLDate::asRFC1123() const
 	return toHTTPDateString (std::string ("%A, %d %b %Y %H:%M:%S GMT"));
 }
 
+std::string LLDate::toLocalDateString (std::string fmt) const
+{
+    LL_PROFILE_ZONE_SCOPED;
+
+    time_t locSeconds = (time_t) mSecondsSinceEpoch;
+    struct tm * lt = localtime (&locSeconds);
+    return toHTTPDateString(lt, fmt);
+}
+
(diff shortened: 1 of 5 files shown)


kylelinden (approved):
Approved. Obvious quality of life improvement.

marchcat (approved):
How about merging this into Maint C?

bennettgoble:
@marchcat merged into maint-c, thanks.

Issue #824, opened 2024-02-17
Replace tab indents with spaces


marchcat:
Code structure improvement: replace tab indents with spaces in the source files.

## Test Plan

Not testable by QA. Build exists = accept.

marchcat:
Moving to Maint X.

AtlasLinden:
QA passed on build with this so setting to Pending Release

PR #1350, opened 2024-04-29
#824 Replace tab indents with spaces

marchcat:


Ansariel:
Is this another attempt of screwing TPVs over for no actual reason or benefit?

PR #1352, opened 2024-04-29
Allow changing debug settings via Lua script

maxim-productengine:
Add the way to change debug settings via Lua script. This change won't persist between sessions.
Also made minor adjustment to 'Debug settings' floater to allow copy the name of the setting to the clipboard.

### DIFF
--- a/indra/newview/llappviewerlistener.cpp
+++ b/indra/newview/llappviewerlistener.cpp
@@ -35,6 +35,7 @@
 // external library headers
 // other Linden headers
 #include "llappviewer.h"
+#include "llviewercontrol.h"
 
 LLAppViewerListener::LLAppViewerListener(const LLAppViewerGetter& getter):
     LLEventAPI("LLAppViewer",
@@ -48,6 +49,10 @@ LLAppViewerListener::LLAppViewerListener(const LLAppViewerGetter& getter):
     add("forceQuit",
         "Quit abruptly",
         &LLAppViewerListener::forceQuit);
+
+    add("setDebugSetting", 
+        "Apply specified [\"value\"] to the debug [\"setting\"] (this change won't persist across sessions)", 
+        &LLAppViewerListener::setDebugSetting, llsd::map("setting", LLSD(), "value", LLSD()));
 }
 
 void LLAppViewerListener::requestQuit(const LLSD& event)
@@ -60,4 +65,13 @@ void LLAppViewerListener::forceQuit(const LLSD& event)
 {
     LL_INFOS() << "Listener requested force quit" << LL_ENDL;
     mAppViewerGetter()->forceQuit();
+}
+
+void LLAppViewerListener::setDebugSetting(const LLSD &event)
+{
+    auto setting_name = event["setting"].asString();
+    auto value = event["value"];
+    LL_WARNS("LLAppViewerListener") << "Changing debug setting \"" << setting_name << "\" to " << value << LL_ENDL;
+    //don't save this change between sesssions
+    gSavedSettings.setUntypedValue(setting_name, value, false);
 }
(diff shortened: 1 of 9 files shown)


nat-goodspeed (changes requested):
I think Lua access to settings should target [`LLViewerControlListener`](https://github.com/secondlife/viewer/blob/release/luau-scripting/indra/newview/llviewercontrollistener.cpp) instead of adding `set|getDebugSetting()` to `LLAppViewerListener`. 

Even if you do add these entry points to `LLAppViewerListener`, `LLViewerControlListener` should still pass the new `saved_value=false` parameter.

I don't mind if the Lua module is still called `LLDebugSettings`.

indra/newview/llappviewerlistener.cpp
```diff
@@ -48,6 +49,14 @@ LLAppViewerListener::LLAppViewerListener(const LLAppViewerGetter& getter):
     add("forceQuit",
         "Quit abruptly",
         &LLAppViewerListener::forceQuit);
+
+    add("setDebugSetting", 
+        "Apply specified [\"value\"] to the debug [\"setting\"] (this change won't persist across sessions)",
```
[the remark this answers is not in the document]
maxim-productengine: Hm, I like it, thanks.

indra/newview/scripts/lua/LLDebugSettings.lua
```diff
@@ -0,0 +1,13 @@
+leap = require 'leap'
+
+local LLDebugSettings = {}
+
+function LLDebugSettings.set(name, value)
+    leap.send('LLAppViewer', {op='setDebugSetting', setting=name, value=value})
```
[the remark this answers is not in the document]
maxim-productengine: Agree, seems reasonable.

indra/newview/scripts/lua/LLDebugSettings.lua
```diff
@@ -0,0 +1,17 @@
+leap = require 'leap'
+
+local LLDebugSettings = {}
+
+function LLDebugSettings.set(name, value)
+    return leap.request('LLViewerControl', {op='set', group='Global', key=name, value=value})
```
nat-goodspeed, line 6: Should `leap.request()` and `leap.generate()` notice an `"error"` key in the returned data and raise the specified error?

If they don't, maybe we should do it here and in `toggle()`.

indra/newview/llappviewerlistener.h
```diff
@@ -47,6 +47,8 @@ class LLAppViewerListener: public LLEventAPI
     void requestQuit(const LLSD& event);
     void forceQuit(const LLSD& event);
 
+    void setDebugSetting(const LLSD &event);
+
```
nat-goodspeed, line 51: Seems like you could revert the changes to both `llappviewerlistener.{h,cpp}`.

indra/newview/llviewercontrollistener.cpp
```diff
@@ -158,7 +159,8 @@ void LLViewerControlListener::toggle(LLSD const & request)
 
 	if (info.control->isType(TYPE_BOOLEAN))
 	{
-		info.control->set(! info.control->get().asBoolean());
+        LL_WARNS("LLViewerControlListener") << "Toggling debug setting " << std::quoted(info.key) << LL_ENDL;
+		info.control->set(! info.control->get().asBoolean(), false);
```
nat-goodspeed, line 163: Let's capture `! info.control->get().asBoolean()` as a local variable first so we can report it in the `LL_WARNS()` before setting it.

nat-goodspeed (changes requested):

indra/newview/scripts/lua/LLDebugSettings.lua
```diff
@@ -0,0 +1,24 @@
+leap = require 'leap'
+
+function check_response(res)
```
nat-goodspeed, line 3: Let's make this a `local` function. I like it. Does it work when you pass an unknown group or control name?

nat-goodspeed (approved):

indra/newview/scripts/lua/LLDebugSettings.lua
maxim-productengine: Yes, I've checked it.

Issue #1354, opened 2024-04-29
Evict LLMutex from coroutines


RunitaiLinden:
LLMutex is not safe to use from inside a coroutine.  Add assertions to LLMutex::lock that prevent the usage of LLMutex from a coroutine and refactor offending code to use LLCoros::Mutex or LL::WorkQueue

RunitaiLinden:
Nothing for QA to do, but this *should* lower the crash rate.

For devs, if you hit an assert in LLMutex::lock about being executed from a coroutine, refactor the offending code to either not be executed from a coroutine (if the stack trace contains non-thread-safe code) or use LLCoros::Mutex instead of LLMutex (if the code is actually thread safe).

PR #1356, opened 2024-04-29
#1354 Make coroutines use LLCoros::Mutex instead of LLMutex

RunitaiLinden:


indra/llmessage/llavatarnamecache.cpp
```diff
@@ -179,19 +180,29 @@ void LLAvatarNameCache::requestAvatarNameCache_(std::string url, std::vector<LLU
...
+            if (workqueue)
+            {
+                workqueue->post([=]()
+                    {
+
+                        if (!success)
+                        {   // on any sort of failure add dummy records for any agent IDs 
+                            // in this request that we do not have cached already
+                            std::vector<LLUUID>::const_iterator it = agentIds.begin();
+                            for (; it != agentIds.end(); ++it)
```
nat-goodspeed, line 195: `for (const auto& agent_id : agent_ids)`

indra/llplugin/llpluginprocessparent.h
```diff
@@ -207,7 +207,7 @@ class LLPluginProcessParent : public LLPluginMessagePipeOwner
 	apr_pollfd_t mPollFD;
 	static apr_pollset_t *sPollSet;
 	static bool sPollsetNeedsRebuild;
-	static LLMutex *sInstancesMutex;
+	static LLCoros::Mutex *sInstancesMutex;
```
nat-goodspeed, line 210: To be honest I'm not sure why `LLPluginProcessParent` isn't derived from `LLInstanceTracker`.

indra/newview/llappviewer.cpp
```diff
@@ -5200,6 +5200,11 @@ void LLAppViewer::updateNameLookupUrl(const LLViewerRegion * regionp)
     }
 }
 
+void LLAppViewer::postToMainThread(const LL::WorkQueue::Work& work)
```
nat-goodspeed, line 5203: Ah - I was looking for this in existing code when I saw your comment referencing it.

But maybe `postToMainCoro()` ?

indra/newview/llviewerregion.cpp
```diff
@@ -2499,10 +2499,15 @@ void LLViewerRegion::setSimulatorFeatures(const LLSD& sim_features)
         };
 
     auto workqueue = LL::WorkQueue::getInstance("mainloop");
+
     if (workqueue)
     {
         LL::WorkQueue::postMaybe(workqueue, work);
```
nat-goodspeed, line 2505: `postMaybe()` restates the `if` on line 2503, so you could just call `post()` here.

indra/newview/llvoicevivox.cpp
```diff
@@ -6437,37 +6437,52 @@ void LLVivoxVoiceClient::notifyStatusObservers(LLVoiceClientStatusObserver::ESta
...
-		it != mStatusObservers.end();
-		)
-	{
-		LLVoiceClientStatusObserver* observer = *it;
-		observer->onChange(status, getAudioSessionURI(), inSpatialChannel());
-		// In case onError() deleted an entry.
-		it = mStatusObservers.upper_bound(observer);
-	}
+    // this function is called from a coroutine, shuttle application hook back to main loop
+    auto mainqueue = LL::WorkQueue::getInstance("mainloop");
```
nat-goodspeed, line 6448: I think I'd prefer to see this `getInstance()` call closer to line 6478 where you test it.

nat-goodspeed (changes requested):
Thanks Dave!

Now that we have Class Template Argument Deduction, should new code even bother with `LLCoros::LockType`? Maybe we should just write:
```
std::scoped_lock lock(mMutex);
```
which should work for any compatible `mMutex` type.

indra/newview/llappviewer.cpp
RunitaiLinden: co-pilot hallucinated the function and named it for me :)

indra/llcommon/llcoros.cpp
```diff
@@ -61,6 +61,28 @@
...
+{
+    if (!LLCoros::instanceExists() || LLCoros::getName().empty())
+    {
+        return true;
+    }
+    
+    // error out all the time
+    // This should only be called from inside llassert or similar
+    // and is low level enough that we can't rely on LLError
+    std::cerr << "LLCoros::on_main_coro(): not on a coroutine\n";
```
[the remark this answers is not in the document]
RunitaiLinden: Removed the assumption that it will only be called from llassert, but not doing assert_on_main_coro.  I suspect llassert(on_main_coro()) will show up in thousands of spots over time, and I want all of those compiled to no-ops.

indra/llcommon/llmutex.cpp
```diff
@@ -44,7 +45,18 @@ LLMutex::~LLMutex()
...
+    LL_PROFILE_ZONE_SCOPED_CATEGORY_THREAD;
+
+    // LLMutex is not coroutine aware and should not be used from a coroutine
+    // If your code is running in a coroutine, you should use LLCoros::Mutex instead
+    // NOTE:  If the stack trace you're staring at contains non-thread-safe code,
+    // you should use LLAppViewer::instance().postToMainThread() to shuttle execution
+    // back to the main loop.
+    // NOTE: If you got here from seeing this assert in your log and you're not seeing
+    // a stack trace that points here, boost and llassert conspired to steal your stack
+    // trace, put a breakpoint in on_main_coro and try again.
```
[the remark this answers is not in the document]
RunitaiLinden: Aye

indra/llcommon/lltracerecording.h
```diff
@@ -631,6 +631,10 @@ namespace LLTrace
 		inline
 		size_t wrapi(size_t i) const
 		{
+            if (mRecordingPeriods.size() == 0)
```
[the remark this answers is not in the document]
RunitaiLinden: This was actually a false error caused by a previous attempt to error out without losing the stack... I was using exit(1), but apparently we have a signal handler somewhere that keeps the program running and just crashes randomly after the stack unwinds?  Don't know, but removing this if.

indra/llmessage/llavatarnamecache.cpp
RunitaiLinden: I just indented it, but sure...

indra/newview/llappviewer.cpp
nat-goodspeed: You know what, why not make the `"mainloop"` `WorkQueue` a member of `LLAppViewer`? Through the magic of `LLInstanceTracker::getInstance()`, it could still be found by its string name, but I like the idea of `LLAppViewer` actually owning it.

nat-goodspeed (changes requested):
One more suggestion.

nat-goodspeed (approved):

Issue #1360, opened 2024-04-30
Avoid of using avatar full names when writing to the log in release builds


LLGuru:
Logging avatar full names could be useful while debugging but is not secure in release builds

Avatar full names could be logged in debug builds but shouldn't be logged in release builds

LLGuru:
Implemented in https://github.com/secondlife/viewer/pull/1368

sergiiiproductengine:
Deferred QA in the scope of https://github.com/secondlife/iqa/issues/258.

Could you please provide a test plan?

sergiiiproductengine:
Failed QA.

Verified on the on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/259.

User name (georgeb.productengine) was observed in following cases:

<details><summary>when share item</summary>
<p>

2024-06-13T08:08:17Z WARNING # newview/lltoastalertpanel.cpp(197) LLToastAlertPanel::LLToastAlertPanel : Alert: Are you sure you want to share the following items:\n\n<nolink>Materials test HUD</nolink>\n\nWith the following Residents:\n\n<nolink>Gdisplay (georgeb.productengine)</nolink>

</p>
</details> 

<details><summary>in teleport request </summary>
<p>

2024-06-13T08:08:06Z WARNING # newview/lltoastalertpanel.cpp(197) LLToastAlertPanel::LLToastAlertPanel : Alert: Request a teleport to Gdisplay (georgeb.productengine) with the following message

</p>
</details> 

<details><summary>in profile url</summary>
<p>

2024-06-13T08:08:00Z INFO # newview/llviewermedia.cpp(2535) LLViewerMediaImpl::navigateInternal : media id= 44d1f139-a879-b914-c70d-70e17343bd67 url=https://my.secondlife.com/georgeb.productengine/, mime_type=text/html
2024-06-13T08:08:00Z INFO # newview/llviewermedia.cpp(1895) LLViewerMediaImpl::loadURI : Asking media source to load URI: https://my.secondlife.com/georgeb.productengine/

</p>
</details> 

<details><summary>when removing friend</summary>
<p>

2024-06-13T08:07:22Z WARNING # newview/lltoastalertpanel.cpp(197) LLToastAlertPanel::LLToastAlertPanel : Alert: Do you want to remove <nolink>Gdisplay (georgeb.productengine)</nolink> from your Friends List?
[...]

PR #1368, opened 2024-04-30
secondlife/viewer#1360 Avoid of using avatar full names when writing to the log in release builds

LLGuru:
For https://github.com/secondlife/viewer/issues/1360

### DIFF
--- a/indra/llcharacter/llcharacter.h
+++ b/indra/llcharacter/llcharacter.h
@@ -119,6 +119,8 @@ public:
 
 	virtual void addDebugText( const std::string& text ) = 0;
 
+	virtual std::string getDebugName() const { return getID().asString(); }
+
 	virtual const LLUUID&	getID() const = 0;
 	//-------------------------------------------------------------------------
 	// End Interface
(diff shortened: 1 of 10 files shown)


indra/llcharacter/llcharacter.h
```diff
@@ -119,6 +119,8 @@ class LLCharacter
 
 	virtual void addDebugText( const std::string& text ) = 0;
 
+	virtual std::string getSecureName() const { return getID().asString(); }
```
akleshchev, line 122: I feel like calling it 'secure name' is a misnormer...

akleshchev (approved):

indra/llcharacter/llcharacter.h
marchcat: I second this. What does "secure" mean in this case? I see just an ID.

akleshchev: It's overriden below, so it isn't just an id, it can return an actual name.

Purpose of this function is purely for debug, so I suggest using getDebugName or getDebInfoName given the definition check.

marchcat (approved):

PR #1373, opened 2024-05-01
#824 Replace tab indents with spaces in Maint X

marchcat:

PR #1374, opened 2024-05-01
Add editorconfig

bennettgoble:
Provide a basic editorconfig file with this repository's defaults for spacing and line ending styles. This can go into any branch: it should only impact new code.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[Makefile]
+indent_style = tab
+
+[*.{yml,yaml}]
+indent_size = 2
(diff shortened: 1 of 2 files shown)


marchcat (approved):
This should work great with the Maint X update from #1373.

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[Makefile]
+indent_style = tab
```
marchcat, line 11: We currently use spaces for makefiles as well. At least, my script did nothing to them except trimming trailing spaces..

nat-goodspeed: ? At least some versions of `make` infer a semantic distinction between tab indent and space indent.

marchcat: Ah, I've been thinking about `cmake`.

PR #1376, opened 2024-05-01
Project/channel by branch

vir-linden:
Use branch name to determine viewer_channel in build.yaml


.github/workflows/build.yaml
```diff
@@ -176,16 +176,22 @@ jobs:
...
-          # viewer channel "Second Life Project Shiny" (ignoring "#hash",
-          # needed to disambiguate tags).
-          if [[ "$GITHUB_REF_TYPE" == "tag" && "${GITHUB_REF_NAME:0:12}" == "Second_Life_" ]]
-          then viewer_channel="${GITHUB_REF_NAME%#*}"
-               export viewer_channel="${viewer_channel//_/ }"
-          else export viewer_channel="Second Life Test"
+
+          # determine the viewer channel from the branch name
+          branch=$AUTOBUILD_VCS_BRANCH
+          echo "branch=$branch" >> "$GITHUB_OUTPUT"
```
nat-goodspeed, line 182: This doesn't need to be put to `"$GITHUB_OUTPUT"` (thus creating `${{ steps.build.outputs.branch }}`) because it's already there as `${{ steps.which-branch.outputs.branch }}` -- that's how we set `$AUTOBUILD_VCS_BRANCH` at [line 108](https://github.com/secondlife/viewer/blob/7f23f160d7520034afd8c3be8a184eac1716227a/.github/workflows/build.yaml#L108).

.github/workflows/build.yaml
```diff
@@ -176,16 +176,22 @@ jobs:
...
+          IFS='/' read -ra ba <<< $branch
+          prefix=${ba[0]}
+          if [ "$prefix" == "project" ]; then
+              proj_name=$(echo ${ba[1]} | sed -e 's/_/ /g' | awk '{for(i=1;i<=NF;i++) $i=toupper(substr($i,1,1)) tolower(substr($i,2));}1')
```
nat-goodspeed, line 186: The reader must already figure out our string-splitting tactic on line 183, so in my opinion it's clearer to apply the same again to the second part of the `project/some_words` branch name.
```
IFS='_' read -ra prj << "${ba[1]}"
# bash can uppercase the first letter of a variable,
# or the first letter of each word of an array
export viewer_channel="Second Life ${prefix^} ${prj[*]^}"
```

.github/workflows/build.yaml
```diff
@@ -176,16 +176,22 @@ jobs:
...
+              export viewer_channel="Second Life Project $proj_name"
+          elif  [ "$prefix" == "release" ] || [ "$prefix" == "main" ];
```
nat-goodspeed, line 188: could also be
```
elif [[ "$prefix" == "release" || "$prefix" == "$main" ]]
```
with a couple fewer child-process forks.

.github/workflows/build.yaml
```diff
@@ -354,7 +360,8 @@ jobs:
   release:
     needs: [build, sign-and-package-windows, sign-and-package-mac]
     runs-on: ubuntu-latest
-    if: github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life_')
+      # Build with a tag like "Second_Life#abcdef0" to generate a release page (used for builds we are planning to deploy).
+    if: github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life')
```
nat-goodspeed, line 364: [Caladbolg is advocating](https://lindenlab.slack.com/archives/C02R43FJP/p1714514147094849?thread_ts=1714500309.255049&cid=C02R43FJP) triggering release builds with a tag that includes the branch name -- not to specify the branch name to the build, but for clarity in navigating.

Maybe something like: `branch/<branchname>#discriminator` ?

`tags:` at line 8 would have to change too.

nat-goodspeed (changes requested):

nat-goodspeed (approved):
I like it. I still think we should adopt Caladbolg's suggestion to trigger on something more informative, which suggests a smaller prefix. Now that we no longer use the content of tag name, I guess we could trigger on a tag like
```
Second_Life_branch_name#discriminator
```
but I think I'd rather see something like "build" or "release" or something short that conveys the intent of the tag.

PR #1393, opened 2024-05-02
Turn on LL_TESTS for Maint B.

nat-goodspeed:
These changes have succeeded when built against `main`. @Nicky-D requests pre-testing the merge into Maint B.


nat-goodspeed:
Looks like missing googlemock? Apparently we do have [a Linux googlemock build](https://github.com/secondlife/3p-googlemock/releases/tag/v1.7.0.2b109d4).

Turning off LL_TESTS for CI builds was an accident. It has always been our intention that CI builds of branches run clean with tests on.

Nicky-D:
I'm on team enable tests, too. Just they should even compile 😟

Re Google mock. Is that even used? I remember vaguely "no" and that those includes can just go.

nat-goodspeed:
Even better!

nat-goodspeed:
Guess that wasn't sufficient.

Nicky-D:
I took a look;
- Yes google mock is indeed needed now
- I did not get the crazy template errors I used to get
- LL has no Linux Google mock. Only win/win64 and that OS we do not speak of

Nicky-D:
I tried to build the google mock 3p, but that fails during test stage with segfaults.

Googling suggests this could be due to the old gmock version in the 3p. As google mock is now part of the gtest suits, this potentially makes updating not any easier

Nicky-D:
@nat-goodspeed  / @bennettgoble  
IF someone can fork and create a release of [googletest](https://github.com/Nicky-D/3p-googletest) please.

This hopefully allows to go forward with this PR

bennettgoble:
@Nicky-D would it be preferred to cherry-pick changes from https://github.com/secondlife/viewer/pull/1077? Googlemock has been removed from DRTVWR-600-maint-A.

bennettgoble:
Ah, nevermind: googleTEST not googleMOCK.

Nicky-D:
@bennettgoble you are right. That other PR is golden. Commencing with cherry picking that first

AiraYumi:
When unused code or libraries are deleted (except for moving dependencies to other libraries), 
I think it would be better to apply it to the main branch, but is that too harsh?

Nicky-D:
I tried LL_TESTS=On after cherry picking #1077 and the tests compile with just one minimal issue. So I think this PR can be closed and the problems dealt with once main gets merged into this branch.

It's crucial though to then merge or cherry pick #1077 too

bennettgoble:
> If library deletion is applied at the project viewer level, other projects will continue to have dependencies until that project viewer is promoted to main.

> It's crucial though to then merge or cherry pick https://github.com/secondlife/viewer/pull/1077 too

Yes, these issues highlight problems with the current branching strategy. As it stands, with few changes being merged into the default branch, it's difficult to know where changes are, or that they exist.

Nicky-D:
> > If library deletion is applied at the project viewer level, other projects will continue to have dependencies until that project viewer is promoted to main.
> 
> > It's crucial though to then merge or cherry pick #1077 too
> 
> Yes, these issues highlight problems with the current branching strategy. As it stands, with few changes being merged into the default branch, it's difficult to know where changes are, or that they exist.

I do not want to derail this PR too much .. but I did wonder this morning to open an Issue for someone to document the branching strategy.

For example we got release/maint-*, but also DRTVWR-...-maint-* and there seems (to me) little rhyme or reason what goes where and when it goes into main.

AiraYumi:
I don't think the current contents of the main branch will resolve the build failure on Mac.
Merge release/maint-x?

marchcat:
> Merge release/maint-x?
Yes, those changes weren't included into the current `main`, but they are needed.

nat-goodspeed:
I think the current state of this PR gets us where we want to be: it runs tests on Windows and Mac and requires that they succeed, while permitting Linux build/test failures so they don't block the whole project.

AiraYumi:
Is it possible to prevent Linux from halting even if a Windows or Mac build does not complete?

nat-goodspeed:
Yes, you could set `continue-on-error: true` and the build would continue regardless of any errors. But this setting serves our purposes until Linux consistently builds reliably.

PR #1398, opened 2024-05-03
Add LLLater class to manage time-delayed tasks.

nat-goodspeed:
The big idea is to reduce the number of per-tick callbacks asking, "Is it time
yet? Is it time yet?" LLEventTimer and LLEventTimeout have historically used that tactic.

LLLater presents doAtTime(LLDate), with doAfterInterval() and doPeriodically()
methods implemented using doAtTime(). All return timer handles. The free functions
doAfterInterval() and doPeriodically() now forward to the corresponding
LLLater methods.

LLLater also presents isRunning(handle) and cancel(handle).

LLLater borrows the tactic of LLEventTimer: while there's at least one running
timer, it registers an LLCallbackList tick() callback to service ready timers.
But instead of looping over all of them asking, "Are you ready?" it keeps them
in a priority queue ordered by desired timestamp, and only touches those whose
timestamp has been reached. Also, it honors a maximum time slice: once the first
of the ready timers have exceeded the time limit, it defers processing other
ready timers to the next tick() call. The intent is to consume fewer cycles
per tick() call, both by the management machinery and the timers themselves.

Revamp LLCallbackList to accept C++ callables in addition to (classic C
function pointer, void*) pairs. Make addFunction() return a handle (different
than LLLater handles) that can be passed to a new deleteFunction() overload,
since std::function instances can't be compared for equality.

In fact, implement LLCallbackList using boost::signals2::signal, which provides
almost exactly what we want.

LLCallbackList continues to accept (function pointer, void*) pairs, but now
we store a lambda that calls the function pointer with that void*. It takes
less horsing around to create a C++ callable from a (function pointer, void*)
pair than the other way around. For containsFunction() and deleteFunction(),
such pairs are the keys for a lookup table whose values are handles.

Instead of having a static global LLCallbackList gIdleCallbacks, make
LLCallbackList an LLSingleton to guarantee initialization. For backwards
compatibility, gIdleCallbacks is now a macro for LLCallbackList::instance().

Move doOnIdleOneTime() and doOnIdleRepeating() functions to LLCallbackList
methods, but for backwards compatibility continue providing free functions.

Reimplement LLEventTimer using LLLater::doPeriodically(). One implication is
that LLEventTimer need no longer be derived from LLInstanceTracker, which we
used to iterate over all instances every tick. Give it start() and stop()
methods, since some subclasses (e.g. LLFlashTimer) used to call its member
LLTimer's start() and stop(). Remove updateClass():
LLCallbackList::callFunctions() now takes care of that.

Remove LLToastLifeTimer::start() and stop(), since LLEventTimer now provides
those.

While at it, make LLEventTimer::tick() return bool instead of BOOL, and change
existing overrides.

Make LLApp::stepFrame() call LLCallbackList::callFunctions() instead of
LLEventTimer::updateClass().

We could have refactored LLEventTimer to use the mechanism now built into
LLLater, but frankly the LLEventTimer API is rather clumsy. You MUST derive a
subclass and override tick(), and you must instantiate your subclass on the
heap because, when your tick() override returns false, LLEventTimer deletes
its subclass instance. The LLLater API is simpler to use, and LLEventTimer is
much simplified by using it.

Merge lleventfilter.h's LLEventTimeoutBase into LLEventTimeout, and likewise
merge LLEventThrottleBase into LLEventThrottle. The separation was for
testability, but now that they're no longer based on LLTimer, it becomes
harder to use dummy time for testing. Temporarily skip tests based on
LLEventTimeoutBase and LLEventThrottleBase.

Instead of listening for LLEventPump("mainloop") ticks and using LLTimer,
LLEventTimeout now uses LLLater::doAfterInterval(). Instead of LLTimer and
LLEventTimeout, LLEventThrottle likewise now uses LLLater::doAfterInterval().

Replace a couple local LLEventTimeout pre-lambda functor classes with lambdas.

Make LLMainThreadTask use the "mainloop" WorkQueue instead of LLEventTimer.

Dignify F64 with a new typedef LLDate::timestamp. LLDate heavily depends on
that as its base time representation, but there are those who question use of
floating-point for time. This is a step towards insulating us from future
changes.


indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+inline
+LLCallbackList::handle_t doOnIdleRepeating(bool_func_t callable)
+{
+    return gIdleCallbacks.doOnIdleRepeating(callable);
+}
+
+/*****************************************************************************
+*   LLLater: callbacks at some future time
+*****************************************************************************/
+class LLLater: public LLSingleton<LLLater>
```
Nicky-D, line 106: I'm really not a fan of this name.
For once: L-L-L-ater. How man L can we stuff in there? YES!

Also it is very unspecific. For the periodic calls it is close to wrong. As they are not strictly called later, but at internals.

Maybe ScheduleFunction? Or split the periodic and one shot functions into "InvokeLater" and "InvokePeriodically"?

(I'm bad at naming too!)

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+
+public:
+    // If tasks that come ready during a given tick() take longer than this,
+    // defer any subsequent ready tasks to a future tick() call.
+    static constexpr F32 TIMESLICE{ 0.005f };
+
+    class handle_t
+    {
+    private:
+        friend class LLLater;
```
Nicky-D, line 149: Friends are good to have in life. But why does handle_t need friendship and not a strict business relationship of having to call `handle_t( value )` and a new `token_t getToken() const` method?

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+        token_t token;
+    public:
+        handle_t(token_t token=0): token(token) {}
+        bool operator==(const handle_t& rhs) const { return this->token == rhs.token; }
+        explicit operator bool() const { return bool(token); }
+        bool operator!() const { return ! bool(*this); }
+    };
 
+    // Call a given callable once at specified timestamp.
+    handle_t doAtTime(nullary_func_t callable, LLDate::timestamp time);
```
Nicky-D, line 159: I propose "scheduleAt"

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+
+    // Call a given callable once after specified interval.
+    handle_t doAfterInterval(nullary_func_t callable, F32 seconds);
```
Nicky-D, line 162: Here I'd propose `scheduleAfter`

Plus create a new type `seconds`. [strong type iz strong!](https://www.fluentcpp.com/2016/12/08/strong-types-for-strong-interfaces/)

This comes with the downside of making it impossible to mix F32 and seconds at will, but with the upside that they cannot just be mixed and misinterpreted

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+
+    // Call a given callable every specified number of seconds, until it returns true.
+    handle_t doPeriodically(bool_func_t callable, F32 seconds);
```
Nicky-D, line 165: In lieu `schedulePeriodically( ..., tSeconds seconds )`

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+
+    // test whether specified handle is still live
+    bool isRunning(handle_t timer) const;
```
Nicky-D, line 168: To me `isRunning` would mean "right now this function runs". I am pretty sure this is not what this method is about to return. `isScheduled` maybe?

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+    // check remaining time
+    F32 getRemaining(handle_t timer) const;
```
Nicky-D, line 170: Here is a good example why I am musing about a specific `seconds` type. `getRemaning` what? Also it is not obvious what this means:
- remaining units till this handle executes?
- remaining units till this handle finishes execution?
- remaining execution counts?

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+
+    // Cancel a future timer set by doAtTime(), doAfterInterval(), doPeriodically().
+    // Return true iff the handle corresponds to a live timer.
```
Nicky-D, line 173: `iff` > `if` I guess

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+    // cancel again.
+    bool cancel(handle_t& timer);
+
+    // Store a handle_t returned by doAtTime(), doAfterInterval() or
+    // doPeriodically() in a temp_handle_t to cancel() automatically on
+    // destruction of the temp_handle_t.
+    class temp_handle_t
+    {
+    public:
+        temp_handle_t() {}
```
Nicky-D, line 185: is `temp_handle_t() = default` an option?

indra/llcommon/llcallbacklist.h
```diff
@@ -27,53 +27,248 @@
...
+    HandleMap mHandles;
+    token_t mToken{ 0 };
+    // While mQueue is non-empty, register for regular callbacks.
+    LLCallbackList::temp_handle_t mLive;
+
+    struct Periodic;
+
+    // internal implementation for doAtTime()
+    HandleMap::iterator doAtTime1(LLDate::timestamp time);
+    handle_t doAtTime2(nullary_func_t callable, HandleMap::iterator iter);
```
Nicky-D, line 256: I'm not sure why we need doAtTime1/2 here, when the compiler is perfectly capable to distinguish them by their arguments. 

Also
- `dotAtTime1` would be in lieu of my suggestion of using `schedule... `
- `doAtTime2 suddenly decides to be totally different, while all previous methods prefixed with doAt accepted some sort of interval or time, now there is an iterator passed.
-

indra/llcommon/llcallbacklist.h
nat-goodspeed: The intention of `handle_t` is to be completely opaque to consumers of `LLLater`. `token_t` is meant to be an internal-only implementation detail, because it might change.

indra/llcommon/llcallbacklist.h
nat-goodspeed: I like that, yes. The names were chosen to be consistent with the existing free functions [`doAfterInterval()` and `doPeriodically()`](https://github.com/secondlife/viewer/blob/main/indra/llcommon/llcallbacklist.h#L71-L75).

indra/llcommon/llcallbacklist.h
nat-goodspeed: The function I'm trying to emulate here is [`F32SecondsImplicit getRemainingTimeF32()`](https://github.com/secondlife/viewer/blob/main/indra/llcommon/lltimer.h#L100). I just think that's an awkwardly long name. I grant that `getRemainingSecondsF32()` would be even clearer, but it's even longer.

Again, the name is based on the idea that a `handle_t` references a timer, and it returns the time until that timer is supposed to fire. What the callback chooses to do at that point is up to the callback.

The second bullet isn't a reasonable interpretation. Even with reflection this class can't know how long until the callback finishes running: you run smack into The Halting Problem.

Since a `doPeriodically()` callback's `bool` return determines whether it will be called again in future, similar remarks apply to the third bullet.

indra/llcommon/llcallbacklist.h
nat-goodspeed: Should I instead have appended "else `false`" ? That just felt redundant to me.

indra/llcommon/llcallbacklist.h
nat-goodspeed: I guess so; is that better? (Sincere question: if that somehow makes the intent clearer to a reader, please explain.)

indra/llcommon/llcallbacklist.h
Nicky-D: Err no, there is a typo "return true iff" with one f too much :)

indra/llcommon/llcallbacklist.h
Nicky-D: Hmm, tough spot. I suppose the questions being how much time it would consume to change that or if it is cheaper as in work and later maintenance to follow bad principles

indra/llcommon/llcallbacklist.h
Nicky-D: Yes, this would make more sense

indra/llcommon/llcallbacklist.h
Nicky-D: In times for auotcompletion everywhere and even AI suggesting us code (to then fix the mistakes the AI made) I'd argue a longer name to make intent clear is worth it

indra/llcommon/llcallbacklist.h
nat-goodspeed: Those internal-only names are because, to implement `doPeriodically()`, I found it necessary to split the original implementation of `doAtTime()` into two other methods. They literally mean "first steps of `doAtTime()`," "second steps of `doAtTime()`."

But it looks like I'm going to have to refactor all that anyway.

indra/llcommon/llcallbacklist.h
Nicky-D: considering the other ctors and operator= are removed via `=delete` at the least I say it is consistent.

indra/llcommon/llcallbacklist.h
nat-goodspeed: Specifically, I want to make it as painless as possible for other maintainers to say, "Hmm, you know, this `LLEventTimer` subclass is way too much boilerplate for this simple use case" and drop in `LL::Timers` (or whatever) instead.

indra/llcommon/llcallbacklist.h
nat-goodspeed: Oh! I was using what I thought was a common abbreviation for "if and only if." Sorry!

maxim-productengine (approved):
Looks good. Let's test it.

marchcat (approved):

indra/llcommon/llcallbacklist.cpp
```diff
@@ -45,186 +44,343 @@ LLCallbackList::~LLCallbackList()
...
-void doPeriodically(bool_func_t callable, F32 seconds)
+bool Timers::tick()
 {
-	new BoolFuncEventTimer(callable, seconds);
+    // Fetch current time only on entry, even though running some mQueue task
+    // may take long enough that the next one after would become ready. We're
+    // sharing this thread with everything else, and there's a risk we might
+    // starve it if we have a sequence of tasks that take nontrivial time.
+    auto now{ LLDate::now().secondsSinceEpoch() };
+    auto cutoff{ now + TIMESLICE };
```
akleshchev, line 305: I suggest exposing this to callbacks/Timers in some fassion. A lot of the 'onIdle' callbacks are declaring own slices. Ex: inventory search and population, which means every visible inventory panel uses own 'slice'.

Instead of making TIMESLICE constant I suggest making it dynamic. Ex: when loading bar is visible (during teleport), may be when viewer is in the background allow larger slices since fps won't matter as much.

nat-goodspeed: Interesting point. Okay. I can support runtime get/set operations, but I don't propose to make those calls in the example places you mentioned.

I'm glad you're not suggesting that LLCallbackList try to vary the timeslice depending on the nature of the tasks in the queue.

nat-goodspeed: Please see [this commit](https://github.com/secondlife/viewer/pull/1555/commits/ce9886e3d374df3017eeaeeb9e30f28343712896) on the lua-timers branch.

nat-goodspeed:
This PR has been subsumed into [this one](https://github.com/secondlife/viewer/pull/1555).

PR #1407, opened 2024-05-04
GCC compatibility

Nicky-D:
GCC will rightfully complain that counterStr is too small. So increase it accordingly.

This function though is peculiar in more than this way, it could run away like a diesel engine on fire. As the for loop possible can never exit. Which granted is very unlikely but could happen none the less

### DIFF
--- a/indra/llfilesystem/tests/lldir_test.cpp
+++ b/indra/llfilesystem/tests/lldir_test.cpp
@@ -434,7 +434,7 @@ namespace tut
       
       for (counter=0, foundUnused=false; !foundUnused; counter++ )
       {
-         char counterStr[3];
+         char counterStr[12] = {};
          sprintf(counterStr, "%02d", counter);
          uniqueDir = dirbase + counterStr;
          foundUnused = ! ( LLFile::isdir(uniqueDir) || LLFile::isfile(uniqueDir) );


nat-goodspeed:
Woof. Please see [NamedTempDir](https://github.com/secondlife/viewer/blob/62a79346af133f06b70cc02daf4f97062d1d61b0/indra/llcommon/tests/llprocess_test.cpp#L242-L266), based on [NamedTempFile::temp_path()](https://github.com/secondlife/viewer/blob/62a79346af133f06b70cc02daf4f97062d1d61b0/indra/test/namedtempfile.h#L97-L114).

Nicky-D:
I unfortunately cannot really make a 'good' fix. main cannot be compiled with Linux yet (and I cannot then run test either)

I agree that whole function is janky, but I'm also a bit hesitant trying to improve it without being able to compile/test.

nat-goodspeed:
> main cannot be compiled with Linux yet

because of the Boost dependency?

Is there a branch on which you could make this fix and then retarget the commit against main?

Nicky-D:
Everything Linux is still in release/main-b Which answer the question where it could be worked on.
I'm unsure though how much that did divert from main and how safe it would be to apply such a fix on branch.

I guess we will never know till tried, I'll give it a shot this week

Nicky-D:
To my dismay have I not found any time yet. But I have not forgotten about this.
But I saw [1461](https://github.com/secondlife/viewer/pull/1461)  and Mac having a blast in Maint-X too. All that said I will close this PR now:
- To not keep PRs open for ages with no feedback from my side.
- In hopes the issue might be tackled with the Mac fixes anyway.

If the latter one won't help here, we can address it with a new PR

PR #1410, opened 2024-05-06
autobuild.xml fix

vir-linden:


nat-goodspeed:
It's weird to me that the Mac build failed because the same change -- same slvoice links -- built clean locally on my Mac without the VPN running.

nat-goodspeed:
That said, the same commit that built for me on Mac failed locally on Windows, with the 400 Bad Request errors, both with and without the VPN connected.

nat-goodspeed:
This has been addressed by another PR.

PR #1419, opened 2024-05-07
Copy xml files to scripts/lua; make Lua debug floater resizable

maxim-productengine:
XML files for demo scripts (test_luafloater_demo.lua, test_luafloater_gesture_list.lua) were not copied to scripts/lua, which we should do.

 Allow resizing Lua debug floater, as Atlas asked.


nat-goodspeed (approved):
Thank you.

Can the viewer's XUI Preview Tool already find an .xml file in scripts/lua? If not, maybe we should enable it to do that?

PR #1422, opened 2024-05-07
Combat2 viewer UI work.

Rider-Linden:
This is viewer work UI for the Combat2 project. It adds two messages sent to the viewer upon avatar death, one for being teleported to a region's telehub and another for teleportation to the parcel landing point.

It also adds an area to the Region tab of the Region/Estate floater for adjusting combat related settings.  On non-combat regions (Enable Combat is unchecked) and regions that don't support Combat2 these new options are grayed out.

image

This is still waiting on #1404


bennettgoble:
@Rider-Linden Can you re-open this when it is ready for incorporation into the develop branch? Doing a bit of stale PR cleanup.

PR #1427, opened 2024-05-08
Enable IME on XWayland desktop environment

AiraYumi:
This change fixes the IME to work even if the desktop environment is running on Wayland.

### DIFF
--- a/indra/llwindow/llwindowsdl.cpp
+++ b/indra/llwindow/llwindowsdl.cpp
@@ -654,11 +654,6 @@ BOOL LLWindowSDL::createContext(int x, int y, int width, int height, int bits, B
         SDL_SetHint( std::get<0>(hint), std::get<1>(hint));
     }
 
-#if LL_LINUX
-    // Enable IME on XWayland
-    setenv("XMODIFIERS", std::string("@im=fcitx;ibus").c_str(), 1);
-#endif
-
     std::initializer_list<std::tuple<uint32_t, char const*, bool>> initList=
             { {SDL_INIT_VIDEO,"SDL_INIT_VIDEO", true},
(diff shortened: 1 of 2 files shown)


indra/llwindow/llwindowsdl.cpp
```diff
@@ -654,6 +654,11 @@ BOOL LLWindowSDL::createContext(int x, int y, int width, int height, int bits, B
         SDL_SetHint( std::get<0>(hint), std::get<1>(hint));
     }
 
+#if LL_LINUX
+    // Enable IME on XWayland
+    setenv("XMODIFIERS", std::string("@im=fcitx;ibus").c_str(), 1);
```
Nicky-D, line 659: This should usually be set in via your DE environment/startup and thus globally and there should be no need for any application to overwrite it.

If it absolutely cannot be set for just the viewer, the best place is the viewer startup shell script.

brad-linden: I agree, putting this into `indra/newview/linux_tools/wrapper.sh` would be a better solution

brad-linden (changes requested):

indra/llwindow/llwindowsdl.cpp
AiraYumi: I see.
I'll move there.

AiraYumi:
The environment variable setting location has been moved.

I checked again and found that ime was enabled when using fcitx, but not when using ibus. I don't know the cause.

I can't change it any further at this point, so please review it and merge if it's okay.

brad-linden (approved):

brad-linden:
thanks for the contribution!

PR #1441, opened 2024-05-09
KHR_materials_emissive_strength support

Geenz:
Also we may have had a bad merge in LLGLTFMaterial.


indra/llprimitive/llgltfmaterial.cpp
```diff
@@ -208,6 +214,27 @@ std::string LLGLTFMaterial::asJSON(bool prettyprint) const
...
 
+// static
+void LLGLTFMaterial::setEmissiveStrengthFromModel(const tinygltf::Material& model, F32& emissive_strength)
+{
+    LL_PROFILE_ZONE_SCOPED;
+
+    const tinygltf::Value::Object &extensions_object = model.extensions;
+    const auto                     emissive_it      = extensions_object.find(GLTF_FILE_EXTENSION_EMISSIVE_STRENGTH);
+
+    LL_INFOS() << "Material extensions: " << model.extensions_json_string << LL_ENDL;
```
cosmic-linden, line 225: Some straggler debug text here

indra/llprimitive/llgltfmaterial.cpp
```diff
@@ -297,6 +327,14 @@ F32 LLGLTFMaterial::floatFromJson(const tinygltf::Value::Object& object, const c
     return (F32)real_json.GetNumberAsDouble();
 }
 
+// static
+void LLGLTFMaterial::writeEmissiveStrength(tinygltf::Material &material, F32 emissive_strength)
+{
+    tinygltf::Value::Object emissive_strength_object;
+    emissive_strength_object[LLGLTFMaterial::GLTF_FILE_EXTENSION_EMISSIVE_STRENGTH_EMISSIVE_STRENGTH] = tinygltf::Value(emissive_strength);
+    material.extensions[LLGLTFMaterial::GLTF_FILE_EXTENSION_EMISSIVE_STRENGTH_EMISSIVE_STRENGTH]      = tinygltf::Value(emissive_strength_object);
```
cosmic-linden, line 335: Typo - this should be `GLTF_FILE_EXTENSION_EMISSIVE_STRENGTH`

indra/llprimitive/llgltfmaterial.cpp
brad-linden, line 225: perhaps `LL_DEBUGS("GLTF")` here? do we expect this to be useful to collect from end users logs in bug reports?

indra/llprimitive/llgltfloader.h
```diff
@@ -79,6 +79,7 @@ class gltf_render_material
     double      normalScale;    // scale applies only to X,Y components of normal
     double      occlusionScale; // strength multiplier for occlusion
     LLColor4    emissiveColor;  // emissive mulitiplier, assumed linear encoding (spec 2.0 is silent)
+    double      emissiveStrength;  // multiplier for emissiveColor
```
brad-linden, line 82: this is confusing to me that it is `double` here but handled as `F32` everywhere else, but seems like that is what we do for all the other fields so ok i guess

indra/newview/app_settings/settings.xml
```diff
@@ -9153,6 +9153,17 @@
       <real>0.00</real>
     </array>
   </map>
+  <key>RenderTransmissionDetail</key>
```
brad-linden, line 9156: getting a jump on the next extension too eh?

brad-linden (approved):

indra/llprimitive/llgltfmaterial.h
```diff
@@ -120,12 +130,6 @@ class LLGLTFMaterial : public LLRefCount
     //       heightmaps cannot currently be described as finite enclosed
     //       volumes.
     // See also LLPanelRegionTerrainInfo::validateMaterials
-    bool mDoubleSided = false;
-
-
-    // These fields are local to viewer and are a part of local bitmap support
-    typedef std::map<LLUUID, LLUUID> local_tex_map_t;
-    local_tex_map_t                  mTrackingIdToLocalTexture;
```
cosmic-linden, line 128: @AndrewMeadows changed the order of the fields to save some bits. That's relevant to resolving this merge conflict. You may be interested in choosing a location for mEmissiveStrength to optimize the packing (although the savings may only be relevant server-side)

indra/llprimitive/llgltfmaterial.cpp
Geenz: Yeah, I'll 'nix it.

indra/newview/skins/default/xui/en/panel_gltf_material.xml
```diff
@@ -339,6 +339,31 @@
...
+      height="10"
+      layout="topleft"
+      left_delta="0"
+      top_pad="5"
+      width="96"
+      name="emissive_strength_lbl">
+      Emissive Strength
+    </text>
+    <spinner
+      decimal_digits="3"
```
cosmic-linden, line 355: I think this will set the limit to 999 or something, which could be quite low for this particular use case.

indra/llprimitive/llgltfloader.h
Geenz: Yeah, I was a bit weirded out by that.  I'm keeping with the established convention more or less.

indra/newview/skins/default/xui/en/panel_gltf_material.xml
Geenz: It's _probably_ fine.  999 on white (or even like 0.1% grey) is gonna be pretty bright - even for HDR use cases.  There shouldn't be any clamping elsewhere beyond what's mandated by the spec in the file.

Geenz: Though this shouldn't be clamped to 1.  Fixing that.

indra/llprimitive/llgltfmaterial.h
Geenz: Yeah - I wasn't sure what the specific ordering was supposed to be here.  I'll look at the git-blame and reorder as needed.

indra/newview/app_settings/settings.xml
Geenz: Params are all getting wired up tomorrow!

indra/llprimitive/llgltfloader.h
cosmic-linden: Looking at the GLTF spec, it does seem there are cases where single precision floating point is enforced. From the section on "Accessors Bounds":

> For floating-point components, JSON-stored minimum and maximum values represent single precision floats and SHOULD be rounded to single precision before usage to avoid any potential boundary mismatches.

Not relevant to this PR, but might be worth keeping in mind for the future. cc: @RunitaiLinden

indra/llprimitive/llgltfmaterial.h
Geenz: I restored the packing from the previous merge - I'm generally assuming there's a very good reason for it.

cosmic-linden (approved):

brad-linden:
note: this PR has lots of merges in it, be sure not to do 'Squash and merge' when landing it.

cosmic-linden (approved):

indra/llprimitive/llgltfmaterial.h
cosmic-linden: Grokking this further, it appears there's a comment/warning further down that `LLGLTFMaterial::getHash` uses `mLocalTexDataDigest` to determine the hash offset of the material object. (see "IMPORTANT: do not move this member down") Any fields prior to `mLocalTexDataDigest` will not get included in the hash. I didn't consider this before, but this probably takes precedence over @AndrewMeadows ' size optimizations.

This also would explain why the merge broke. There was a fix to GLTF hashing (and therefore render batching) in a maintenance branch.

bennettgoble:
@Geenz Can you re-open this when it is ready for incorporation into the develop branch? Doing a bit of stale PR cleanup.

PR #1446, opened 2024-05-10
Fix broken merge and BOOL/bool issues

Ansariel:
@RunitaiLinden Fix for broken merge (doesn't compile) and remaining BOOL/bool issues.

Also: The GLTF tests are once again broken...


marchcat:
Do we need to merge this into Maint A as well?

Ansariel:
> Do we need to merge this into Maint A as well?

No. These are issues specific to this branch.

PR #1473, opened 2024-05-14
brad/mac shader fix

brad-linden:
fix error: 'Redeclaration of sized array 'terrain_texcoord' not allowed'


brad-linden (commented):
the first declaration appears here: L202-209

cosmic-linden (approved):
Oops! Thanks for the catch.

PR #1481, opened 2024-05-15
Add trusted flag to UI callbacks, so not everything is accessible from the script

maxim-productengine:
We have lots of UI callbacks defined in the viewer. So, to avoid lots of changes at once, I suggest adding separate registry, mostly (but not only) for functions we allow to be called from the script.
Added UNTRUSTED_THROTTLE similar to the one used in LLCommandHandlerRegistry::dispatch(..). We'll need to double check what is allowed to be called (in llviewermenu.cpp and) what should be throttled.


indra/newview/lluilistener.cpp
```diff
@@ -59,43 +60,57 @@ LLUIListener::LLUIListener():
 
 void LLUIListener::call(const LLSD& event) const
 {
-    LLUICtrl::commit_callback_t* func =
-        LLUICtrl::CommitCallbackRegistry::getValue(event["function"]);
-    if (! func)
+    static F64 last_throttle_time = 0.0;
```
nat-goodspeed: I'd prefer that `last_throttle_time` be a non-static member of `LLUIListener`.

indra/newview/lluilistener.cpp
```diff
@@ -59,43 +60,57 @@ LLUIListener::LLUIListener():
...
+    if (info->mHandleUntrusted == LLUICtrl::LLCommitCallbackInfo::UNTRUSTED_THROTTLE)
+    {
+        F64 cur_time = LLTimer::getElapsedSeconds();
+        if (cur_time < last_throttle_time + THROTTLE_PERIOD)
+        {
+            LL_WARNS("LLUIListener") << "Throttled function " << std::quoted(event["function"].asString()) << LL_ENDL;
+            return;
+        }
+        last_throttle_time = cur_time;
+    }
```
nat-goodspeed, line 90: Are we going to need to replicate support for `UNTRUSTED_BLOCK` and `UNTRUSTED_THROTTLE` in other places, or is this the one place where that's applicable?

There's another instance of new throttle code in a recent PR. I find myself musing about a wrapper like:
```
// either execute 'lambda' and return 'true', or don't and return 'false'
// shared throttle
template <typename LAMBDA>
bool throttle(F64& lasttime, F64 limit, LAMBDA&& lambda);
// function-static throttle (distinct instance for each LAMBDA)
template <typename LAMBDA>
bool throttle(F64 limit, LAMBDA&& lambda)
{
    static F64 lasttime = 0.0;
    return throttle(lasttime, limit, std::forward<LAMBDA>(lambda));
}
```

indra/newview/scripts/lua/util.lua
```diff
@@ -41,4 +41,11 @@ function util.equal(t1, t2)
     return util.empty(temp)
 end
 
+function util.check_response(res)
+  if res.error then
+    error(res.error)
+  end
+  return res
+end
```
nat-goodspeed, line 49: It's a good idea to centralize that logic. I decided to [bake it into `leap.lua`](https://github.com/secondlife/viewer/pull/1555/files#diff-4edaba8867aa1260f8c65602f02cb56cd59d5114829d78faa133d11a98e1af14), which would mean `UI.lua` can just call `leap.request()`. Should we cherry-pick that `leap.lua` commit?

indra/newview/llviewermenu.cpp
```diff
@@ -9395,16 +9395,19 @@ void initialize_edit_menu()
 
 }
 
+typedef LLUICtrl::LLCommitCallbackInfo cb_info;
+#define COMMIT_ADD(func_name, func) LLUICtrl::SharedCommitCallbackRegistry::currentRegistrar().add(func_name, cb_info(func))
+#define COMMIT_ADD_TRUSTED(func_name, func) LLUICtrl::SharedCommitCallbackRegistry::currentRegistrar().add(func_name, cb_info(func, cb_info::UNTRUSTED_BLOCK))
```
nat-goodspeed, line 9400: Instead of two macros, how about one (inline?) function with a default `EUntrustedCall` value? That lets you pass `UNTRUSTED_THROTTLE` too.

indra/newview/llviewermenu.cpp
```diff
@@ -9395,16 +9395,19 @@ void initialize_edit_menu()
...
+
 void initialize_spellcheck_menu()
 {
-    LLUICtrl::CommitCallbackRegistry::Registrar& commit = LLUICtrl::CommitCallbackRegistry::currentRegistrar();
```
nat-goodspeed, line 9400: So at least in this source, we're not using `CommitCallbackRegistry` at all any more, only `SharedCommitCallbackRegistry`. That makes me feel more positive about simply changing the value type of `CommitCallbackRegistry` to `cb_info`, instead of having a whole separate `SharedCommitCallbackRegistry`.

nat-goodspeed (changes requested):
I'm willing to be told I don't yet understand the relationship between `CommitCallbackRegistry` and `SharedCommitCallbackRegistry`. But as far as I can tell right now, you could consolidate the two into `CommitCallbackRegistry`.

indra/newview/scripts/lua/util.lua
maxim-productengine: Sounds good, I'll cherry-pick those changes.

maxim-productengine:
On the first sight, there a lot of changes, but most of them are just consequences of changing CommitCallbackRegistry and removing SharedCommitCallback Registry.

nat-goodspeed (approved):
Maxim, **thank you** for a VERY thorough review of use cases!

indra/newview/lluilistener.cpp
```diff
@@ -48,54 +50,75 @@ LLUIListener::LLUIListener():
...
-        // nose in such an error, crump rather than quietly ignoring it.
-        LL_WARNS("LLUIListener") << "function '" << event["function"] << "' not found" << LL_ENDL;
+        return response.error(stringize("Function ", std::quoted(event["function"].asString()), " was not found"));
+    }
+    if (info->handle_untrusted == cb_info::UNTRUSTED_BLOCK) 
+    {
+        return response.error(stringize("Function ", std::quoted(event["function"].asString()), " is not allowed to be called from the script"));
+    }
+    F64 cur_time = LLTimer::getElapsedSeconds();
+    bool is_untrusted_throttle = info->handle_untrusted == cb_info::UNTRUSTED_THROTTLE;
```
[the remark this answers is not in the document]
maxim-productengine: Agree, it looks cleaner.
(Probably, if (or when) we have one more throttle timer for Lua calls, I'll use your idea with separate wrapper function)

indra/newview/lluilistener.cpp
```diff
@@ -48,54 +50,75 @@ LLUIListener::LLUIListener():
...
     }
     else
+    {
+        mLastMinThrottle = cur_time;
+    }
+
+
+    LLUICtrl::commit_callback_t func = info->callback_func;
+
+    if (info->callback_func)
```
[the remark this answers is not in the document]
maxim-productengine: Currently, I don't see a real use case with an unspecified function, so let's move it above as you said.

indra/newview/llviewermenu.cpp
```diff
@@ -9428,8 +9430,8 @@ void initialize_menus()
         bool mMult;
     };
 
+    LLUICtrl::CommitRegistrarHelper registrar(LLUICtrl::CommitCallbackRegistry::defaultRegistrar());
```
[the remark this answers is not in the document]
maxim-productengine: Thanks, I'll restore `currentRegistrar()`, although I'm not sure there is a difference for that case.

indra/llui/lluictrl.cpp
```diff
@@ -283,13 +283,13 @@ LLUICtrl::commit_signal_t::slot_type LLUICtrl::initCommitCallback(const CommitCa
...
         std::string function_name = cb.function_name;
         setFunctionName(function_name);
-        commit_callback_t* func = (CommitCallbackRegistry::getValue(function_name));
-        if (func)
+        LLUICtrl::CommitCallbackInfo *info = (CommitCallbackRegistry::getValue(function_name));
+        if (info && info->callback_func) 
         {
             if (cb.parameter.isProvided())
-                return boost::bind((*func), _1, cb.parameter);
+                return boost::bind((info->callback_func), _1, cb.parameter);
```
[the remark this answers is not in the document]
maxim-productengine: Last time we had fix_whitespace.py. Now we need remove_boost_bind.py.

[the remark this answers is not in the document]
nat-goodspeed: Ha! Of course a change to `std::bind()` would be straightforward but distressingly pervasive. It does make me wonder if an AI coding assistant could correctly convert `boost::bind()` calls to lambdas. I strongly doubt it. If I'm proved wrong, my respect for AI coding assistants will get a big Boost.

PR #1486, opened 2024-05-15
Experiment with getting manually triggered workflow builds to optionally do a release

brad-linden:
Screenshot 2024-05-15 at 2 54 12 PM

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,7 +27,7 @@ jobs:
       # When you want to use a string variable as a workflow YAML boolean, it's
       # important to ensure it's the empty string when false. If you omit || '',
       # its value when false is "false", which is interpreted as true.
-      RELEASE_RUN: ${{ github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life') && 'Y' || '' }}
+      RELEASE_RUN: ${{ (github.event.inputs.release_run || github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life')) && 'Y' || '' }}
     steps:
       - name: Set Variable
         id: setvar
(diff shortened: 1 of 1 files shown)


nat-goodspeed (approved):

nat-goodspeed:
@vir-linden , check this out!

nat-goodspeed:
image

brad-linden:
we may also want to add a string field or combo box or something for configuring the channel for manually triggered workflow_dispatch events

brad-linden:
documentation for using inputs is here:
https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

.github/workflows/build.yaml
```diff
@@ -22,7 +27,7 @@ jobs:
       # When you want to use a string variable as a workflow YAML boolean, it's
       # important to ensure it's the empty string when false. If you omit || '',
       # its value when false is "false", which is interpreted as true.
-      RELEASE_RUN: ${{ github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life') && 'Y' || '' }}
+      RELEASE_RUN: ${{ (github.event.inputs.release_run || github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life')) && 'Y' || '' }}
```
nat-goodspeed, line 30: I think (but am not sure, in workflow YAML expression land) that `&&` binds more closely than `||`. That means your parens are correct as written -- but might be clearer to the casual reader if you additionally parenthesized the original `(... == 'tag' && startsWith(...))` expression.

nat-goodspeed (approved):
I like this change.

vir-linden:
This is very cool, thanks!

brad-linden:
so, unfortunately it looks like more work will be needed to make this checkbox actually functional.  currently it is failing to run the release action when a tag is not present

nat-goodspeed:
Ah - like [this run](https://github.com/secondlife/viewer/actions/runs/9117271875/job/25069778581).

Also, I think the logic isn't quite correct yet: when I asked to run [this build](https://github.com/secondlife/viewer/actions/runs/9117271875/job/25067737711), I did _not_ check the `release_run` checkbox.

PR #1490, opened 2024-05-16
Attempt to speed up which-branch's `pip3 install PyGithub`

brad-linden:


brad-linden:
hmm this doesn't do what I want, bottleneck actually appears to be in the which-branch.py script itself

brad-linden:
that doesn't sound that inefficient.  there are a small enough number of active branches at a given time (<500) that we should be able to iterate over them,  I still don't think it should take 3 minutes to do that

nat-goodspeed:
Oho, but now that we have a preliminary `setvar` job, we could move branch detection there and do it once for all platforms!

nat-goodspeed:
https://github.com/secondlife/viewer/commit/7e381d62a8e7f64f1bae96a8ef6b38316427381b

brad-linden:
this approach seems unlikely to move the needle on build times, closing.

PR #1492, opened 2024-05-16
Update threejs to use the platform agnostic package

marchcat:


nat-goodspeed (approved):
Ha, thank you! I was just trying a local build of the same change, but go for it!

marchcat:
My local build finished first 😹

PR #1496, opened 2024-05-16
Attempt to diagnose stack overflow in test programs. Also, improve build.yaml logic.

nat-goodspeed:


indra/llcommon/tests/llleap_test.cpp
```diff
@@ -35,7 +35,7 @@
 // causes Windows abdominal pain such that it later fails code-signing in some
 // mysterious way. Entirely suppressing these LLLeap tests pushes the failure
 // rate MUCH lower. Can we re-enable them with a smaller data size on Windows?
-const size_t BUFFERED_LENGTH =  100*1024;
+const size_t BUFFERED_LENGTH = 1023*1024;
```
brad-linden, line 38: although that comment is disturbing, possibly this would be better consolidated outside of the `#if defined(LL_WINDOWS)` block?

.github/workflows/build.yaml
```diff
@@ -21,20 +21,32 @@ jobs:
...
+      # Even though inputs.release_run is specified with type boolean, which
+      # correctly presents a checkbox, its *value* is a GH workflow string
+      # 'true' or 'false'. If you simply test github.event.inputs.release_run,
+      # it always evaluates as true because it's a non-empty string either way.
       # When you want to use a string variable as a workflow YAML boolean, it's
       # important to ensure it's the empty string when false. If you omit || '',
-      # its value when false is "false", which is interpreted as true.
-      RELEASE_RUN: ${{ (github.event.inputs.release_run || github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life')) && 'Y' || '' }}
+      # its value when false is "false", which (again) is interpreted as true.
+      RELEASE_RUN: ${{ (github.event.inputs.release_run != 'false' || (github.ref_type == 'tag' && startsWith(github.ref_name, 'Second_Life'))) && 'Y' || '' }}
```
brad-linden, line 36: good catch, sorry about that

brad-linden (approved):
seems worth a shot.

.github/workflows/build.yaml
nat-goodspeed: I don't mean to throw shade! Even if you echo `github.event.inputs.relase_run` to stdout, you see `false` either way. There's no way to directly ask the workflow engine for the _type_ of that variable, and since you explicitly state that it's `boolean`, it's entirely reasonable to think that type might persist. I only got here because empirically it wasn't working, and empirically this change made it better.

nat-goodspeed:
n.b. [This is what I was shooting for](https://github.com/secondlife/viewer/actions/runs/9128885333/job/25102161528#step:9:1221):
```
[llleap, 9: unknown] exception: Windows_SEH_exception: 'test threw c00000fd (Stack Overflow)'
```

indra/test/test.cpp
```diff
@@ -520,6 +524,64 @@ void wouldHaveCrashed(const std::string& message)
...
+    {
+        // This is a non-C++ exception, e.g. hardware check.
+        // By the time the handler gets control, the stack has been unwound,
+        // so report the stack trace now at filter() time.
+        // Sadly, even though, at the time of this writing, stack overflow is
+        // the problem we would most like to diagnose, calling another
+        // function when the stack is already blown only terminates us faster.
+        if (code != STATUS_STACK_FULL)
+        {
+            std::cerr << boost::stacktrace::stacktrace() << std::endl;
```
brad-linden, line 555: probably not worth holding this up, but is there something else minimal we could do?  is  there a way to grab just the top function name of the stack trace without blowing the stack further?

brad-linden (approved):

.github/workflows/build.yaml
brad-linden: yeah no offense taken, just lamenting how hard it is to test this CI stuff before it lands

indra/test/test.cpp
nat-goodspeed: I don't know what part of a stack frame Boost.Stacktrace needs to touch to derive that lowest-level entry. What I got without that `if` was an Access Violation, which I think means Boost.Stacktrace was trying to look at a part of the frame that overlapped beyond the end of the available stack. And it would be the lowest-level function that overlaps. So... I don't know if there's anything we can safely do there.

brad-linden (approved):
👍

marchcat (approved):

PR #1503, opened 2024-05-17
fix bad indentation and modernise

beqjanus:
The indentation here is very misleading (in fact, it causes GCC and Clang to fail with warnings)
This fixes the indentation and modernises the code.
Arguably, the loops are redundant as the underlying resources should be automatically deleted.


marchcat (approved):
This is great, thank you!

PR #1504, opened 2024-05-17
#1386 Remove Buttons from mapping options

LLGuru:
Additional change to ignore input from inappropriate channels
when entering a new channel mapping using the game controller


AndrewMeadows (approved):
It is good!

PR #1518, opened 2024-05-17
Server Side Translation

vir-linden:
Second Life Project SST


bennettgoble:
Will re-open when work resumes.

PR #1540, opened 2024-05-21
viewer#1529 Mesh Uploader should permit lowering material count

akleshchev:
1. Allow having less materials/faces for lower lods
2. Fix UI not updating when selected combo-option changes


akleshchev:
I'm not sure total_submeshes check is needed at all since it doesn't work right with submodels (if model A and B have 4 and 5 faces at high lod, viewer would have let a case of 5 and 4 at a lower lod pass).

maxim-productengine (approved):

Issue #1510, opened 2024-05-17
LLGameControl - Implement Save/Load operations for device specific settings for each device separately (save only differences from the default settings: storing default settings should just exclude this section from the settings)


AndrewMeadows:

LLGuru:
Implemented in https://github.com/secondlife/viewer/pull/1544

PR #1544, opened 2024-05-22
#1510 LLGameControl: Implement Save/Load operations for device settings

LLGuru:


indra/llcommon/llsd.cpp
```diff
@@ -840,6 +840,23 @@ void LLSD::clear()						{ Impl::assignUndefined(impl); }
...
+    if (t == TypeUndefined)
+        return true;
+    if (t == TypeBinary)
+        return !asBoolean();
+    if (t == TypeInteger)
+        return !asInteger();
+    if (t == TypeReal)
+        return !asReal();
+    if (t == TypeString || t == TypeMap || t == TypeArray)
+        return !size();
```
akleshchev, line 855: Shouldn't this check UUIDs and dates? And why not 'switch'?

P.S. Shouldn't this be safe(impl).isEmpty() to warranty all types get covered?

indra/llcommon/llsd.cpp
```diff
@@ -840,6 +840,23 @@ void LLSD::clear()						{ Impl::assignUndefined(impl); }
...
+    return false;
```
akleshchev, line 856: I suggest using a `switch` since Type is an enum. Ideally to use `safe(impl).isEmpty()` to make sure no type gets missed

indra/llcommon/llsd.h
```diff
@@ -311,6 +311,20 @@ class LL_COMMON_API LLSD
             LL_PROFILE_ZONE_SCOPED_CATEGORY_LLSD;
             return (*this)[String(c)];
         }
+
+        template<typename T>
+        LLSD(const std::map<String, T>& map, bool include_empty = false)
```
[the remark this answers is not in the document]
LLGuru: Not sure. In this case I don't want to add empty entries to a settings element.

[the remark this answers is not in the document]
akleshchev: std::map's constructors in my opinion are good examples how to do it right, and there you pass a comparator, not a hidden condition.

akleshchev (approved):

indra/llwindow/llgamecontrol.cpp
```diff
@@ -585,28 +851,54 @@ void LLGameControllerManager::initializeMappingsByDefault()
...
 
+void LLGameControllerManager::addController(SDL_JoystickID id, const std::string& guid, const std::string& name)
+{
     llassert(id >= 0);
-    llassert(controller);
 
     if (findDevice(id) != mDevices.end())
     {
-        LL_WARNS("GameController") << "device already added" << LL_ENDL;
+        LL_WARNS("SDL2") << "device id=" << id << " was already added" << LL_ENDL;
```
akleshchev, line 892: I suggest reporting guid and name here just in case. I'm not sure how those ids work, but I doubt id is helpfull for figuring out what happened.

LLGuru: This logging is being done immediately before the call `addController`

indra/llwindow/llgamecontrol.cpp
```diff
@@ -441,57 +620,144 @@ std::string LLGameControl::Options::AxisOptions::saveToString() const
...
+        else if (depth) // Assume '}' here
+        {
+            depth--;
+            offset = pos + 1;
+        }
+        else
+        {
+            return false; // Extra '}' found
+        }
     }
```
AndrewMeadows, line 692: Since you're performing the operation `depth--`, that variable should be `S32` instead of `U32` , and the loop should `return false` immediately when `depth` goes negative.

indra/llwindow/llgamecontrol.cpp
```diff
@@ -309,11 +446,17 @@ namespace
 
     constexpr U8 MAX_AXIS = NUM_AXES - 1;
     constexpr U8 MAX_BUTTON = NUM_BUTTONS - 1;
+    constexpr U8 MAX_AXIS_DEAD_ZONE = 255;
+    constexpr U8 MAX_AXIS_OFFSET = 255;
```
AndrewMeadows, line 450: The raw axis value is a `S16`, so `MAX_AXIS_DEAD_ZONE` should be `S16` and probably higher than `255`... maybe even as high as 2^14 = `16384`.  Similarly, `MAX_AXIS_OFFSET` should be `S16`.

indra/llwindow/llgamecontrol.cpp
```diff
@@ -441,57 +620,144 @@ std::string LLGameControl::Options::AxisOptions::saveToString() const
...
+        S32 number = std::stoi(offset);
+        if (abs(number) > MAX_AXIS_OFFSET || std::to_string(number) != offset)
+        {
+            LL_WARNS("SDL2") << "Invalid offset value: '" << offset << "'" << LL_ENDL;
+        }
+        else
+        {
+            mOffset = (U8)number;
+        }
+    }
```
AndrewMeadows, line 749: `offset` should be allowed to be negative.  The logic will have to clamp a `max` and a `min`.

AndrewMeadows (changes requested):

indra/llwindow/llgamecontrol.cpp
LLGuru: I gave these values randomly, hoping that you provide the correct values :-)

AndrewMeadows: Sure, DEAD_ZONE can be explicitly positive but applied symmetrically on the negative side, however, since we'll be doing **math** with S16 axis measurement the DEAD_ZONE should also be S16.

An axis measurement will range positive and negative and that channel may very well need a negative OFFSET adjustment.

LLGuru: The value of MAX_AXIS_DEAD_ZONE is also always positive
It doesn't make sense to set a negative limit for mDeadZone

Moreover, I would like to be sure that mDeadZone is positive
That's why I use the unsigned integer type (U8 replaced with U16)

indra/llwindow/llgamecontrol.cpp
```diff
@@ -386,28 +530,63 @@ bool LLGameControl::State::onButton(U8 button, bool pressed)
...
 {
-    if (axis >= 0 && axis < mAxisOptions.size())
+    if (axis > MAX_AXIS)
+    {
+        LL_WARNS("SDL2") << "Invalid input axis: " << axis << LL_ENDL;
+    }
+    else
     {
         const AxisOptions& options = mAxisOptions[axis];
         value += options.mOffset;
```
AndrewMeadows, line 592: To avoid wraparound effects I think `options.mOffset` should be `S32`.  Then we could do the math in `S32` space, clamp to `S16` limits, and cast the final result to `S16`:

```
constexpr S32 MAX_AXIS_VALUE = 32767;
constexpr S32 MIN_AXIS_VALUE = -32768;
value = (S16)(std::clamp((S32)(value) + options.mOffset, MIN_AXIS_VALUE, MAX_AXIS_VALUE));
```

AndrewMeadows (changes requested):

scripts/messages/message_template.msg
```diff
@@ -9144,7 +9144,7 @@ version 2.0
 // message will be resent a few times in the hopes the server finally receives it.
 //
 {
-	GameControlInput Medium 431 NotTrusted Zerocoded
+	GameControlInput High 32 NotTrusted Zerocoded
```
AndrewMeadows, line 9147: Thanks for making these `message_template.msg` changes.  I will be cutting a corresponding PR on the **server** code later today.

AndrewMeadows (approved):
This looks good.

PR #1546, opened 2024-05-22
Adjust Copy3rdPartyLibs.cmake for Visual Studio 2022 17.10.0 and greater

Ansariel:
@nat-goodspeed The current Visual Studio 2022 Version 17.10.0 update comes with an updated C++ compiler that identifies as follows:
```
-- The C compiler identification is MSVC 19.40.33808.0
-- The CXX compiler identification is MSVC 19.40.33808.0
```
This will currently break builds. To make builds work with the current VS version and assuming that future updates will be in the range 17.40 - 17.49, I adapted the CMake file for copying the VC++ runtime accordingly.


akleshchev (approved):

marchcat (approved):

marchcat:
Thank you!

PR #1548, opened 2024-05-22
Fix line endlings

Ansariel:
@marchcat I have no idea how that happened, but apparently Git decided to change all line endings to CR-LF when running that glorious commands mentioned in https://github.com/secondlife/viewer/wiki/2024-Consistent-Whitespace

This should hopefully fix all the mangled up files...


marchcat (approved):
@Ansariel, that's interesting, I had all these files with LF only, so it indeed looks like your local git had changed them.
Thanks for noticing that!

PR #1549, opened 2024-05-22
Add Git attributes file to prevent accidental line ending changes

Ansariel:
@marchcat After the disastrous line endings screw up, maybe add a .gitattributes file to prevent that from happening again...


marchcat (approved):
Thank you! Let's add this.

Issue #1511, opened 2024-05-17
LLPanelPreferenceGameControl - Add a tab control to separate common settings (action-to-channel mappings) from device-specific settings (channel-to-button mappings etc.)


AndrewMeadows:

LLGuru:
Implemented in https://github.com/secondlife/viewer/pull/1559

PR #1559, opened 2024-05-24
#1511 LLPanelPreferenceGameControl: Add a tab for device settings

LLGuru:


indra/newview/skins/default/xui/en/panel_preferences_game_control.xml
```diff
@@ -10,198 +10,206 @@
...
+        width="380"
+        height="18">
         <combo_box.item
-            label="AXIS_0-"
+            label="AXIS_0- (Left stick: push left)"
             name="AXIS_0-"
             value="AXIS_0-"/>
         <combo_box.item
-            label="AXIS_0+"
+            label="AXIS_0+ (Left stick: push right)"
```
AndrewMeadows, line 79: Left is positive and right is negative, so swap the two descriptions here, and also for the left-right cases below.

indra/newview/skins/default/xui/en/panel_preferences_game_control.xml
```diff
@@ -10,198 +10,206 @@
...
         <combo_box.item
-            label="AXIS_1-"
+            label="AXIS_1- (Left stick: push up)"
             name="AXIS_1-"
             value="AXIS_1-"/>
         <combo_box.item
-            label="AXIS_1+"
+            label="AXIS_1+ (Left stick: push down)"
             name="AXIS_1+"
             value="AXIS_1+"/>
```
AndrewMeadows, line 89: Up is positive and down is negative, so swap the two descriptions here, and also for the up-down cases below.

AndrewMeadows (changes requested):

AndrewMeadows (approved):
This is good enough to move forward.

PR #1560, opened 2024-05-24
viewer#1408 Change misleading content removal wanring

akleshchev:
Not the only issue here, ex: viewer permits renaming but server refuses to rename. But should be good enough for now.

### DIFF
--- a/indra/newview/llpanelobjectinventory.cpp
+++ b/indra/newview/llpanelobjectinventory.cpp
@@ -307,20 +307,27 @@ BOOL LLTaskInvFVBridge::isItemRenameable() const
 BOOL LLTaskInvFVBridge::renameItem(const std::string& new_name)
 {
     LLViewerObject* object = gObjectList.findObject(mPanel->getTaskUUID());
-    if(object)
+    if(!object)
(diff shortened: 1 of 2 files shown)


akleshchev:
Edit: included 'rename', still needs a pass from UX or some review.

indra/newview/llpanelobjectinventory.cpp
```diff
@@ -307,20 +307,27 @@ BOOL LLTaskInvFVBridge::isItemRenameable() const
...
-            object->updateInventory(
-                new_item,
-                TASK_INVENTORY_ITEM_KEY,
-                false);
-        }
+        return false;
+    }
+    if (!object->permModify())
+    {
+        LLNotificationsUtil::add("CantModifyContentInNoModTask");
```
maxim-productengine, line 316: We won't get this notification because of the previous check.

akleshchev: Missed it, thanks...

maxim-productengine (approved):

marchcat (approved):

PR #1547, opened 2024-05-22
Add support for sending messages to Nearby chat from Lua script

maxim-productengine:
Add support for sending messages to Nearby chat via Lua scripts.
Update UI to distinguish between messages sent by the script.

PR #1565, opened 2024-05-24
Nat's ideas from PR #1547

nat-goodspeed:


maxim-productengine (approved):
I like these new util functions.

PR #1571, opened 2024-05-27
Try copying the VC++ runtime files from the redistributable package first

Ansariel:
...instead of clobbering the files from some Windows system directory.


marchcat (approved):
Thank you!

PR #1593, opened 2024-05-30
viewer#1583 [Win] Viewer fails to detect more than 4Gb of video memory

akleshchev:
Half cleanup, half remake.

1. Removed pre-dx9 code (initD3D) since it is now pointless, dx9 is supposed to be built in and d3d isn't aware of total memory, only free memory.
2. Restored and fixed 'budget' code (it wasn't cast correctly in release)
3. Made dx9 adapter to no longer persists since it only needs to grab memory once, so also renamed it from initDX to checkDXMem
4. debugEnumerateGraphicsAdapters was basically doing the same work initDX was doing, but for all adapters, so I merged the two

### DIFF
--- a/indra/llwindow/llwindowwin32.cpp
+++ b/indra/llwindow/llwindowwin32.cpp
@@ -361,14 +361,8 @@ struct LLWindowWin32::LLWindowWin32Thread : public LL::ThreadPool
         mGLReady = true;
     }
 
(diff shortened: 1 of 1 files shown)


maxim-productengine (approved):

marchcat (approved):
🔥

PR #1604, opened 2024-05-31
Add support for RequestTaskInventory capability

SaladDais:
See https://feedback.secondlife.com/feature-requests/p/move-object-inventory-listing-to-http-capability

Apparently there's been a capability for fetching LLSD object inventory added recently, use it in the viewer to make object inventory loads faster. Make the viewer prefer to load inventory via the capability rather than over the deprecated Xfer system, though both are still supported. Inventories that took several minutes to load now load in 0.5 seconds or so.

I don't totally understand some of the existing object inventory code related to `obj->mInventorySerialNum` and  `obj->mExpectedInventorySerialNum` mismatches, but this code seems to behave correctly, might deserve some extra scrutiny.

I'm not sure if `release/maint-b` is the correct base branch to use here, but that's the one that builds Linux correctly for me, so that's the one I use :)

### DIFF
--- a/indra/newview/llviewerobject.cpp
+++ b/indra/newview/llviewerobject.cpp
@@ -3036,24 +3036,33 @@ void LLViewerObject::fetchInventoryFromServer()
         delete mInventory;
         mInventory = NULL;
 
-        // Results in processTaskInv
-        LLMessageSystem* msg = gMessageSystem;
-        msg->newMessageFast(_PREHASH_RequestTaskInventory);
-        msg->nextBlockFast(_PREHASH_AgentData);
-        msg->addUUIDFast(_PREHASH_AgentID, gAgent.getID());
-        msg->addUUIDFast(_PREHASH_SessionID, gAgent.getSessionID());
-        msg->nextBlockFast(_PREHASH_InventoryData);
-        msg->addU32Fast(_PREHASH_LocalID, mLocalID);
-        msg->sendReliable(mRegionp->getHost());
-
         // This will get reset by doInventoryCallback or processTaskInv
         mInvRequestState = INVENTORY_REQUEST_PENDING;
+
+        if (mRegionp && !mRegionp->getCapability("RequestTaskInventory").empty())
+        {
+            LLCoros::instance().launch("LLViewerObject::fetchInventoryFromCapCoro()",
+                                       boost::bind(&LLViewerObject::fetchInventoryFromCapCoro, mID));
+        }
+        else
+        {
+            LL_WARNS() << "Using old task inventory path!" << LL_ENDL;
+            // Results in processTaskInv
+            LLMessageSystem *msg = gMessageSystem;
+            msg->newMessageFast(_PREHASH_RequestTaskInventory);
+            msg->nextBlockFast(_PREHASH_AgentData);
+            msg->addUUIDFast(_PREHASH_AgentID, gAgent.getID());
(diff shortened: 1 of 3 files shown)


indra/newview/llviewerobject.cpp
```diff
@@ -3466,6 +3555,52 @@ BOOL LLViewerObject::loadTaskInvFile(const std::string& filename)
     return TRUE;
 }
 
+BOOL LLViewerObject::loadTaskInvLLSD(const LLSD& inv_result)
```
Ansariel, line 3558: Could have just return void since the result is never actually use in some way (or at least return true booleans, since the fake BOOL is on its way out with Maint-A)

indra/newview/llviewerobject.cpp
```diff
@@ -3466,6 +3555,52 @@ BOOL LLViewerObject::loadTaskInvFile(const std::string& filename)
...
+        }
+
+        // Synthesize the "Contents" category, the viewer expects it, but it isn't sent.
+        LLPointer<LLInventoryObject> inv = new LLInventoryObject(mID, LLUUID::null, LLAssetType::AT_CATEGORY, "Contents");
+        mInventory->push_front(inv);
+
+        const LLSD& inventory = inv_result["contents"];
+        auto iter = inventory.beginArray();
+        auto end = inventory.endArray();
+        for(; iter != end; ++iter)
```
Ansariel, line 3578:
```
for (const auto& inv_entry : llsd::inArray(inventory))
{
    if (inv_entry.has("item_id"))
```

indra/newview/llviewerobject.cpp
```diff
@@ -3084,6 +3093,72 @@ void LLViewerObject::fetchInventoryDelayedCoro(const LLUUID task_inv, const F64
...
+            // have mutated the task inventory since we kicked off the request, and those mutations may have not
+            // been taken into account yet. Of course, those mutations may have actually failed which would result
+            // in the inv serial never increasing.
+            //
+            // When we detect this case, set the expected inv serial to the inventory serial we actually received
+            // and kick off a re-request after a slight delay.
+            S16 serial = (S16)result["inventory_serial"].asInteger();
+            potentially_stale = serial < obj->mExpectedInventorySerialNum;
+            LL_INFOS() << "Inventory loaded for " << task_inv << LL_ENDL;
+            obj->mInventorySerialNum = serial;
```
akleshchev, line 3132: `getAndSuspend` can take significant time, `obj` might become invalid by that point, crashes in this place are likely. I suggest either repeating `gObjectList.findObject` after `getAndSuspend` or at least using an LLPointer

indra/newview/llviewerobject.cpp
SaladDais: Aha, was hoping there was an API like that, thanks!

indra/newview/llviewerobject.cpp
SaladDais: I'll go with repeating `gObjectList.findObject()` + a null check. I'll also remove my pointless `->isDead()` since it seems an object shouldn't be accessible through `findObject()` once dead anyway.

SaladDais:
Rebased on top of current `release/maint-b` and squashed in fixes to address review feedback

marchcat (approved):

akleshchev (approved):

PR #1635, opened 2024-06-05
viewer-private#244 Reconnect to voice after 'tuning'

akleshchev:
activate() does nothing if call already started so either suspending should mark call as suspended or resuming should mark it as ready


marchcat (approved):

roxanneskelly (commented):
LGTM

PR #1636, opened 2024-06-05
An alternative in getting CPU frequency on Apple Silicon M-series

erikkundiman:
This solution was retrieved from [https://listman.redhat.com/archives/libvir-list/2022-February/228217.html](https://listman.redhat.com/archives/libvir-list/2022-February/228217.html)

`sysctl hw.cpufrequency` would result as empty on Apple Silicon (at least) M1 & M2, when run natively.

Ironically (and that's why it's been working with viewers relying on Rosetta 2),
`arch -x86_64 /bin/bash -c 'sysctl hw.cpufrequency'`
run on an Apple Silicon would result in the correct number.

And just maybe M3 or M4 implements `cpufrequency`, which should be more accurate than the alternative (cause they give different results at least on my Intel Mac), hence `tbfrequency * clockrate.hz` is used only when the result from the first (original) attempt is 0, instead of applying it to all Apple Silicon (preprocessed with __arm64__) regardless.


erikkundiman:

marchcat:

erikkundiman:
I'm glad if I can be of any help!

marchcat (approved):
Going to merge this after #1656.

Issue #1638, opened 2024-06-05
Add permissions check for Develop->GLTF->Save As/Upload...


RunitaiLinden:
You should only be able to Upload or Save an object as GLTF if you are the owner and creator of the object according to SL permissions.

When Upload/Save is disabled, the corresponding buttons in Develop->GLTF menu should be disabled.

brad-linden:
looks like this should be in state integrated, is that right?

RunitaiLinden:
No QA needed.

Dan-Linden:
Fractal on Aditi has GLTF content to test with. Develop->render metadata->GLTF Nodes will highlight the GLTF content.

Dan-Linden:
This should get some QA attention anyway.

AtlasLinden:
Setting status to Pending Release.
Negative test cases performed. That is, no visible changes spotted outside of what can be accessed via the Develop menu. GLTF->Save As/Upload menu items are only enabled on the PMFP channel on aditi and using the 2024.06 Atlasaurus builds. Tested on 7.1.9.9865199239 Win10/MacOS

PR #1653, opened 2024-06-05
#1638 Add permissions checks to GLTF Save As and Upload buttons

RunitaiLinden:


indra/llinventory/llpermissions.h
```diff
@@ -353,6 +338,11 @@ bool LLPermissions::allowMoveBy(const LLUUID& agent) const
     return allowOperationBy(PERM_MOVE, agent, LLUUID::null);
 }
 
+bool LLPermissions::allowExportBy(const LLUUID& agent) const
+{
+    return agent == mOwner && agent == mCreator;
```
akleshchev, line 343: Is it fine that applying to an object that you have permissions to edit won't work?

ex: agent A shares permissions with agent B to edit All A's objects and then B tries to apply a material?

akleshchev (approved):

indra/llinventory/llpermissions.h
RunitaiLinden: Yes.... modifiable does not mean exportable.  We don't have a permissions flag for "can export," so keeping the permissions for exporting as restrictive as possible pending feedback from residents.

brad-linden (approved):

PR #1670, opened 2024-06-09
Fix issues resulting from b42f9d836b4c0f7fbd4bdae1734021e2a09fdbe8

Ansariel:
@marchcat Apparently some weird old cruft was causing issues with the size_t changes. Had to "re-cruft" some of these changes for a quick and easy fix after going through the entire set of changes twice. Hopefully these are all issues.


marchcat (approved):

marchcat:
@Ansariel, thank you!

PR #1675, opened 2024-06-10
Fix for the laggy tracker beacon on Mac Apple Silicon

mobserveur:
This fix comes in 2 commits : 

commit 1: 
I first pushed out of the loop the gl.begin() and gl.end() to avoid multiple draw calls.
I also changed the BEACON_VERTS value for DARWIN, but since only one draw call was made it wasn't so much needed anymore.

commit 2:
I switched to using TRIANGLES instead of TRIANGLES_STRIP because the beacon is basically some sort of grid and to draw a grid using a triangles strip properly, in 1 call, i would have needed to use degenerate triangles, and indices. 
The way llrender deals with those formal immediate calls (and buffering them) also made me choose this method.
I also adjusted the subdivisions according to the actual height of the DOWN and UP beacons, with a constant defining the maximum of subdivisions.
The code remains compatible with the debug setting : CheesyBeacon which adds an animated effect.

the issue : https://github.com/secondlife/viewer/issues/999

### DIFF
--- a/indra/newview/lltracker.cpp
+++ b/indra/newview/lltracker.cpp
@@ -489,10 +489,21 @@ void draw_shockwave(F32 center_z, F32 t, S32 steps, LLColor4 color)
     gGL.end();
 }
 
+
 void LLTracker::drawBeacon(LLVector3 pos_agent, std::string direction, LLColor4 fogged_color, F32 dist)
 {
+#if LL_DARWIN
+    const U32 BEACON_VERTS = 16;
+#else
     const U32 BEACON_VERTS = 256;
+#endif
+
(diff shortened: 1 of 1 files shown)


mobserveur:

marchcat (commented):

brad-linden:
I'm sorry I didn't expect this to get closed. The project/gltf_development branch has been renamed to develop. Please reopen this and base it on that branch

erikkundiman:
Oops! I'm sorry, I should have force pushed the branch. I hope it's okay that mobserveur will have to recreate the PR.

PR #1678, opened 2024-06-10
Fix incorrect use of VX/VY/VZ/VW indices when color components are actually accessed

Ansariel:
@marchcat More fixes 😄


marchcat (approved):
Awesome, thank you!

RunitaiLinden (commented):
Not a fan of VRED/VGREEN/VBLUE or VX/VY/VZ generally, so if you see 0,1,2,3 indexing, DO NOT substitute these constants.

If you really want to get into making the code more readable and explicit, you should check out how glm does its unions (&foo.x == &foo.r == &foo[0]).

For future work, expect to see more direct usage of glm and less usage of LLVector/Color etc.

PR #1680, opened 2024-06-10
Fix possible null pointer crash

Ansariel:
@marchcat Since Github created the PR for the wrong branch before: More fixes...


marchcat (approved):
Thank you!

PR #1684, opened 2024-06-10
update message_template.msg to agree with master

AndrewMeadows:
The **master-message-template** repo hasn't been updated since 2017... until today. A PR has just been created to update that with a cleaned up copy of what is being used by the server develop branch. This PR aims to provide the same cleaned up version. No messages should have changed, only whitespace (a lot of it) and a few comments.


SaladDais:
FYI this does change messages, but only to add blocks / new messages, that should be safe. no-whitespace diff view is probably best for this: https://github.com/secondlife/viewer/pull/1684/files?diff=unified&w=1.

Also, this should probably use `develop` as the merge target now.

akleshchev (approved):

AndrewMeadows:
Now that the **GameControl** message is in both server and viewer codebases this PR has been updated.  It is ready to go.

BTW, it introduces the Combat message template changes to develop early, but that should be ok.

PR #1685, opened 2024-06-10
Attempted workaround for actions/runner-images#10004 build failures.

brad-linden:


brad-linden:
once I have confirmed that we can produce working builds from this I will probably retarget it at main

brad-linden:
got a successful test build:
https://github.com/secondlife/viewer/releases/tag/Second_Life_Test%232e41f5a-pmfp

.github/workflows/build.yaml
```diff
@@ -312,7 +312,7 @@ jobs:
       AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
       AZURE_TENANT_ID:     ${{ secrets.AZURE_TENANT_ID }}
     needs: build
-    runs-on: windows
+    runs-on: windows-large
```
brad-linden, line 315: may want to leave this change out

nat-goodspeed: ? but if it makes the difference between hanging forever and running, we should keep it.

nat-goodspeed (approved):
I'm good with these changes, although I'll have to take the MUTEX_CONSTRUCTOR one on faith. I just updated the PR from the newly updated main; let's see a clean GH build from that.

.github/workflows/build.yaml
brad-linden: yeah, for now seems required to guarantee getting a runner

brad-linden:
> I'm good with these changes, although I'll have to take the MUTEX_CONSTRUCTOR one on faith. I just updated the PR from the newly updated main; let's see a clean GH build from that.

yeah, my understanding is that that feature is the only ABI difference between the 2 recent builds of msvcrt.dll and adding that `#define` prevents using the ABI breaking features so you can run against either deployed dll

PR #1688, opened 2024-06-10
Re-enable compiler warnings C4127, C4512 & C4706

Ansariel:
Disable particular CRT and WinSock API warnings for functions Microsoft considers unsafe/deprecated

@marchcat And some more changes... 😄


marchcat (approved):
Thank you, @Ansariel!

PR #1714, opened 2024-06-11
Fix some apparent previous merge error

Ansariel:
@marchcat And another fix...


marchcat (approved):
Yeah, auto-merge does that from time to time. Thank you for noticing this!

PR #1715, opened 2024-06-11
Quick fix for a possible merge bug from featurettes.

Geenz:


indra/newview/pipeline.cpp
```diff
@@ -778,12 +778,9 @@ bool LLPipeline::allocateScreenBuffer(U32 resX, U32 resY, U32 samples)
     { // hacky -- allocate auxillary buffer
 
         gCubeSnapshot = TRUE;
-        mReflectionMapManager.initReflectionMaps();
-        mHeroProbeManager.initReflectionMaps();
 
         if (sReflectionProbesEnabled)
         {
-            gCubeSnapshot = TRUE;
```
Ansariel, line 786: Does gCubeSnapshot has to be true for the hero probes? If not it seems to be correct inside the if block.

Geenz: It *should* be fine in this case.  The hero probe manager does a lot of the same stuff as the standard reflection probe manager.

Ansariel: Here it was moved into the if-block: 42623ad81142f93290b31b380cc0f888a1c3c3b6

Geenz: This fix seems to work fine.  I'm not entirely sure why `gCubeSnapshot` was moved into that branch - but that changeset exists prior to the mirrors work being merged into featurettes.  In this case, as initialization for both the reflection probe and hero probe managers is extremely similar and require similar setups all around, I'd argue that this change is likely to be "more correct" in the given context.  Some light refactor work could go in here to make this more clear.  But for the intents and purposes of "fixing a potential merge bug" - I think this suffices.

akleshchev (approved):

indra/newview/pipeline.cpp
marchcat: So should that line just be placed before the if-block?
In any case, this change seems more appropriate for the `develop` branch to avoid more BOOL/bool changes in the future.

Ansariel: I would suggest we go with Geenz's version.

marchcat (approved):

indra/newview/pipeline.cpp
Geenz: Per @marchcat's point, I've changed the target branch to develop.

Issue #1687, opened 2024-06-10
Add support for GLTF KHR_texture_transform


RunitaiLinden:
### Test plan:
1. Download each of the assets from the Test model: https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/TextureTransformTest/glTF
2. Create a box in Second Life.
3. While the box is selected in edit mode, click Develop menu > GLTF > Open....
4. Choose the "TextureTransformTest.gltf" file
5. **Verify** the model appears in-world and the arrows and green boxes match the Screenshot on https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/TextureTransformTest

### Description

See [KHR_texture_transform](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_transform/README.md)

Test model:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/TextureTransformTest

brad-linden:
is this complete?  should it me marked integrated?

katerynamproductengine:
Passed QA.

Verified on the Second Life Release 7.1.9.9620320242 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/266.

PR #1717, opened 2024-06-11
#1687 Add support for KHR_texture_transform

RunitaiLinden:


indra/newview/app_settings/shaders/class1/gltf/pbrmetallicroughnessV.glsl
```diff
@@ -53,13 +54,37 @@ in vec3 normal;
...
+        -sin(Rotation), cos(Rotation), 0,
+        0, 0, 1);
+
+    mat3 scale = mat3(Scale.x,0,0, 0,Scale.y,0, 0,0,1);
+
+    mat3 matrix = translation * rotation * scale;
+
+    vec2 uvTransformed = ( matrix * vec3(texcoord.xy, 1) ).xy;
+
+    uvTransformed.y = 1.0 - uvTransformed.y;
```
cosmic-linden, line 84: I'm pretty sure that the offset of 1 is not needed for pure GLTF. I think the offset used in the PBR-on-Blinn-Phong texture transform is meant to accommodate texture animations. Case in point, in textureUtilV.glsl, I have removed the offset for `terrain_texture_transform`

indra/newview/gltf/buffer_util.h
```diff
@@ -816,6 +816,39 @@ namespace LL
...
+        template<>
+        inline bool copy(const Value& src, vec2& dst)
+        {
+            if (src.is_array())
+            {
+                const boost::json::array& arr = src.as_array();
+                if (arr.size() == 2)
+                {
+                    std::error_code ec;
+                    vec3 t;
```
cosmic-linden, line 829: t is vec3 but dst is vec2

cosmic-linden (approved):

indra/newview/app_settings/shaders/class1/gltf/pbrmetallicroughnessV.glsl
RunitaiLinden: if you mean the 1.0-, it is necessary.  There's a corresponding vertical flip in primitive.cpp.  I'm  trying to keep the UVs in CPU memory in GLTF space and not OpenGL space, and I'm trying to keep the glsl shaders as close to stock GLTF as possible.

indra/newview/gltf/buffer_util.h
RunitaiLinden: good catch, will fix in the next PR

indra/newview/app_settings/shaders/class1/gltf/pbrmetallicroughnessV.glsl
cosmic-linden: What I mean is, this could work instead:

```
uvTransformed.y = -uvTransformed.y;
```

cosmic-linden: Assuming that's correct, looking at `Primitive::prep` : This implies we will also need to remove the uv offsets of 1 there too.

PR #1725, opened 2024-06-12
WIP: UI-related Lua API work

nat-goodspeed:
Add `login.lua`, `logout.lua` with a test script for each. (Add `"userQuit"` operation to `"LLAppViewer"` listener to engage `LLAppViewer::userQuit()`, which is the entry point that pops up "Are you sure?")

Add some `UI.lua` mouse and keyboard functions.

Also add `popup.lua` and `test_popup.lua`. Defend `LLFloaterLUADebug` handler methods against obscure recursive calls via the viewer's UI.

Add `LL.atexit()` entry point and `test_atexit.lua` test script.

Lua "require" modules have been moved to the `indra/newview/scripts/lua/require` subdirectory, and the `require()` logic has been changed to look there for them. This makes it a little easier to identify test scripts versus the modules they're testing.

In that connection, the `LL.source_path()`, `LL.source_dir()` and `LL.abspath()` viewer builtins (all based on `lluau::source_path()`) have been changed to start at the directory in which the top-level script is found, rather than the directory in which the calling module is found.

`mapargs()` is now slightly easier to code: it now accepts a single comma-separated string of `'param,names,...'` as well as the `{'param', 'names', ...}` table form.

In addition to `mapargs()`, this PR introduces another new Python-like wrinkle: `util.classctor()`, whose [explanation is here](https://github.com/secondlife/viewer/commit/18c4dcc5998e061fe3ab54607665c775dd18c826). Most of the Lua module changes are adapting to the slightly simpler class constructor API permitted by `util.classctor()`.

There are a [couple bug fixes](https://github.com/secondlife/viewer/commit/f2020bff30808d28aec06cce5fed61717fcde7fc) in `startup.lua`.

Add an "ok" response to the LUA Debug Console when you execute a LUA string that returns no result. Before, the code would execute with absolutely no response, leaving the user wondering if the Enter key or Execute button had in fact done anything.

On the C++ side, introduce `lua_emplace<T>()` to push a new `userdata` object on the Lua stack containing an instance of C++ class `T`, and `lua_toclass<T>()` to obtain a `T*` pointer from a specified `userdata` object on the Lua stack. A `userdata` object created by `lua_emplace<T>()` is guaranteed to destroy its contained `T` object at script termination. Use `lua_emplace<LuaListener>()` when a Lua script requires a `LuaListener` instance. This means we no longer need the two `LuaState::getListener()` methods, and `LuaListener` is no longer derived from `LLInstanceTracker`. Its constructor no longer needs to invent a random int key.

Add `tempset.h` containing a generic `TempSet` template class. Declaring an instance of `TempSet`, binding a reference to some variable and a desired temporary value, saves the current value, sets the temporary value and, on scope exit (`~TempSet()`), restores the original value.

Try to improve diagnostic output for Lua errors discovered by the `require()` machinery.

`LL.help()` output was failing to identify the function names belonging to some of its descriptions. Fixed.


nat-goodspeed:
The `popup.lua` behavior is a little mysterious. It logs that it's engaging the selected notification, then immediately exits without displaying UI, and a few seconds later crashes the viewer.

Turning on DEBUG level logging for the Lua key clarifies somewhat: the Lua script C++ coroutine calls LLNotifications to put up a modal dialog, but moments later it notices that the Lua script has been looping too many times, and suspends. Somehow this interacts badly with LLNotifications - the Lua script exits, and it appears that something in the viewer is confused.

@maxim-productengine, please help me track down what's going on here? This may be the first case in which LLNotifications puts up a modal dialog on a C++ coroutine other than the main one.

maxim-productengine:
It looks good. So I'm waiting for test_login.lua update before approving. Or is there anything else planned here?
Also I'd like to see a script test example with 'atexit' usage,

nat-goodspeed:
Thanks for reviewing!

I hope to submit this for real later today, with login working properly.

The implementation of `lua_emplace<T>()` calls `LL.atexit()` (from C++) to ensure that the new `T` instance is properly destroyed on script termination. `LuaState::obtainListener()` now uses `lua_emplace<LuaListener>()` to track the `LuaListener` associated with the `lua_State`. So to me, the presence of `~LuaListener()` log messages proves that `LL.atexit()` is working properly.

That said, I'll add a test script to illustrate Lua-side usage.

maxim-productengine (approved):
Ship it!

maxim-productengine (approved):

PR #1727, opened 2024-06-12
Replace LL_X11 and LL_FLTK to LL_LINUX

AiraYumi:
I think LL_FLTK and LL_X11 can be replaced with LL_LINUX since they are currently only used by Linux.


marchcat (approved):
Thank you!

Issue #1611, opened 2024-05-31
[PBR] Regression in anti-flipping mechanism for camera


Martin-Pitt:
### Environment

Second Life Release 7.1.8.9103842320 (64bit) - Materials Featurettes

### Description

The code related to up/down camera limits were changed which causes a regression and causes a error where the user can end up flipping their camera upside down and also causes issues with the limits being applied at the wrong place.

Here is an unlisted video recording reproducing the issue: https://www.youtube.com/watch?v=ObZnP_t8PpM
My tank is a linden vehicle of type car with flags (VEHICLE_FLAG_LIMIT_MOTOR_UP|VEHICLE_FLAG_NO_DEFLECTION_UP that allows for tilting the camera relative to the vehicle.

The relevant commit might be #f815b015cecda18098dd2d16f65682a37e1bff7c after going through previous revisions and trying to find any changes to the camera clamping behaviour. 

### Reproduction steps

1. Vehicle which can tilt your camera, such driving a car or tank in mouselook and you are tilted on a very steep ramp/surface
2. Try to keep looking up until you flip

akleshchev:
Related: https://github.com/secondlife/viewer/pull/1055

akleshchev:
There is a newer fix in [maint-c](https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%23e8de17d-c) which might affect this https://github.com/secondlife/viewer/pull/1327

LLGuru:
Fixed in https://github.com/secondlife/viewer/pull/1740

katerynamproductengine:
Failed QA.

**On Mac:** 
1) There is a [new issue #3357](https://github.com/secondlife/viewer/issues/3357) with camera modes (Mouse Look, Object View, Orbit View) is affecting movement, zoom, and rotation, making the test impossible to execute.
2) There is a [new issue #3359](https://github.com/secondlife/viewer/issues/3359) regarding the rendering of custom objects on macOS is affecting testing, as it prevents visibility of certain objects on the location, including the tanks needed for the Vertical Sim test.

**On Windows:**

**Vertical Sim**: Test **deferred.** Currently, there is no available tank on the location that can be used for the ride. Could you please specify the exact location for testing?

**Kart 1.0:** Test **failed**. After rezzing and sitting in a vehicle with a tilted sit target (e.g., Kart 1.0 from the Library), the camera flips when looking upwards in Mouse Look mode.

[...]

PR #1740, opened 2024-06-12
#1611 Regression in anti-flipping mechanism for mouselook camera

LLGuru:


akleshchev:
I rezzed a Kart 1.0, rode it and when I loke up camera flips to the other orientation. I think that isn't intended?

But Kart and tank are working fine when on a vertical surface. Was able to raise tank's barrel without issues.

akleshchev (approved):

indra/newview/llagent.cpp
```diff
@@ -1459,69 +1459,64 @@ LLVector3 LLAgent::getReferenceUpVector()
...
-            check_viewer_camera = true;
+            // Camera sight relative to agent frame (focus - offset)
+            LLVector3 camera_offset(gAgentCamera.getCameraOffsetInitial());
+            LLVector3 camera_focus(gAgentCamera.getFocusOffsetInitial());
+            LLVector3 camera_sight(camera_focus - camera_offset);
+            // 2D projection of the camera sight to the XZ plane
+            LLVector2 camera_sight_2d_vert(1, camera_sight[VZ]);
+            camera_sight_2d_vert.normalize();
+            // Cosine of the 2D angle between initial camera At and X axis (in the XZ plane)
+            F32 camera_sight_to_at_2d_vert_cos = camera_sight_2d_vert * LLVector2(LLVector3::x_axis);
```
akleshchev, line 1500: I'm not sure XZ plane is the right chose. 'plane' is relative. Ex: If you are flying a vehicle vertically or upside down...

marchcat (approved):

indra/newview/llagent.cpp
LLGuru: I've meant an XZ plane in agent coordinate system, not in world
In the THIRD_PERSON mode the viewer camera and the agent camera have the constant vertical angle between each other
For example, if it is 10 degrees, then when agent looks up 80 degrees over horizont, the viewer camera looks at zenith
This code just calculates this vertical angle between the viewer camera and the agent camera

marchcat (commented):
This looks much cleaner now.

akleshchev:
Personaly I keep a 'do not touch' policy with math classes. They work and they work well and had worked for years, changing anything, especially en masse always has a chance to break something.

marchcat:
Right, it's usually no point in fixing what isn't broken. Not to mention that it can (and often does) introduce some new bugs.

PR #1696, opened 2024-06-11
Remove orphaned code apparently left over from testing

Ansariel:
@RunitaiLinden @brad-linden This seems to be a left-over from when Geenz added mirrors. There is no such a control in the corresponding XUI file.

PR #1746, opened 2024-06-13
Fix a few merge issues

Ansariel:
Fixed a few merge issues and the changes from #1696


brad-linden (approved):
thanks for the contributions!

PR #1747, opened 2024-06-13
Fix for lltracker beacon performance issue

mobserveur:
This fix comes in 2 commits because i made a first fix that was working but i improved it. In the second commit i got rid of the IF DARWIN preprocessor. The beacon is basically a grid that goes up and down (so there are 2 steps). I maintained compatibility with the cheesyBeacon debug option. The number of subdivisions is calculated dynamically depending on the UP or DOWN part of the beacon, and their actual height. The fix also benefits AMD and NVidia gpus. The main original issue was the usage of TRIANGLES_STRIP and the lack of degenerate triangles support in llrender, so i pushed out the begin and end calls (that are managed by llrender to redirect those old immediate draw calls to a more optimized buffer based method), and converted to TRIANGLES (which is not a performance problem here since the grid is rather simple). The original method was causing a flush() after each call, which resulted is atrocious performance on Apple GPU, and also lowered the performance on other gpus (when moving or looking around).

### DIFF
--- a/indra/newview/lltracker.cpp
+++ b/indra/newview/lltracker.cpp
@@ -489,10 +489,21 @@ void draw_shockwave(F32 center_z, F32 t, S32 steps, LLColor4 color)
     gGL.end();
 }
 
+
 void LLTracker::drawBeacon(LLVector3 pos_agent, std::string direction, LLColor4 fogged_color, F32 dist)
 {
+#if LL_DARWIN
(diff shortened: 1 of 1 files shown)


brad-linden (approved):

brad-linden:
thanks for the contribution!  sorry for the mixup on the previous PR #1675

brad-linden:
fixes #999

PR #1767, opened 2024-06-14
Add nearby chat listener

maxim-productengine:
Allow Lua script to listen to the Nearby chat, receiving the message text and other related info.

### DIFF
--- a/indra/newview/llfloaterimnearbychatlistener.cpp
+++ b/indra/newview/llfloaterimnearbychatlistener.cpp
@@ -33,6 +33,7 @@
 
 #include "llagent.h"
 #include "llchat.h"
+#include "llluamanager.h"
 #include "llviewercontrol.h"
 #include "stringize.h"
 
@@ -49,6 +50,35 @@ LLFloaterIMNearbyChatListener::LLFloaterIMNearbyChatListener()
         "[\"channel\"] chat channel number [default = 0]\n"
         "[\"type\"] chat type \"whisper\", \"normal\", \"shout\" [default = \"normal\"]",
         &LLFloaterIMNearbyChatListener::sendChat);
+
+    add("listen",
+        "Start listening to the Nearby chat, chat messages will be resent to the script event pump",
+        &LLFloaterIMNearbyChatListener::listenChat,
+        llsd::map("coro_name", LLSD(), "reply", LLSD()));
+
+    add("stopListening",
+        "Stop listening to the Nearby chat",
+        &LLFloaterIMNearbyChatListener::stopListeningChat,
+        llsd::map("coro_name", LLSD()));
+ 
+    mOutConnection = LLEventPumps::instance().obtain("LLNearbyChat").listen("LLFloaterIMNearbyChatListener", [this](const LLSD &data)
+    {
+        std::map<std::string, std::string> reply_pumps = mReplyPumps;
+        std::map<std::string, std::string> scripts = LLLUAmanager::getScriptNames();
+        for (auto &it : reply_pumps)
+        {
+            //check if listener script is still running
+            if (scripts.find(it.first) != scripts.end())
+            {
+                LLEventPumps::instance().obtain(it.second).post(data);
+            }
+            else
+            {
+                mReplyPumps.erase(it.first);
+            }
+        }
+        return false;
+    });
 }
 
 
@@ -103,3 +133,13 @@ void LLFloaterIMNearbyChatListener::sendChat(LLSD const & chat_data)
                                               gSavedSettings.getBOOL("PlayChatAnim"));
 }
 
+
+void LLFloaterIMNearbyChatListener::listenChat(LLSD const &chat_data)
+{
+    mReplyPumps[chat_data["coro_name"].asString()] = chat_data["reply"].asString();
+}
+
+void LLFloaterIMNearbyChatListener::stopListeningChat(LLSD const &chat_data)
+{
+    mReplyPumps.erase(chat_data["coro_name"].asString());
+}
--- a/indra/newview/llfloaterimnearbychatlistener.cpp
+++ b/indra/newview/llfloaterimnearbychatlistener.cpp
@@ -33,7 +33,6 @@
 
 #include "llagent.h"
 #include "llchat.h"
-#include "llluamanager.h"
 #include "llviewercontrol.h"
 #include "stringize.h"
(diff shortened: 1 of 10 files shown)


indra/newview/llfloaterimnearbychatlistener.cpp
```diff
@@ -49,6 +50,34 @@ LLFloaterIMNearbyChatListener::LLFloaterIMNearbyChatListener()
...
+            {
+                LLEventPumps::instance().obtain(it.second).post(data);
+            }
+            else
+            {
+                mReplyPumps.erase(it.first);
+            }
+        }
+        return false;
+    });
```
nat-goodspeed, line 80: `LLLeapListener` (accessed via the script's "command pump") already supports this functionality via the "listen" and "stoplistening" operations. Since "listen" stores the new connection in the `LLLeapListener` instance itself, and since each `LuaState` has its own `LuaListener`, and each `LuaListener` its own `LLLeapListener` instance, the connection is guaranteed to disconnect on script termination (`~LuaState()`).

indra/newview/llviewermessage.cpp
```diff
@@ -2749,6 +2749,11 @@ void process_chat_from_simulator(LLMessageSystem *msg, void **user_data)
             msg_notify["from_id"] = chat.mFromID;
             msg_notify["source_type"] = chat.mSourceType;
             on_new_message(msg_notify);
+
+
+            msg_notify["chat_type"] = chat.mChatType;
+            msg_notify["message"] = mesg;
+            LLEventPumps::instance().obtain("LLNearbyChat").post(msg_notify);
```
nat-goodspeed, line 2756: Is one of these fields the chat channel? That seems useful for a listener to know.

indra/newview/scripts/lua/LLChatListener.lua
```diff
@@ -0,0 +1,41 @@
...
+local LLChatListener = {}
+local waitfor = {}
+
+function LLChatListener:new()
+    local obj = setmetatable({}, self)
+    self.__index = self
+    obj.name = 'Chat_listener'
+
+    return obj
+end
```
nat-goodspeed, line 13: Heh - I want this to use `util.classctor()`, but that's still isolated on [my lua-login branch](https://github.com/secondlife/viewer/pull/1725/files#diff-867be21d2425e7bb99ec3554b881808693cb4d296bcbd55e3b04d173039f7215). We'll get there.

indra/newview/scripts/lua/LLChatListener.lua
```diff
@@ -0,0 +1,41 @@
...
+
+function LLChatListener:handleMessages(event_data)
+    --print(inspect(event_data))
+    return true
+end
+
+function LLChatListener:start()
+    waitfor = leap.WaitFor:new(-1, self.name)
```
nat-goodspeed, line 21: The fact that `leap.WaitFor` resolves without a `local leap = require 'leap'` statement means I must have inadvertently tossed `leap` into the global namespace. I'd still rather see the `require` statement in this module.

indra/newview/scripts/lua/LLChatListener.lua
```diff
@@ -0,0 +1,41 @@
...
+function LLChatListener:new()
+    local obj = setmetatable({}, self)
+    self.__index = self
+    obj.name = 'Chat_listener'
+
+    return obj
+end
+
+function LLChatListener:handleMessages(event_data)
+    --print(inspect(event_data))
```
nat-goodspeed, line 16: Since the anticipated usage is to override `handleMessages()`, maybe leave this uncommented? If the coder forgets to do that, at least they'll get a visual reminder.

Oh, I see: your example override method forwards the call to this base-class method. Okay.

indra/newview/scripts/lua/LLChatListener.lua
```diff
@@ -0,0 +1,41 @@
...
+    return true
+end
+
+function LLChatListener:start()
+    waitfor = leap.WaitFor:new(-1, self.name)
+    function waitfor:filter(pump, data)
+        return data
```
nat-goodspeed, line 23: Using `LLLeapListener`'s "listen" operation, `pump` should arrive as `"LLNearbyChat"`. Your `filter()` function should return only those -- otherwise your chat listener will echo all incoming events, e.g. user floater clicks, ...

indra/newview/scripts/lua/LLChatListener.lua
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    fiber.launch(self.name, function()
+        event = waitfor:wait()
+        while event and self:handleMessages(event) do
+          event = waitfor:wait()
+        end
+    end)
```
nat-goodspeed, line 31: Maybe this pattern should be encapsulated in a `leap` convenience method. I'll think about that.

nat-goodspeed (changes requested):

indra/newview/llfloaterimnearbychatlistener.cpp
maxim-productengine: I see, I totally missed that and tried to reinvent the wheel.

indra/newview/llviewermessage.cpp
maxim-productengine: As far as I know, only channel 0 messages are sent to the avatars. All other channels can only be listened by lsl scripts.

indra/newview/scripts/lua/LLChatListener.lua
maxim-productengine: Agree

indra/newview/scripts/lua/LLChatListener.lua
maxim-productengine: Uncommented and updated example script too.

indra/newview/llfloaterimnearbychatlistener.cpp
nat-goodspeed: Part of the problem here is that I have never written any documentation about the available `LLEventAPI` listeners and the operations they support, so discoverability is low. In my defense, I always expected the suite of listeners and their operations to keep growing (and to diverge across branches and viewer forks), so that the only reliable information is really the source code. That's why every listener and every operation is defined with queryable metadata.

nat-goodspeed (approved):

PR #1779, opened 2024-06-14
Example / proof of concept viewer code using IM history

simon-linden:
This is sample / proof-of-concept code and not a full design or implementation, thus a draft PR for now.

Shows how the viewer can use the "ChatSessionRequest" cap and new "fetch im history" command to query the region and SL back-end for the chat history in a direct IM session.


vir-linden (commented):
I would recommend finding an alternate name to session_id. We already use session_id as an identifier for a specific viewer session, which is unrelated to the usage here. Maybe chat_id or something? Whatever we pick, should look for conflicting existing usages.

simon-linden:
> I would recommend finding an alternate name to session_id. We already use session_id as an identifier for a specific viewer session, which is unrelated to the usage here. Maybe chat_id or something? Whatever we pick, should look for conflicting existing usages.

The viewer actually uses 'session_id' in this context already, such as functions like bool LLIMModel::newSession(const LLUUID& session_id, ...      "session_id" shows up in various forms almost 350 times in llimview.cpp so it's usage here is following both viewer and server conventions.

Coming at the term without the context of IM sessions is definitely confusing, but I think that ship has sailed

simon-linden:
> > This is sample / proof-of-concept code and not a full design or implementation, thus a draft PR for now.
> 
> Is this supposed to have a viewer issue or be assigned to a specific project? It hangs in a void with no dependencies to indicate rediness state or it's future.

Understood, it is in limbo right now.   I talked with Vir about it Wednesday to he's up to speed with the work.   It will need some work with Product and UI to sort out design and backwards-compatible issues before we know what we really want to do with the desktop viewer and adding the feature.

akleshchev:
Thanks for the info, was just confused about it. I suggest at least assigning a milestone or project.

Issue #1080, opened 2024-03-28
BugSplat Crash: Second Life Release!LLAppViewer::initStrings()::$_84::operator()(int*) const(3014)


sl-service-account:
**BugSplat** [Key Crash 91971](https://app.bugsplat.com/v2/keycrash?stackKeyId=91971&database=SecondLife_Viewer_2018)

**Some Customer Responses:**
Log:
2024-03-28T10:47:09Z ERROR # newview/llappviewer.cpp(3012) initStrings : Viewer failed to find localization and UI files. Please reinstall viewer from https://secondlife.com/support/downloads and contact https://support.secondlife.com if issue persists after reinstall.<br/>

**Sample Version:**
7.1.5.8443777128

**Sample Callstack:**
[...]

PR #1782, opened 2024-06-17
#1080 BugSplat Crash: LLAppViewer::initStrings() - show crash reason

LLGuru:
The root cause seems to be on the client side

For helping the user to understand the actual crash reason
we could include it's short description into the crash message


akleshchev (approved):
There is no point to call `LL_WARNS() << crash_reason << LL_ENDL;`, LL_ERRS will already print that to logs.

P.S. Also might be a good idea to remove 'Please reinstall' part of the LL_ERRS, user doesn't see that unless user goes to check logs, showMissingFiles() already asks user to reinstall.

akleshchev:
>  Also might be a good idea to remove 'Please reinstall' part of the LL_ERRS

Thanks. I added that part a long time ago when I though that user is capable of seeing those LL_ERRS 😅, but it's invisible to users.

PR #1811, opened 2024-06-20
Revert LLLuaFloater "idle" events in favor of Lua timers.Timer().

nat-goodspeed:


maxim-productengine (approved):
I like that moving to new timers requires minimal changes.
Also LLNotifications.lua will be redundant because we'll have popup.lua

nat-goodspeed:
Yes, I figure that once lua-speedometer-demo is merged into release/luau-scripting, I'll merge it into lua-login and then replace LLNotification.lua with popup.lua there.

PR #1819, opened 2024-06-21
brad/tag release

brad-linden:
Allow manually dispatched workflow form to create tags for releases.

Screenshot 2024-06-20 at 5 22 51 PM


brad-linden:
this isn't quite working yet.  It seems that it can create tags, but those tags don't trigger a build because they aren't "pushed"

nat-goodspeed:
> this isn't quite working yet.

Still true, or resolved?

nat-goodspeed (approved):
I'm good with this.

Can `build.yaml` triggers be extended to notice the creation of a new tag?

brad-linden:
> Can `build.yaml` triggers be extended to notice the creation of a new tag?

yes.  I have begun researching `workflow_run` triggers which I think can be used for that.  I'll do that experiment under another PR

PR #1848, opened 2024-06-26
viewer#1666 Allow switching 2k textures off in bulk uploads

akleshchev:


indra/newview/llfloaterbulkupload.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (!mAllow2kTextures)
+    {
+        new_height -= mWarningPanel->getRect().getHeight();
+    }
+    reshape(getRect().getWidth(), new_height, false);
+}
+
+void LLFloaterBulkUpload::onUpload2KCheckBox()
+{
+    mAllow2kTextures = !mCheckboxUpload2K->getValue().asBoolean();
```
maxim-productengine, line 114: Is it intended that "BulkUpload2KTextures" debug setting is not updated?

marchcat (approved):

indra/newview/llfloaterbulkupload.cpp
akleshchev: Fixed, thanks!

maxim-productengine (approved):

PR #1849, opened 2024-06-26
Increase texture discard bias if system memory gets low

Ansariel:
Requested by @RunitaiLinden : Increase texture discard level if system memory gets low. However, this might need an additional change from @beqjanus so it works at least on macOS the same way: https://github.com/FirestormViewer/phoenix-firestorm/commit/3b074ba4af5e303125db606dd69eb4282a91f957


brad-linden (approved):
Looks good.  Thanks!

PR #1855, opened 2024-06-27
realign system ram functions

beqjanus:
Related commit to Ansariel's PR.
make the system ram function align across all supported platforms. Taken from https://github.com/FirestormViewer/phoenix-firestorm/commit/3b074ba4af5e303125db606dd69eb4282a91f957
+ clean up FS specific comment markers and upstream code retention


beqjanus:
this PR has pulled in some commits that should be on a separate PR. I'll try to cancel this and start over

PR #1878, opened 2024-06-28
Add LuaAutorunPath, LuaCommandPath and LuaRequirePath settings.

nat-goodspeed:
Remove AutorunLuaScriptName and the `LLLUAmanager::runScriptOnLogin()` method that checked it.

Instead, iterate over LuaAutorunPath directories at viewer startup, iterate over *.lua files in each and implicitly run those.

Make require() implementation honor LuaRequirePath.

I wasn't intending to implement LuaCommandPath in this PR.

### DIFF
--- a/indra/llcommon/lua_function.h
+++ b/indra/llcommon/lua_function.h
@@ -118,11 +118,12 @@ private:
 *   LuaPopper
 *****************************************************************************/
 /**
- * LuaPopper is an RAII struct whose role is to pop some number of entries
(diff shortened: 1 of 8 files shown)


nat-goodspeed:
Best to review with "ignore whitespace changes": I eliminated tab characters in settings.xml. If that's a problem, I can reapply the substantive deltas to an unmodified settings.xml.

maxim-productengine (approved):

PR #1891, opened 2024-07-01
Loop fixes

Ansariel:
LLFlatListView::detachItems was changed a while ago and now had a scoping issue for the "iter" variable, so I changed that and all the other strange loops introduced in the original commit.

Also removed the include of the format library, which a) isn't used, b) creates a compiler warning and c) can't be used without C++20 support anyway


marchcat (approved):
Thank you!

marchcat:
Please update the source branch to pass the pre-commit check.

Issue #1111, opened 2024-04-02
Remove xmlrpc-epi


bennettgoble:
The xmlrpc-epi library is 15 years old and only used in two small areas: login and when buying land. Let's replace it with basic HTTP calls so that we no longer have to maintain it.

## Test Plan

Run test plans for the following:

- Login behavior
- Buying land flows

bennettgoble:
Removing this dependency will simplify updating expat --as xmlrpc-epi is a dependent.

akleshchev:
We can implement a LLSDSerialize::toXMLRPC() instead, but won't it be better to just get rid of XMLRPC entirely? What benefits does it have over LLSD or JSON that cause viewer<-> to use XMLRPC?

bennettgoble:
@akleshchev I agree, it would be better to eliminate the need to do XML-RPC completely. I wonder if we add an alternative login method that used a basic JSON or LLSD POST over HTTP. We'd need to keep XML-RPC on the server side to support existing viewers, but at least the XML-RPC behavior could be eliminated from the client.

akleshchev:
Created https://github.com/secondlife/server/issues/1195

brad-linden:
I'm not sure that login is the only xmlrpc dependency, aren't there also some buy currency operations that go to an xmprpc endpoint?  probably will want something like what @akleshchev mentiond ( https://github.com/secondlife/viewer/issues/1111#issuecomment-2045691021 ) and serialize "RPC structured" llsd to xml and then manually send HTTP post to the endpoint

LLGuru:
Done in https://github.com/secondlife/viewer/pull/1893

akleshchev:
Fixing issue's state since it didn't go through testing (not that there is much to test)

Test plan:
1. Confirm login orks
2. Confirm buying land works

igorlproductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10516785745 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/316.

PR #1893, opened 2024-07-01
#1111 Remove xmlrpc-epi

LLGuru:


indra/llcommon/llsd.cpp
```diff
@@ -872,6 +916,155 @@ const LLSD::Binary& LLSD::asBinary() const  { return safe(impl).asBinary(); }
 
 const LLSD::String& LLSD::asStringRef() const { return safe(impl).asStringRef(); }
 
+LLSD::String LLSD::asXMLRPCValue() const { return "<value>" + safe(impl).asXMLRPCValue() + "</value>"; }
```
akleshchev, line 919: This is a wrong place for this logic, I suggest following LlsdFromJson and LlsdToJson example and move this to a separate llsdxmlrpc file or follow LLSDXMLParser's example and do an LLSDXMLRPCParser.

indra/llfilesystem/lldir.cpp
```diff
@@ -864,6 +864,25 @@ std::vector<std::string> LLDir::findSkinnedFilenames(const std::string& subdir,
     return results;
 }
 
+std::string LLDir::getFileContents(const std::string& filename) const
```
akleshchev, line 867: This is a very confusing name and in a confusing place. How is content of LLFile is related to LLDir?
Suggestion: some variant of LLFile::fileAsString(filename)?

akleshchev (approved):
Looks good, but I strongly suggest to rename and move that file reading related code out of directory code. Keep file reading to files.

indra/llfilesystem/lldir.cpp
akleshchev: Also could have easily been static instead of needing gDirUtilp instance.

indra/llcommon/llsd.cpp
LLGuru: I would like to re-use the existing type hierarchy implemented in LLSD instead of creating the mirror outside

akleshchev: But it isn't a type, it's xml parsing. 

P.S. I think llsd files are shared between viewer and server so it might be better to minimize changes.

marchcat (approved):

indra/newview/llcurrencyuimanager.cpp
```diff
@@ -160,32 +161,29 @@ void LLCurrencyUIManager::Impl::updateCurrencyInfo()
...
-    // With GitHub builds, the build number is too big to fit in a 32-bit int,
-    // and XMLRPC_VALUE doesn't deal with integers wider than int. Use string.
-    keywordArgs.appendString("viewerBuildVersion", stringize(LLVersionInfo::instance().getBuild()));
-
-    LLXMLRPCValue params = LLXMLRPCValue::createArray();
-    params.append(keywordArgs);
+    const LLVersionInfo& vi(LLVersionInfo::instance());
+
+    LLSD params = LLSD::emptyMap()
+        .with("agentId", gAgent.getID().asString())
```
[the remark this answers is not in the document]
LLGuru: Done

maxim-productengine (approved):

maxim-productengine:
Looks good. But I'd like to be see mac build finished successfully.

PR #1903, opened 2024-07-02
Reduce string temporaries from frequently called llui find and get functions

RyeMutt:
This set of changes reduces the number of string temporaries created from calling various find and get functions in llui by modernizing to use std::string_view


marchcat (approved):
Thank you!

PR #1912, opened 2024-07-02
viewer#1666 Allow switching 2k textures off in bulk uploads

akleshchev:
Reposting with an adjustment from steeltoe.


marchcat (approved):
🚢 it!

PR #1929, opened 2024-07-04
SL-18721 Move window shutdown further down

akleshchev:
Multiple people have deadlocks. According to Ansariel's callstack it deadlocks at http, so moved this to be after http gets stopped (and after some of the caching happens)


marchcat (approved):

maxim-productengine (approved):

Ansariel:
Seems good so far - didn't get a stuck shutdown in like 10 attempts yet. I am just wondering where these arbitrary 200ms sleep come from?

akleshchev:
Ugh... that was for testing, forgot to remove it.

PR #1930, opened 2024-07-05
Introduce move support and string_view to LLSD

RyeMutt:
This set of changes introduces move operator and constructors as well as string_view support to LLSD to reduce temporaries


marchcat (approved):
Thank you!

Issue #1931, opened 2024-07-05
BugSplat Crash: SecondLifeViewer!LLRenderTarget::release(387)


sl-service-account:
**BugSplat** [Key Crash 74323](https://app.bugsplat.com/v2/keycrash?stackKeyId=74323&database=SecondLife_Viewer_2018)

**Sample Version:**
7.0.0.578792 or 7.0.0.578161

**First Report Date:**
03-30-2024

**Sample Callstack:**
<table><tr><td>SecondLifeViewer!LLRenderTarget::release</td><td>d:/work/viewer/viewer_w64/latest/indra/llrender/llrendertarget.cpp(387)</td></tr><tr><td>SecondLifeViewer!LLRenderTarget::~LLRenderTarget</td><td>d:/work/viewer/viewer_w64/latest/indra/llrender/llrendertarget.cpp(74)</td></tr><tr><td>SecondLifeViewer!__ArrayUnwind</td><td>d:/agent/_work/2/s/src/vctools/crt/vcstartup/src/eh/ehvecdtr.cpp(119)</td></tr><tr><td>SecondLifeViewer!``eh vector destructor iterator''::`1'::fin$0</td><td></td></tr><tr><td>VCRUNTIME140!__C_specific_handler</td><td>D:/a/_work/1/s/src/vctools/crt/vcruntime/src/eh/riscchandler.cpp(443)</td></tr><tr><td>ntdll!RtlpExecuteHandlerForUnwind</td><td></td></tr><tr><td>ntdll!RtlUnwindEx</td><td></td></tr><tr><td>ntdll!_C_specific_handler</td><td></td></tr><tr><td>ntdll!RtlpExecuteHandlerForException</td><td></td></tr><tr><td>ntdll!RtlDispatchException</td><td></td></tr><tr><td>ntdll!KiUserExceptionDispatch</td><td></td></tr></table>

LLGuru:
- Fixed by https://github.com/secondlife/viewer/pull/1933

AtlasLinden:
Reopening for testing

valeriiaaproductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10516785745 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/316.

PR #1933, opened 2024-07-05
#1931 BugSplat Crash: SecondLifeViewer!LLRenderTarget::release(387)

LLGuru:


akleshchev:
Shouldn't this trigger an LL_ERRS or something similar? Looks like app can't continue.

marchcat (commented):
llappviewer.cpp part should probably have some comments.

LLGuru:
> Shouldn't this trigger an LL_ERRS or something similar? Looks like app can't continue.

I believe all works fine

All crash reports contain the logs with the last line "WinMain : Application init failed."
https://app.bugsplat.com/v2/keycrash?database=SecondLife_Viewer_2018&stackKeyId=74323

So we know that init() returns unsucceeded, but we don't know which part is done
Cancelling of the problem method running on vector destroy seems to be enough

akleshchev (approved):

marchcat (approved):

PR #1934, opened 2024-07-05
Add GameControl behind feature-flag

AndrewMeadows:
This PR adds the "experimental" GameControl project but leaves it behind a feature flag (menu option) and is disabled by default.


Ansariel:
May I ask the question how a feature flag protects from potential catastrophic bugs caused by merging features in project state?

AndrewMeadows:
@Ansariel the "put it behind a feature flag" practice doesn't fully protect against "catstrophic bugs" but it allows for a easier/faster development process while providing some protections against bugs.  It is part of our transition to a [gitflow development strategy](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow).

Rather than cut a project branch and allow it to mature for a long time in its own corner of the tree until it is "done" we instead make progress with smaller deliverable changes that can be merged back into the **develop** branch sooner rather than later.  Most projects need to be put behind a "feature flag" (a switch of some sort that can be used to enable/disable it) to reduce the likelihood the changes will interfere with normal operation.  Such features can be enabled for testing, and will be more easily accessible to anyone who wants to test it.  It is especially important to put **GameControl** behind a feature flag because it is a WIP (work in progress) and it will eventually replace the legacy game controller support provided by the **ndof** driver used for **3DConnexion** devices (**SpaceMouse**, etc).

Our transition to workflow is newer than the **GameControl** project, which is why it has lived in its own branch for so long and why the preparation of this PR was so complicated (lots of conflicts during the last rebase).  Going forward we can expect smaller PRs and an easier dev process.

The "feature flag" isn't the only protection we have against "catastrophic bugs".  We also have a review process for PRs like this and we will be testing a pre-release installable that will be associated with this PR before merging it.

Ansariel:
Oh I was just asking because this more or less looks like going back to how it was in the past - and where pathfinding stalled development for a long time since it caused a serious bug in the development branch. As a result, the development was split into the way it was until now. The "new" strategy however exposes the same risk as years before again.

AndrewMeadows:
If **Pathfinding** had been properly isolated by a feature flag then it could have been workflowed into the codebase in a more incremental way without stalling development.  That is the theory anyway; the word "properly" is doing a lot of work in that statement.  In any case, we will try workflow for a while to see how it works for us.

indra/llwindow/llkeyboardsdl.cpp
```diff
@@ -24,7 +24,7 @@
  * $/LicenseInfo$
  */
 
-#if LL_SDL
+#if LL_LINUX
```
AndrewMeadows, line 27: It used to be the case only the linux build used **SDL**, but going forward with **SDL2** it is used for ALL platforms (well... except for DARWIN which restricts game controllers at the OS level), so I'm replacing `LL_SDL `with `LL_LINUX`.

indra/llwindow/llwindow.cpp
```diff
@@ -27,14 +27,14 @@
...
-#include "llwindowsdl.h"
-#elif LL_WINDOWS
+#if LL_WINDOWS
 #include "llwindowwin32.h"
 #elif LL_DARWIN
 #include "llwindowmacosx.h"
+#elif LL_MESA_HEADLESS
+#include "llwindowmesaheadless.h"
+#elif LL_LINUX
+#include "llwindowsdl.h"
```
AndrewMeadows, line 37: I'm sorting the OS types: WINDOWS, DARWIN, MESA, LINUX.  Here and other spots.

indra/newview/llviewerjoystick.cpp
```diff
@@ -314,18 +316,11 @@ void LLViewerJoystick::HotPlugRemovalCallback(NDOF_Device *dev)
 
 // -----------------------------------------------------------------------------
 LLViewerJoystick::LLViewerJoystick()
-:   mDriverState(JDS_UNINITIALIZED),
-    mNdofDev(NULL),
-    mResetFlag(false),
-    mCameraUpdated(true),
-    mOverrideCamera(false),
-    mJoystickRun(0)
```
AndrewMeadows, line 322: These moved to implicit initializations.

indra/newview/llviewermenu.cpp
```diff
@@ -9779,6 +9808,9 @@ void initialize_menus()
     view_listener_t::addMenu(new LLAdvancedToggleFeature(), "Advanced.ToggleFeature");
     view_listener_t::addMenu(new LLAdvancedCheckFeature(), "Advanced.CheckFeature");
 
+    view_listener_t::addMenu(new LLAdvancedToggleExperiment(), "Advanced.ToggleExperiment");
+    view_listener_t::addMenu(new LLAdvancedCheckExperiment(), "Advanced.CheckExperiment");
+
```
AndrewMeadows, line 9813: I'm adding an "Experimental Features" category in the Advanced menu.  So far it has only one item: "GameControl".

indra/newview/llviewermessage.cpp
```diff
@@ -3139,87 +3139,77 @@ void process_crossed_region(LLMessageSystem* msg, void**)
...
 
 
-
-// Sends avatar and camera information to simulator.
-// Sent roughly once per frame, or 20 times per second, whichever is less often
-
-const F32 THRESHOLD_HEAD_ROT_QDOT = 0.9997f;    // ~= 2.5 degrees -- if its less than this we need to update head_rot
-const F32 MAX_HEAD_ROT_QDOT = 0.99999f;         // ~= 0.5 degrees -- if its greater than this then no need to update head_rot
-                                                // between these values we delay the updates (but no more than one second)
-
```
AndrewMeadows, line 3149: The logic around when to send an AgentUpdate message was split between llappviewer.cpp::idle() and this file.  I'm moving all of the decision logic out of llappviewer.cpp and into the implementation of send_agent_update().  Basically the logic can be summed up as: **send if we need a keep-alive message OR (we are not throttled AND something changed)**.

indra/newview/llviewermessage.cpp
```diff
@@ -3228,171 +3218,150 @@ void send_agent_update(bool force_send, bool send_reliable)
...
-        }
-//      if (drag_delta_chg.magVec() > TRANSLATE_THRESHOLD)
-//      {
-//          LL_INFOS("Messaging") << "drag delta " << drag_delta_chg.magVec() << LL_ENDL;
-//      }
-        if (control_flag_change)
-        {
-            LL_INFOS("Messaging") << "msg " << msg_number << ", frame " << LLFrameTimer::getFrameCount() << ", dcf = " << control_flag_change << LL_ENDL;
-        }
-*/
```
AndrewMeadows, line 3280: This commented out debugging logic is removed for cruft.

indra/llui/llscrolllistctrl.cpp
```diff
@@ -1741,6 +1712,16 @@ void LLScrollListCtrl::setEnabled(bool enabled)
...
+
+    // when the table is disabled also disable its items
+    for (LLScrollListItem* item : mItemList)
+    {
+        item->setEnabled(enabled);
+        if (!enabled)
+        {
+            item->setSelected(false);
+        }
+    }
```
AndrewMeadows, line 1724: This is new behavior across all tables in the UI: when the table itself is disabled/enabled then so are the items in the table.

indra/newview/llagentcamera.cpp
```diff
@@ -2204,10 +2204,6 @@ void LLAgentCamera::changeCameraToMouselook(bool animate)
         mCameraMode = CAMERA_MODE_MOUSELOOK;
         const U32 old_flags = gAgent.getControlFlags();
         gAgent.setControlFlags(AGENT_CONTROL_MOUSELOOK);
-        if (old_flags != gAgent.getControlFlags())
-        {
-            gAgent.setFlagsDirty();
-        }
```
AndrewMeadows, line 2210: The "dirty flag" info in `LLAgent `has been removed.  It is no longer helpful.  Instead, the `send_agent_update()` logic checks the current value against the last one sent to decide if something changed.

indra/newview/llfloaterjoystick.cpp
```diff
@@ -78,15 +79,22 @@ BOOL CALLBACK di8_list_devices_callback(LPCDIDEVICEINSTANCE device_instance_ptr,
...
         std::string product_name = utf16str_to_utf8str(llutf16string(device_instance_ptr->tszProductName));
-        S32 size = sizeof(GUID);
-        LLSD::Binary data; //just an std::vector
-        data.resize(size);
-        memcpy(&data[0], &device_instance_ptr->guidInstance /*POD _GUID*/, size);
-
-        LLFloaterJoystick * floater = (LLFloaterJoystick*)pvRef;
-        LLSD value = data;
-        floater->addDevice(product_name, value);
+        bool include_device = LLViewerJoystick::is3DConnexionDevice(product_name) || !LLGameControl::isEnabled();
```
AndrewMeadows, line 85: When `LLGameControl::isEnabled()` we  disable the generic game-controller support portion of the old LLViewerJoystick stuff, because otherwise there would be a feature conflict when trying to drive the avatar or camera using a game controller.

indra/llwindow/llgamecontrol.cpp
```diff
@@ -0,0 +1,2101 @@
...
+    g_manager.setFlycamMappings(flycamMappings);
+
+    // Load device-specific settings
+    g_deviceOptions.clear();
+    LLSD options = s_loadObject(SETTING_KNOWNCONTROLLERS);
+    for (auto it = options.beginMap(); it != options.endMap(); ++it)
+    {
+        g_deviceOptions.emplace(it->first, it->second);
+    }
+    g_manager.loadDeviceOptionsFromSettings();
```
akleshchev, line 2083: might also be a good idea to subscrive to changes

akleshchev (approved):

marchcat (approved):

indra/newview/llfloaterpreference.cpp
```diff
@@ -632,13 +631,12 @@ void LLFloaterPreference::cancel(const std::vector<std::string> settings_to_skip
...
-        iter != tabcontainer->getChildList()->end(); ++iter)
+    for (LLView* view : *tabcontainer->getChildList())
     {
-        LLView* view = *iter;
-        LLPanelPreference* panel = dynamic_cast<LLPanelPreference*>(view);
-        if (panel)
-            panel->cancel(settings_to_skip);
+        if (LLPanelPreference* panel = dynamic_cast<LLPanelPreference*>(view))
+        {
+            panel->cancel();
```
maxim-productengine, line 638: Probably `settings_to_skip` is missed here.

maxim-productengine (approved):

indra/llui/lltabcontainer.cpp
```diff
@@ -2173,32 +2173,23 @@ S32 LLTabContainer::getTotalTabWidth() const
...
             break;
         }
     }
-
-    if( foundTab )
-        this->setVisible( true );
-    else
-        this->setVisible( false );
-
-    updateMaxScrollPos();
```
AndrewMeadows, line 2203: LLTabContainer::setTabVisibility() was walking mTabList twice when it only needed to walk it once, so I cleaned up its implementation.  I added one slight change in behavior where it first checks to see if the visibility is actually changing before applying the setting.  Note: this subtly avoids some side-effects of the `this->selectTab(); this->setVisible(true);` where UI which hasn't actually changed can nevertheless become selected/focused.  I don't expect this to cause any UX problems.

Ansariel:
It's probably better to merge this into `release/maint-b` because of the Linux/SDL2 changes in there btw

AndrewMeadows:
I was just chatting with Brad today about "Where is the Linux viewer?" and learned about **maint-b**.  My view is... whoever merges first into **develop** wins and the loser must resolve the conflicts.  That said, I don't want to make someone else sort through my conflicting code and chance them getting it wrong.  I will try to take a look at **maint-b** to form an opinion about what path might be easiest.

indra/llui/lltexteditor.h
```diff
@@ -313,12 +313,11 @@ class LLTextEditor :
     template <typename STRINGTYPE>
     void            pasteTextWithLinebreaks(const STRINGTYPE& clean_string)
     {
-        pasteTextWithLinebreaks<LLWString>(ll_convert(clean_string));
+        pasteTextWithLinebreaksInternal(ll_convert(clean_string));
     }
-    template <>
-    void            pasteTextWithLinebreaks<LLWString>(const LLWString & clean_string);
```
AndrewMeadows, line 319: This "explicit specialization in non-namespace context" does not build on linux.  My solution is to supply an explicit internal method for handling all input converted to `LLWString`.

indra/llwindow/llkeyboardheadless.h
```diff
@@ -33,20 +33,15 @@ class LLKeyboardHeadless : public LLKeyboard
...
-#ifndef LL_SDL
-    /*virtual*/ bool    handleKeyUp(const U16 key, MASK mask) { return false; }
-    /*virtual*/ bool    handleKeyDown(const U16 key, MASK mask) { return false; }
-#else
-    /*virtual*/ bool    handleKeyUp(const U32 key, MASK mask) { return false; }
-    /*virtual*/ bool    handleKeyDown(const U32 key, MASK mask) { return false; }
-#endif
-    /*virtual*/ void    resetMaskKeys();
-    /*virtual*/ MASK    currentMask(bool for_mouse_event);
-    /*virtual*/ void    scanKeyboard();
```
AndrewMeadows, line 47: The `/*virtual*/` comments here are artifacts from the olden days before the `override` keyword was added to C++.  They were used to signal the reader that these were virtual methods in the base class that the derived class was trying to override.  These days, please use the `override` keyword because it allows the compiler to enforce the override signature is correct and corresponds to a base method that really is virtual.

indra/llwindow/llwindowsdl.cpp
```diff
@@ -633,64 +634,17 @@ void LLWindowSDL::tryFindFullscreenSize( int &width, int &height )
...
-    LL_INFOS() << " Running against SDL "
-               << int(r_sdl_version.major) << "."
-               << int(r_sdl_version.minor) << "."
-               << int(r_sdl_version.patch) << LL_ENDL;
-
-    if (width == 0)
-        width = 1024;
-    if (height == 0)
-        width = 768;
-
```
AndrewMeadows, line 693: We are now using **SDL2** on ALL platforms rather than just linux (for game controller input) so this `SDL_Init` logic has been moved to **llsdl.cpp**, out of this `#ifdef LL_LINUX` context.

indra/llwindow/llwindow.h
```diff
@@ -63,16 +63,16 @@ class LLWindow : public LLInstanceTracker<LLWindow>
...
+    virtual bool getMinimized() const = 0;
+    virtual bool getMaximized() const = 0;
     virtual bool maximize() = 0;
     virtual void minimize() = 0;
     virtual void restore() = 0;
-    bool getFullscreen()    { return mFullscreen; };
-    virtual bool getPosition(LLCoordScreen *position) = 0;
-    virtual bool getSize(LLCoordScreen *size) = 0;
-    virtual bool getSize(LLCoordWindow *size) = 0;
+    virtual bool getFullscreen() const { return mFullscreen; };
```
AndrewMeadows, line 72: Note: `getFullscreen()` was incorrectly left non-virtual.  I fixed it and removed the confuzled comment about it in **llwindowheadless.h**.

indra/llwindow/llgamecontrol.cpp
```diff
@@ -0,0 +1,2126 @@
...
+#include "SDL2/SDL_joystick.h"
+
+#include "indra_constants.h"
+#include "llfile.h"
+#include "llgamecontroltranslator.h"
+#include "llsd.h"
+
+namespace std
+{
+    string to_string(const char* text)
```
brad-linden, line 44: i believe this duplicates `ll_safe_string()` do we really want this?  is it just so we can return literals below?  wouldn't it be better to just return string object literals?

brad-linden (approved):
I'm fine with merging this as is.  but have some thoughts about future improvements

indra/llui/llcombobox.cpp
```diff
@@ -256,6 +251,11 @@ bool LLComboBox::itemExists(const std::string& name)
     return mList->getItemByLabel(name);
 }
 
+std::vector<LLScrollListItem*> LLComboBox::getAllData() const
+{
+    return mList->getAllData();
```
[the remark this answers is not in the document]
AndrewMeadows: Good question.  I performed an audit and learned:

`LLComboBox::mList` is set in the **ctor** via `LLUICtrlFactory::create<LLScrollListCtrl>(params)` which eventually invokes `LLUICtrlFactory::createWidgetImpl<LLScrollListCtrl>(params)` which never returns `nullptr`.  On failure it provides an empty list.

Meanwhile, there are a few places in `LLComboBox` where we check for null `mList`.  Those checks should probably be removed since it will never be null and it misleads future readers by suggesting it could be.

indra/llui/lltexteditor.h
AndrewMeadows: That may be the case but but I think the explicit impl is easier to read/understand.

indra/llwindow/llkeyboard.h
```diff
@@ -55,10 +55,12 @@ class LLWindowCallbacks;
 class LLKeyboard
 {
 public:
-#ifndef LL_SDL
```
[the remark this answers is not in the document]
AndrewMeadows: All platforms now compile with SDL, so LL_SDL would effectively always be true, however we don't necessarily want to use U32 for the NATIVE_KEY_TYPE in all platforms.  If any Mac build wants to use SDL for keyboard input then they'll have to tweak their code to accommodate under the new paradigm.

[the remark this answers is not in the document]
brad-linden: I don't understand why that is preferable to simply continuing to let `LL_SDL` represent what we use it for:  is SDL the provider for our LLWindow implementation and non-game related input events.

indra/newview/llfloaterpreference.cpp
AndrewMeadows: Thanks, good catch.

indra/llwindow/llkeyboard.h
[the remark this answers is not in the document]
AndrewMeadows: It used to be the case that SDL was not a dependency for all platforms, so `LL_SDL` would only be defined for some but not all.  Going forward SDL is a dependency on all platforms so `LL_SDL` is always defined and would no longer provide a real preprocessor logic fork.  If preprcessor forking is necessary then it should pivot on a variable name that more clearly conveys the condition that matters (for example: USE_SDL_FOR_KEYPRESS_EVENTS).

indra/newview/llappviewer.cpp
```diff
@@ -1135,6 +1138,15 @@ bool LLAppViewer::init()
...
     }
 
+    LLGameControl::init(gDirUtilp->getExpandedFilename(LL_PATH_APP_SETTINGS, "gamecontrollerdb.txt"),
+        [&](const std::string& name) -> bool { return gSavedSettings.getBOOL(name); },
+        [&](const std::string& name, bool value) { gSavedSettings.setBOOL(name, value); },
+        [&](const std::string& name) -> std::string { return gSavedSettings.getString(name); },
+        [&](const std::string& name, const std::string& value) { gSavedSettings.setString(name, value); },
+        [&](const std::string& name) -> LLSD { return gSavedSettings.getLLSD(name); },
+        [&](const std::string& name, const LLSD& value) { gSavedSettings.setLLSD(name, value); },
+        [&]() { LLPanelPreferenceGameControl::updateDeviceList(); });
```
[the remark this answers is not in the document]
AndrewMeadows: I agree.  This was introduced by another developer.  I already tried to clean it up but then got lost and ended up with something that didn't work, so I abandoned  the effort.  I will add a TODO comment.

indra/llwindow/llkeyboard.h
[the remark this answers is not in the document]
brad-linden: yeah that makes sense.  probably something like `LL_USE_SDL_WINDOW` I think it is required that the LLWindow implementation be tightly coupled to the standard keyboard & mouse input events.

indra/newview/app_settings/settings.xml
```diff
@@ -2612,6 +2612,94 @@
...
+    <map>
+        <key>Comment</key>
+        <string>Map of controller device-specific options.</string>
+        <key>Persist</key>
+        <integer>1</integer>
+        <key>Type</key>
+        <string>LLSD</string>
+        <key>Value</key>
+        <map/>
+    </map>
```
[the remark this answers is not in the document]
AndrewMeadows: No, not that I can think of.  This all relates to viewer-side config.

indra/llcommon/llinitparam.h
```diff
@@ -1771,8 +1771,13 @@ namespace LLInitParam
 
         // implicit conversion
         operator const container_t&() const { return mValues; }
+        operator container_t&() { return mValues; }
```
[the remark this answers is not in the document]
AndrewMeadows: The compiler says this method is not necessary.  When I test GameControl UI/config everything appears to work correctly without it.  I will remove.

.github/workflows/build.yaml
```diff
@@ -44,10 +44,14 @@ jobs:
           fi
   build:
     needs: setup
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         runner: [windows-large, macos-12-xl, linux-large]
```
RyeMutt, line 50: `linux-large` should probably be removed from the main runner array since it's included below.

brad-linden: I kind of prefer to have it in both places. not sure what's best.  this seems to be the only way to override some matrix variables for a single platform unfortunately

PR #1941, opened 2024-07-08
Introduce move assignment and construction to LLPointer

RyeMutt:
This introduces move assignment and construction to LLPointer to reduce temporary ref/unref calls


marchcat (approved):
Thank you!

PR #1944, opened 2024-07-08
Re-enable compiler warnings C4244 and C4396 except for lltracerecording.h and llunittype.h for now

Ansariel:
@marchcat Had to disable C4244 for lltracerecording. and llunittype.h since the templates defined there are somewhat a mess and will need a closer look why and where exactly C4244 is still happening in them.

Since this is is an OS build, I can't say if there are still occurrances in llkdu


marchcat (approved):
Thank you, @Ansariel!

marchcat:
This builds fine for me (non-OS); merging.

PR #1945, opened 2024-07-08
Remove ancient MSVC7 code

Ansariel:


marchcat (approved):
Thank you! This code is definitely not needed any more.

PR #1946, opened 2024-07-08
Update tracy profiler to 0.10

RyeMutt:
Update tracy to version 0.10 for improvements and bug fixes


RunitaiLinden (changes requested):
Don't put ; in the macro.  We want to force people to put the ; where the macro is used for consistency (avoids a mishmash of ; and no ; and makes auto-indent consistent between IDEs).

RunitaiLinden (approved):

PR #1949, opened 2024-07-08
Fix for crash in XMLRPC reply decoding on login with large inventories

sldevel:
Commit 2ea5ac0c43e3e28d2b1774f5367d099271a1da32 introduced a crash bug due to the recursive construction of the XMLTreeNode wrapper class. The constructor of the said class typically recurses twice as many times as there are entries in the user's inventory list.

This commit:
- Moves the fromXMLRPCValue() method and its helper functions from the LLSD class/module to the LLXMLNode class, where it belongs, thus making LLSD::TreeNode (which was a wrapper class to avoid making llcommon dependant on llxml, which is still the case after this commit) totally moot; the fromXMLRPCValue() call is now done directly on the LLXMLNode.
- Moves the XML and XMLRPC decoding code out of the HTTP coroutine LLXMLRPCTransaction::Handler (coroutines got an even smaller and fixed stack), and into LLXMLRPCTransaction::Impl::process().
- Removes XMLTreeNode entirely, fixing the crash as a result.


indra/newview/llxmlrpctransaction.cpp
```diff
@@ -104,6 +100,8 @@ class LLXMLRPCTransaction::Impl
 
     std::string         mResponseText;
     LLSD                mResponseData;
+	bool                mHasResponse;
+	bool                mResponseParsed;
```
marchcat, line 104:
```suggestion
    bool                mHasResponse;
    bool                mResponseParsed;
```

Please fix this to make pre-commit happy, and then I'll gladly accept the change.
```
Check Indentation (Spaces)...............................................Failed
- hook id: indent-with-spaces
- exit code: 1

indra/newview/llxmlrpctransaction.cpp: 2 lines starting with tabs found
```

sldevel: Here you go...

marchcat: Thank you very much!

marchcat (approved):

PR #1961, opened 2024-07-09
Try using viewer-build-util@bugsplat-911 for symbol-upload fix.

nat-goodspeed:


maxim-productengine (approved):

nat-goodspeed:
This is a PR so that GitHub will build it. If it resolves this morning's consistent failure to upload symbols to BugSplat, I'll update viewer-build-util@v2 with the fix, so that all viewer builds will pick it up. Then we can discard this viewer `build.yaml` change.

nat-goodspeed:
The viewer-build-util change worked, so I updated its v2 tag, which updates all current consumers. This PR has served its purpose.

Issue #1826, opened 2024-06-21
LLError hangs viewer


akleshchev:
### Environment

Second Life Test 7.1.9.9604758944 (64bit)
Release Notes

CPU: AMD Ryzen 7 3800X 8-Core Processor              (3892.68 MHz)
Memory: 32683 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4529)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2060 SUPER/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.5244
OpenGL Version: 4.6.0 NVIDIA 552.44

Window size: 1696x1100
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 7238MB
Disk cache: Max size 1638.4 MB (99.9% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Not Connected
June 21 2024 09:05:17

### Description

Viewer hangs for half a minute than closes.
Report doesn't reach bugplat.

'second instance' doesn't hang, but doesn't send a crash report either.

### Reproduction steps

1. Eneable develop menu from settings
2. Click develop -> Force an error -> LLError and Crash

Only LLError is affected, other crashes work normally.

nat-goodspeed:
@brad-linden , could this be another instance where the compiler is too clever, detects that we're doing something illegal and discards the code?

brad-linden:
It is possible something came in to change this behavior in a recent visual studio update.  do we have indications that this is happening since June 18?  https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes#17103--visual-studio-2022-version-17103

brad-linden:
actually anytime since the beginning of May could be suspect, there have been a bunch of updates recently

RunitaiLinden:
[...]

PR #1967, opened 2024-07-09
#1826 Fix for LL_ERRS crashing in mutex lock

RunitaiLinden:


indra/llcommon/llerror.cpp
```diff
@@ -1044,7 +1045,7 @@ namespace LLError
             return;
         }
         SettingsConfigPtr s = Globals::getInstance()->getSettingsConfig();
-        LLCoros::LockType lock(s->mRecorderMutex);
+        std::unique_lock<boost::fibers::recursive_mutex> lock(s->mRecorderMutex);
```
nat-goodspeed, line 1048: C++17 has Class Template Argument Deduction. Let's try declaring just `std::unique_lock` here. If that works, I'd change `LLCoros::LockType` to just `std::unique_lock`.

brad-linden (approved):

PR #1970, opened 2024-07-10
Remove orphaned constants gMinVideoRam and gMaxVideoRam

Ansariel:


marchcat (approved):
Thank you, @Ansariel!

PR #1999, opened 2024-07-11
GLTF Transmission + IOR + Dispersion + Volume

Geenz:


RunitaiLinden:
Will need to do some A/B tests vs develop to ensure no performance or visual regression for existing content before pulling to develop... looks like pbrBaseLight likely got a lot more expensive.

Issue #1500, opened 2024-05-17
Better out of disk space handling


sl-service-account:
Viewer often fails to write cache due to not having enough free disk space which can negatively affect viewer's performance, additionally OS might fail to dynamically allocate page file, causing 'out of memory' crashes. Yet viewer stays silent despite issues.

Potential solutions:

1. Make sure viewer warns user about space running out runtime (any runtime warning shouldn't be obstructive since missing cache isn't a critical issue, viewer can live with some textures not cached).
2. Make sure viewer warns user about low space on startup (ex: compare free disk space to expected cache size).
3. Temporary reduce cache size if possible and trigger cache cleanup

Exaple: user crashed out of memory, but log is full of "There is not enough space on the disk." in logs:
**BugSplat** [Crash 1473685](https://app.bugsplat.com/v2/crash?id=1473685&database=SecondLife_Viewer_2018)

akleshchev:
User has a lot of "There is not enough space on the disk." which causes issues with texture fetching and other systems.

I'm not sure if this is the cause of the crash, but if user had dynamically allocating page file, it could have caused user to run out of memory.

igorlproductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10622905308 (64bit) on Win10/Win11 in the scope of https://github.com/secondlife/iqa/issues/316.

The log shows that the Viewer encounters issues when there is insufficient disk space and records these errors in the log. Importantly, the Viewer also notifies the user about these issues, which meets the requirement of providing real-time alerts.

Actual: 
![Capture](https://github.com/user-attachments/assets/009bb7ad-8264-40cb-85f7-d4cdb2d85474)

PR #2004, opened 2024-07-11
#1500 BugSplat Crash when out of disk space

LLGuru:
In case of out-of-disk-space error the viewer will display the following notification:
image
(not more often than once per 60 seconds)

### DIFF
--- a/indra/llcommon/llapr.cpp
+++ b/indra/llcommon/llapr.cpp
@@ -28,6 +28,7 @@
 
 #include "linden_common.h"
 #include "llapr.h"
+#include "llapp.h"
 #include "llmutex.h"
 #include "apr_dso.h"
 
@@ -606,7 +607,11 @@ S32 LLAPRFile::writeEx(const std::string& filename, const void *buf, S32 offset,
         apr_status_t s = apr_file_write(file_handle, buf, &bytes_written);
         if (s != APR_SUCCESS)
         {
-            LL_WARNS("APR") << " Attempting to write filename: " << filename << LL_ENDL;
+            LL_WARNS("APR") << "Attempting to write filename: " << filename << LL_ENDL;
+            if (APR_STATUS_IS_ENOSPC(s))
+            {
+                LLApp::notifyOutOfDiskSpace();
+            }
             ll_apr_warn_status(s);
             bytes_written = 0;
         }
--- a/indra/newview/skins/default/xui/en/notifications.xml
+++ b/indra/newview/skins/default/xui/en/notifications.xml
@@ -3032,6 +3032,15 @@ This is usually a temporary failure. Please customize and save the wearable agai
   </notification>
 
   <notification
+   icon="notifytip.tga"
+   name="OutOfDiskSpace"
+   type="notifytip">
+The system is out of disk space. You will need to free up some space on your computer or clear the cache.
+<tag>fail</tag>
+    <unique/>
+  </notification>
+
+  <notification
    icon="alertmodal.tga"
    name="YouHaveBeenLoggedOut"
    type="alertmodal">
--- a/indra/llcommon/llapp.cpp
+++ b/indra/llcommon/llapp.cpp
@@ -487,6 +487,33 @@ int LLApp::getPid()
 #endif
 }
 
+// static
+void LLApp::notifyOutOfDiskSpace()
+{
+    static const U32Seconds min_interval = U32Seconds(60);
+    static U32Seconds min_time_to_send = U32Seconds(0);
+    U32Seconds now = LLTimer::getTotalTime();
+    if (now < min_time_to_send)
+        return;
+
+    min_time_to_send = now + min_interval;
+
+    if (LLApp* app = instance())
+    {
+        app->sendOutOfDiskSpaceNotification();
+    }
+    else
+    {
+        LL_WARNS() << "No app instance" << LL_ENDL;
+    }
+}
+
+// virtual
+void LLApp::sendOutOfDiskSpaceNotification()
+{
+    LL_WARNS() << "Should never be called" << LL_ENDL; // Should be overridden
+}
+
 #ifndef LL_WINDOWS
 void setup_signals()
 {
--- a/indra/llcommon/llapp.h
+++ b/indra/llcommon/llapp.h
@@ -202,6 +202,8 @@ public:
     static bool isExiting(); // Either quitting or error (app is exiting, cleanly or not)
     static int getPid();
 
+    static void notifyOutOfDiskSpace();
+
     //
     // Sleep for specified time while still running
     //
@@ -300,6 +302,8 @@ protected:
       * @brief This method is called once a frame to do once a frame tasks.
       */
     void stepFrame();
+
+    virtual void sendOutOfDiskSpaceNotification();
 
 private:
     // Contains the filename of the minidump file after a crash.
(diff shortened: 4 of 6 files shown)


indra/llcommon/llapr.cpp
```diff
@@ -606,7 +606,11 @@ S32 LLAPRFile::writeEx(const std::string& filename, const void *buf, S32 offset,
         apr_status_t s = apr_file_write(file_handle, buf, &bytes_written);
         if (s != APR_SUCCESS)
         {
-            LL_WARNS("APR") << " Attempting to write filename: " << filename << LL_ENDL;
```
akleshchev, line 609: That seems really ill advised, why crash viewer when disk is out of space whne viewer can keep working? It should instead notify user, may be clear a bit of cache. Switching to read only mode might be an option too.

akleshchev (changes requested):

indra/llcommon/llapr.cpp
akleshchev: And the biggest issue: LL_ERRS is a silent crash Without warning user, user is the only one in position to fix the issue, so user should be informed, preferably at startup, not runtime.

marchcat: Well, this will appear in the log. But the user can just keep trying running it over and over - there should definitely be some more explicit way for them to know about the problem.

indra/llcommon/llcoros.cpp
```diff
@@ -297,7 +297,8 @@ std::string LLCoros::launch(const std::string& prefix, const callable_t& callabl
         // Out of memory on stack allocation?
         LLError::LLUserWarningMsg::showOutOfMemory();
         printActiveCoroutines();
-        LL_ERRS("LLCoros") << "Bad memory allocation in LLCoros::launch(" << prefix << ")!" << LL_ENDL;
+        LL_WARNS("LLCoros") << "Bad memory allocation in LLCoros::launch(" << prefix << ")!" << LL_ENDL;
```
akleshchev, line 300: This already shows an showOutOfMemory() notification.

indra/llcommon/llapr.cpp
```diff
@@ -606,7 +607,11 @@ S32 LLAPRFile::writeEx(const std::string& filename, const void *buf, S32 offset,
...
+            LL_WARNS("APR") << "Attempting to write filename: " << filename << LL_ENDL;
+            if (APR_STATUS_IS_ENOSPC(s))
+            {
+                LLApp::error("There is not enough space on the disk", false);
```
akleshchev, line 613: you can use LLUserWarningMsg::show(const std::string& message)

LLGuru: > But the bigest proble: this is going to spam, viewer handles a lot of textures.

LLAppViewer::showError proposes to terminate the program in case of non-critical error
User can press OK if these message become bothering him and terminate the viewer

indra/llcommon/llapr.cpp
```diff
@@ -574,6 +575,9 @@ S32 LLAPRFile::readEx(const std::string& filename, void *buf, S32 offset, S32 nb
 S32 LLAPRFile::writeEx(const std::string& filename, const void *buf, S32 offset, S32 nbytes, LLVolatileAPRPool* pool)
 {
     LL_PROFILE_ZONE_SCOPED;
+    if (LLApp::isFileReadOnlyMode())
```
akleshchev, line 578: Operations will just fail with no reason yet resources for writing will keep allocating? Also this isn't the only way of writing to cache, textures use Apr because it's supposed to be thread friendly.

indra/llcommon/llcoros.cpp
```diff
@@ -297,7 +297,11 @@ std::string LLCoros::launch(const std::string& prefix, const callable_t& callabl
...
+        if (!LLApp::isStopped() && !LLApp::isExiting())
+        {
+            LLApp::setQuitting();
+        }
```
akleshchev, line 304: I updated task description, I originally picked that crash report specifically for out of disk issues, ticked has zero relation to Coro's allocation. Please keep out of memory and out of disk handling separately.

indra/llcommon/llerror.cpp
```diff
@@ -1602,10 +1602,15 @@ namespace LLError
         return out << boost::stacktrace::stacktrace();
     }
 
-    // LLOutOfMemoryWarning
-    std::string LLUserWarningMsg::sLocalizedOutOfMemoryTitle;
-    std::string LLUserWarningMsg::sLocalizedOutOfMemoryWarning;
-    LLUserWarningMsg::Handler LLUserWarningMsg::sHandler;
+    // LLOutOfMemoryWarning, LLOutOfDiskWarning
+    // need to be preallocated before viewer runs out of memory/disk
+    static std::string sLocalizedOutOfMemoryTitle;
```
akleshchev, line 1607: sLocalizedOutOfMemoryTitle and showOutOfMemory() were preallocated (arguably should have been inline as well) because it's out of memory handling, to not allocate once we run out. There is no point prerecording these strings for everything else.

indra/llcommon/llerror.cpp
```diff
@@ -1623,6 +1628,21 @@ namespace LLError
         }
     }
 
+    void LLUserWarningMsg::showOutOfDisk()
+    {
```
akleshchev, line 1632: If you call this from main thread it will put thread on pause and will prevent viewer from sending 'keep alive' packets. Running out of disk space is a non critical issue, it should not abruptly interrupt user's experience, instead it should be a popup in the corner that recommend to free some space.

indra/llcommon/llapr.cpp
akleshchev: It's called from a thread, there is more than one.

indra/llcommon/llapr.cpp
LLGuru: What do you mean writing "read only"? The firld LLAppViewer::mSecondInstance ?

mSecondInstance is not used exclusively to prevent writing to the disk,
it seems to be used in absolutely different algorithms

And writing to the disk could happen in many cases without testing this flag

Also I'm not sure that we may turn it on in the middle of the viewer session
Viewer with mSecondInstance uses other directories for writing files

That's why I decided that we'd better introduce a new separate flag,
semantically corresponding to the general purpose: prevent storing data to the disk

And yes, there are many points where viewer writes files,
so the current commit is just a declaration of the idea, but not a final it's implementation

It  is still a draft

indra/newview/skins/default/xui/en/notifications.xml
```diff
@@ -3031,6 +3031,14 @@ This is usually a temporary failure. Please customize and save the wearable agai
   <tag>fail</tag>
   </notification>
 
+  <notification
+   icon="notifytip.tga"
+   name="OutOfDiskSpace"
+   type="notifytip">
+The system is out of disk space. You will need to free up some space on your computer or clear the cache.
+<tag>fail</tag>
+  </notification>
```
akleshchev, line 3040: suggestion: mark as unique

akleshchev (approved):

indra/newview/skins/default/xui/en/notifications.xml
vir-linden: How about making the "should never be called" base class method pure virtual?

vir-linden: I don't have a strong feeling about the wording, think it's OK as-is.

PR #2009, opened 2024-07-12
Show description and actual value of LLSD type setting

maxim-productengine:
Quick fix to show value/description of LLSD type setting. (not sure if we want to provide ability to change LLSD inside Debug settings floater - if so, we may add it later)
 Also added the fix for 'Run' button. (`onBtnRun ` calls `runSelectedScript`, where the same check occurs).


nat-goodspeed (approved):
I like it! Thank you!

PR #2010, opened 2024-07-12
Make image_list_t iterator const

zi-ree:
gcc complains about this iterator not being const


zi-ree:

marchcat (approved):
Thank you!
This builds fine on Visual Studio as well.

PR #2011, opened 2024-07-12
Initialize array before use

zi-ree:
gcc detects usage of this array as uninitialized, so make sure it's initialized. I am not sure if this is a legitimate warning, or if the code guarantees the array gets initialized before use in lines 140-142 and this is a performance optimization.


zi-ree:

marchcat (approved):
Thank you!

PR #2017, opened 2024-07-12
Fix: Stop textures from being set as inactive if still on a face.

TommyTheTerrible:
Textures were being set to Inactive even if still in scene, causing them to be deleted and re-decoded in a loop.


TommyTheTerrible:

marchcat (commented):

marchcat (approved):

Issue #2020, opened 2024-07-14
BugSplat Crash #1473685: SecondLifeViewer!<lambda_19108505e810ff43b1e426e9cce4dbc7>::operator()(52)


sl-service-account:
**BugSplat** [Crash 1473685](https://app.bugsplat.com/v2/crash?id=1473685&database=SecondLife_Viewer_2018)

**Notes:**
The key stack frame is `SecondLifeViewer!`LLCoros::launch'::`1'::catch$13`

**Customer Response:**
Bad memory allocation in LLCoros::launch(LLViewerMediaImpl::mimeDiscoveryCoro)! - mailto:KonigWolf<br/>

**Application Version:**
7.1.7.8974243247

**Callstack:**
[...]

PR #2022, opened 2024-07-14
#2020 Crash #1473685: LLCoros::launch - out of memory

LLGuru:


marchcat (changes requested):
```
/Users/runner/work/viewer/viewer/indra/test/llapp_tut.cpp:49:24: error: allocating an object of abstract class type 'tut::application::LLTestApp'
            mApp = new LLTestApp;
                       ^
In file included from /Users/runner/work/viewer/viewer/indra/test/llapp_tut.cpp:31:
/Users/runner/work/viewer/viewer/indra/llcommon/llapp.h:293:18: note: unimplemented pure virtual method 'forceQuit' in 'LLTestApp'
    virtual void forceQuit() = 0; // Puts the viewer into 'shutting down without error' mode.
                 ^
/Users/runner/work/viewer/viewer/indra/llcommon/llapp.h:294:18: note: unimplemented pure virtual method 'requestQuit' in 'LLTestApp'
    virtual void requestQuit() = 0; // Request a quit. A kinder, gentler quit.
                 ^
1 error generated.
```

indra/llcommon/llapp.cpp
```diff
@@ -414,6 +414,22 @@ void LLApp::writeMiniDump()
 {
 }
 
+// static
+void LLApp::quit(bool force)
```
marchcat, line 418: That looks nice. But shouldn't any other `app->forceQuit()` and `app->requestQuit()` be replaced with the wrapper as well?

LLGuru: They could, but not should :-)

indra/newview/llappviewer.cpp
```diff
@@ -3980,6 +3981,7 @@ void LLAppViewer::fastQuit(S32 error_code)
     _exit(final_error_code);
```
akleshchev, line 3981: If you are doing _exit to not report to bugsplat, you also need to remove SecondLife.exec_marker and logout_marker. Otherwise viewer will report "previoues viewer sessuion crashed" to statiscs server and we will have a hard time tracing where those percents are comming from.

indra/llcommon/llapp.cpp
```diff
@@ -414,6 +414,22 @@ void LLApp::writeMiniDump()
...
+{
+    if (LLApp* app = instance())
+    {
+        if (force)
+        {
+            app->forceQuit(); // Puts the viewer into 'shutting down without error' mode.
```
akleshchev, line 424: If you are doing _exit to not report to bugsplat, you also need to remove SecondLife.exec_marker and logout_marker. Otherwise viewer will report "previous viewer session crashed" to statiscs server and we will have a hard time tracing where those crash-percents are comming from. Or alternatively provide a way for statistics to distinguish out of memory from everything else.

LLGuru: LLAppViewer::removeMarkerFiles() is being called from ~LLAppViewer

akleshchev: Ok, thanks for checking.

indra/llcommon/llapp.cpp
```diff
@@ -414,6 +414,22 @@ void LLApp::writeMiniDump()
...
+        }
+        else
+        {
+            app->requestQuit(); // Request a quit. A kinder, gentler quit.
+        }
+    }
```
nat-goodspeed, line 430: I think we can assume that any caller of `LLApp::quit()` might behave badly if the call were to return. What if `instance()` returns `nullptr`? I think that case should be an `LL_ERRS()` hard stop, or at least a `std::exit()` call.

indra/llcommon/llcoros.cpp
```diff
@@ -297,7 +297,8 @@ std::string LLCoros::launch(const std::string& prefix, const callable_t& callabl
         // Out of memory on stack allocation?
         LLError::LLUserWarningMsg::showOutOfMemory();
         printActiveCoroutines();
-        LL_ERRS("LLCoros") << "Bad memory allocation in LLCoros::launch(" << prefix << ")!" << LL_ENDL;
+        LL_WARNS("LLCoros") << "Bad memory allocation in LLCoros::launch(" << prefix << ")!" << LL_ENDL;
+        LLApp::quit();
```
nat-goodspeed, line 301: I stipulate that in this situation, the viewer has always before just terminated without explanation. But if we're trying to improve that situation, maybe put up a notification? It seems worth a try: inability to allocate a contiguous block of `mStackSize` bytes doesn't mean it would be impossible to allocate something smaller.

As coded, this new `quit()` call will call `LLApp::requestQuit()`. Unless the user has previously clicked "I never want to see this confirmation prompt again," `requestQuit()` will pop up a message asking the user if they're sure they want to quit. As a user, I would be confused by that prompt, and would click "No, of course I didn't ask to quit!" That would probably return from `quit()` (I haven't tested this scenario), leading to internal confusion and a likely crash.

I would rather see a modal popup saying something about the out-of-memory condition and apologetically informing the user that the viewer must shut down -- much like the `LL_ERRS()` popup in a developer build. (Maybe reuse that notification?) On return, we could unconditionally call `forceQuit()`.

indra/llcommon/llapp.h
```diff
@@ -288,6 +290,8 @@ class LL_COMMON_API LLApp
     string_map mOptionMap;  // Contains all command-line options and arguments in a map
 
 protected:
+    virtual void forceQuit() {} // Puts the viewer into 'shutting down without error' mode.
+    virtual void requestQuit() {} // Request a quit. A kinder, gentler quit.
```
nat-goodspeed, line 294: I don't think the default behavior should be to return. I think if control actually reaches either of these base-class methods, it should call `std::exit()` or something comparable.

indra/newview/llappviewer.h
```diff
@@ -90,9 +90,9 @@ class LLAppViewer : public LLApp
 
     // Application control
     void flushLFSIO(); // waits for lfs transfers to complete
-    void forceQuit(); // Puts the viewer into 'shutting down without error' mode.
+    virtual void forceQuit(); // Puts the viewer into 'shutting down without error' mode.
```
nat-goodspeed, line 93: Declare these `override`, not `virtual`.

indra/newview/llappviewer.cpp
```diff
@@ -3958,6 +3958,7 @@ void LLAppViewer::removeDumpDir()
     }
 }
 
+// virtual
```
nat-goodspeed, line 3961: moot if the `.h` file declares these `override`

nat-goodspeed (changes requested):

indra/llcommon/llcoros.cpp
akleshchev: Viewer currently calls showOutOfMemory(); which informs user of what's going on.

nat-goodspeed: Thank you, I did overlook that.

I still think it would be odd to then put up a second prompt asking the user if they really want to quit, and returning normally if they click No.

LLGuru: > Viewer currently calls showOutOfMemory() at string 298 which informs user of what's going on.
Sure, this looks as follows:
image

LLGuru: > Unless the user has previously clicked "I never want to see this confirmation prompt again," `requestQuit()` will pop up a message asking the user if they're sure they want to quit

I simulate this error, and I observe only one message box (shown above).
After clicking "OK" button I see the viewer is terminated silently in a few seconds.

indra/llcommon/llapp.h
LLGuru: > I don't think the default behavior should be to return

This behavior is not default, first I made these functions pure virtual,
but tests had failed due to no implementation for them (see above).
So in fact these methods are always overridden in the viewer itself.

LLGuru: As an alternative, I can implement these functions in LLTestApp (with empty bodies)
and make them pure virtual in the base class LLApp again

indra/llcommon/llapp.cpp
LLGuru: Done

indra/newview/llappviewer.h
LLGuru: Done

indra/llcommon/llcoros.cpp
akleshchev: requestQuit doesn't show any notifications, it just does a sendLogoutRequest then forceQuit

indra/newview/llappviewer.h
LLGuru: > Declare these `override`, not `virtual`.

It seems we shouldn't mark some functions as `override`
```
In file included from /Users/runner/work/viewer/viewer/indra/newview/tests/lllogininstance_test.cpp:223:
/Users/runner/work/viewer/viewer/indra/newview/tests/../llappviewer.h:87:18: error: 'init' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
    virtual bool init();            // Override to do application initialization
```
Clang requires using the same rule for all overridden functions in the class:
either mark all of them as override or do not use this keyword at all

Issue #2029, opened 2024-07-15
OSMessageBoxWin32 shows message box behind the viewer window


LLGuru:
### Environment

SecondLife viewer any win64 build

### Description

In some cases (e.g., runtime error message and user click the window)
OSMessageBoxWin32 shows message box behind the viewer window

As a result, the user doesn't see this message box and can't interact with it until switches between windows

### Reproduction steps

- Start viewer
- Double click top menu `Develop` | `Force an Error` | `Force LLError And Crash`
![image](https://github.com/user-attachments/assets/c3cad4a3-408a-4041-a3e5-81465dea3433)

LLGuru:
- Fixed in https://github.com/secondlife/viewer/pull/2031

AtlasLinden:
Reopening for testing

valeriiaaproductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10582490681 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/316.

PR #2031, opened 2024-07-15
#2029 OSMessageBoxWin32 shows message box behind the viewer window

LLGuru:


marchcat (approved):

indra/llwindow/llwindowwin32.cpp
```diff
@@ -1702,10 +1709,15 @@ void LLWindowWin32::recreateWindow(RECT window_rect, DWORD dw_ex_style, DWORD dw
...
+    {
+        sWindowHandleForMessageBox = NULL;
+    }
+
     // zero out mWindowHandle and mhDC before destroying window so window
     // thread falls back to peekmessage
-    mWindowHandle = 0;
-    mhDC = 0;
+    mWindowHandle = NULL;
+    mhDC = NULL;
```
akleshchev, line 1720: Handle isn't a pointer, it contains an integer. Ex: -1 equals invalid, While NULL isn't technically invalid to use, it often implies a pointer, plus Microsoft uses 0. I suggest using Microsoft's standard and leave it as a 0.

indra/llwindow/llwindowwin32.cpp
```diff
@@ -3684,7 +3698,7 @@ S32 OSMessageBoxWin32(const std::string& text, const std::string& caption, U32 t
         break;
     }
 
-    int retval_win = MessageBoxW(NULL, // HWND
+    int retval_win = MessageBoxW(sWindowHandleForMessageBox, // HWND
```
akleshchev, line 3701: This looks a bit convoluted, and why have two variables do the same thing?
Suggestions:
LLWindowWin32 *window = static_cast<LLWindowWin32*>(mWindow);
HWND handle = window ? window->mHandle : 0;
MessageBoxW(handle ,...)

akleshchev, line 3701: Window handle can be invalid (-1), are you certain message box will work correctly if supplies with an invalid handle?

indra/llwindow/llwindowwin32.cpp
```diff
@@ -970,6 +972,11 @@ void LLWindowWin32::close()
 
     LL_DEBUGS("Window") << "Destroying Window" << LL_ENDL;
 
+    if (sWindowHandleForMessageBox == mWindowHandle)
+    {
+        sWindowHandleForMessageBox = NULL;
+    }
+
```
akleshchev, line 979: If window somehow gets closed without without a call for close (was the norm untill recently), sWindowHandleForMessageBox will stay invalid. I suggest checking it again in ~LLWindowWin32()

indra/llwindow/llwindowwin32.cpp
LLGuru: HWND is a type of pointer to a struct HWND__
image
Semantically it is a pointer, but NULL is defined as 0
Of course, there is almost no difference between 0 and NULL
But if we use NULL everywhere in this class and 0 omly once,
I suggest replace 0 with NULL for better readability

akleshchev: Sorry, you are correct.

indra/llwindow/llwindowwin32.cpp
LLGuru: > Suggestions:
> LLWindowWin32 _window = static_cast<LLWindowWin32_>(mWindow);

The function `OSMessageBox` is not a member of the class `LLWindow`
and doesn't have access neither to an instance of this class nor to it's field `mWindow`

akleshchev: It should have access to gViewerWindow

LLGuru: > Window handle can be invalid (-1)

It can't be NULL
This value is returned by CreateWindowEx
```
Return value
Type: HWND

If the function succeeds, the return value is a handle to the new window.

If the function fails, the return value is NULL. To get extended error information, call GetLastError
```
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowexa

indra/llwindow/llwindowwin32.cpp
LLGuru: Done

indra/llwindow/llwindowwin32.cpp
akleshchev: static_cast<LLWindowWin32>(gViewerWindow->getWindow())

LLGuru: > It should have access to gViewerWindow->getWindow(), it's global

Maybe it should, but it can't :-)
The variable gViewerWindow is not accessible from llwindow library

I have already gone through all these ideas myself :-)

akleshchev: You are correct, there is bunch of accessibility issues. Still not a nice solution, fills like it should have been a class with  setHWND and IsWindow(handle) checks, but this code being platform specific makes it's complicated...

akleshchev (approved):

indra/llwindow/llwindowwin32.cpp
nat-goodspeed: > I have already gone through all these ideas myself :-)

When I've tried several things that didn't work out, even if I realized they wouldn't before writing or testing the relevant code, and I write something that might make a subsequent maintainer raise their eyebrows, I often leave a comment of the form "This is why I'm doing it this way, instead of what you would think would be more obvious..."

LLGuru: > I often leave a comment

Done

PR #2032, opened 2024-07-16
Fix: Update calcDataSizeJ2C to pyramid-base file size estimation

TommyTheTerrible:
Used the loop from the previous LayerFactored method to create a more accurate file size estimation by walking up the pyramid tiles.

Sizes are much larger in many cases and eliminate partial decoder issues with OpenJPEG.

KDU not tested but expected to produce better files as well.

Should also stop decode failures on tiny or very rectangular dimensions.

### DIFF
--- a/indra/llimage/llimagej2c.cpp
+++ b/indra/llimage/llimagej2c.cpp
@@ -283,7 +283,7 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
     while (surface > s)
     {
         if (nb_layers <= (5 - discard_level))
-            totalbytes += s * comp * precision * rate;
+            totalbytes += (S32)(s * comp * precision * rate);
         nb_layers++;
         s *= 4;
     }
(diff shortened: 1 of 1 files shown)


marchcat (commented):

I'll test it with KDU as well before further integration.

indra/llimage/llimagej2c.cpp
```diff
@@ -278,13 +278,20 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
     S32 nb_layers = 1;
     S32 surface = w*h;
     S32 s = 64*64;
+    S32 precision = 8; // assumed bitrate per component channel, might change in future for HDR support
+    S32 totalbytes = s * comp * precision * rate; // first level computed before loop
```
marchcat, line 282:
```suggestion
    S32 totalbytes = (S32)(s * comp * precision * rate); // first level computed before loop
```

indra/llimage/llimagej2c.cpp
```diff
@@ -278,13 +278,20 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
...
     while (surface > s)
     {
+        if (nb_layers <= (5 - discard_level))
+            totalbytes += s * comp * precision * rate;
```
marchcat, line 286:
```suggestion
            totalbytes += (S32)(s * comp * precision * rate);
```

marchcat (commented):
Tried to build on Visual Studio and got two `conversion from 'F32' to 'S32', possible loss of data` warnings (treated as errors). The suggested changes suppress them.

Otherwise, the code looks good.

marchcat (commented):
It looks like I can directly apply those suggestions inside the PR. Great, this makes things faster.

marchcat (approved):
Thank you very much for your contribution!

Issue #2039, opened 2024-07-16
BugSplat Crash #1494687: Second Life Release!LLPipeline::calcNearbyLights(LLCamera&)


sl-service-account:
**BugSplat** [Crash 1494687](https://app.bugsplat.com/v2/crash?id=1494687&database=SecondLife_Viewer_2018)

**Application Version:**
7.1.9.9688089989

**Callstack:**
[...]

PR #2040, opened 2024-07-16
#2039 BugSplat Crash #1494687: LLPipeline::calcNearbyLights(LLCamera&)

LLGuru:


marchcat (approved):

akleshchev:
I double checked and this isn't a shutdown crash. Log is corrupted (from a wrong session, see timestamps) so it isn't an indicator of shutdown going on.

PR #2047, opened 2024-07-17
Lua api for adding new menu items to the Top menu

maxim-productengine:


nat-goodspeed (approved):
I think this will be huge!

I almost feel like changing the signatures and the protocol for these menu-building operations so a caller can engage [LLEventAPI's "batch" mode](https://github.com/secondlife/viewer/blob/lua-top-menu/indra/llcommon/lleventdispatcher.h#L810-L837), passing an array of operations to be performed in order. What I mean by "signature and protocol" is:

* make each menu-building method return `LLSD`, instead of a `void` return and an embedded `Response` object
* throw DispatchError(message) instead of calling `response.error()`

Then the Lua side could bundle the data to construct a whole menu in one call.

-----

I think this would be even more immensely powerful if a Lua script could add a menu item that would invoke a Lua script, coded as a top-level function receiving an argument. This idea needs a bit of caution: I don't think we should register a Lua script as a named callback function. Given the "UI" "call" operation, that would empower a Lua script to run an arbitrary number of other Lua scripts, potentially drowning the viewer.

Nonetheless, if we allowed a Lua script to configure some number of menu items to let a user easily run other Lua scripts, that would be tremendous.

maxim-productengine:
I see your point. I was thinking about a similar thing: adding UI callback for invoking a Lua script, which name is defined in the param. This way it will be possible to create a handy menu with quick access to the scripts you want. This UI callback won't be triggered by the script as we have 'block' flag.

PR #2058, opened 2024-07-18
viewer#1997 Warn user if closing appearance with unsaved changes

akleshchev:


indra/newview/skins/default/xui/en/notifications.xml
```diff
@@ -4252,6 +4252,20 @@ Unable to add estate manager [AGENT] to banned list.
...
 
+  <notification
+   icon="alertmodal.tga"
+   name="UsavedWearableChanges"
+   type="alertmodal">
+You have unsaved changes.
+      <tag>group</tag>
+    <tag>confirm</tag>
+    <usetemplate
+     ignoretext="Confirm before I discard unsaved wearable changes"
```
marchcat, line 4263:
```suggestion
     ignoretext="Are you sure you want to discard unsaved wearable changes?"
```

This will probably fit better.

marchcat (approved):

indra/newview/skins/default/xui/en/notifications.xml
akleshchev: That's not a visible part, it's for preferences so it should cater to preferences. Popup shows a basic 'don't ask again' text.
image

marchcat: Ah, that makes sense, thank you.

indra/newview/llfloatersidepanelcontainer.cpp
```diff
@@ -90,6 +91,46 @@ void LLFloaterSidePanelContainer::closeFloater(bool app_quitting)
...
+        }
+    }
+
+    closeFloater();
+}
+
+void LLFloaterSidePanelContainer::onCloseMsgCallback(const LLSD& notification, const LLSD& response)
+{
+    S32 option = LLNotificationsUtil::getSelectedOption(notification, response);
+    if (0 == option)
```
maxim-productengine, line 128: "Keep editing" button will close the floater this way. 
So probably button names should be switched, or you can adjust the check here.

akleshchev: You are correct, thanks for noticing.

maxim-productengine (approved):
I'm a bit surprised, that we didn't have that notification before.

Issue #1873, opened 2024-06-28
glTF Mesh Open Regression


aiaustin:
### Environment

Second Life Release 7.1.9.9620320242 (64bit)
Release Notes

You are at 149.9, 144.9, 23.0 in Rumpus Room 2048 located at simhost-01129fe6e38889eb3.aditi
SLURL: secondlife://Aditi/secondlife/Rumpus%20Room%202048/150/145/23
(global coordinates 36,501.9, 14,224.9, 23.0)
PMFP 2024-06-25.9663125755
Release Notes

CPU: Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz (3491.59 MHz)
Memory: 32713 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4529)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 980/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.5612
OpenGL Version: 4.6.0 NVIDIA 556.12

Window size: 1171x772
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 256m
Bandwidth: 2400kbit/s
LOD factor: 2
Render quality: 6
Texture memory: 4095MB
Disk cache: Max size 1638.4 MB (99.3% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 11/25,972 (0.0%)
June 28 2024 02:51:58

### Description

Many of my glTF test meshes show issues in Atlasaurus (Second Life Release 7.1.9.9620320242). The issues are present in many meshes of various sizes from small to large. They include missing mesh parts, out of place and wrongly scaled or rotated mesh parts (see example Pisten Bully), single faces not displaying (front textured face of Blast wall), etc. The meshes display wrongly whether freshly opened onto a prim, or whether uploaded, saved to inventory and rezzed from there. 

[...]

PR #2063, opened 2024-07-18
Fix for loss of GLTF override data due to objects eviction from the rendered scene

vldevel:
Commit 64c541f40191dee0fb6b3b0c09d0d9bfacb454a5 introduced a bug causing the indue removal of GLTF override data whenever an object is 'killed' when not rendered any more (e.g. an object in a connected neighbour region finding itself out of draw distance when the user moves or cams around).

Since this data is not resent by the sim server until the next reconnection of the said sim, the object may re-rez later with missing overrides.

This commit simply comments out the culprit line.


sldevel:
Just a note that, after the code is fixed and when you are ready to run the fixed viewer, you need to clear the objects cache to see the fix working in already visited regions (else you pick up corrupted/missing caches object data); maybe an object cache version increment would be a good thing to do, so that users running the future fixed viewer will automatically have their old corrupted cache wiped out...

marchcat:

indra/newview/llviewerregion.cpp
```diff
@@ -1162,8 +1162,14 @@ void LLViewerRegion::killCacheEntry(LLVOCacheEntry* entry, bool for_rendering)
             child = entry->getChild();
         }
     }
-    // Kill the assocaited overrides
+#if 0    // Nope: do not do this !  The object may re-rez later in the same
+         // session (e.g. for neighbour region objects, depending on draw
+         // distance, while moving or caming around) and its override may
+         // not be re-sent by the simulator, causing a loss in the override
+         // data !  HB
```
brad-linden, line 1169: I need to test this more to fully understand these cases, but the object cache probably needs a larger redesign to properly handle the neighbor region object cases.  the main correctness criteria after cache eviction should be that if the overrides get evicted from the cache then the object itself will also be evicted from the cache, and when the viewer sends the cache miss message to the simulator the full object update and overrides will be re-sent.

sldevel: This fix, which has been in place in my viewer for weeks, resulted from the direct observation and reliable reproduction of the problem by me. Since the bug I observed happened just after the inclusion of that line in my code base, I could easily identify it as the culprit (the benefit of weekly releases is that such bugs are quickly identified and easy to trace back to a culprit code change).

However, I did not try and diagnose what happened at the cache level, but could also see issues with this line active, after relogs and missing overrides.

brad-linden (changes requested):
if an object is removed from the main VOCache, it needs to have its overrides removed from the override cache too, otherwise the cache can grow without bound.  if there is a case where the object comes back into view later and Overrides are not being sent down from the server, then we need to fix the Cache Probe mechanism.  we probably want to file an issue with a repro to get that fixed

brad-linden:
>Since this data is not resent by the sim server until the next reconnection of the said sim, the object may re-rez later with missing overrides.
to be clear, this is not intentional.  the overrides are expected to be re-sent when the object re-rezes later

aiaustin:
This issue and problem is almost certainly one cause of the issue "glTF Mesh Open Regression". See https://github.com/secondlife/viewer/issues/1873

Full perm examples that display incorrectly are still available on 19-Feb-2025 at
secondlife://Aditi/secondlife/Rumpus%20Room%202048/150/145/23

beqjanus:
I can't say with full certainty that this "fix" is wrong, but it seems to me very likely that it is fixing a side-effect of the deeper problem in the way that the gltf-extras are being updated and the disfunctional cache in general. The function the that one-line change is in removes all traces of an entry, disassociating itself from the parent and killing all the children. It makes no sense to me that one small aspect (the so-called gltf extras) should be left untouched. The net result will be a gradual accumulation of these leftovers. I may well be missing a subtle interaction here but I feel that the correct fix for this issue is deeper inside the cache.

aiaustin:
In case its helpful, I reached out to "@Rye" who is the  remaining contributor that has not signed the CLA.  https://github.com/rye  - this was his response.

I do not recall being involved in that fix as I've never heard of this project. I'm not sure what logic the CLA bot chose to pick my account as a contributor, but it seems to have picked wrong. (I have been pretty good about signing my commits and none of the commits on that PR with the keyword "rye" bear any markers I signed them.

brad-linden:
those commits are associated with the email rye@lindenlab.com which has never been associated with the github account rye.  we have attempted to add an exception to the CLA bot to prevent it from scanning those commits because of these errors, but apparently something went wrong with that.  please disregard the errors for that address

brad-linden:
I am closing this pull request.  I think we need to start from scratch on this issue.  I have filed https://github.com/secondlife/viewer/issues/3688 and attempted to capture the repro you described here.  Please add more details there to help get this bug solved the right way

PR #2066, opened 2024-07-18
Enable SSE2 instruction support in glm for optimized codegen

RyeMutt:
Enable SSE2 instruction use in GLM


Geenz (approved):
LTGM.

marchcat (approved):
Thank you!

PR #2072, opened 2024-07-19
viewer#2071 Properly handle 'out of memory' for meshes

akleshchev:


Ansariel:
What is the point of showing an out of memory error message then then LL_ERR out anyway? Just abort whatever what was going on and give the user the chance to shutdown the viewer gracefully!

marchcat (approved):

akleshchev:
>Just abort whatever what was going on and give the user the chance to shutdown the viewer gracefully!

I wanted to do that originally as well, but most crashes happen during teleports, login or during situations where viewer rapidly allocates more memory. Short of a disconect there is little chance of a 'gracefull' shutdown and that would still have chances of crashing. It is technically possible to achive, but it would require some time to implement and cover all cases and is far too low as a priority to do.

>What is the point of showing an out of memory error message then then LL_ERR out anyway?
Situation isn't recoverable, thus it is an LL_ERR. Instead of crashing silently (which viewer normnally does earlier or later when out of memory) notification tells user what happened and gives some advice what to do to avoid this situation next run.

sldevel:
> > Just abort whatever what was going on and give the user the chance to shutdown the viewer gracefully!
> 
> I wanted to do that originally as well, but most crashes happen during teleports, login or during situations where viewer rapidly allocates more memory. Short of a disconect there is little chance of a 'gracefull' shutdown and that would still have chances of crashing.

I do not agree.

First: the viewer does not allocate more memory *during* teleports; if you properly wrote it, it even *frees* a lot of memory at the TELEPORT_START_ARRIVAL step, by cancelling stale textures fetches, cleaning up pipeline rebuild groups, force-flushing old images... But it is true that LL's viewer does not have this code, so forget it. It however starts allocating a lot of memory *after* TELEPORT_ARRIVING; at this point, if you get an out of memory issue, you could freeze the texture fetcher (IIRC there's already a variable to do this), or simply disconnect the viewer cleanly and pop up an alert to let the user know that the viewer ran out of free memory but did disconnect appropriately.

Second: LLERRS currently crashes the viewer and therefore causes a timeout disconnection on the server side, which in turn may cause attachments states to be lost or reverted to a former state, as well as resulting with the user finding their avatar located in the sim where they last logged in before the crash (instead of the TP departure sim) when they relog: this can be an issue if their outfit changed and the rating is no more appropriate; e.g. they wore a "Moderate" outfit on login in a Moderate sim, but then moved around, changed to an "Adult" outfit, and crashed as they were TPing from an Adult sim to another and they will consequently find themselves rezzed with an Adult outfit in a Moderate sim on next relog.

Third: a relog immediately after a LLERRS (i.e. before the presence server timed out) will cause all sorts of problems (failure to get valid caps URLs, and consequently to retrieve inventory, or even to connect to neighbour sims), forcing the user to log out again and relog (which is very user-unfriendly, to say the least).

> It is technically possible to achive, but it would require some time to implement and cover all cases and is far too low as a priority to do.

If you lack time, you could still do what I did for the Cool VL Viewer. on llerrs (i.E. LLERRS for LL's viewer), my viewer first attempts a clean disconnection from the sim server and only *then* it displays an OSMEssageBox to make the user aware of what just happened and, *finally*, it crashes (abort() actually) the viewer since no valid fallback code path has been setup in such a case.

akleshchev:
> First attempts a clean disconnection from the sim server and only then it displays an OSMessageBox to make the user aware of what just happened

I'm not sure I like the order of events, but it is reasonable to make a clean disconnect if possible, so will do that.

akleshchev:
Did a variant of a soft quit. It's still liable to crash and user can technically abort it, but figured it's much better than what we had either way.

PR #2073, opened 2024-07-19
calcDataSizeJ2C adjusted to use maximum possible components

TommyTheTerrible:
Previous pyramid walking calculation assumed the incoming components variable can be accurate but unfortunately the needs_aux is only set to true if the face has an alpha mask setting.

Without this information we must assume the J2C files have the maximum component size of four so that alpha channels are found when decoding both the color and aux textures.


marchcat (approved):
Thanks!

marchcat:
Previous PR: #2032

PR #2075, opened 2024-07-20
Add newer MSVC redist libs to copy action

RyeMutt:
This copies and packages additional redist libs for full compatibility with the newer MSVC versions on systems that do not have it installed.


marchcat (approved):
Thank you!

PR #2077, opened 2024-07-21
Fix failure to join or detach threads causing rare shutdown termination

RyeMutt:
This fixes a behavior where LLThreads were never detached or joined and could result in a crash on shutdown in rare instances .

### DIFF
--- a/indra/llcommon/llthread.cpp
+++ b/indra/llcommon/llthread.cpp
@@ -269,6 +269,7 @@ void LLThread::shutdown()
             mStatus = STOPPED;
             return;
         }
+        delete mThreadp;
         mThreadp = NULL;
     }
 
@@ -299,6 +300,7 @@ void LLThread::start()
     {
         mThreadp = new std::thread(std::bind(&LLThread::threadRun, this));
         mNativeHandle = mThreadp->native_handle();
+        mThreadp->detach();
     }
     catch (std::system_error& ex)
     {


akleshchev (approved):

indra/llcommon/llthread.cpp
```diff
@@ -269,6 +269,7 @@ void LLThread::shutdown()
             mStatus = STOPPED;
             return;
         }
+        delete mThreadp;
```
nat-goodspeed, line 272: I'd rather see `mThreadp` be a `std::unique_ptr<std::thread>`, with a `reset()` call here and in `start()`.

indra/llcommon/llthread.cpp
```diff
@@ -299,6 +300,7 @@ void LLThread::start()
     {
         mThreadp = new std::thread(std::bind(&LLThread::threadRun, this));
         mNativeHandle = mThreadp->native_handle();
+        mThreadp->detach();
```
nat-goodspeed, line 303: Given that we have explicit logic to wait for thread termination, but time out and forcibly kill the thread, this makes sense, yes. In effect we manually "try to `join()`."

nat-goodspeed (changes requested):

indra/llcommon/llthread.cpp
akleshchev: Thank you, that's probably outside of a 'contribution' work, I will do an issue for that later.

PR #2078, opened 2024-07-22
Fix two rare shutdown crashes in gCacheName and gObjectList

RyeMutt:
This fixes two rare shutdown crashes I've observed caused by static construction/destruction order issues.


akleshchev (approved):

marchcat (approved):
Thank you!

PR #2079, opened 2024-07-22
Use thread pool to create plugin processes to reduce created/destroyed threads

RyeMutt:
This fixes a new thread being created and destroyed for every plugin process launched by utilizing the threadpool workers. It also fixes a few crashes I observed over the last few months.


akleshchev (approved):

marchcat (approved):
Thank you!

callumlinden (approved):

PR #2080, opened 2024-07-22
Cache various frequently accessed settings

RyeMutt:
Cache various frequently accessed settings to reduce potential frame stutter.


akleshchev (approved):

marchcat (approved):
Thank you!

PR #2085, opened 2024-07-22
Fix various nullptr crashes

RyeMutt:
Fixes multiple nullptr crashes noted in Alchemy's crash logs over the past few months


marchcat (approved):
Thank you very much!

akleshchev (approved):

PR #2093, opened 2024-07-24
viewer#2065 Region restart schedule floater

akleshchev:
image

### DIFF
--- a/indra/newview/llfloaterregionrestartschedule.cpp
+++ b/indra/newview/llfloaterregionrestartschedule.cpp
@@ -68,12 +68,18 @@ bool LLFloaterRegionRestartSchedule::postBuild()
     mPMAMButton = getChild<LLButton>("am_pm_btn");
     mPMAMButton->setClickedCallback([this](LLUICtrl*, const LLSD&) { onPMAMButtonClicked(); });
 
-    bool use_24h_format = gSavedSettings.getBOOL("Use24HourClock");
-    if (use_24h_format)
+    // By default mPMAMButton is supposed to be visible.
+    // If localized xml set mPMAMButton to be invisible, assume
+    // 24h format and prealligned "UTC" label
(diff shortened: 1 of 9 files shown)


marchcat (approved):
This looks awesome!

indra/newview/llfloaterregionrestartschedule.cpp
```diff
@@ -0,0 +1,377 @@
...
+        LL_WARNS("Region") << "Saving region schedule, but RegionSchedule capability is not available" << LL_ENDL;
+    }
+}
+
+void LLFloaterRegionRestartSchedule::onCommitHours(const LLSD& value)
+{
+    S32 hours = value.asInteger();
+    if (mPMAMButton->getVisible())
+    {
+        llclamp(hours, 1, 12);
```
akleshchev, line 199: I'm not sure if I should translate 0 to 12 here.

maxim-productengine (approved):

Ansariel:
My clock as 24 hours - I don't need an AM/PM selector!

akleshchev:
> My clock has 24 hours - I don't need an AM/PM selector!

Mine too, but am/pm button is part of requested UI design and unfortunately viewer has no locale detection.

Ansariel:
Time to file this as a bug because I as a user have no idea what AM/PM means! 😂

akleshchev:
There is a debug setting to switch it

marchcat:
BTW, what about the clock we have in the upper right corner of the main viewer window? I don't remember if it allows changing its time format, but if it does, that setting could also be used.

Ansariel:
> BTW, what about the clock we have in the upper right corner of the main viewer window? I don't remember if it allows changing its time format, but if it does, that setting could also be used.

Not sure the LL viewer can, but Firestorm for sure can show 24h clock.

PR #2101, opened 2024-07-24
#2100 BugSplat Crash #1497210: LLVOAvatar::updateImpostors()(10616)

LLGuru:


marchcat (approved):

indra/newview/llvoavatar.cpp
```diff
@@ -10763,17 +10763,18 @@ void LLVOAvatar::updateImpostors()
...
     {
-        LLVOAvatar* avatar = (LLVOAvatar*) *iter;
-        if (!avatar->isDead()
-            && avatar->isVisible()
-            && avatar->isImpostor()
-            && avatar->needsImpostorUpdate())
+        if (LLVOAvatar* avatar = dynamic_cast<LLVOAvatar*>(instance))
         {
-            avatar->calcMutedAVColor();
-            gPipeline.generateImpostor(avatar);
```
akleshchev, line 10776: What is this going to change crash wise? This doesn't looke like it will change anything at all and shouldn't be accosiated with a crash fix.

marchcat: The `dynamic_cast` ensures type safety here.

akleshchev: It does not, LLVOAvatar is the only use for LLAvatarAppearance/LLCharacter

LLGuru: > What is this going to change crash wise?

This is just a kind off null-check

akleshchev: sInstances can not have null inside, it gets cleared in LLCharacter's destructor so it is pointless to do a null check.

LLGuru: What could be a reason for the signal SIGSEGV on the line 10616?
image
image
I see the only explanation that `avatar` is NULL (`avatar->mDrawable.mPointer` is being checked)
Your thoughts?

marchcat: There's nothing wrong with a null check. But the question is, why is it a nullptr at this point?

LLGuru: > P.S. other places also don't do null checks for sInstances, sInstances is intended to be safe in this regard.

It seems there are some critical places where this check is being done
image
image
image

akleshchev: > What could be a reason for the signal SIGSEGV on the line 10616?
Invalid pointer, perhaps you should make sure characters aren't created/destroyed outside of main thread?

Issue #2103, opened 2024-07-24
BugSplat Crash #1497033: Second Life Release!LLViewerTextureList::getImage(LLUUID const&, FTType, int, LLGLTexture::EBoostLevel, signed char, int, unsigned int, LLHost)


sl-service-account:
**BugSplat** [Crash 1497033](https://app.bugsplat.com/v2/crash?id=1497033&database=SecondLife_Viewer_2018)

**Application Version:**
7.1.9.9688089989

**Callstack:**
[...]

PR #2104, opened 2024-07-24
#2103 BugSplat Crash #1497033: LLViewerTextureList::getImage

LLGuru:


marchcat (approved):

indra/newview/llviewertexturelist.cpp
```diff
@@ -561,9 +561,9 @@ LLViewerFetchedTexture* LLViewerTextureList::getImage(const LLUUID &image_id,
                                                    LLHost request_from_host)
 {
     LL_PROFILE_ZONE_SCOPED_CATEGORY_TEXTURE;
-    if(!mInitialized)
```
akleshchev, line 564: Crash time: 2024-07-20T22:49:39Z (according to plCrashReporter.crashlog)
Log time: 2024-07-20T23:03:30Z
This looks like a log from a different session.

Why do you think isExiting() will fix it?
Call hapens from withing (!LLApp::isExiting()) check and if error was set by something texture cach should still be valid.

PR #2105, opened 2024-07-24
Fix warning from misnamed button image name reference

RyeMutt:
Fix widgets referencing PushButton_Disabled_Selected instead of PushButton_Selected_Disabled

Plus cache a few more frequently accessed settings


marchcat (approved):
Thank you!

PR #2114, opened 2024-07-25
Reinstate Texture Deletion Grace Period

TommyTheTerrible:
Adjust logic of LLViewerFetchedTexture::setInactive to allow inactive state to be flipped and allow LLViewerTextureList::updateImageDecodePriority to offer a grace period of at least 60 seconds in inactive status before deletion.

Once those 60 seconds have passed, the texture is set to deletion candidate status.

While in deletion candidate status, the texture may be deleted if a further timeout is set or if VRAM has reached 5% remaining free memory.

The value for deletion timeout while a delete candidate is set as zero in this code to stop the timed deletion, but I hope to persuade stakeholders in a later PR to set this value to 60 seconds or at least a user defined setting.

Stakeholders: @Ansariel @vldevel @RunitaiLinden 

Previous discussion: https://github.com/secondlife/viewer/commit/27a2531c5bf375ed208d8fd5764a8797f08ce831

### DIFF
--- a/indra/newview/llviewertexture.cpp
+++ b/indra/newview/llviewertexture.cpp
@@ -1252,13 +1252,21 @@ void LLViewerFetchedTexture::setDeletionCandidate()
 }
 
 //set the texture inactive
-void LLViewerFetchedTexture::setInactive()
+// <TS:3T> Adjusted function to allow mTextureState to be flipped back to Active if sent a true bool.
+void LLViewerFetchedTexture::setInactive(bool found)
 {
-    if(mTextureState == ACTIVE && mGLTexturep.notNull() && mGLTexturep->getTexName() && !mGLTexturep->getBoundRecently())
+    if(mTextureState > DELETED && mTextureState != NO_DELETE && mGLTexturep.notNull() && mGLTexturep->getTexName() && !mGLTexturep->getBoundRecently())
     {
-        mTextureState = INACTIVE;
+        if (found) {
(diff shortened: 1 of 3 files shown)


Ansariel:
Way too complicated! And why would should a texture get removed after 60 seconds if it's cached in memory and memory isn't low? As described in the other discussion that doesn't make sense!

The other part of removing if memory gets low can be easily achieved by doing something like this or alternatively by a percentage-based approach:
```
void LLViewerFetchedTexture::destroyTexture()
{

    static LLCachedControl<F32> MinVRAMTextureDestroyThreshold(gSavedSettings, "MinVRAMTextureDestroyThreshold");
    if (sFreeVRAMMegabytes >= MinVRAMTextureDestroyThreshold)
    {
        return;
    }

    if (mNeedsCreateTexture)//return if in the process of generating a new texture.
    {
        return;
    }

    //LL_DEBUGS("Avatar") << mID << LL_ENDL;
    destroyGLTexture();
    mFullyLoaded = false;
}
```

TommyTheTerrible:
Updated code to use LLViewerTexture::sFreeVRAMMegabytes. I do not see "MinVRAMTextureDestroyThreshold" in the repository, so I believe that would require a debug setting to be added but I think that might be a good idea.

I believe @Ansariel is pushing a non-performant preference with their code suggestion to place the check into the LLViewerFetchedTexture::destroyTexture function itself. Putting the check there will not allow any way around their preference.

PR #2115, opened 2024-07-26
Reduce utf8 to wstring conversion and llwstring temporaries during text draw

RyeMutt:
This reduces the number of LLWString temporaries and needless utf8 to wstring conversions during various text draw calls.


marchcat (approved):
Thank you!

PR #2117, opened 2024-07-26
viewer-private#255 p2p outgoing calls did not work correctly

akleshchev:
Issue:
P2P was catching STATUS_LEFT_CHANNEL meant for nearby chat and adopting channel info for itself

Solution:
- Moved one of notifyStatusObservers calls so that it would have uri data instead of firing with no channel info
- Made p2p sessions init with uri data, like it was before webrtc. Which is used to distinguish observer notifications.
- Removed mAudioSessionChanged. It was unused yet confusing.

### DIFF
--- a/indra/newview/llvoicechannel.cpp
+++ b/indra/newview/llvoicechannel.cpp
@@ -122,7 +122,13 @@ void LLVoiceChannel::onChange(EStatusType type, const LLSD& channelInfo, bool pr
 {
     LL_DEBUGS("Voice") << "Incoming channel info: " << channelInfo << LL_ENDL;
     LL_DEBUGS("Voice") << "Current channel info: " << mChannelInfo << LL_ENDL;
-    if (mChannelInfo.isUndefined() || (mChannelInfo.isMap() && mChannelInfo.size() == 0))
+    if (mChannelInfo.has("channel_uri")
+        && (!channelInfo.has("channel_uri") || mChannelInfo["channel_uri"] != channelInfo["channel_uri"]))
+    {
+        return;
+    }
+    if (mChannelInfo.isUndefined()
+        || (mChannelInfo.isMap() && mChannelInfo.size() <= 1)) // p2p will have uri beforehand
     {
         mChannelInfo = channelInfo;
     }
@@ -768,6 +774,11 @@ LLVoiceChannelP2P::LLVoiceChannelP2P(const LLUUID      &session_id,
     mReceivedCall(FALSE),
     mOutgoingCallInterface(outgoing_call_interface)
 {
+    std::string sip_uri = LLVoiceClient::getInstance()->sipURIFromID(other_user_id);
+    if (!sip_uri.empty())
+    {
+        mChannelInfo["channel_uri"] = sip_uri;
+    }
 }
 
 void LLVoiceChannelP2P::handleStatusChange(EStatusType type)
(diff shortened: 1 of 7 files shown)


marchcat (approved):

indra/newview/llvoicechannel.cpp
```diff
@@ -122,7 +122,13 @@ void LLVoiceChannel::onChange(EStatusType type, const LLSD& channelInfo, bool pr
 {
     LL_DEBUGS("Voice") << "Incoming channel info: " << channelInfo << LL_ENDL;
     LL_DEBUGS("Voice") << "Current channel info: " << mChannelInfo << LL_ENDL;
-    if (mChannelInfo.isUndefined() || (mChannelInfo.isMap() && mChannelInfo.size() == 0))
+    if (mChannelInfo.has("channel_uri")
```
roxanneskelly, line 125: Because you'd also need to compare channel type (webrtc, vivox, or default) and such, this should probably use LLVoiceClient::compareChannels as used below.  You may be able to get away with moving the mChannelInfo assignment below the compareChannels or something.

indra/newview/llvoicechannel.cpp
```diff
@@ -768,6 +774,11 @@ LLVoiceChannelP2P::LLVoiceChannelP2P(const LLUUID      &session_id,
     mReceivedCall(false),
     mOutgoingCallInterface(outgoing_call_interface)
 {
+    std::string sip_uri = LLVoiceClient::getInstance()->sipURIFromID(other_user_id);
```
roxanneskelly, line 777: This works, but the intent is to keep mChannelInfo opaque as far as anything outside of llvoiceclient.cpp and the voice providers. 

Would probably be better to change sipURIFromID into a different call which returns a full LLSD mChannelInfo blob (including voice server type, etc.)

roxanneskelly (changes requested):

indra/newview/llvoicechannel.cpp
akleshchev: The problem is: there are no full credencials yet so a 'full' blob doesn't appear to be an option. I will double check that, but I likely will go with some kind of `LLVoiceClient::getInstance()->getTemplateChannelInfo()`

roxanneskelly: That works for me.

indra/newview/llvoicechannel.cpp
akleshchev: For what cases is that channel assigment needed? Is it webrtc specific?

roxanneskelly: There is a case where you may transition from Vivox to WebRTC when you're in a p2p call, resulting in a different channel uri.  Not quite working yet, but that's the type of stuff I was shooting for.

akleshchev: Done. Works for Vivox. Webrtc had no such issue so I'm returning LLSD() there. 

>There is a case where you may transition from Vivox to WebRTC when you're in a p2p call

That seems like it should be calling setChannelInfo, not be assigned on a general listener that processes events from multiple channes.

Milestone was retarretargeted to webrtc, so I rebased to web rtc's branch.

roxanneskelly (approved):
Okay, I'm okay with this, as long as it works with webrtc on the beta grid.

akleshchev:
> This works, but the intent is to keep mChannelInfo opaque as far as anything outside of llvoiceclient.cpp

Should I make that info private to voice client? At the moment it's easy to access, if we limit access potential maintainers will check why instead of just using it.

PR #2120, opened 2024-07-26
Fix a bunch of XUI warnings

Ansariel:
* PBR region terrain creates dummy widgets and thus never hooking up the commit callback properly to call LLPanelRegionInfo::onChangeAnything
* Telehub checkbox doesn't exist on the world map for quite a while
* Cancel button doesn't exist for group profiles

The region terrain issues also exists in Atlasaurus.


akleshchev (approved):

marchcat (approved):
Thank you, @Ansariel!

PR #2122, opened 2024-07-26
#2100 BugSplat Crash #1497210: LLVOAvatar::updateImpostors()(10616)

LLGuru:
Key ideas
1. Make `LLCharacter::sInstances` a `std::list` instead of a `std::vector`
2. Avoid of using a temporary copy of the global static array (crash fix)

### DIFF
--- a/indra/llcharacter/llcharacter.cpp
+++ b/indra/llcharacter/llcharacter.cpp
@@ -38,7 +38,7 @@
 
 LLStringTable LLCharacter::sVisualParamNames(1024);
 
-std::vector< LLCharacter* > LLCharacter::sInstances;
+std::list< LLCharacter* > LLCharacter::sInstances;
 bool LLCharacter::sAllowInstancesChange = true ;
 
 //-----------------------------------------------------------------------------
@@ -53,7 +53,6 @@ LLCharacter::LLCharacter()
     mSkeletonSerialNum( 0 )
 {
     llassert_always(sAllowInstancesChange) ;
-    sInstances.push_back(this);
 
     mMotionController.setCharacter( this );
     mPauseRequest = new LLPauseRequestHandle();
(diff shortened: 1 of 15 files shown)


akleshchev:
My impression is that LLVO classes aren't supposed to be threaded

> BugSplat Crash #1497210
> Protect the access to LLCharacter::sInstances using LLSharedMutex

I found no signs of threading on a superficial check. How is this supposed to fix the crash?

marchcat (approved):
The improvements look reasonable.

LLGuru:
> How is this supposed to fix the crash? Have you found the cause?

This allows us to implement the step 3:
```
3. Avoid of using a temporary copy of the global static array (crash fix)
```

akleshchev:
> Avoid of using a temporary copy of the global static array (crash fix)

This implies that something modifes list from inside the loop? It looks like write mutex would just dead lock waiting for the loop to free the shared mutex.

LLGuru:
> It looks like write mutex would just dead lock waiting for the loop to free the shared mutex.

No dead-lock is expected in case of using an exclusive (modifying) lock in the same thread
If (theoretically) an exclusive lock will be used in other thread, it should lead to a safe waiting
How it will work in coroutines, QA will have the final say

akleshchev:
> No dead-lock is expected in case of using an exclusive (modifying) lock in the same thread

As far as I can tell this list is exclusive to main thread. There is no need for thread protections. For mutexes to work correctly with coroutines you need to use fibers::mutex.

Ansariel:
Why all these dynamic_casts to LLVOAvatar? Pretty sure we know all instances are LLVOAvatar and we can use a static_cast

akleshchev:
> Pretty sure we know all instances are LLVOAvatar and we can use a static_cast

That is correct, LLCharacter is virtual, it's always LLVOAvatar

LLGuru:
> Why all these dynamic_casts to LLVOAvatar? Pretty sure we know all instances are LLVOAvatar and we can use a static_cast

What is fair today can be an erroneous statement tomorrow

If we create in future the other class of characters and add these instances to the same container,
we will add these checks everywhere

The correctly written source code shouldn't depend of such situations

Ansariel:
If at any point in the future there will be a change that could require a dynamic downcast it can be changed accordingly. Since it is not used now and in the past 20 years, it is a much better way to use a static cast in terms of performance reasons.

LLGuru:
> it's always LLVOAvatar since that's the only derived class.

To be sure in that I moved `push_back` and `erase` to constructor and destructor of `LLVOAvatar`
Bonus: objects are being added only after they are fully constructed, and they are being removed before they started destroying

akleshchev:
> No dead-lock is expected in case of using an exclusive (modifying) lock in the same thread

I tested it and 'deadlock' is possible, but implementation is good and throws an LL_ERRS

Ansariel:
So we have the choice between a random crash and an LL_ERRS crash with decreased performance due to acquiring a lock on a per-frame basis in worst case - is that what we are talking about?

akleshchev (approved):

akleshchev:
> Do we know if there is a different thread other than the main thread modifying the list and which one it is?

Guru removed mutexes. As far as I know there is no theads which is why I'm against mutexes in this case.

Ansariel:
> > Do we know if there is a different thread other than the main thread modifying the list and which one it is?
> 
> Guru removed mutexes. As far as I know there is no theads which is why I'm against mutexes in this case.

It's still defined in LLCharacter though

PR #2125, opened 2024-07-27
Fix warning for Type mismatch in LLMsgVarData::addData for SpaceIP

RyeMutt:
Fixes warning about type mismatch when constructing FindAgent packet from finding another avatar on map

`2024-07-27T00:23:22Z WARNING # llmessage/llmessagetemplate.cpp(42) LLMsgVarData::addData : Type mismatch in LLMsgVarData::addData for SpaceIP`

Correct type is IPAddr as shown in message template https://github.com/secondlife/viewer/blob/develop/scripts/messages/message_template.msg#L5689


marchcat (approved):
Thank you!

PR #2126, opened 2024-07-27
Fix rapidly triggering gestures can make them get stuck

Ansariel:
See https://jira.firestormviewer.org/browse/FIRE-5819 for details


marchcat (approved):
Thank you!

PR #2127, opened 2024-07-27
Fix running multiple viewer instances stealing sound from first instance

Ansariel:
See https://jira.firestormviewer.org/browse/FIRE-480 for details


marchcat (approved):
Thanks!

PR #2128, opened 2024-07-27
Fix trying to bind to non-existent texunit sampler during water draw

RyeMutt:
This fixes trying to bind to a texture sampler that doesn't exist in the shader. Noticed it in debug log spam.

`2024-07-27T16:41:17Z DEBUG # llrender/llrender.cpp(1500) LLRender::getTexUnit : Non-existing texture unit layer requested: 4294967295`


marchcat (approved):
Thank you!

cosmic-linden (approved):
I notice there is still a call to `shader->disableTexture(LLShaderMgr::DIFFUSE_MAP);`. Is that still needed? (See also `beginPostDeferredPass`)

RyeMutt:
Good catch, cleaned that up too.

cosmic-linden (approved):

PR #2132, opened 2024-07-28
Fix callbacks for PBR region terrain panel not getting wired up properly

Ansariel:


akleshchev (approved):

akleshchev:
Build fails and blocks merge due to trailing whitespace

```
Fixing indra/llwebrtc/llwebrtc.h
Fixing indra/llwebrtc/llwebrtc_impl.h
Fixing indra/llwebrtc/llwebrtc.cpp
```

I'm pulling a fix into webrtc then will pull changes from webrtc to atlasaurus. Sorry for the inconvinience, you can either cherry-pick [fix](https://github.com/secondlife/viewer/commit/88f52f1b90af1813dc70525f2e08b16cc5939af5) in manually or wait and pick it up once it gets to atlasaurus.

Ansariel:
How can the check complain about files I a) didn't touch, b) must have been failing before when WebRTC got merged last time and c) if I check them, Notepad++ tells me there are no trailing whitespaces?

akleshchev:
> How can the check complain about files I a) didn't touch and

Unfortuantely the check is new, unpolished and bit too intrusive as a result. I agree that it shouldn't be checking everything.

> must have been failing before when WebRTC got merged last time?

WebRTC doesn't verify whitespaces yet, but that's where the files are from, so I'm pushing a fix to webrtc first..

marchcat (approved):
Thank you!

marchcat:
The check verifies the entire codebase, but it's called "pre-commit" for a reason — it blocks committing (unless forced) and, in the case of trailing whitespaces, fixes all the issues right away. You just need to add these changes to the commit.

The problem is, not all the developers are using it properly yet..

marchcat:
@brad-linden can we still merge into atlasaurus?

akleshchev:
> can we still merge into atlasaurus?

There is a couple serious issues inherited from webrtc so I assume at minimum there will be one more merge.

Ansariel:
> > you rather want to run this as a local pre-commit script
> 
> Yes, that's exactly what is intended. Please see https://github.com/secondlife/git-hooks for more info. You only need to install it once when cloning a repo. In this case, it will check only your changes, which is also helpful. However, it's easy to add `pre-commit run --all-files` into any integration script you're using.

Nice that it is mentioned like nowhere in any instruction document! 😉

marchcat:
> Nice that it is mentioned like nowhere in any instruction document! 😉

Thanks for mentioning that, time to review the docs...

PR #2133, opened 2024-07-28
Add ability to compile and run tests on Linux

Nicky-D:
- Previously this failed due to errors in lldir_test.cpp
- test_httprequest.hpp cannot run on Linux for the same reasons it was already disabled on Mac.


akleshchev (approved):

marchcat (approved):
Thank you, @Nicky-D!

PR #2135, opened 2024-07-28
Update to Boost 1.85

RyeMutt:
This updates to Boost 1.85 and fixes related deprecation warnings.

This also reverts part of https://github.com/secondlife/viewer/commit/eaa04bddeb6ca3b0db14854c216c6c4ffe0319f4 and outputs the lltest binary to the sharedlibs staging dirs to use the correct MSVC redist to fix tests with new boost.

This will need verifying with a GHA build run with tests to ensure tests work correctly in that environment as I can only use free runners currently.

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -104,11 +104,11 @@
             <key>archive</key>
             <map>
               <key>hash</key>
-              <string>d8d9e1e15ec09c81acfa9ffb80c3f20435373543</string>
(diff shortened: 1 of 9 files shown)


akleshchev (approved):

akleshchev:
Thank you! Looks like this needs to be done regardless of tests, so pulling in as is, but will wait with url change until I'm sure this is working.

PR #2147, opened 2024-07-29
build.yaml: Fix FROM_FORK value, ReleaseOS PR builds

bennettgoble:
Fix up builds from PRs by adding the missing `FROM_FORK` environment variable which must have gone missing during a rebase.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -156,8 +156,6 @@ pre_build()
 
         if [[ "$arch" == "Darwin" ]]
         then
-            RELEASE_CRASH_REPORTING=ON
-            HAVOK=ON
             SIGNING=("-DENABLE_SIGNING:BOOL=YES" \
                           "-DSIGNING_IDENTITY:STRING=Developer ID Application: Linden Research, Inc.")
         fi
--- a/build.sh
+++ b/build.sh
@@ -149,8 +149,10 @@ pre_build()
     RELEASE_CRASH_REPORTING=OFF
     HAVOK=OFF
     SIGNING=()
-    if [[ "$variant" == "Release" ]]
+    if [[ "$variant" != *OS ]]
     then
+        # Proprietary builds
+
         RELEASE_CRASH_REPORTING=ON
         HAVOK=ON
 
--- a/build.sh
+++ b/build.sh
@@ -146,12 +146,21 @@ pre_build()
     && [ -r "$master_message_template_checkout/message_template.msg" ] \
     && template_verifier_master_url="-DTEMPLATE_VERIFIER_MASTER_URL=file://$master_message_template_checkout/message_template.msg"
 
-    RELEASE_CRASH_REPORTING=ON
-    HAVOK=ON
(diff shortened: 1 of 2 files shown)


bennettgoble:
Looks like macOS ReleaseOS may be broken? The ostensibly open source build is attempting to install a proprietary library, havok-source.

akleshchev (approved):

akleshchev:
For some reason -DHAVOK:BOOL=ON' is set, but shouldn't be.

> '-DCMAKE_BUILD_TYPE:STRING=Release'

Shouldn't this be ReleaseOS?

marchcat:
It's being set at [build.sh#L150](https://github.com/secondlife/viewer/blame/5809b391434245e9c15688a6d4678a827eff72df/build.sh#L150). However, I don't see any code there to disable it. How do we build ReleaseOS on windows?"

akleshchev:
That `HAVOK=ON` does look like it should be conditional instead of always ON.

marchcat:
That's right. Probably something like this at [build.sh#L152](https://github.com/secondlife/viewer/blob/5809b391434245e9c15688a6d4678a827eff72df/build.sh#L152)
```
if [[ "$arch" == "Darwin" ]]
then
    if [[ "$variant" == "Release" ]]
    then
        SIGNING=("-DENABLE_SIGNING:BOOL=YES" \
                 "-DSIGNING_IDENTITY:STRING=Developer ID Application: Linden Research, Inc.")
    else
        # disable havok for non-release builds
        HAVOK=OFF
    fi
fi

```

build.sh
```diff
@@ -146,12 +146,21 @@ pre_build()
     && [ -r "$master_message_template_checkout/message_template.msg" ] \
     && template_verifier_master_url="-DTEMPLATE_VERIFIER_MASTER_URL=file://$master_message_template_checkout/message_template.msg"
 
-    RELEASE_CRASH_REPORTING=ON
-    HAVOK=ON
+    RELEASE_CRASH_REPORTING=OFF
+    HAVOK=OFF
```
marchcat, line 150: That's even better than I thought.

build.sh
```diff
@@ -146,12 +146,21 @@ pre_build()
...
-                  "-DSIGNING_IDENTITY:STRING=Developer ID Application: Linden Research, Inc.")
+    if [[ "$variant" == "Release" ]]
+    then
+        RELEASE_CRASH_REPORTING=ON
+        HAVOK=ON
+
+        if [[ "$arch" == "Darwin" ]]
+        then
+            RELEASE_CRASH_REPORTING=ON
+            HAVOK=ON
```
marchcat, line 160: These two look redundant here, but it's fine.

marchcat (approved):
I see `'-DHAVOK:BOOL=OFF'` in the log, so the fix works.

build.sh
bennettgoble: Whoops. Fixed, thanks.

akleshchev (approved):

PR #1843, opened 2024-06-25
1836 dont store texture in system memory unless absolutely necessary

RunitaiLinden:

PR #2150, opened 2024-07-30
1836 dont store texture in system memory unless absolutely necessary (#1843) local crash fix

LLGuru:
This llassert can raise a crash (in dev build) when rotating the world moving camera far from the target

In fact the map `sTextureAllocs` is never used - so any algorithm replacing llassert will fix the crash well

### DIFF
--- a/indra/llrender/llimagegl.cpp
+++ b/indra/llrender/llimagegl.cpp
@@ -76,10 +76,18 @@ void LLImageGLMemory::alloc_tex_image(U32 width, U32 height, U32 pixformat)
     llassert(size >= 0);
 
     sTexMemMutex.lock();
-    llassert(sTextureAllocs.find(texName) == sTextureAllocs.end());
 
-    sTextureAllocs[texName] = size;
-    sTextureBytes += size;
+    auto iter = sTextureAllocs.find(texName);
+    if (iter == sTextureAllocs.end())
+    {
+        sTextureBytes += size;
+        sTextureAllocs.emplace(texName, size);
+    }
+    else
+    {
+        sTextureBytes += size - iter->second;
+        iter->second = size;
+    }
 
     sTexMemMutex.unlock();
 }


akleshchev (commented):
This assert triggers from `LLImageGL::scaleDown` and is a sign that `free_tex_image(mTexName);` didn't work. Since it didn't work viewer is doing a 'false flag' `free_tex_image(mTexName);` then your code is cleaning again. It shouln't be cleaning twice.

From my point of view is that mTexName is storing a wrong value thus viewer is clening wrong texture.

marchcat (dismissed):

akleshchev:
> If the assert is invalid, should we just be removing that?

@vir-linden Asset is valid. Issue: `free_tex_image` gets called for a wrong id, thus `alloc_tex_image` asserts due to memory not being freed. Proper fix: call free_tex_image for right id (not calling it doesn't appear to be an option since current id isn't nessesary identical to current one).

RunitaiLinden (commented):
This isn't a good fix... it's a pre-condition of alloc_tex_image that no allocation exists in sTextureAllocs.  The root cause is somewhere else and needs to be addressed.

This might be a race condition... 

In LLImageGL::deleteTextures, there's room for the GL to re-issue a texName before free_tex_images is called for that texname:

```glDeleteTextures((GLsizei) sFreeList[idx].size(), sFreeList[idx].data());
free_tex_images((GLsizei) sFreeList[idx].size(), sFreeList[idx].data());
```

Maybe flip the order of those calls or lock sTexMemMutex for the duration of deleteTextures.  Maybe also add an llassert(glIsTexture(texName) to free_tex_image and make it a precondition that the texName hasn't been released back to GL before it has been removed from sTextureAllocs

akleshchev (changes requested):
Agree with Davep, this fixes a symptom and removing an assert mask an actual issue.

PR #2154, opened 2024-07-30
Represent the many "LLAgent" "setCameraParams" args in an array.

nat-goodspeed:
This encapsulates the boilerplate associated with passing each distinct parameter to its corresponding LLFollowCamMgr method.


maxim-productengine (approved):
I like that helper function!

maxim-productengine (approved):

PR #2176, opened 2024-08-02
Update viewer-font and emojibase packages for Twemoji with Unicode 15

RyeMutt:
This updates the Twemoji font and related emojibase package with support for Unicode 14/15 emoji such as 🪼 🫛 🪇 🫚 🫎


marchcat (approved):
Thank you!

PR #2177, opened 2024-08-02
Remove unused pcre library dependency

RyeMutt:
Remove unused PCRE library dependency it was dropped with https://github.com/secondlife/3p-colladadom/pull/7


marchcat (approved):
Yay for the cleanup!

Thanks!

PR #2178, opened 2024-08-02
Disable EmojiHelper if Digit before Colon

TommyTheTerrible:
Typing in times with the new emoji helper pop-up can be frustrating, so I would like to propose disabling it when there is a number before the colon.


indra/llui/llemojihelper.cpp
```diff
@@ -76,7 +76,9 @@ bool LLEmojiHelper::isCursorInEmojiCode(const LLWString& wtext, S32 cursorPos, S
         shortCodePos--;
     }
 
-    bool isShortCode = (L':' == wtext[shortCodePos - 1]) && (cursorPos - shortCodePos >= 2);
+    bool isShortCode = (cursorPos - shortCodePos >= 2) && (L':' == wtext[shortCodePos - 1]);
+    if(isShortCode && (shortCodePos >= 2) && isdigit(wtext[shortCodePos - 2])) // <TS:3T> Add qualifier to avoid emoji pop-up when typing times.
+        isShortCode = false;
```
marchcat, line 81:
```suggestion
        isShortCode = false;
```

marchcat (approved):

PR #2183, opened 2024-08-02
Use dedicated property to determine if build is called from a fork

marchcat:


marchcat:
The test build from the fork succeeded as well, starting the ReleaseOS builds.

nat-goodspeed (approved):

bennettgoble (approved):
TIL! Nice.

PR #2184, opened 2024-08-02
Fix integer underflow causing issues with wearables

Ansariel:
@marchcat Can fixes still be merged into Atlasaurus?

The statement `size-max_per_type` in line 2139 could cause an integer underflow and result in incorrect wearables getting filtered out (usually all clothing layers). This will cause issues with every feature using LLAgentWearables::getWearableFromItemID(), e.g. LLAgentWearables::isWearingItem()


marchcat (approved):
There's probably still a chance. Worst case - the fix will just go to `develop`.

Thank you!

marchcat:
@brad-linden, what do you think?

brad-linden (approved):
I'm fine with this going into atlasaurus

PR #2185, opened 2024-08-02
Add 'LLPanelLogin' 'login', 'savedLogins' operations.

nat-goodspeed:
'login' accepts optional 'username', 'slurl', 'grid'.

'savedLogins' returns the list of saved usernames in both display form and
internal form.

Make LLPanelLogin::getUserName() accept (const LLPointer<LLCredential>&).
There's a whole separate discussion pending as to whether const LLPointer<T>
should provide access to non-const T methods.

Similarly, make LLCredential::getIdentifier() a const method. These two
changes enable read-only access to credentials.

Make LLPanelLogin methods capture and reuse LLGridManager::instance() as
appropriate.

Add require/login.lua and test_login.lua.


nat-goodspeed:
Maxim, I'm having trouble with two things.

The first is that switching grids in the 'login' operation doesn't work. I've cribbed lots of code from LLPanelLogin, but clearly I still don't have the interactions right.

The second is that the 'savedLogins' operation returns an empty 'logins' array (which translates into Lua as the 'logins' key missing altogether) -- even when I'm at a login screen whose pulldown shows multiple saved logins.

I would appreciate your insight as to how to improve each of these.

Finally, login.lua still needs a bit of additional code to defend against startup state beyond 'STATE_LOGIN_WAIT'. Once we get past the login screen, the LLPanelLogin instance goes away, and so does its LLPanelLoginListener instance. When that happens, posting to 'LLPanelLogin' produces _no_ response, so a calling Lua script just hangs there. I can do that, but not today.

maxim-productengine:
So, grid selection works correctly. You just used the wrong name. For example, I tested with: `grid='util.aditi.lindenlab.com'` and `grid='util.aditi.lindenlab.com'`.

indra/newview/llpanelloginlistener.cpp
```diff
@@ -43,9 +53,177 @@ LLPanelLoginListener::LLPanelLoginListener(LLPanelLogin* instance):
...
+    std::string grid = (args.has("grid")? args["grid"].asString()
+                        : LLGridManager::instance().getGrid());
+    if (! gSecAPIHandler->hasCredentialMap("login_list", grid))
+    {
+        LL_INFOS() << "no creds for " << grid << LL_ENDL;
+        return llsd::map("error",
+                         stringize("No saved logins for grid ", std::quoted(grid)));
+    }
+    LLSecAPIHandler::credential_map_t creds;
+    gSecAPIHandler->loadCredentialMap("login.list", grid, creds);
```
maxim-productengine, line 214: Should be "login_list". Oof, I spent so much time before noticing it...

maxim-productengine (approved):

PR #2187, opened 2024-08-03
Reduce Avatar Loading Checks

TommyTheTerrible:
This simple trick stops checking if the avatar is loaded if the avatar is loaded, decreasing calculation times and increasing FPS.


marchcat (approved):
Thank you!

PR #2188, opened 2024-08-03
Feature/warning cleansweep

Nicky-D:
- Remove what seems obsolete warning suppression of -Wno-dangling-pointer (suppressing this warning is not a good idea anyway)
- Fix some python warnings of normal strings containing unknown/illegal escape sequences
- Do not put a whole object of type LLRandLagFib2281 into the TLS space, one some systems this leads to TLS space exhaustion


marchcat (approved):
Thank you!

indra/llcommon/llrand.cpp
```diff
@@ -58,9 +58,26 @@
...
  */
 
-// gRandomGenerator is a stateful static object, which is therefore not
+// pRandomGenerator is a stateful static object, which is therefore not
 // inherently thread-safe.
-static thread_local LLRandLagFib2281 gRandomGenerator(LLUUID::getRandomSeed());
+//We use a pointer to not construct a huge object in the TLS space, sadly this is necessary
+// due to libcef.so on Linux being compiled with TLS model initial-exec (resulting in
+// FLAG STATIC_TLS, see readelf libcef.so). CEFs own TLS objects + LLRandLagFib2281 then will exhaust the
+// available TLS space, causing media failure.
```
nat-goodspeed, line 66: How does exhausted TLS space manifest? Would there be anything in the viewer log, or would we have to look in system logs?

We're investigating a Mac crash bug that may (or may not) be related to TLS consumption.

PR #2198, opened 2024-08-05
Lua UI visibility api

maxim-productengine:


indra/llui/llcommandmanager.cpp
```diff
@@ -189,3 +189,13 @@ bool LLCommandManager::load()
...
 }
+
+LLSD LLCommandManager::getCommandNames()
+{
+    LLSD cmd_names;
+    for (auto &it : mCommands)
+    {
+        cmd_names.append(it->name());
+    }
+    return cmd_names;
```
nat-goodspeed, line 200: or
```
return llsd::toArray(mCommands, [](const auto& cmd){ return cmd.name(); });
```

indra/llui/llfloaterreg.cpp
```diff
@@ -607,3 +607,13 @@ U32 LLFloaterReg::getVisibleFloaterInstanceCount()
...
 }
+
+LLSD LLFloaterReg::getFloaterNames()
+{
+    LLSD names;
+    for (auto &it : sGroupMap)
+    {
+        names.append(it.first);
+    }
+    return names;
```
nat-goodspeed, line 618: or
```
return llsd::toArray(sGroupMap, [](const auto& pair){ return pair.first; });
```

indra/newview/lluilistener.cpp
```diff
@@ -97,6 +99,41 @@ LLUIListener::LLUIListener():
...
         &LLUIListener::addMenuSeparator,
         llsd::map("parent_menu", LLSD(), "reply", LLSD()));
+
+    add("setMenuVisible",
+        "Set menu [\"name\"] visibility to [\"visible\"]",
+        &LLUIListener::setMenuVisible,
+        llsd::map("name", LLSD(), "visible", LLSD(), "reply", LLSD()));
+
+    add("defaultToolbars",
+        "todo: defaultToolbars desc",
```
nat-goodspeed, line 109: ...of course this...

indra/newview/lluilistener.cpp
```diff
@@ -97,6 +99,41 @@ LLUIListener::LLUIListener():
...
+    add("clearToolbars",
+        "Clear all buttons off the toolbars",
+        &LLUIListener::clearAllToolbars);
+
+    add("addToolbarBtn",
+        "Add [\"btn_name\"] toolbar button to the [\"toolbar\"]\n"
+        "[1 (left toolbar), 2 (right toolbar), 3 (bottom toolbar)]\n"
+        "Position of the command in the original list can be specified as [\"rank\"]",
+        &LLUIListener::addToolbarBtn,
+        llsd::map("btn_name", LLSD(), "reply", LLSD()));
```
nat-goodspeed, line 121: How are `"btn_name"` buttons normally defined? Can we add a `"btn_name"` that invokes a Lua script?

Since `"toolbar"` is optional, mention that the default is bottom -- that's not at all obvious.

Wouldn't it be better to accept a string `"left"`, `"right"`, `"bottom"`?

It's worth mentioning (especially for Lua callers) that `rank` starts at 0.

indra/newview/lluilistener.cpp
```diff
@@ -280,3 +317,60 @@ void LLUIListener::addMenuSeparator(const LLSD&event) const
...
+{
+    Response response(LLSD(), event);
+
+    typedef LLToolBarEnums::EToolBarLocation ToolBarLocation;
+    ToolBarLocation toolbar = ToolBarLocation::TOOLBAR_BOTTOM;
+    if (event.has("toolbar"))
+    {
+        toolbar = llclamp((ToolBarLocation)event["toolbar"].asInteger(), ToolBarLocation::TOOLBAR_NONE, ToolBarLocation::TOOLBAR_BOTTOM);
+    }
+    S32 rank = event.has("rank") ? event["rank"].asInteger() : - 1;
```
nat-goodspeed, line 352: Use `RANK_NONE` instead of `-1`.

indra/newview/lluilistener.cpp
```diff
@@ -97,6 +99,41 @@ LLUIListener::LLUIListener():
...
+    add("addToolbarBtn",
+        "Add [\"btn_name\"] toolbar button to the [\"toolbar\"]\n"
+        "[1 (left toolbar), 2 (right toolbar), 3 (bottom toolbar)]\n"
+        "Position of the command in the original list can be specified as [\"rank\"]",
+        &LLUIListener::addToolbarBtn,
+        llsd::map("btn_name", LLSD(), "reply", LLSD()));
+
+    add("removeToolbarBtn",
+        "Remove [\"btn_name\"] toolbar button off the toolbar,\n"
+        "return [\"rank\"] (old position) of the command in the original list",
```
nat-goodspeed, line 125: Mention that rank `-1` means that `btn_name` was not found.

indra/newview/lluilistener.cpp
```diff
@@ -97,6 +99,41 @@ LLUIListener::LLUIListener():
         "Add menu separator to the [\"parent_menu\"] within the Top menu.",
         &LLUIListener::addMenuSeparator,
         llsd::map("parent_menu", LLSD(), "reply", LLSD()));
+
+    add("setMenuVisible",
+        "Set menu [\"name\"] visibility to [\"visible\"]",
```
nat-goodspeed, line 104: Question: I called the top-menu query method `getParents()` because the only existing use for those names was to pass `parent_menu` parameters to other methods. This method makes me wonder whether I should instead have called it `getMenus()` or `getTopMenus()`?

indra/newview/scripts/lua/require/UI.lua
```diff
@@ -150,6 +150,10 @@ function UI.addMenu(...)
     return leap.request('UI', args)
 end
 
+function UI.setMenuVisible(name, visible)
+    return leap.request('UI', {op='setMenuVisible', name=name, visible=visible})
+end
```
nat-goodspeed, line 155: I'm guessing we don't need to extract a particular item because the only reason for calling `request()` is to handle `"error"`, yes?

indra/newview/lluilistener.cpp
```diff
@@ -97,6 +99,41 @@ LLUIListener::LLUIListener():
...
+        &LLUIListener::setMenuVisible,
+        llsd::map("name", LLSD(), "visible", LLSD(), "reply", LLSD()));
+
+    add("defaultToolbars",
+        "todo: defaultToolbars desc",
+        &LLUIListener::restoreDefaultToolbars);
+
+    add("clearToolbars",
+        "Clear all buttons off the toolbars",
+        &LLUIListener::clearAllToolbars);
```
nat-goodspeed, line 114: Hmm, okay. There's no requirement to name the C++ method the same as its registered string name, and you do use them consistently. It made me pause and scroll back to cross-check when I saw `UI.clearToolbars()`, but this is correct.

indra/newview/scripts/lua/require/UI.lua
```diff
@@ -169,4 +173,61 @@ function UI.addMenuSeparator(...)
...
+    leap.send('UI', {op='clearToolbars'})
+end
+
+function UI.defaultToolbars()
+    leap.send('UI', {op='defaultToolbars'})
+end
+
+-- UI.addToolbarBtn{btn_name=btn_name
+--          [, toolbar= 3] -- 1 [TOOLBAR_LEFT], 2 [TOOLBAR_RIGHT], 3 [TOOLBAR_BOTTOM]
+--          [, rank=1]} -- position on the toolbar
```
nat-goodspeed, line 190: but `rank=1` would put the new button in the second position, yes?

nat-goodspeed (changes requested):

indra/newview/lluilistener.cpp
maxim-productengine: Toolbar buttons are defined in app_settings\commands.xml. It depends. Adding just one button and define which script it invokes - trivial. Ability to add multiple new buttons that will not persist between sessions and have custom callbacks - may be tricky. (I need to think about it).
Agree. Using "left", "right", "bottom" seems to be better.

indra/newview/scripts/lua/require/UI.lua
maxim-productengine: Yep, similar to `addMenuSeparator`

indra/newview/lluilistener.cpp
maxim-productengine: I'll change it to `clearAllToolbars` for consistency anyway, thanks.

indra/newview/scripts/lua/require/UI.lua
maxim-productengine: Yes, that's right. I'll mention it here.

indra/newview/lluilistener.cpp
maxim-productengine: I think I like `getTopMenus()` more. `getParents()` is quite general name.

indra/newview/lluilistener.cpp
nat-goodspeed: Problem of scale?

nat-goodspeed (approved):

PR #2200, opened 2024-08-05
Ignore VmallocTotal from /proc/meminfo

Nicky-D:
Ignore VmallocTotal as it's fixed, huge Number.
Not ignoring it will cause a lot of log spam due to conversion/parse errors.

### DIFF
--- a/indra/llcommon/llsys.cpp
+++ b/indra/llcommon/llsys.cpp
@@ -1107,6 +1107,14 @@ LLSD LLMemoryInfo::loadStatsMap()
                 LLSD::String key(matched[1].first, matched[1].second);
                 LLSD::String value_str(matched[2].first, matched[2].second);
                 LLSD::Integer value(0);
+
+                // Skip over VmallocTotal. It's just a fixed and huge number on (modern) systems. "34359738367 kB"
+                // https://unix.stackexchange.com/questions/700724/why-is-vmalloctotal-34359738367-kb
+                // If not skipped converting it to a LLSD::integer (32 bit) will fail and spam the logs (this function
+                // is called quite frequently).
+                if( key == "VmallocTotal")
+                    continue;
+
                 try
                 {
                     value = boost::lexical_cast<LLSD::Integer>(value_str);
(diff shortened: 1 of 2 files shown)


indra/llcommon/llsys.cpp
```diff
@@ -1106,10 +1106,10 @@ LLSD LLMemoryInfo::loadStatsMap()
                 // e.g. "MemTotal:      4108424 kB"
                 LLSD::String key(matched[1].first, matched[1].second);
                 LLSD::String value_str(matched[2].first, matched[2].second);
-                LLSD::Integer value(0);
+                U64 value(0);
```
nat-goodspeed, line 1109: The problem with this change is that the destination for `value` is the `stats.add()` call at line 1122. `stats` is defined as `Stats` at line 936. Class `Stats`, defined on line 699, stores every statistic in `LLSD` blob `mStats` (line 731). Passing `U64` as `value` engages the `add()` method at line 708, which casts `value` to `LLSD::Integer` to store into `mStats`.

When we faced a similar problem with enormous GitHub build numbers, we switched to string instead. You would need to add an `add()` method accepting `std::string`.

I don't think any consumers of this information actually depend on it being `LLSD::Integer` versus `LLSD::String`.

Nicky-D: You're right. I misread that code under the bias and assumption that Windows was already using U64 there.

I would really, really prefer not to change the type, as it's hard to predict in which dark corner of the code that could fall flat on the face.
Especially if those stats should ever get send to some LLs witches wizardry and those expects ints (We don't want to get cursed by the witches, do we?)

I inspected my machine where this conversion error constantly floods the logs:
VmallocTotal:   34359738367 kB

That's actually a bogus stat anyway https://unix.stackexchange.com/questions/700724/why-is-vmalloctotal-34359738367-kb

I'm inclined to play it safe there and not risk any deeper changes. Ignore VmallocTotal (because it is useless anyway) and be done minimal impact.

What do you think?

nat-goodspeed: > I'm inclined to play it safe there and not risk any deeper changes. Ignore VmallocTotal (because it is useless anyway) and be done minimal impact.

I think you mean close this PR and call it good?

> this conversion error constantly floods the logs

Maybe only log the `LL_WARNS()` if `(key != "VmallocTotal")` ? Or just skip the `meminfo` line altogether if `(key == "VmallocTotal")` ?

Nicky-D: > I think you mean close this PR and call it good?

Kinda. Revert the U64 change back to Integer.
Then just test for "VmallocTotal" as you wrote and skip over it. And obviously comment why it gets skipped.

nat-goodspeed (approved):

marchcat (approved):

PR #2202, opened 2024-08-06
Convert LLUI and LLRender2D to LLSimpleton to reduce overhead during UI draw

RyeMutt:
Convert LLUI and LLRender2D to LLSimpleton to reduce overhead during ui draw and construction


nat-goodspeed (approved):
Since most `LLRender2D` methods have been `static` since before this change, it's not clear why the calls you changed were using the `LLRender2D::getInstance()` incantation anyway.

I'm a little unhappy about having to add a `deleteSingleton()` call back into the static list of cleanup calls. I get that you want to avoid locking overhead on every `getInstance()` call, but `LLSingleton` dependency tracking only incurs overhead on construction and destruction. Makes me want a middle class that does the dependency tracking without retrieval locking (but with an `llassert()` verifying that every `getInstance()` call comes from the same thread).

Since we don't yet have that, let's go with this.

marchcat (approved):

PR #2209, opened 2024-08-06
Introduce a custom coroutine/fiber scheduler to prioritize UI.

nat-goodspeed:
The viewer's main thread's main fiber is responsible for coordinating just about everything. With the default round_robin fiber scheduling algorithm, launching too many additional fibers could starve the main fiber, resulting in visible lag.

This custom scheduler tracks when it switches to and from the main fiber, and at each context switch, how long it's been since the last time the main fiber ran. If that exceeds a certain timeslice, it jumps the main fiber to the head of the queue and resumes that instead of any other ready fiber.


nat-goodspeed:
Maxim, I'm really interested in your careful scrutiny of the business logic here. Because it depends so strongly on the Boost.Fibers scheduler API, and on the default `round_robin` base scheduler, please let me know if you want to pair so I can talk you through it.

The [Boost.Fiber scheduling doc](https://www.boost.org/doc/libs/1_85_0/libs/fiber/doc/html/fiber/scheduling.html) covers the `use_scheduling_algorithm()` API, the [`algorithm` abstract base class](https://github.com/boostorg/fiber/blob/88b8304e78dfab4baee5d6beca09938d2826061a/include/boost/fiber/algo/algorithm.hpp) and its methods, the [`round_robin` default scheduler](https://github.com/boostorg/fiber/blob/88b8304e78dfab4baee5d6beca09938d2826061a/include/boost/fiber/algo/round_robin.hpp) and the [`context` class](https://github.com/boostorg/fiber/blob/88b8304e78dfab4baee5d6beca09938d2826061a/include/boost/fiber/context.hpp) passed into an `algorithm` implementation.

The [`round_robin` implementation](https://github.com/boostorg/fiber/blob/88b8304e78dfab4baee5d6beca09938d2826061a/src/algo/round_robin.cpp) is also of potential interest.

The `context` class includes `boost::intrusive::list` hooks for several lists. The one used by `round_robin`, and therefore by `llcoro::scheduler`, is the ready list. I consulted the Boost.Intrusive [overview](https://www.boost.org/doc/libs/1_85_0/doc/html/intrusive/usage.html) and [list documentation](https://www.boost.org/doc/libs/1_85_0/doc/html/intrusive/list.html).

As noted in the comments, it's really hard to empirically verify correct behavior of this scheduler, which is why I'd like to ask you for a thorough review. Does this code do what I think it should? It might help to ignore the logging stuff, which should be totally orthogonal. The rest isn't a lot of lines of code.

Also - is that a reasonable place to put the `llcoro::scheduler::use()` call? Is there a better place very near the start of the viewer's main thread?

indra/newview/llstartup.cpp
```diff
@@ -400,10 +401,12 @@ bool idle_startup()
         static bool first_call = true;
         if (first_call)
         {
+            first_call = false;
             // Other phases get handled when startup state changes,
             // need to capture the initial state as well.
             LLStartUp::getPhases().startPhase(LLStartUp::getStartupStateString());
-            first_call = false;
+            // Use our custom scheduler for coroutine scheduling.
+            llcoro::scheduler::use();
```
maxim-productengine, line 409: Maybe move the call somewhere here: https://github.com/secondlife/viewer/blob/ec092bb626ecb2f293343e35209b31e363de6a5d/indra/newview/llappviewer.cpp#L768

nat-goodspeed: Thanks, I like that suggestion.

maxim-productengine: Well, I think I understood scheduling enough to say that it looks good.

On the side note.. With these changes, it will be interesting to disable our 'infinite-loop-protection' and try something like that in Lua script:
```
while true do
 print("1")
end
```

nat-goodspeed: Thank you!

I definitely value the interrupt feature that ensures other viewer coroutines get a chance even when a Lua C++ coroutine is compute-bound. I think that integrates nicely with this feature.

I still think it's a good idea to kill off a Lua coroutine that's compute-bound for too long, but maybe we can raise the limit?

maxim-productengine (approved):

indra/newview/llstartup.cpp
maxim-productengine: I suggest leaving the current limit as it is. We can raise it if we ever encounter a real 'non-synthetic' script that exceeds that limit.

indra/newview/scripts/lua/test_flycam.lua
```diff
@@ -0,0 +1,38 @@
...
+        -- usual case
+        LLAgent.setCamera{ camera_pos=cameraPos(os.clock() - start), camera_locked=true }
+        return nil
+    else
+        -- last time
+        LLAgent.removeCamParams()
+        LLAgent.setFollowCamActive(false)
+        return true
+    end
+end
```
nat-goodspeed, line 34: I think I still don't understand follow cam behavior very well. This script does indeed fly the camera in a circle around the avatar, as long as `camera_locked=true`. The trouble with that is that the avatar can walk away from the center of the circle, leaving the camera still describing the original circle while watching the avatar walk away.

I had assumed that `camera_locked=false` would mean that the camera would continue to fly in circles around the avatar even as the avatar walks around. Instead it engages in a behavior I find very weird. It hops to some position, comes closer and then moves away, then hops to a different position, comes closer and then moves away, returns to the first position, ...

I'm having trouble explaining this to myself.

indra/newview/llstartup.cpp
nat-goodspeed: Heh - I've hit that already. You can see in [test_callables.lua](https://github.com/secondlife/viewer/blob/release/luau-scripting/indra/newview/scripts/lua/test_callables.lua) that the script explicitly iterates through the table returned by `UI.callables()`. That's because the simpler construct
```
print(inspect(UI.callables()))
```
exceeded the timeout during `inspect()` processing.

maxim-productengine: Ok, now I completely agree with raising the limit.

indra/newview/scripts/lua/test_flycam.lua
maxim-productengine: Unfortunately, I don't have a quick answer, because 'follow cam' is quite new for me too. So I need to dig and test this script myself.

nat-goodspeed: > the avatar can walk away from the center of the circle, leaving the camera still describing the original circle while watching the avatar walk away.

Of course there's a simple fix for that, which I've pushed as ["Ensure that the flycam stays near moving avatar."](https://github.com/secondlife/viewer/pull/2209/commits/b4fe47a5c0abac02d161640e04a9a78afb1c5987)

PR #2213, opened 2024-08-07
Replace boost::shared_ptr with std in voice classes

akleshchev:
Three comits:
 - Replace boost::shared_ptr with std in voice classes
 - Minimized some getInstance() calls on draw and idle
 - fixed a crash on exit

### DIFF
--- a/indra/llcommon/llerrorcontrol.h
+++ b/indra/llcommon/llerrorcontrol.h
@@ -198,7 +198,7 @@ namespace LLError
     };
 
     /**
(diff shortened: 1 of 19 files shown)


maxim-productengine (approved):

marchcat (commented):
Wow, looks like we missed some BOOL->bool changes during the main pass. Thank you for noticing that!

akleshchev:
This is webrtc, not atlasaurus, I wanted to push that 'bool function() const' to atlasaurus, but decided that webrtc is the right place for it and potential conflict is simple enough to resolve.

PR #2219, opened 2024-08-07
viewer#2043 bugsplat symbol upload update

akleshchev:
Cherrypick to decode crashes


maxim-productengine (approved):

akleshchev:
strangely webrtc doesn't need it, but might move to atlasaurus

PR #2226, opened 2024-08-08
Auto-scaling amortisation of dynamic BB calcs

beqjanus:
Contributing code that has been in use in TPVs (FS and Alchemy, maybe others) for a number of years.

Limits the overhead of the dynamic BB calcs to
AvatarExtentRefreshMaxPerBatch per AvatarExtentRefreshPeriodBatch frames
default is 5 avatar per 4 frames. Thus a busy region with 25 avatars would
take 20 frames to refresh the all the BBs, keeping the per frame impact within a reasonable time quanta

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -1874,6 +1874,28 @@
     <key>Value</key>
     <real>64.0</real>
   </map>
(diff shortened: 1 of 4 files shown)


marchcat (approved):
Thank you!

PR #1561, opened 2024-05-24
Import Japanese translation for webRTC Voice from PR #1558

marchcat:
Original author: @YuzuruJewell

PR #743, opened 2024-02-06
Fix Japanese Translation

sakuranoelfayray:

PR #713, opened 2024-02-01
Update the XML files in the ja folder without changing the name attribute values of the XML files in the en folder.

YuzuruJewell:
Update the XML files in the ja folder without changing the name attribute values of the XML files in the en folder.

The files below could not be replaced with Japanese without changing the duplicate name attribute values using my tools.
In order to apply the Japanese translation without changing the XML files in the en folder, I temporarily changed the name attribute value of the xml files, translated the corresponding English to Japanese, and then changed it back to the original name attribute value.
I would like someone to fix the duplicate name attribute values on these files.

floater_about_land.xml
floater_beacons.xml
floater_model_preview.xml
floater_pathfinding_linksets.xml
floater_preferences_graphics_advanced.xml
floater_scene_load_stats.xml
floater_snapshot.xml
floater_test_layout_stacks.xml
floater_test_line_editor.xml
floater_test_text_editor.xml
floater_test_textbox.xml
floater_test_widgets.xml
floater_translation_settings.xml
floater_world_map.xml
fonts.xml
menu_navbar.xml
menu_settings_gear.xml
menu_topinfobar.xml
mime_types.xml
panel_group_creation_sidetray.xml
panel_notification_list_item.xml
panel_performance_preferences.xml
panel_preferences_advanced.xml
panel_preferences_sound.xml
panel_profile_pick.xml
panel_volume_pulldown.xml
sidepanel_item_info.xml

PR #558, opened 2023-12-04
BUG-234686 Japanese Translation Update (Inventory expansion and PBR)

logue:
This time, I translated GLTF and inventory expansion, corrected the expressions in the text (「です。」、「ます。」tone), and corrected translation mistakes.

https://jira.secondlife.com/browse/BUG-234686

PR #2232, opened 2024-08-08
Japanese translation update 2024/08

marchcat:
This PR mostly consists of the contents of PRs #558. #713. #743 processed automatically;

Other related PR: #1561


Ansariel:
Seeing a lot of attributes in the localized version that should not be there, e.g. `enabled`, `type` or `control_name`. The only attributes that should be included apart from the ones containing localized strings are the positioning attributes if special handling is needed to make the localization fit.

Issue #2229, opened 2024-08-08
BugSplat Crash #1502471: SecondLifeViewer!LLPolySkeletalDistortion::apply(196)


sl-service-account:
**BugSplat** [Crash 1502471](https://app.bugsplat.com/v2/crash?id=1502471&database=SecondLife_Viewer_2018)

**Application Version:**
7.1.9.10084807842

**Callstack:**
[...]

PR #2235, opened 2024-08-08
#2229 BugSplat Crash #1502471: SecondLifeViewer!LLPolySkeletalDistortion::apply(196)

LLGuru:


indra/newview/llvoavatar.cpp
```diff
@@ -3081,30 +3081,15 @@ void LLVOAvatar::idleUpdateAppearanceAnimation()
         }
         else
         {
-            F32 morph_amt = calcMorphAmount();
-            LLVisualParam *param;
-
             if (!isSelf())
             {
+                F32 morph_amt = calcMorphAmount();
```
akleshchev, line 3086: May be make it `calcMorphAmount() const` while at it?

indra/llcharacter/llcharacter.cpp
```diff
@@ -65,7 +65,9 @@ LLCharacter::LLCharacter()
 //-----------------------------------------------------------------------------
 LLCharacter::~LLCharacter()
 {
-    for (const auto& it : mVisualParamIndexMap)
+    visual_param_index_map_t local_map;
+    std::swap(local_map, mVisualParamIndexMap);
+    for (const auto& it : local_map)
```
akleshchev, line 70: Please add a meaningfull comment, why is this needed in destrucor? Fix implies that visual param destructor can alter mVisualParamIndexMap, but that doesn't appear to be true. And I see even less relation to the crash report.

akleshchev (approved):

PR #2240, opened 2024-08-09
Lua api for sending group messages

maxim-productengine:


bennettgoble:
Hey, @maxim-productengine. Thanks for the PR --could you include an example of how this is used in your PR description? This would help reviewers quickly understand what the new API looks like!

indra/newview/scripts/lua/require/LLChat.lua
```diff
@@ -14,4 +18,25 @@ function LLChat.sendShout(msg)
     leap.send('LLChatBar', {op='sendChat', type='shout', message=msg})
 end
 
+-- 0 is public nearby channel, other channels are used to communicate with LSL scripts
+function LLChat.sendChannel(msg, channel)
+    leap.send('LLChatBar', {op='sendChat', channel=channel, message=msg})
+end
```
nat-goodspeed, line 24: As far as I'm concerned, you could just add a `channel` parameter to `sendNearby()` and add `channel=channel` to the event table at line 10. If the caller passes only the `msg` parameter, `channel` will be `nil`, and setting a Lua table key to `nil` deletes it from the table -- making it exactly like the present `sendNearby()` event table.

I like the comment, though. Definitely keep that.

indra/newview/scripts/lua/test_group_chat.lua
```diff
@@ -0,0 +1,16 @@
+LLChat = require 'LLChat'
+inspect = require 'inspect'
+LLAgent = require 'LLAgent'
+popup = require 'popup'
+
+local OK = 'OK_okcancelbuttons'
```
nat-goodspeed, line 6: Should we provide aliases in the `popup` module? To be honest, I'm not sure why the notification response strings are so awkward.

nat-goodspeed (approved):
Nice. See also the Throttle proposal in #2265.

indra/newview/scripts/lua/test_group_chat.lua
maxim-productengine: Yep, I was thinking about it too. It would be better to provide handy built-in check, instead of using this 'magic' `OK_okcancelbuttons`.

indra/newview/scripts/lua/require/LLChat.lua
maxim-productengine: Sure, it makes sense.

indra/newview/groupchatlistener.h
```diff
@@ -26,15 +26,22 @@
...
+class LLGroupChatListener: public LLEventAPI
 {
 public:
-    GroupChatListener();
+    LLGroupChatListener();
+
+private:
+    void startGroupChat(LLSD const &data);
+    void leaveGroupChat(LLSD const &data);
+    void sendGroupIM(LLSD const &data);
```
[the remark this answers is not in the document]
maxim-productengine: We can skip `startGroupChat `step, if group chat has been already opened manually.
We don't have 'group chat listener' for now. So, if/when we add one, it's probably a good idea to provide a list of current opened group conversations.

PR #2247, opened 2024-08-11
Fix avatar lod corruption issue 

RyeMutt:
Fix avatar lod getting stuck/corrupted from https://github.com/secondlife/viewer/pull/2238 where the isAvatar check was mistakenly inverted


marchcat (approved):
Thank you!

PR #2249, opened 2024-08-11
Fix sculpt discard level being discarded due to shadowing outer scope discard level

### LINKED ISSUE #2024: Sculpts getting corrupted by texture downres
### Environment

User report, don't have this info

### Description

Arton Rotaru reports that some Davep recent fixes to develop are causing sculpt maps to drop res, never come back to full rez. Says they drop after 10 minutes or so.

### Reproduction steps

Test in a region with sculpted prims

RyeMutt:
This may fix #2024. Noticed this while fixing warnings.


marchcat (approved):
Thank you!

PR #2252, opened 2024-08-12
CI: Upload ReleaseOS app

bennettgoble:
Upload the built app on ReleaseOS builds. Not sure why we wouldn't want this.


akleshchev (approved):

nat-goodspeed:
In the weeks following migration from TeamCity to GitHub, shock & horror were manifested by those who pay bills at how much more GH CI was costing us than anticipated. Some short cuts were implemented in hopes of making a difference, without careful analysis of the actual impact.

PR #2265, opened 2024-08-12
Add Throttle and LogThrottle classes to manage throttled APIs.

nat-goodspeed:


nat-goodspeed:
Seeing the pattern of functions being throttled, I've been thinking about such a wrapper for a while now. The usage is a little more awkward than I'd hoped. Is it worth pursuing this direction?

I'd like even better to have `LogThrottle` (or some other subclass) wrap a whole `LLEventAPI` subclass method, so we could directly pass the `LogThrottle` data member to `LLEventAPI::add()`. I haven't yet tried to make that work.

Another `Throttle` subclass I'd like to see would be one that knows about the `void(const LLSD&)` signature and uses `sendReply()` to notify the LEAP invoker that the call has been throttled.

maxim-productengine (approved):
I like this wrapper. I'm sure it will be handy with future apis too.

PR #2269, opened 2024-08-13
Shutdown crashes

akleshchev:
1. Made removeObserver static because floater wasn't removed from observers. Voice singleton gets removed first, so made it call rtc/vivox singletons directly (probably better to just move observers into LLVoiceClient or get rid of those internal singletons, but this is a crash fix, not a refactoring change)

2. sessions survived for too long after singleton was destroyed and cleanup got issues, fixed by cleaning up along with the singleton. Also renamed it from mSessions to sSession.


marchcat (approved):

roxanneskelly (approved):
LGTM

Issue #2271, opened 2024-08-13
BugSplat Crash #1504299: SecondLifeViewer!LLFocusableElement::~LLFocusableElement(79)


sl-service-account:
**BugSplat** [Crash 1504299](https://app.bugsplat.com/v2/crash?id=1504299&database=SecondLife_Viewer_2018)

**Application Version:**
7.1.9.10084807842

**Callstack:**
[...]

PR #2273, opened 2024-08-13
#2271 BugSplat Crash #1504299: LLFocusableElement::~LLFocusableElement(79)

LLGuru:


akleshchev (changes requested):
If it was allocated it should be deallocated, this merely introduces a memory leak.

marchcat (changes requested):
Right. It should not depend on isExiting() and should probably looks like this:
```
        if (mFocusLostCallback)
        {
            delete mFocusLostCallback;
            mFocusLostCallback = nullptr;
        }
        // and so on...
```

akleshchev:
Aditionally this crash can happen without shutdown:
https://app.bugsplat.com/v2/crash?database=SecondLife_Viewer_2018&id=1496385

marchcat (approved):

akleshchev (approved):
I don't see how this going to fix the crash. But it won't hurt and might actually reveal more info.

PR #2277, opened 2024-08-13
viewer-private#260 The 'Speak' button looks disabled during IM voice chat

akleshchev:
p2p channels was reusing obsolete channel info which resulted in challe considered to be inactive due to handle mismatch


roxanneskelly (approved):
LGTM

PR #2283, opened 2024-08-13
Reduce thread contention between main thread and asset fetch threads

RyeMutt:
Reduce thread contention between the main thread and all asset fetch threads due to frequent access to the singleton master dependency list by utilizing simpletons and statics.

Found while profiling:
image

### DIFF
--- a/indra/llfilesystem/lldiskcache.cpp
+++ b/indra/llfilesystem/lldiskcache.cpp
@@ -39,15 +39,25 @@
 
 #include "lldiskcache.h"
 
+ /**
+  * The prefix inserted at the start of a cache file filename to
+  * help identify it as a cache file. It's probably not required
+  * (just the presence in the cache folder is enough) but I am
+  * paranoid about the cache folder being set to something bad
+  * like the users' OS system dir by mistake or maliciously and
+  * this will help to offset any damage if that happens.
+  */
+static const std::string CACHE_FILENAME_PREFIX("sl_cache");
+
+std::string LLDiskCache::sCacheDir;
+
 LLDiskCache::LLDiskCache(const std::string cache_dir,
                          const uintmax_t max_size_bytes,
                          const bool enable_cache_debug_info) :
-    mCacheDir(cache_dir),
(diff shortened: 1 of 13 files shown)


indra/llfilesystem/llfilesystem.cpp
```diff
@@ -325,3 +320,64 @@ bool LLFileSystem::remove()
 
     return true;
 }
+
+void LLFileSystem::updateFileAccessTime(const std::string file_path)
```
Ansariel, line 324: Forgot to also change it to a const ref or even string_view?

RyeMutt: Good catch! Fixed!

indra/llfilesystem/lldiskcache.cpp
```diff
@@ -236,97 +246,17 @@ const std::string LLDiskCache::assetTypeToString(LLAssetType::EType at)
...
-    {
-        boost::filesystem::last_write_time(file_path, cur_time, ec);
-    }
-#endif
-
-    if (ec.failed())
-    {
-        LL_WARNS() << "Failed to update last write time for cache file " << file_path << ": " << ec.message() << LL_ENDL;
-    }
+    return llformat("%s%s%s_%s_0.asset", sCacheDir.c_str(), gDirUtilp->getDirDelimiter().c_str(), CACHE_FILENAME_PREFIX.c_str(), id.c_str());
```
Ansariel, line 251: Could probably omit the "_0" entirely and increase the cache version to trigger a cache clear.

marchcat (approved):
Thank you!

RyeMutt:
Build error on mac should be fixed now.

RyeMutt:
Windows failure appears to be CI-script related

indra/llfilesystem/lldiskcache.cpp
RunitaiLinden: If I'm reading the code right, that _0 was always there and "extra_info" was always empty, so _0 is just the status quo.

Issue #2289, opened 2024-08-14
BugSplat Crash #1496385: SecondLifeViewer!LLFocusableElement::~LLFocusableElement(79)


sl-service-account:
**BugSplat** [Crash 1496385](https://app.bugsplat.com/v2/crash?id=1496385&database=SecondLife_Viewer_2018)

**Notes:**
1) The instance of LLRootView is being destroyed as a child view of some LLTextBox
2) It seems to be the second attempt of destroying the same object (signal is already deleted)

**Application Version:**
7.1.8.9375512768

**Callstack:**
[...]

PR #2290, opened 2024-08-14
#2289 BugSplat Crash #1496385: SecondLifeViewer!LLFocusableElement::~LLFocusableElement(79)

LLGuru:


akleshchev:
I'm against modifying this class without solid reason, it works well as is, please don't try fixing wohat ain't broke.

LLGuru:
> I'm against modifying LLLayoutStack class without solid reason, it works well as is and very widespread, please don't try fixing what ain't broke.

Analyzing multiple cases with this stack id I found that they could ve different, the classes on the top of the stack are different, but all these use-cases contain that `for_each` deleting from `LLLayoutStack` destructor

I suppose that this abnormal way to delete children views could lead to a double deleting of the same view
I can't prove this right now, but I'm sure that we'd better use the standard way to delete children when possible

akleshchev:
> all these use-cases contain this dangerous for_each deleting from LLLayoutStack destructor

1. That for_each looks ugly, probably called from a wrong place. But it looks safer than your change in a sense that some some class with nonstandard resize logic can trigger resize on exit and will trigger a potentially dead mResizeBar. I think at minimum you should have modified deleteAllChildren to clean panels prior to cleaning everything else and should have set embedded_panelp->mResizeBar to null after deleting it.

2. I tested this class, surrounded destructors with logging and no 'double deleting' happened. Judging by how deletion happens, 'double deletion' is not possible. First deletion will clear child list so second one won't happen.

3. ~LLFocusableElement is confined to a dozen specific floaters, but LLLayoutStack is widespread, if LLLayoutStack was causing issues it would be causing them everywhere, not to a limited set of floaters.

akleshchev (approved):
This change now fixes a potential issue with deleteAllChildren call, I'm a bit warry of changes to this class, but it's widespread and simple enough that we will see any side effects fairly fast.

PR #2314, opened 2024-08-15
BugSplat Crash #1504587: std::vector::empty()

LLGuru:


akleshchev (approved):
This might be a wrong approach, it looks like threads are supposed to have a recorder, maybe add an assert?

PR #2331, opened 2024-08-18
Change Clang-Format settings to adhere to the used coding style

Ansariel:
* Fix whitespaces for initializer lists (`type x {value};` to `type x{ value };`)
* Change pointer alignment from right to left (`type *name` to `type* name`)


marchcat (approved):
Thank you!

PR #2333, opened 2024-08-18
Clean up excessive dynamic casts in LLAccordionCtrl

RyeMutt:
Clean up excessive dynamic casts in LLAccordionCtrl and other light cleanup such as uninitialized member variables


marchcat (commented):
Yay for the cleanup! Thanks!

marchcat (approved):

Issue #2339, opened 2024-08-19
BugSplat Crash #1502099: SecondLifeViewer!LLRenderPass::pushUntexturedBatch(656)


sl-service-account:
**BugSplat** [Crash 1502099](https://app.bugsplat.com/v2/crash?id=1502099&database=SecondLife_Viewer_2018)

**Application Version:**
7.1.9.10084807842

**Callstack:**
[...]

PR #2340, opened 2024-08-19
#2339 BugSplat Crash #1502099: LLRenderPass::pushUntexturedBatch(656)

LLGuru:


indra/newview/lldrawpool.cpp
```diff
@@ -651,6 +651,10 @@ void LLRenderPass::pushUntexturedBatch(LLDrawInfo& params)
         return;
     }
 
+    llassert(params.mVertexBuffer);
+    if (!params.mVertexBuffer)
+        return;
+
```
akleshchev, line 657: This situation doesn appear to be possible.

1. LLDrawInfo gets initialized with the buffer so it always has this buffer from the start and I see no resets or assigments outside of constructor.
2. In each buffer gets used directly before constructing LLDrawInfo, as far as I can tell supplied buffer can't be null otherwise viewer would have crashed before contructing LLDrawInfo.
3. Even if I missed something and null does get passed, based on context it looks like right solution would be to not make LLDrawInfo and just return.
4. Asserting this far is not helpfull, if you need an assert it should be in place where null buffer gets assigned, where we can unwind it to see what's going on.

akleshchev (changes requested):

vir-linden:
I agree with Caladbolg's note here - we need to understand the underlying problem and try to fix it. Skipping over the immediate cause of the crash with a low level check might prevent the crash, but isn't useful if it leaves us in a bad state.

PR #2348, opened 2024-08-19
Remove orphaned LLAllocator

Ansariel:
LLAllocator was a non-functional left-over from when tcmalloc was used


marchcat (approved):
Yay, cleanup!
Thanks!

PR #2353, opened 2024-08-19
Re-introduce llappearanceutility to viewer & add 2k bake support

rohvani:
This PR re-introduces the `llappearanceutility` to the viewer, allowing us to keep it in sync with future changes to the viewer, as well as allowing support for 2k avatar bakes.


indra/llappearanceutility/llprocesstexture.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    if ("eyes" == mInputData["slot_id"].asString())
+    {
+        mBakeSize = 512;
+    }
+    else
+    {
+        // optimization: maybe we could use maxTextureDecodedWidth and 
+        //   maxTextureDecodedHeight for new LLBakingWindow below?
```
rohvani, line 171: Maybe we could use `maxTextureDecodedWidth` and `maxTextureDecodedHeight` for the `LLBakingWindow` below?

Ansariel:
It is probably a good idea to convert all those fake BOOLs into actual booleans since they are on their way out and would reduce work when merging develop into maint-b.

There is also still a reference to llappearance_utility in autobuild.xml that probably can go then.

indra/llappearanceutility/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+      ${BOOST_COROUTINE_LIBRARY}
+      ${BOOST_CONTEXT_LIBRARY}
+      ${BOOST_SYSTEM_LIBRARY}
+      ${TCMALLOC_LIBRARIES}
+      )
+  
+  add_custom_target(appearance-utility-headless-bin-target ALL
+                    DEPENDS appearance-utility-headless-bin)
+endif (BUILD_HEADLESS)
+
```
Nicky-D, line 87: The whole file need some love, to be adapted to the newer cmake style.  IT needs to drop tcmalloc and add the appropriate targets rather than the old style variables.

indra/llappearanceutility/llbakingavatar.h
```diff
@@ -0,0 +1,142 @@
...
+ **                                                                            **
+ **                    STATE
+ **/
+public:
+    /*virtual*/ bool    isSelf() const { return true; }
+    /*virtual*/ bool    isValid() const { return TRUE; }
+    /*virtual*/ bool    isUsingBakedTextures() const { return TRUE; }
+
+/**                    State
+ **                                                                            **
```
Nicky-D, line 122: All those /*virtual*/ method really should use "override"

indra/llrender/llimagegl.cpp
```diff
@@ -1052,7 +1052,7 @@ U32 type_width_from_pixtype(U32 pixtype)
 bool should_stagger_image_set(bool compressed)
 {
 #if LL_DARWIN || LL_LINUX
-    return false;
+    return !compressed && on_main_thread() && gGLManager.mIsAMD;
```
rohvani, line 1055: This change was necessary on Linux, from what I can remember, this was causing an issue with apputil that @RunitaiLinden helped me with. Davep, do you by chance remember why we made this change? IIRC, something was really slow prior to making this change -- we were spending a lot of time around here.

cc: @brad-linden

rohvani: here's the commit: https://github.com/secondlife/viewer/pull/2353/commits/0f90b5698557a3045064728df422ba38cd3ab30a

brad-linden: yeah I just automatically resolved the conflict here, but now that you mention it, perhaps the better solution would be to spit out these cases now:
```suggestion
#if LL_LINUX
    return false;
#elif LL_DARWIN
    return !compressed && on_main_thread() && gGLManager.mIsAMD;
#else
```

PR #2356, opened 2024-08-19
Follow up for 4fa41c5c784c6e478d49c6641893e694aa7ed5c3 breaking inbox search/hiding

RyeMutt:
Fixes inbox panel not hiding when switching modes or searching

Fix by setting the two cached pointers for the panel after floater construction has finished.


marchcat (approved):
Thanks!

Issue #2354, opened 2024-08-19
Enable Tracy profiling by default on Test viewers and local builds


RunitaiLinden:
### Description

Now that our Tracy integration defaults to on demand profiling, we should default any Test viewer to have Tracy enabled.

Dan-Linden:
Closed. Not QA'd.

PR #2357, opened 2024-08-19
#2354 Enable Tracy on test builds

RunitaiLinden:
This should make it so any viewer that uses the Second Life Test channel can be profiled on-demand with Tracy.


Geenz (approved):
Seems simple enough.

Issue #2349, opened 2024-08-19
Avatar/animesh rendering correctness pass


cosmic-linden:
Tracy code investigation item.

Note: Performance impact of this change is likely not significant, however some changes will fix bugs.

## Test plan

- Keep an eye out for regressions in avatar loading reliability, and the quality of loading avatars.
- Prims with GLTF materials should render as expected

katerynamproductengine:
Passed QA.

Verified on the Second Life Release 7.1.11.10947335521 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/348.

PR #2364, opened 2024-08-20
secondlife/viewer#2349: Avatar/animesh rendering correctness pass

cosmic-linden:


indra/newview/lldrawpoolmaterials.cpp
```diff
@@ -247,21 +249,28 @@ void LLDrawPoolMaterials::renderDeferred(S32 pass)
...
+            }
 
-                mShader->uniformMatrix3x4fv(LLViewerShaderMgr::AVATAR_MATRIX,
-                    count,
-                    false,
-                    (GLfloat*)&(mpc.mGLMp[0]));
+            if (skipLastSkin)
+            {
+                //skin info not loaded yet, don't render
+                continue;
```
cosmic-linden, line 273: I replaced a return statement with a continue statement. Seems not quite correct to me if one skin fails to load for some reason. Pointing this out in case this is a Chesterton's fence. (@RunitaiLinden wrote this line according to the git logs)

indra/newview/lldrawpoolmaterials.cpp
```diff
@@ -247,21 +249,28 @@ void LLDrawPoolMaterials::renderDeferred(S32 pass)
...
         if (rigged && params.mAvatar.notNull())
         {
-            if (params.mAvatar != lastAvatar)
+            if (params.mAvatar != lastAvatar || params.mSkinInfo != lastSkinInfo)
             {
+                llassert(params.mSkinInfo);
                 const LLVOAvatar::MatrixPaletteCache& mpc = params.mAvatar->updateSkinInfoMatrixPalette(params.mSkinInfo);
                 U32 count = static_cast<U32>(mpc.mMatrixPalette.size());
+                skipLastSkin = !bool(count);
+                lastAvatar = params.mAvatar;
```
cosmic-linden, line 258: lastAvatar appeared to always be null, preventing this optimization from working

indra/newview/lldrawpool.cpp
```diff
@@ -688,6 +671,81 @@ bool LLRenderPass::uploadMatrixPalette(LLVOAvatar* avatar, LLMeshSkinInfo* skinI
     return true;
 }
 
+// Returns true if rendering should proceed
+//static
+bool LLRenderPass::uploadMatrixPalette(LLVOAvatar* avatar, LLMeshSkinInfo* skinInfo, const LLVOAvatar*& lastAvatar, U64& lastMeshId, bool& skipLastSkin)
```
cosmic-linden, line 676: Inspection of `LLDrawPoolMaterials` indicated that there are several important things to check when drawing a rigged mesh:

- Is the avatar null? If so, don't render.
    - Note: Some cases of null avatar are more suspicious than others. Some may be bugs. I previously added `llassert(avatar)` here for a bit to catch some bugs, but I'm not convinced all cases of null avatar are bugs.
- Is the matrix palette not defined? If so, don't render.
- During alpha draw passes, shaders are interleaved, so don't assume the skinning is up-to-date if the shader changes

indra/newview/lldrawpoolmaterials.cpp
```diff
@@ -245,23 +247,11 @@ void LLDrawPoolMaterials::renderDeferred(S32 pass)
...
-                {
-                    //skin info not loaded yet, don't render
-                    return;
-                }
-
-                mShader->uniformMatrix3x4fv(LLViewerShaderMgr::AVATAR_MATRIX,
-                    count,
-                    false,
-                    (GLfloat*)&(mpc.mGLMp[0]));
+                continue;
```
cosmic-linden, line 254: This is now a `continue`, but was previously a `return`. The `return` appeared to not be correct, as it would interfere with the rendering of other objects in the draw pool.

RunitaiLinden (approved):

PR #2365, opened 2024-08-20
3p-lib updates

marchcat:
- tinyexr to v1.0.9 - secondlife/3p-tinyexr#5;
- freetype to v2.13.3 - secondlife/3p-freetype#14;
- boost to v1.86.0 - secondlife/3p-boost#13;
- (dependency) secondlife/3p-colladadom#11.

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -99,6 +99,16 @@
       </map>
       <key>boost</key>
       <map>
+        <key>copyright</key>
+        <string>(see individual source files)</string>
+        <key>description</key>
+        <string>Boost C++ Libraries</string>
+        <key>license</key>
+        <string>boost 1.0</string>
+        <key>license_file</key>
+        <string>LICENSES/boost.txt</string>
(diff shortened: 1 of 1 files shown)


autobuild.xml
```diff
@@ -99,18 +99,28 @@
       </map>
       <key>boost</key>
       <map>
+        <key>copyright</key>
```
Ansariel, line 102: Pretty sure the next one doing library updates is going to shuffle all these keys to the end again... :joy:

marchcat: It depends on the autobuild version, but yes, it's a known problem.

RyeMutt:
You'll also need to update boost in https://github.com/secondlife/3p-colladadom and rebuild that

marchcat:
> You'll also need to update boost in https://github.com/secondlife/3p-colladadom and rebuild that

Right, working on that now.

akleshchev (approved):

PR #2369, opened 2024-08-20
Incremented viewer version after release/2024.08-DeltaFPS branch creation

brad-linden:
I'm considering trying something new here for the new git-flow process.  since release branches now try to avoid merging from main when a viewer is promoted (as they should already contain all the relevant changes by way of them being in develop as well).  it seems that the right time to increment the viewer version is immediately after freezing a new release branch.  that way nightly builds of develop will now have an incremented version number from that point onward.


marchcat (approved):

nat-goodspeed:
> the right time to increment the viewer version is immediately after freezing a new release branch

To make sure I understand: I think you mean right after creating a new release branch from develop? Or do you mean "freezing" in the sense of no additional changes, just before promoting? From context, I think the former.

Issue #2237, opened 2024-08-09
Lua: require\\fiber.lua stack overflow ((null))


AtlasLinden:
### Environment

Second Life Release 7.1.9.10286089860 (64bit)
Release Notes

You are at 121.0, 110.0, 23.1 in Pasta Cake located at simhost-0a7de8a08a5d18b0e.agni
SLURL: http://maps.secondlife.com/secondlife/Pasta%20Cake/121/110/23
(global coordinates 332,153.0, 305,518.0, 23.1)
Second Life Server 2024-06-11.9458617693
Release Notes

CPU: 12th Gen Intel(R) Core(TM) i9-12900HK (2918.4 MHz)
Memory: 32440 MB
OS Version: Microsoft Windows 10 64-bit (Build 19044.3086)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 3050 Ti Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.4659
OpenGL Version: 4.6.0 NVIDIA 546.59

Window size: 1278x1368
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 10000kbit/s
LOD factor: 1.75
Render quality: 5
Texture memory: 4095MB
Disk cache: Max size 1638.4 MB (36.6% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: FMOD Studio 2.02.20
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/1,148 (0.0%)
August 09 2024 06:35:38

### Description

This error was noticed in the logs after noticing a lua scripted floater stop updating.
`WARNING #Lua# llcommon/lua_function.cpp(517) LuaState::~LuaState : atexit() function error: [string "C:\\Program Files\\SecondLifeLua\\scripts/lua/require\\fiber.lua"]:326: stack overflow ((null))`

### Reproduction steps

Not the most ideal steps but it does reproduce fairly consistently with the following:
1. Run the test_luafloater_speedometer.lua script
2. Leave it running for a few minutes
3. Use other apps outside of the viewer in the meantime
[...]

PR #2373, opened 2024-08-21
Fix for #2237: intermittent Lua data stack overflow.

nat-goodspeed:
Use a static `unordered_map` to allow a function receiving `(lua_State* L)` to
look up the `LuaState` instance managing that `lua_State`. We've thought about
this from time to time already. `LuaState`'s constructor creates the map entry;
its destructor removes it; the new static `getParent(lua_State* L)` method
performs the lookup.

Migrate `lluau::set_interrupts_counter()` and `check_interrupts_counter()` into
`LuaState` member functions. Add a new `mInterrupts` counter for them.

Importantly, `LuaState::check_interrupts_counter()`, which is indirectly called
by a `lua_callbacks().interrupt` function, no longer performs any Lua stack
operations. Empirically, it seems the Lua engine is capable of interrupting
itself at a moment when re-entry confuses it.

Change previous `lluau::set_interrupts_counter(L, 0)` calls to
`LuaState::getParent(L).set_interrupts_counter(0)`.

Also add `LuaStackDelta` class, and a `lua_checkdelta()` helper macro, to verify
that the Lua data stack depth on exit from a block differs from the depth on
entry by exactly the expected amount. Sprinkle `lua_checkdelta()` macros in
likely places.

### DIFF
--- a/indra/llcommon/scope_exit.h
+++ b/indra/llcommon/scope_exit.h
@@ -1,0 +1,34 @@
+/**
+ * @file   scope_exit.h
+ * @author Nat Goodspeed
+ * @date   2024-08-15
+ * @brief  Cheap imitation of std::experimental::scope_exit
+ * 
+ * $LicenseInfo:firstyear=2024&license=viewerlgpl$
+ * Copyright (c) 2024, Linden Research, Inc.
+ * $/LicenseInfo$
+ */
+
+#if ! defined(LL_SCOPE_EXIT_H)
+#define LL_SCOPE_EXIT_H
+
+#include <functional>
+
+namespace LL
+{
+
+class scope_exit
+{
+public:
+    scope_exit(const std::function<void()>& func): mFunc(func) {}
+    scope_exit(const scope_exit&) = delete;
+    scope_exit& operator=(const scope_exit&) = delete;
+    ~scope_exit() { mFunc(); }
+
+private:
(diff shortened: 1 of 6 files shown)


nat-goodspeed:
hmm, mention of #2237 in the PR title wasn't sufficient to link it.

indra/llcommon/scope_exit.h
```diff
@@ -0,0 +1,34 @@
...
+
+#if ! defined(LL_SCOPE_EXIT_H)
+#define LL_SCOPE_EXIT_H
+
+#include <functional>
+
+namespace LL
+{
+
+class scope_exit
```
maxim-productengine, line 20: Did I miss something or `scope_exit` is not used for now?

maxim-productengine (approved):
Thanks!
I like the idea of `LuaStackDelta`, it should help to avoid data stack issues in the future.

indra/llcommon/scope_exit.h
nat-goodspeed: `scope_exit` is not used right now, correct. I introduced it in a previous round of changes to set a Lua recursion flag in `check_interrupts_counter()` and clear it on exit, whether or not it exited via exception.

But since I have often considered the question of whether to introduce a specific class to ensure that some cleanup happens on exit from a block, I added it anyway.

PR #2374, opened 2024-08-21
Japanese Translation Update 

logue:
This mainly covers the differences from the pull request made in July of last year to today.
Includes corrections to many mistranslations and sentence corrections.


marchcat:

This looks much cleaner than the combined result I have in #2232 so far.

PR #2375, opened 2024-08-21
More changes to Clang-Format to match the current coding standard

Ansariel:
Trying to use the Clang-Format definition provided results in a shocking discrepancy to how the code is currently formated. Here another attempt to align the definition closer to the current code:

* Fix the access modifier offset to -4 instead of -2
* Allow lambdas and functions only for inline definitions
* Always add line break after a template declaration
* Don't add spacing after a C-style cast
* Don't add spacing after template keyword (both variants are currently used a lot, but the variants without space are more, so I opted for this one)
* Only add 1 space before trailing comments

This kinda makes me wonder if somebody actually spent time on configuring this properly or just dropped in a version downloaded somewhere off the Internet... 😉

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,7 +2,7 @@
 # clang-format version 10.0.0+
 Language:        Cpp
 # BasedOnStyle:  Microsoft
-AccessModifierOffset: -2
+AccessModifierOffset: -4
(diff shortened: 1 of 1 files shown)


marchcat (approved):
Thank you!

PR #2377, opened 2024-08-21
Reenable Linux build

Nicky-D:
- Make sure <memory> gets included for std::unique_ptr. This probably got included as a lucky accident by some other file before and then got refactored.
- Add zlib-ng ng as a dependency to llwindow. The new libpng16.a needs zlib (adler32). Strangely adding it directly to the dependencies of libpng did not work.
- libexpat.so* is gone due to latest package update, remove it from packaging
- Renable Linx build for GHA


marchcat (approved):
Thank you very much!

PR #2379, opened 2024-08-21
Fix formatting and add some more constexpr

Ansariel:
The formatting in llpanelprofile.cpp got mangled up in the most recent merge with Atlasaurus. This PR fixes it and also adds more constexpr for constants I found while going through broken formatting.


akleshchev (approved):

akleshchev:
Thank you.

Issue #2325, opened 2024-08-17
WebRTC: p2p voice calling option sometimes becomes disabled after calling another agent back too soon


maestrolinden:
### Environment

backend version: utilx:584130

Atlasaurus 7.1.9.10326512121 (64bit)
Release Notes

You are at 128.8, 128.7, 23.0 in webRTC1 located at simhost-07c23d76c83837d55.aditi
SLURL: secondlife://Aditi/secondlife/webRTC1/129/129/23
(global coordinates 259,969.0, 245,377.0, 23.0)
WebRTC-Voice 2024-08-15.10397494850
Release Notes

CPU: Intel(R) Core(TM) i7-4930K CPU @ 3.40GHz (3402.15 MHz)
Memory: 32704 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4780)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.3623
OpenGL Version: 4.6.0 NVIDIA 536.23

Window size: 1911x1321
Font Size Adjustment: 96pt
UI Scaling: 0.75
Draw distance: 128m
Bandwidth: 10000kbit/s
LOD factor: 1.75
Render quality: 5
Texture memory: 7234MB
Disk cache: Max size 1638.4 MB (63.3% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 20/3,969 (0.5%)
August 16 2024 16:49:21

### Description

If 2 agents just ended a p2p voice call in webRTC mode, then one agent attempts to call back too quickly (on the order of 1 second) after the call ended, the viewer can get into a state where the "Voice call" UI option to call the other agent (in either the right-click context menu or the top icon in "Conversations" is permanently disabled.  This condition persists even after the conversation is closed an reopened.

While although the UI to call the other avatar is disabled, the users are still able to voice chat spatially, and are still able to voice call other nearby avatars.

### Reproduction steps

[...]

PR #2381, opened 2024-08-21
#2325 WebRTC: p2p voice calling option sometimes becomes disabled

LLGuru:


indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -433,16 +434,34 @@ void LLFloaterIMSessionTab::draw()
 
 void LLFloaterIMSessionTab::enableDisableCallBtn()
 {
-    if (LLVoiceClient::instanceExists() && mVoiceButton)
+    if (!mVoiceButton)
+        return;
+
+    BOOL enable = FALSE;
```
marchcat, line 440: `bool`?

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -1038,8 +1076,16 @@ void LLFloaterIMSessionTab::reloadEmptyFloaters()
 
 void LLFloaterIMSessionTab::updateCallBtnState(bool callIsActive)
 {
+    static int prev_state = 0;
+    int new_state = callIsActive ? 1 : 2;
     mVoiceButton->setImageOverlay(callIsActive? getString("call_btn_stop") : getString("call_btn_start"));
     mVoiceButton->setToolTip(callIsActive? getString("end_call_button_tooltip") : getString("start_call_button_tooltip"));
+    mVoiceButtonHangUpMode = callIsActive;
```
akleshchev, line 1083: This seems odd. Voice is in a specific state and button should corelate to that, not try and tracks states on its own. Letting user click ''hang up" multiple times if voice got stuck shutting down or initializing isn't helpfull.

I think instead button should check voice's current state and represent that state accurately.

indra/newview/llfloaterimsessiontab.cpp
LLGuru: This branch is still using BOOL

akleshchev:
Can you explain why button was stuck as disabled?

Also this doesn't cover right-click menu.

indra/newview/llfloaterimsessiontab.cpp
LLGuru: This mode variable just means what exactly user sees on the button,
what he expects from the click - start call or hang up

indra/newview/llfloaterimsessiontab.cpp
marchcat: if you pass `bool` value as an argument is should work as well, right?

indra/newview/llfloaterimsessiontab.cpp
LLGuru: The comment is added to the variable declaration
image

akleshchev: Not what I mean. If `mVoiceButtonHangUpMode` you are setting `enable = TRUE` regardless of voice state. Why are you tracking state independently? Why not check voice's state instead?

Besides won't this break if there are multiple LLFloaterIMSessionTab?

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -433,16 +434,34 @@ void LLFloaterIMSessionTab::draw()
...
+
+    if (mSessionID.notNull() && mSession && mSession->mSessionInitialized && mSession->mCallBackEnabled)
     {
-        mVoiceButton->setEnabled(
-            mSessionID.notNull()
-            && mSession
-            && mSession->mSessionInitialized
-            && LLVoiceClient::getInstance()->voiceEnabled()
-            && LLVoiceClient::getInstance()->isVoiceWorking()
-            && mSession->mCallBackEnabled);
```
akleshchev, line 444: If voiceEnabled() and isVoiceWorking() are somehow insuficient, then LLAvatarActions::canCall() won't work too. Or is it mSessionInitialized that is false?

WebRTC is fresh and has issues, most likely problem is not in LLFloaterIMSessionTab, but likely WebRTC got stuck in a weird state.

indra/newview/llfloaterimsessiontab.cpp
LLGuru: Because the voice state can be changed later than we enable/disable the button
The final check should be done after the user clicks the button and callback is executed

indra/newview/llfloaterimsessiontab.cpp
roxanneskelly: release/webrtc-voice is kinda a dead branch.  Development has been moved to atlasaurus, and after to develop.

indra/newview/llfloaterimsessiontab.cpp
roxanneskelly: Agreed.  There is also, in fact, a chance that the issue is server-side.

roxanneskelly (changes requested):
Should move to merging into atlasaurus or develop, depending on where we want to get this issue in.

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -1038,8 +1076,16 @@ void LLFloaterIMSessionTab::reloadEmptyFloaters()
 
 void LLFloaterIMSessionTab::updateCallBtnState(bool callIsActive)
 {
+    static int prev_state = 0;
```
akleshchev, line 1079: Won't prev_state break if there are multiple LLFloaterIMSessionTab?

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -433,16 +434,34 @@ void LLFloaterIMSessionTab::draw()
...
-            mSessionID.notNull()
-            && mSession
-            && mSession->mSessionInitialized
-            && LLVoiceClient::getInstance()->voiceEnabled()
-            && LLVoiceClient::getInstance()->isVoiceWorking()
-            && mSession->mCallBackEnabled);
+        if (mVoiceButtonHangUpMode)
+        {
+            // We allow to hang up from any state
+            enable = TRUE;
```
akleshchev, line 447: Why isn't this checking voice state to make sure there is something to hung up? Like an outgoing attempt of a call.

indra/newview/llfloaterimsessiontab.cpp
akleshchev: > There is also, in fact, a chance that the issue is server-side.

Issue is that viewer's UI gets stuck in a state where user can't disable outgoing call.

LLGuru:
> Also this doesn't cover right-click menu.

It does
image

indra/newview/llfloaterimsessiontab.cpp
LLGuru: > Agreed. There is also, in fact, a chance that the issue is server-side.

The bug described in the issue https://github.com/secondlife/viewer/issues/2325 was reproduced on the build version `Atlasaurus 7.1.9.10326512121`

But on the build compiled from the latest source code this bug was not reproduced

There was an other variant of the bug observed: the button became disabled during the early recall,
but it returned to a normal state after hang up made via the button `Cancel` on the outgoing call floater

This bug seems to be 100% UI-related, the voice channel works well with no issues

indra/newview/llfloaterimsessiontab.cpp
LLGuru: > Development has been moved to atlasaurus, and after to develop.

Retargeted to develop

marchcat: nice, so there shouldn't be any fake bools there

akleshchev (commented):
Menu apparently works in newer builds, static variable issue got resolved. I don't like that 'disconnect' button gets enabled unconditionally, but that can be worked on later if it becomes an issue and QA can test it on vivox.

LLGuru:
> Should move to merging into atlasaurus or develop, depending on where we want to get this issue in.

Retargeted to develop

LLGuru:
> I don't like that 'disconnect' button gets enabled unconditionally

We shouldn't disable such commands as Dosconnrct
If there is nothing to disconnect, the button should look as Connect (enabled or disabled)

In fact, this button should accept only three states:
- Connect (disabled)
- Connect (enabled)
- Disconnect (enabled)

indra/newview/llfloaterimsessiontab.cpp
LLGuru: > Why isn't this checking voice state to make sure there is something to hung up? Like an outgoing attempt of a call.

See the comment below https://github.com/secondlife/viewer/pull/2381#issuecomment-2304226886

akleshchev (approved):

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -434,16 +435,34 @@ void LLFloaterIMSessionTab::draw()
...
+        else
+        {
+            // We allow to start call from this state only
+            if (mSession->mVoiceChannel->getState() == LLVoiceChannel::STATE_NO_CHANNEL_INFO &&
+                LLVoiceClient::instanceExists())
+            {
+                LLVoiceClient* client = LLVoiceClient::getInstance();
+                if (client->voiceEnabled() && client->isVoiceWorking())
+                {
+                    enable = TRUE;
```
marchcat, line 459:
```suggestion
                    enable = true;
```

indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -434,16 +435,34 @@ void LLFloaterIMSessionTab::draw()
...
-            mSessionID.notNull()
-            && mSession
-            && mSession->mSessionInitialized
-            && LLVoiceClient::getInstance()->voiceEnabled()
-            && LLVoiceClient::getInstance()->isVoiceWorking()
-            && mSession->mCallBackEnabled);
+        if (mVoiceButtonHangUpMode)
+        {
+            // We allow to hang up from any state
+            enable = TRUE;
```
marchcat, line 448:
```suggestion
            enable = true;
```

marchcat (changes requested):

akleshchev:
> We shouldn't disable such commands as Disconnrct
> If there is nothing to disconnect, the button should look as Connect (enabled or disabled)

Disagree. If disconnect for some reason takes long, that just gives user a chance to click it again which can range from doing nothing to causeing all sorts of fun bugs. Better to not permit it.

indra/newview/llfloaterimsessiontab.cpp
LLGuru: Fixed

marchcat (changes requested):
[llfloaterimsessiontab.cpp](https://github.com/secondlife/viewer/pull/2381/files#diff-edcec7dbbb4747cdcb59ef2f3f896c3ecf1da3a03916485702681e6cf3c6a6d2) L448

LLGuru:
> [llfloaterimsessiontab.cpp](https://github.com/secondlife/viewer/pull/2381/files#diff-edcec7dbbb4747cdcb59ef2f3f896c3ecf1da3a03916485702681e6cf3c6a6d2) L448

Fixed

marchcat (approved):

roxanneskelly (approved):

PR #2383, opened 2024-08-21
Fix local resetting of skeleton

FelixWolf:
Not entirely sure if this is the right repo to push this to, but there is a bug that forbids resetting of the skeleton locally. This adds a check to verify if the source object is the target object *OR* if the effect was created locally.


marchcat (approved):

While Maint C is still fine, new PRs can target `develop` branch directly, unless they affect any branch-specific features.

Issue #2345, opened 2024-08-19
Turn off mirrors by default


Dan-Linden:
### Test plan:
1. Delete the settings.xml file from C:\Users\[USERNAME]\AppData\Roaming\SecondLife\user_settings
2. Install the Featurettes viewer from https://releasenotes.secondlife.com/viewer/7.1.8.9375512768.html
3. Set Preferences > Graphics > Quality & Speed to High
4. Click the "Advanced settings" button and verify Mirrors checkbox is checked.
5. Quit the viewer.
6. Install the viewer-in-test
7. Open Preferences > Graphics > Quality & Speed and click the "Advanced settings" button
8. Verify the Mirrors checkbox is unchecked.
9. Delete the settings.xml file from C:\Users\[USERNAME]\AppData\Roaming\SecondLife\user_settings
10. Run the viewer in test
11. Open Preferences > Graphics > Quality & Speed and click the "Advanced settings" button
12. Verify the Mirrors checkbox is unchecked.

### Environment

.

### Description

Turn off mirrors by default, at least up to High+ graphics setting.

### Reproduction steps

1. Set Preferences > Graphics > Quality & Speed to High and High+
2. Click the "Advanced settings" button and verify Mirrors checkbox is unchecked.

Dan-Linden:
Reopened for QA.

Dan-Linden:
Reopened.
We need this to also turn off mirrors for people who had mirrors enabled in the Featurettes viewer. For the data.

dmytrohproductengine:
Failed QA.
Verified on the Second Life Release 7.1.9.10479800881 (64bit) on Win10/OSX in the scope of [secondlife/iqa#311](https://github.com/secondlife/iqa/issues/311).

Win10 - Passed.
The "Mirror" checkbox is unchecked for all graphics presets.

OSX - Failed.
The "Mirror" checkbox is checked by default for the "High" graphics preset (see "OSX_High_preset_7.1.9.10479800881.png").

**OSX_High_preset_7.1.9.10479800881.png**
![OSX_High_preset_7 1 9 10479800881](https://github.com/user-attachments/assets/2e306fff-f8e5-42eb-a762-8390d290db07)

sergiiiproductengine:
Passed QA.

Verified on the Second Life Release 7.1.9.10495892574 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/312.

PR #2386, opened 2024-08-21
Bump featuretable versions to fix secondlife/viewer#2345

brad-linden:
also fixed mac High entry for RenderMirrors that got missed


Geenz (approved):
LGTM

PR #2392, opened 2024-08-22
Remove or suppress left over debug messages in hot paths

RyeMutt:
Remove or suppress left over debug messages in hot paths found in tracy

### DIFF
--- a/indra/llcommon/llcallstack.h
+++ b/indra/llcommon/llcallstack.h
@@ -79,9 +79,13 @@ struct LLContextStatus
 
 LL_COMMON_API std::ostream& operator<<(std::ostream& s, const LLContextStatus& context_status);
 
+#ifndef LL_RELEASE_FOR_DOWNLOAD
 #define dumpStack(tag)                          \
     LL_DEBUGS(tag) << "STACK:\n"                \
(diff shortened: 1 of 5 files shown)


marchcat (approved):
Thank you!

nat-goodspeed (commented):
I like the `dumpStack()` change, especially since I don't know to what extent `LLCallStack` has ever been useful. (Even less now, presumably, given `boost::stacktrace`.)

I get the point of removing unconditional extra calls to `getRiggedVolume()`, `llformat()` and `getEstTrianglesMax()`.

If we never turn on DEBUG logging for `"TEXTUREENTRY"`, `"RiggedBox"`, `"DynamicBox"` or `"AnimatedObjectsLinkset"`, then sure, those `LL_DEBUGS()` constructs are cruft, and the source code will be more readable without them.

I have to admit, though, those `LL_DEBUGS()` removals make me wonder whether disabled logging blocks are more expensive than our predecessors obviously expected? That is, were these removed because they had a noticeable effect on frame rate, even disabled?

If _that's_ true, then presumably we need to reexamine our use of logging statements in many more areas.

Issue #2393, opened 2024-08-22
BugSplat Crash #1504586: SecondLifeViewer!LLRenderTarget::getTexture(483)


sl-service-account:
**BugSplat** [Crash 1504586](https://app.bugsplat.com/v2/crash?id=1504586&database=SecondLife_Viewer_2018)

**Application Version:**
7.1.9.10084807842

**Callstack:**
[...]

PR #2394, opened 2024-08-22
#2393 BugSplat Crash #1504586: LLRenderTarget::getTexture(483)

LLGuru:


indra/llrender/llrendertarget.cpp
```diff
@@ -475,7 +475,7 @@ void LLRenderTarget::clear(U32 mask_in)
 
 U32 LLRenderTarget::getTexture(U32 attachment) const
 {
-    if (attachment > mTex.size()-1)
```
akleshchev, line 478: What's the point of this change?

LLGuru: > What's the point of this change?

It doesn't work in case of an empty container

akleshchev: Right, unsigned int under the hood, so zero minus one won't work correctly and looks like it can be empty.

akleshchev (approved):

indra/llrender/llrendertarget.cpp
akleshchev: But why does it crash?
Even if condition doesn't work, execution would check mTex.empty() and return zero, not crash on getTexture().

LLGuru: Not sure how the optimizer transforms such ugly code

RunitaiLinden (approved):

Issue #2385, opened 2024-08-21
LLChat message types have different senders


AtlasLinden:
### Environment

Lua Viewer: Second Life Release 7.1.9.10286089860 (64bit)

### Description

Sender is either the script or the user when using LLChat function. 
This should be consistent and all should have the same sender.
Following image shows use of LLChat.sendWhisper(), LLChat.sendNearby() and LLChat.sendShout() respectively
![image](https://github.com/user-attachments/assets/0290036c-328f-433f-862d-bc5d77557cda)

### Reproduction steps

Here is a modified test_LLChat.lua script that produces the above output
```
LLChat = require 'LLChat'
local timers = require 'timers'

function generateRandomWord(length)
    local alphabet = "abcdefghijklmnopqrstuvwxyz"
    local wordTable = {}
    for i = 1, length do
        local randomIndex = math.random(1, #alphabet)
        table.insert(wordTable, alphabet:sub(randomIndex, randomIndex))
    end
    return table.concat(wordTable)
end

local msg = {'AI says:'}
math.randomseed(os.time())
for i = 1, math.random(1, 10) do
    table.insert(msg, generateRandomWord(math.random(1, 8)))
end

local id = "5fea458b-30d9-b032-ddde-d1b5276f2949"
LLChat.sendWhisper(table.concat(msg, ' '))
timers.Timer(1, 'wait')
LLChat.sendNearby(table.concat(msg, ' '))
timers.Timer(1, 'wait')
LLChat.sendShout(table.concat(msg, ' '))
```

PR #2400, opened 2024-08-22
Fix for #2385: say, shout and whisper messages from the script should be displayed consistently

maxim-productengine:
Fix for [#2385](https://github.com/secondlife/viewer/issues/2385)

### DIFF
--- a/indra/newview/llviewermessage.cpp
+++ b/indra/newview/llviewermessage.cpp
@@ -2646,15 +2646,19 @@ void process_chat_from_simulator(LLMessageSystem *msg, void **user_data)
         else
         {
             chat.mText = "";
-            std::string chat_text(mesg);
             auto [msg_without_prefix, is_lua] = LLStringUtil::withoutPrefix(mesg, LUA_PREFIX);
+            std::string prefix;
+            if (is_lua)
+            {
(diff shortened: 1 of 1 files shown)


indra/newview/llviewermessage.cpp
```diff
@@ -2646,19 +2646,21 @@ void process_chat_from_simulator(LLMessageSystem *msg, void **user_data)
         else
         {
             chat.mText = "";
+            std::string chat_text(mesg);
+            auto [msg_without_prefix, is_lua] = LLStringUtil::withoutPrefix(mesg, LUA_PREFIX);
```
nat-goodspeed, line 2650: what about after this `withoutPrefix()` call:
```
std::string prefix;
if (is_lua)
{
    prefix = LUA_PREFIX;
}
```
...
```
case CHAT_TYPE_WHISPER:
    prefix += LLTrans::getString("whisper") + " ";
    break;
case CHAT_TYPE_SHOUT:
    prefix += LLTrans::getString("shout") + " ";
    break;
```
...
```
(line 2673)
chat.mText = prefix + msg_without_prefix;
```
which would make line 2648 moot.

nat-goodspeed (approved):

Issue #2398, opened 2024-08-22
Remove SSAO From Reflection Probes


RunitaiLinden:
### Description

Reflection probes shouldn't use GPU resources on fine details like that.  Remove the SSAO pass from probe render passes.

igorlproductengine:
Need more info.
Verified on the Second Life Release 7.1.11.10947335521 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/348.

Could you please provide more details about this issue? Specifically, could you explain the expected behavior for reflection probes after the SSAO pass is removed? Additionally, it would be helpful to have a test plan to ensure the functionality is thoroughly validated and meets performance expectations.

AtlasLinden:
Unfortunately there's not any ideal way to test this.
Instead perform part 1.1 of the [Material UI Test Plan](https://lindenlab.atlassian.net/wiki/spaces/QA/pages/2730917904/Material+Viewer+UI+Test+Plan#1.1-Screen-Space-Ambient-Occlusion) both in and around some reflection probes

igorlproductengine:
Passed QA.

Verified on Second Life Release 7.1.11.10947335521 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/348.

According to the part 1.1 of the [Material UI Test Plan](https://lindenlab.atlassian.net/wiki/spaces/QA/pages/2730917904/Material+Viewer+UI+Test+Plan#1.1-Screen-Space-Ambient-Occlusion) and some reflection probes.

PR #2405, opened 2024-08-23
Add dedicated sun probe shader and remove SSAO and shadow smoothing from probe generation passes

RyeMutt:
* Disable SSAO and shadow smoothing and add dedicated sun shader without ssao for probes
* Remove reference to dead sunLightNoFragCoordV shader

#2398


RyeMutt:

Ansariel:
Are the shader files missing?

RyeMutt:
The sun shader files are reused for this. I created a variant that always loads the non-ssao sun shader.

Ansariel:
Weird! I can't find sunLightNoFragCoordV.glsl anywhere...

RyeMutt:
Good catch. Looks like that was removed a while ago.

Ansariel:
And the other ones are only availble on shader level >= 2 - not sure if you check for that, so just FYI

RunitaiLinden (approved):

PR #2406, opened 2024-08-23
Update LLImageJ2C::calcDataSizeJ2C for better 2k image support

TommyTheTerrible:
Adjusted calculations based on dimensions and assumed maximum block size so that higher discards (4-5) of 2048x2048 images can be decoded with aux/alpha. (Should also work for dimensions larger than 2048.)

This function will now return a reliable discard 5 data size for unknown dimensions (w and/or h equals 0), which could be used in LLTextureFetch::createRequest to skip the header fetch and go right to a discard 5 decode.

Tested on OpenJPEG 2.5 with partial decode support (opj_decoder_set_strict_mode set to false).

Should work on KDU fine but might be a good idea to test.

Renamed variables and added comments to help explain the process better.

### DIFF
--- a/indra/llimage/llimagej2c.cpp
+++ b/indra/llimage/llimagej2c.cpp
@@ -276,16 +276,20 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
     // Estimate the number of layers. This is consistent with what's done for j2c encoding in LLImageJ2CKDU::encodeImpl().
     constexpr S32 precision = 8; // assumed bitrate per component channel, might change in future for HDR support
     constexpr S32 max_components = 4; // assumed the file has four components; three color and alpha
-    S32 nb_layers = 1;
-    const S32 surface = w*h;
-    S32 s = 64*64;
-    S32 totalbytes = (S32)(s * max_components * precision * rate); // first level computed before loop
-    while (surface > s)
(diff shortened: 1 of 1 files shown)


marchcat (approved):
Thank you!

I've tested this change very briefly on KDU, it seems fine.

indra/llimage/llimagej2c.cpp
```diff
@@ -276,16 +276,20 @@ S32 LLImageJ2C::calcDataSizeJ2C(S32 w, S32 h, S32 comp, S32 discard_level, F32 r
...
     constexpr S32 precision = 8; // assumed bitrate per component channel, might change in future for HDR support
     constexpr S32 max_components = 4; // assumed the file has four components; three color and alpha
-    S32 nb_layers = 1;
-    const S32 surface = w*h;
-    S32 s = 64*64;
-    S32 totalbytes = (S32)(s * max_components * precision * rate); // first level computed before loop
-    while (surface > s)
+    // Use MAX_IMAGE_SIZE_DEFAULT (currently 2048) if either dimension is unknown (zero)
+    S32 width  = (w > 0) ? w : 2048;
+    S32 height = (h > 0) ? h : 2048;
```
marchcat, line 281:
```suggestion
    S32 width  = (w > 0) ? w : 2048;
    S32 height = (h > 0) ? h : 2048;
```
To pass the pre-commit check.

Issue #2388, opened 2024-08-21
Blurry moving pictures


AtlasLinden:
### Environment

DeltaFPS viewer: Second Life Test 7.1.10.10478834997 (64bit)

### Description

Texture animation that involves switching via different textures is breaking
Actual: https://gyazo.com/9b069035dd3b8f602590a8d0b9e79f7c
Expected (Atlasaurus): https://gyazo.com/7d8261776278b52ddbda3023f7217f1f

### Reproduction steps

Moving picture object temporarily located at: http://maps.secondlife.com/secondlife/TextureTest/146/246/23
See also Galaxy Arcade sign at location it was first spotted: http://maps.secondlife.com/secondlife/Rizal/129/106/23

RunitaiLinden:
Fixed.

Also:
- Don't downres textures until video memory or system memory is low
- Do downres textures whenever the viewer loses focus for more than 10 seconds.

dmytrohproductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10582490681 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/316.

PR #2414, opened 2024-08-23
#2388 Texture downrezzing tune up.

RunitaiLinden:
- Hold onto unreferenced textures for 30 seconds.
- Don't downres unless memory is low
- Downres when viewer is backgrounded.


RyeMutt (approved):
Looks good! 👍

PR #2415, opened 2024-08-24
Introduce Contrast Adaptive Sharpening post process effect

RyeMutt:
This introduces Contrast Adaptive Sharpening from https://github.com/GPUOpen-Effects/FidelityFX-CAS

Setting range is 0.0 - 1.0 with 0.0 disabling the effect/shader pass

0.0:
Snapshot_004
0.4:
Snapshot_002
1.0:
Snapshot_003

### DIFF
--- a/indra/newview/pipeline.cpp
+++ b/indra/newview/pipeline.cpp
@@ -213,7 +213,6 @@ S32 LLPipeline::RenderBufferVisualization;
 bool LLPipeline::RenderMirrors;
 S32 LLPipeline::RenderHeroProbeUpdateRate;
 S32 LLPipeline::RenderHeroProbeConservativeUpdateMultiplier;
-F32 LLPipeline::RenderCASSharpness;
 LLTrace::EventStatHandle<S64> LLPipeline::sStatBatchSize("renderbatchsize");
 
 const U32 LLPipeline::MAX_BAKE_WIDTH = 512;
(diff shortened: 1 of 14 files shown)


Geenz (approved):
There's definitely some eyebrow raising stuff in LLPipeline, but it appears that this is how AMD wants you to integrate shaders like this.  So it's fine, and will probably take more time to separate the AMD specific methods from their GLSL than it's probably worth.

Tested this locally - seems to work as I'd expect.

indra/newview/pipeline.cpp
```diff
@@ -7133,6 +7147,50 @@ void LLPipeline::generateGlow(LLRenderTarget* src)
...
+    // Bind setup:
+    dst->bindTarget();
+
+    sharpen_shader->bind();
+
+    {
+        static LLStaticHashedString cas_param_0("cas_param_0");
+        static LLStaticHashedString cas_param_1("cas_param_1");
+        static LLStaticHashedString out_screen_res("out_screen_res");
+
```
RunitaiLinden, line 7169: Not a blocker for a merge, but should probably just use a static LLCachedControl here for RenderCASSharpness.  "refreshCachedSettings" has a lot of side effects and is very slow, so unless a shader reload or a RenderTarget allocation is required, just use LLCachedControl wherever you want to access the saved setting.

RunitaiLinden (approved):

indra/newview/pipeline.cpp
```diff
@@ -117,6 +117,17 @@
...
+#define A_GCC 1
+#pragma GCC diagnostic ignored "-Wunused-function"
+#pragma GCC diagnostic ignored "-Wunused-variable"
+#if LL_LINUX
+#pragma GCC diagnostic ignored "-Wrestrict"
+#endif
+#endif
+#define A_CPU 1
+#include "app_settings/shaders/class1/deferred/CASF.glsl" // This is also C++
+
```
RunitaiLinden, line 130: Fine for now, but we should probably break this into its own .cpp file so it doesn't pollute all of pipeline.cpp.

RyeMutt: Fixed!

RyeMutt: Yeah I agree on that. Went with the easiest initial integration.

indra/newview/pipeline.cpp
RyeMutt: Fixed.

PR #2425, opened 2024-08-27
Larger font batch

akleshchev:
1. Increased font batch size from 30 to 512 glyphs. This reduces buffer searches and allocations in flush().

Still WIP on moving buffers into viewes, but that likely won't cover everythings, so this change should be usefull regardless.

2. Noticed an inefficiency in renderDebug(), it was running over a huge list of bridges pointlesly. Performance gain is tiny (from 7-16μs to 1-5μs), but couldn't ignore it.


marchcat (approved):

akleshchev:
Doesn't do what I expected it to, still figuring some things about fonts.

Issue #2428, opened 2024-08-27
Crash occurs after trying to apply PBR material to one face of a cube


sl-service-account:
**Steps to reproduce:**
1. Login to the [viewer](https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%231adf2d0-DeltaFPS).
2. Rez a cube > switch build tools to 'Select Face'.
3. Click any face of the cube.
4. Go to the 'Texture' tab > select 'PBR Metallic roughtness' in Material dropdown.
5. Select any material (no matter what way - using Material picker or 'Choose from inventory' button).
6. Click 'OK' button in 'Pick: texture' floater.

**Actual result:** The viewer crashes.

**Expected result:** The viewer should not crash.

**Additional info:**
The issue was found during https://github.com/secondlife/viewer/issues/1922 testing.
The issue is not reproduced on the current release viewer (7.1.9.10515727195).
The issue is not reproduced after applying PBR material to a cube without 'Select Face' option.

The issue is reproduced on Win10/OSX:
[Win10 BugSplat Key Crash 95817](https://app.bugsplat.com/v2/keycrash?stackKeyId=95817&database=SecondLife_Viewer_2018)
[OSX BugSplat Key Crash 95819](https://app.bugsplat.com/v2/keycrash?stackKeyId=95819&database=SecondLife_Viewer_2018)

<details><summary>Win10 environment</summary>
<p>

Second Life Release 7.1.10.10516785745 (64bit)
Release Notes

You are at 128.4, 84.3, 21.8 in Product Engine 2 located at simhost-0353d13a62153db1f.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine%202/128/84/22
(global coordinates 337,024.0, 307,284.0, 21.8)
Second Life Server 2024-07-22.10048683488
Release Notes

CPU: 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz (3302.4 MHz)
Memory: 16076 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4780)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 3060 Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6094
OpenGL Version: 4.6.0 NVIDIA 560.94

Window size: 1920x1027
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
[...]

PR #2430, opened 2024-08-27
#2428 Fix for crash when applying PBR material

RunitaiLinden:
Also attempt to fix some occasional bad texture memory tracking.


brad-linden (approved):
👍

PR #2433, opened 2024-08-27
Make LLPipeline::renderDebug() a bit cheaper

akleshchev:
Each frame renderDebug() was pointlesly going over a large list of partitions and bridges.

### DIFF
--- a/indra/newview/pipeline.cpp
+++ b/indra/newview/pipeline.cpp
@@ -4532,33 +4532,51 @@ void LLPipeline::renderDebug()
     }
 
     // Debug stuff.
-    for (LLWorld::region_list_t::const_iterator iter = LLWorld::getInstance()->getRegionList().begin();
-            iter != LLWorld::getInstance()->getRegionList().end(); ++iter)
+    if (gPipeline.hasRenderDebugMask(LLPipeline::RENDER_DEBUG_OCTREE |
+        LLPipeline::RENDER_DEBUG_OCCLUSION |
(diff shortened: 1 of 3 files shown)


indra/newview/pipeline.cpp
```diff
@@ -4532,33 +4532,53 @@ void LLPipeline::renderDebug()
...
+        LLPipeline::RENDER_DEBUG_AVATAR_VOLUME |
+        LLPipeline::RENDER_DEBUG_AVATAR_JOINTS |
+        LLPipeline::RENDER_DEBUG_AGENT_TARGET |
+        LLPipeline::RENDER_DEBUG_SHADOW_FRUSTA |
+        LLPipeline::RENDER_DEBUG_TEXEL_DENSITY))
+    {
+        LL_PROFILE_ZONE_NAMED_CATEGORY_DISPLAY("partitions");
+
+        for (LLWorld::region_list_t::const_iterator iter = LLWorld::getInstance()->getRegionList().begin();
             iter != LLWorld::getInstance()->getRegionList().end(); ++iter)
```
marchcat, line 4555: Since we're making changes anyway, something like this might be better:
```
for (const auto& region : LLWorld::getInstance()->getRegionList())
```

marchcat (approved):

indra/newview/pipeline.cpp
akleshchev: Will do.

marchcat (approved):

PR #2436, opened 2024-08-28
Fix background folder fetch pushing requests to wrong queue

Ansariel:
Actual fix is lines 388 and 395

### DIFF
--- a/indra/newview/llinventorymodelbackgroundfetch.cpp
+++ b/indra/newview/llinventorymodelbackgroundfetch.cpp
@@ -201,7 +201,7 @@ LLInventoryModelBackgroundFetch::LLInventoryModelBackgroundFetch():
 
 LLInventoryModelBackgroundFetch::~LLInventoryModelBackgroundFetch()
 {
-    gIdleCallbacks.deleteFunction(&LLInventoryModelBackgroundFetch::backgroundFetchCB, NULL);
+    gIdleCallbacks.deleteFunction(&LLInventoryModelBackgroundFetch::backgroundFetchCB, nullptr);
 }
 
 bool LLInventoryModelBackgroundFetch::isBulkFetchProcessingComplete() const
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

marchcat (approved):
Thank you!

indra/newview/llinventorymodelbackgroundfetch.cpp
```diff
@@ -254,35 +254,35 @@ bool LLInventoryModelBackgroundFetch::folderFetchActive() const
...
 }
 
-void LLInventoryModelBackgroundFetch::addRequestAtFront(const LLUUID & id, bool recursive, bool is_category)
+void LLInventoryModelBackgroundFetch::addRequestAtFront(const LLUUID& id, bool recursive, bool is_category)
 {
     EFetchType recursion_type = recursive ? FT_RECURSIVE : FT_DEFAULT;
     if (is_category)
     {
-        mFetchFolderQueue.push_front(FetchQueueInfo(id, recursion_type, is_category));
+        mFetchFolderQueue.emplace_front(FetchQueueInfo(id, recursion_type, is_category));
```
brad-linden, line 262: shouldn't all these `emplace()` variants usages avoid calling the constructor and just directly take the constructor's args?

brad-linden (commented):
I'm thinking that we will likely want to merge this into `develop` rather than DeltaFPS for now

indra/newview/llinventorymodelbackgroundfetch.cpp
Ansariel: That definitely makes sense, yes. I updated the PR accordingly.

Ansariel:
> I'm thinking that we will likely want to merge this into `develop` rather than DeltaFPS for now

Isn't DeltaFPS targetted to become next release? That's the reason I created the PR against this branch, especially since this bug is currently out in the wild. If you prefer it going into develop, I can retarget the PR though.

brad-linden:
> Isn't DeltaFPS targetted to become next release? That's the reason I created the PR against this branch, especially since this bug is currently out in the wild. If you prefer it going into develop, I can retarget the PR though.

Yes, but it is frozen for stabilization.  In general with the new git flow process we won't be committing things directly to the `release/foo` branches unless the issue being fixed is specifically flagged as a showstopper by QA or product review.  We are hoping to get DeltaFPS quickly and have at least one more performance focused release to follow.

brad-linden (approved):

PR #2439, opened 2024-08-28
Add simple metrics of Lua usage

maxim-productengine:
Add two counter metrics: total scripts and auto-run scripts.

### DIFF
--- a/indra/newview/llluamanager.h
+++ b/indra/newview/llluamanager.h
@@ -58,7 +58,7 @@ public:
     // same semantics as script_result_fn parameters
     typedef std::pair<int, LLSD> script_result;
 
-    static void runScriptFile(const std::string &filename, script_result_fn result_cb = {},
+    static void runScriptFile(const std::string &filename, bool autorun = false, script_result_fn result_cb = {},
                               script_finished_fn finished_cb = {});
     // Start running a Lua script file, returning an LLCoros::Future whose
     // get() method will pause the calling coroutine until it can deliver the
@@ -83,6 +83,9 @@ public:
     static script_result waitScriptLine(const std::string& chunk);
 
     static const std::map<std::string, std::string> getScriptNames() { return sScriptNames; }
+
+    static S32 sAutorunScriptCount;
+    static S32 sScriptCount;
 
  private:
    static std::map<std::string, std::string> sScriptNames;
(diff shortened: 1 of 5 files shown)


indra/newview/llluamanager.h
```diff
@@ -84,6 +84,9 @@ class LLLUAmanager
 
     static const std::map<std::string, std::string> getScriptNames() { return sScriptNames; }
 
+    static S32 sAutorunScriptCount;
+    static S32 sScriptCount;
```
nat-goodspeed, line 88: Maybe a stats struct? I can think of a few additional Lua-related statistics of potential interest, even in the short term:
* the high-water mark for number of concurrently running Lua scripts (Lua-targeted C++ coroutines)
* stats based on totaling interrupts: low, median, high interrupt count per script; total Lua interrupts for the viewer session.

Lua interrupts don't correlate with real time, or even with Lua bytecode steps, but they're sort of like gauges measuring "engine hours" for heavy equipment: they indicate (in an admittedly unspecified way) the number of cycles the Lua engine spends doing script work. We might capture "kilointerrupts" if the numbers get absurdly large.

indra/newview/llviewerstats.cpp
```diff
@@ -673,6 +674,10 @@ void send_viewer_stats(bool include_preferences)
 
     system["shader_level"] = shader_level;
 
+    LLSD &scripts = body["scripts"];
+    scripts["lua_scripts"] = LLLUAmanager::sScriptCount;
+    scripts["lua_auto_scripts"] = LLLUAmanager::sAutorunScriptCount;
```
nat-goodspeed, line 679: or `body["scripts"] = llsd::map(...);`

nat-goodspeed (approved):

indra/newview/llluamanager.h
maxim-productengine: Yep, it makes sense to use a struct if we have more than 2-3 stats.
Agree, those two may be useful. We should discuss it.
 But I can imagine that a script with an infinite loop can spoil total Lua interrupts for the viewer session.

nat-goodspeed: Okay, how about: we count interrupts for individual scripts, deferring contributing to the viewer session total until script termination. But if we terminate a script with the "infinite loop" error, we zero its interrupt statistic so it doesn't throw off stats.

maxim-productengine: Yes, it makes sense.

Issue #2360, opened 2024-08-20
Incorrect Day Cycle name in Edit Day Cycle floater after clicking the "Customize" button


valeriiaaproductengine:
### Environment

Second Life Test 7.1.9.10164740217 (64bit)
Release Notes

You are at 186.3, 55.9, 21.5 in Product Engine located at simhost-0b8a048fe6055a4b1.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine/186/56/21
(global coordinates 336,826.0, 307,256.0, 21.5)
Second Life Server 2024-06-11.9458617693
Release Notes

CPU: 11th Gen Intel(R) Core(TM) i5-11260H @ 2.60GHz (2611.2 MHz)
Memory: 16111 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4780)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 3050 Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.2879
OpenGL Version: 4.6.0 NVIDIA 528.79

Window size: 1920x1001
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 4095MB
Disk cache: Max size 1638.4 MB (0.9% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Not Connected
Packets Lost: 6/53,749 (0.0%)
August 20 2024 04:10:45

### Description

See Reproduction steps

### Reproduction steps

1. Login to the Second Life Test 7.1.9.10164740217 viewer.
2. Select Me-> Inventory.
3. Create New Settings ->  New Day Cicle.
4. Without opening the New Day Cicle setting, right-click it and select Rename.
5. Being on your own parcel, right-click and select "About Land".
6. Select Environment tab, click the Use inventory button, select the previously created  New Day Cicle and click ok.
7. In about land floater -> Environment tab click the "Customize" button.
 

[...]

PR #2446, opened 2024-08-28
#2360 Incorrect Day Cycle name in Edit Day Cycle floater

LLGuru:


indra/newview/llfloatereditextdaycycle.cpp
```diff
@@ -433,7 +433,7 @@ void LLFloaterEditExtDayCycle::refresh()
     {
         LLLineEditor* name_field = getChild<LLLineEditor>(TXT_DAY_NAME);
         name_field->setText(mEditDay->getName());
-        name_field->setEnabled(mCanMod);
+        name_field->setEnabled(false);
```
marchcat, line 436: #2360 doesn't seem to mention anything about disabling the field. Is this line correct?

LLGuru: I think Yes

This name should be modified another waym using Save As or directly in the Library
Moreover, editing this text doesn't take effect because this value is never used later

marchcat (approved):

indra/newview/llfloatereditextdaycycle.cpp
akleshchev: I suggest checking how it works for notecards and scripts and making the same here. Changing this field appears to trigger onCommitName and apply correctly to inventory.

LLGuru: > Changing this field appears to trigger onCommitName and apply correctly to inventory.

onCommitName modifies the unpacked copy of the object
The modified name stays unused after applying changes

akleshchev (approved):
This looks better.

But arguably region environments should be renamable as well, users would expect them to be.

Issue #2445, opened 2024-08-28
Force an Error callstacks should show an accurate source location in bugsplat


brad-linden:
### Environment

Second Life Test 7.1.10.242411739 (64bit)
Release Notes
Build Configuration RelWithDebInfo

CPU: 12th Gen Intel(R) Core(TM) i9-12900HK (2918.4 MHz)
Memory: 32440 MB
OS Version: Microsoft Windows 11 64-bit (Build 22631.4037)
Graphics Card Vendor: Intel
Graphics Card: Intel(R) Iris(R) Xe Graphics

Windows Graphics Driver Version: 31.0.101.5186
OpenGL Version: 4.6.0 - Build 31.0.101.5186

### Description

currently any deliberate crashes from LL_ERRS are all showing up as a lambda object in gltfscenemanager.cpp.  the accurate source location of the error should be shown instead.

### Reproduction steps

1. Log in
2. From Develop menu, select Force An Error -> Force LLError and Crash
3. Find the crash report in bugsplat

Expected Behavior
the crash report should show LLAppViewer::forceErrorLLError() as the source location for the crash

georgebproductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10622905308 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/316.

PR #2447, opened 2024-08-28
#2445 revive old LLERROR_CRASH implementation

brad-linden:
this removes the indirect crashdriver() method of crashing in LL_ERRS and restores the old method of dereferencing a bad pointer except using 0xDEADBEEFDEADBEEF instead of NULL, which by experience seems better supported by modern compilers for the time being

should fix secondlife/viewer#2445

### DIFF
--- a/indra/llcommon/llerror.cpp
+++ b/indra/llcommon/llerror.cpp
@@ -1641,19 +1641,3 @@ namespace LLError
         sLocalizedOutOfMemoryWarning = message;
     }
 }
-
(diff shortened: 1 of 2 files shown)


marchcat (approved):

brad-linden:
if this stops working at some point in the future due to even better Undefined Behavior detection, then it is probably worth looking at replacing this with inline assembly to force the crash.

akleshchev (approved):

Issue #2421, opened 2024-08-26
Do not calculate and store silhouette edges for nearly every geometric prim with a corner


cosmic-linden:
`LLVolumeFace::mEdge` is calculated for many prims at region load time (depending on the flags), then stored long-term until the prim is unloaded. This is not necessary because the values of `mEdge` are only used for calculating silhouette edges, which are part of the selection outlines. Selection outlines can be calculated on-demand, so the silhouette edges should as well.

This should decrease time spent in `LLVolumeFace::createSide`, and reduce the memory overhead of cubes and some other prims.

## Test plan

Selection outlines should look the same as before this change.

igorlproductengine:
Need more info.
Verified on the Second Life Release 7.1.11.10947335521 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/348.

Could you please provide a test plan?

igorlproductengine:
Passed QA.

Verified on Second Life Release 7.1.11.10947335521 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/348.

Hello, @cosmic-linden. I believe these changes should be reflected in the statistics bar for better visibility and tracking.

PR #2449, opened 2024-08-29
secondlife/viewer#2421: Do not calculate and store silhouette edges for nearly every geometric prim with a corner

cosmic-linden:
Removes `LLVolumeFace::mEdge` in favor of calculating the "edge" array on-demand in `LLSelectMgr::generateSilhouette`.

Most of the contents of the "edge" array are used only for ifdef'd out debugging code - seems like low-hanging fruit for code removal, assuming we don't need that anymore.

### DIFF
--- a/indra/llmath/llvolume.h
+++ b/indra/llmath/llvolume.h
@@ -918,6 +918,15 @@ public:
     // Get a reference to the octree, which may be null
     const LLVolumeOctree* getOctree() const;
 
+    // Part of silhouette generation (used by selection outlines)
+    // Populates the provided edge array with numbers corresponding to
+    // *partial* logic of whether a particular index should be rendered
+    // as a silhouette edge. -1 indicates the index should be rendered as a
+    // silhouette edge. See generateSilhouetteVertices for the full logic.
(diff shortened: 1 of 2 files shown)


indra/llmath/llvolume.h
```diff
@@ -1113,9 +1120,9 @@ class LLVolume : public LLRefCount
 
 public:
     virtual void setMeshAssetLoaded(bool loaded);
-    virtual bool isMeshAssetLoaded();
+    virtual bool isMeshAssetLoaded() const;
     virtual void setMeshAssetUnavaliable(bool unavaliable);
-    virtual bool isMeshAssetUnavaliable();
+    virtual bool isMeshAssetUnavaliable() const;
```
brad-linden, line 1125: it looks like these methods are not overridden in any subclasses (I found LLModel and LLRiggedVolume)?  so this won't break them.  maybe in the future we should make these non-virtual?

brad-linden (approved):
👍

indra/llmath/llvolume.h
cosmic-linden: Yea, I'm not sure why these are virtual.

PR #2451, opened 2024-08-29
Give certain `LLInventoryListener` queries an API based on result sets.

nat-goodspeed:
Add `LL.setdtor()` function to set a "destructor" for any Lua object.

`setdtor('description', object, function)` returns a proxy userdata object
referencing object and function. When the proxy is garbage-collected, or at
the end of the script, its destructor calls `function(object)`.

The original object may be retrieved as `proxy._target`, e.g. to pass it to
the `table` library. The proxy also has a metatable with metamethods
supporting arithmetic operations, string concatenation, length, table
indexing and table iteration. For other operations, retrieve `proxy._target`.

Add a test script to verify `setdtor()` proxy behavior.

Inject into the viewer's Lua global namespace a replacement `next()` that
recognizes a `setdtor()` proxy and passes its underlying `_target` to the
built-in `next()` function. This permits iterating over a table wrapped by
a `setdtor()` proxy.

Similarly, but more broadly, inject a replacement `pairs()` function that
recognizes when the table of interest has an `__iter()` metamethod - the
built-in `pairs()` function does not. Also replace `ipairs()` with a function
that recognizes when the table of interest has an `__index()` metamethod.

Fix bugs in our `lua_setfieldv()`, `lua_rawgetfield()` and `lua_rawsetfield()`.

Add C++ functions `lua_destroyuserdata()` to explicitly destroy a
`lua_emplace<T>()` userdata object, plus `lua_destroybounduserdata()`. The
latter can bind such a userdata object as an upvalue to pass to `LL.atexit()`.

Introduce `LLIntTracker<T>`, a template `LLInstanceTracker<T, int>` subclass
that implicitly generates a distinct `int` key for each instance. An `int` key, unlike
a pointer, can be passed to a script via `LLSD`.

Introduce abstract base class `LL::ResultSet`, derived
from `LLIntTracker` so each instance has a distinct `int` key. `ResultSet` supports
virtual `getLength()` and `getSlice()` operations. `getSlice()` returns an `LLSD`
array limited to `MAX_ITEM_LIMIT` result set entries. It permits retrieving a "slice" of the
contained result set starting at an arbitrary index. A sequence of `getSlice()`
calls can eventually retrieve a whole result set.

In llinventorylistener.cpp,
introduce `ResultSet` subclasses `CatResultSet` containing `cat_array_t`, and
`ItemResultSet` containing `item_array_t`. Each implements a virtual method that
produces an `LLSD` map from a single array item.

Make `LLInventoryListener::getItemsInfo()`, `getDirectDescendants()` and
`collectDescendantsIf()` instantiate heap `CatResultSet` and `ItemResultSet` objects
containing the resultant `LLPointer` arrays, and return their int keys and total lengths for
categories and items.

Add `LLInventoryListener::getSlice()` and `closeResult()` methods that accept the
int keys of result sets. `getSlice()` returns the requested partial `LLSD` array to its
caller, while `closeResult()` is fire-and-forget.

Because bulk data transfer is now performed by `getSlice()` rather than by
`collectDescendantsIf()`, change the latter's "limit" default to unlimited.
Allow the C++ code to collect an arbitrary number of `LLPointer` array entries,
as long as `getSlice()` limits retrieval overhead.

Spell "descendants" correctly, unlike the "descendents" spelling embedded in
the rest of the viewer... sigh. Make the Lua module provide both spellings.

Make `MAX_ITEM_LIMIT` an `S32` instead of `F32`.

Introduce result_view.lua. `result_view()` accepts and stores a result-set key
and total length, and provides metamethods that support both indexed access
and traversal over the whole result set -- even though it only fetches one slice
at a time using the `getSlice()` operation. The table returned by `result_view()`
has a `close()` method that invokes `closeResult()` with the bound int key.
Add test_result_view.lua to verify its behavior.

In LLInventory.lua, wrap `result_view()` instances for categories and for items
from 'getItemsInfo', 'getDirectDescendants' and 'collectDescendantsIf'
in an outer table with `categories` and `items` fields. It is this outer table
that provides a `close()` function releasing result sets for both categories and
items, and that is wrapped with `setdtor()`.

Also make the "LLFloaterReg" listener's "getFloaterNames" operation return a
`result_view()` wrapped in `setdtor()`.

Tweak test scripts. Since `LLInventory.collectDescendantsIf()` now returns a `setdtor()`
proxy, `inspect(landmarks)` now reports "<userdata 1>" rather than enumerating the table.
But iterating over `pairs(landmarks.items)` works.

Instead of copying names of calling cards to a separate array and then picking randomly
from that, directly select a random calling card from the returned result set. Do it several
times to illustrate random access to the result set.

Add Lua `debug.traceback()` information to errors encountered in `LuaState::expr()`.

Make `lua_what()` and `lua_stack()` report the name, if known, for a function object.

Add script name to most Lua log messages.

Make `LL.help()` and `LL.leaphelp()` help text include the `LL.` prefix.


indra/llcommon/lua_function.cpp
```diff
@@ -974,6 +1002,289 @@ lua_function(
...
+    int rc = lua_pcall(L, 1, 0, 0);
+    if (rc != LUA_OK)
+    {
+        // TODO: we don't really want to propagate the error here.
+        // If this setdtor_refs instance is being destroyed by
+        // LL.atexit(), we want to continue cleanup. If it's being
+        // garbage-collected, the call is completely unpredictable from
+        // the consuming script's point of view. But what to do about this
+        // error?? For now, just log it.
+        LL_WARNS("Lua") << "setdtor(" << std::quoted(desc) << ") error: "
```
maxim-productengine, line 1280: Maybe, we should show script name here too, if the lua error is not thrown.

maxim-productengine (approved):
Looks good (at least I don't see any issues). 
I'm going to play with this new function after the merge to make sure it works as I think,

indra/llcommon/lua_function.cpp
nat-goodspeed: Really good point! I've added it to other Lua log messages too.

indra/newview/llinventorylistener.cpp
```diff
@@ -61,78 +62,190 @@ LLInventoryListener::LLInventoryListener()
         &LLInventoryListener::getBasicFolderID,
         llsd::map("ft_name", LLSD(), "reply", LLSD()));
 
-    add("getDirectDescendents",
-        "Return the direct descendents(both items and folders) of the [\"folder_id\"]",
-        &LLInventoryListener::getDirectDescendents,
+    add("getDirectDescendants",
```
maxim-productengine, line 65: I just copied the func name from llinventorymodel without thinking about spelling. 
Somehow most of our inventory-related functions have incorrect spelling (collectDescendEnts).

indra/newview/scripts/lua/require/LLInventory.lua
```diff
@@ -1,12 +1,66 @@
...
+                    -- with the same name but an underscore prefix.
+                    local resultkey = t['_' .. key]
+                    -- TODO: This only ever fetches the FIRST slice. What we
+                    -- really want is to return a table with metamethods that
+                    -- manage indexed access and table iteration.
+                    -- Remember our C++ entry point uses 0-relative indexing.
+                    local slice = leap.request(
+                        'LLInventory',
+                        {op='getSlice', result=resultkey, index=0}).slice
+                    print(`getSlice({resultkey}, 0) => {slice} ({#slice} entries)`)
```
maxim-productengine, line 42: Probably this line should be commented out.

indra/newview/llinventorylistener.cpp
nat-goodspeed: I'm sorry, I did not mean to imply that the misspelling was your fault!

maxim-productengine: No worries, I was just surprised that I hadn't noticed it before.

nat-goodspeed:
I'm pretty excited about the `LL::ResultSet` abstract base class and the Lua `result_view()` function, allowing us to leverage result-set functionality for more than `LLInventory` -- as exemplified by [the change to `"getFloaterNames"`](https://github.com/secondlife/viewer/pull/2451/commits/a8dd7135f0423384dbbb1e3b98514149c6a69e6b). Feel free to disagree with that particular use case, but it does show that it's relatively straightforward on both sides of the language boundary.

maxim-productengine:
II really like it and it looks clean!

Not related - I've just remembered that we have some inventory functions in LLAppearance that actually can be updated and moved to LLInventory.

maxim-productengine (approved):

PR #2452, opened 2024-08-29
update OpenJPEG 2.5.2

AiraYumi:
Update OpenJPEG to 2.5.2.

If the build is successful and there are no problems, please merge it.
If the build fails, fix the build errors.


marchcat (approved):

marchcat:
Thank you!

PR #2455, opened 2024-08-29
#2360 Incorrect Day Cycle name (remove unused key CANMOD)

LLGuru:


marchcat (approved):

akleshchev:
I think it's better to fix `mCanMod = true;` at string 307 to not override value from the key.

akleshchev (approved):

Issue #1856, opened 2024-06-27
BugSplat Crash: SecondLifeViewer!LLRender::flush(1637)


sl-service-account:
### Test plan:
1. Verify the viewer starts and can log in on a Windows system with any of the following video cards.
Intel(R) HD Graphics (this one may be unsupported on 64bit Windows)
Intel(R) HD Graphics Family (this one may be unsupported on 64bit Windows)
Intel(R) HD Graphics 4000
Intel(R) HD Graphics 4400
Intel(R) HD Graphics 4600
Intel(R) HD Graphics 5500
Intel(R) HD Graphics 6000
2. Please note which video card was tested.

Expected: Log in without crash

**BugSplat** [Key Crash 94199](https://app.bugsplat.com/v2/keycrash?stackKeyId=94199&database=SecondLife_Viewer_2018)

**Sample Version:**
7.1.9.9620320242

**Sample Callstack:**
[...]

PR #2456, opened 2024-08-29
error handling for intel crashes from GLTF Scene shader

brad-linden:
should fix secondlife/viewer#1856


akleshchev (approved):

brad-linden:
ok dan says the fix works, the other issue he was running into is unrelated.  also I fixed it so the notification only shows once when attempting to open a scene, rather than every frame (oops)

PR #2463, opened 2024-08-29
#2428 Fix for crash when applying PBR material to a single face (take 2)

RunitaiLinden:
Also fix for some objects randomly disappearing until they're selected again.


brad-linden (approved):
looks good

PR #2470, opened 2024-08-30
Add Lua api to start/stop playing animation

maxim-productengine:


indra/newview/llagentlistener.cpp
```diff
@@ -591,3 +609,60 @@ void LLAgentListener::removeFollowCamParams(LLSD const & event) const
 {
     LLFollowCamMgr::getInstance()->removeFollowCamParams(gAgentID);
 }
+
+LLViewerInventoryItem* get_anim_item(LLEventAPI::Response &response, const LLSD &event_data)
+{
+    LLViewerInventoryItem* item = gInventory.getItem(event_data["item_id"].asUUID());
+    if (!item || (item->getInventoryType() != LLInventoryType::IT_ANIMATION))
+    {
+        response.error(stringize("Animation item ", std::quoted(event_data["item_id"].asString()), " was not found"));
```
nat-goodspeed, line 618: fwiw `LLEventAPI`'s base class `LLDispatchListener` catches exceptions, and returns to the invoker a response LLSD with an "error" key containing the exception message. So throwing an exception for an error case is also a reasonable thing to do, especially in a helper function that wouldn't otherwise need access to the Response object.

nat-goodspeed (approved):

indra/newview/llagentlistener.cpp
maxim-productengine: Thanks, I didn't know about that.

PR #2477, opened 2024-09-02
Font Rendering optimisations

akleshchev:
Store and draw buffers directly instead of doing vertex generation and cache search.


marchcat (approved):

indra/llui/llbutton.cpp
```diff
@@ -396,6 +421,30 @@ bool LLButton::postBuild()
...
+}
+
+void LLButton::reshape(S32 width, S32 height, bool called_from_parent)
+{
+    S32 delta_width = width - getRect().getWidth();
+    S32 delta_height = height - getRect().getHeight();
+
+    if (delta_width || delta_height || sForceReshape)
+    {
+        LLUICtrl::reshape(width, height, called_from_parent);
```
maxim-productengine, line 437: Won't it call `LLButton::dirtyRect()` indirectly, so font buffer will be reset twice?

indra/llui/llbutton.cpp
```diff
@@ -1025,11 +1081,13 @@ bool LLButton::toggleState()
 void LLButton::setLabel( const std::string& label )
 {
     mUnselectedLabel = mSelectedLabel = label;
+    mFontBuffer.reset();
```
maxim-productengine, line 1084: Maybe adding simple check that a label is actually new will be useful?

maxim-productengine (approved):

indra/llui/llbutton.cpp
akleshchev: You are correct, will remove this.

RunitaiLinden (approved):

PR #2479, opened 2024-09-02
#2403 Reset Skeleton issues

LLGuru:


marchcat (approved):

indra/newview/llviewermenu.cpp
```diff
@@ -6748,57 +6748,63 @@ class LLAvatarToggleSearch : public view_listener_t
...
     }
 };
 
+class LLAvatarResetSelfSkeleton : public view_listener_t
+{
+    bool handleEvent(const LLSD& userdata)
+    {
+        if (LLVOAvatar* avatar = find_avatar_from_object(LLSelectMgr::getInstance()->getSelection()->getPrimaryObject()))
+        {
+            avatar->resetSkeleton(false);
```
akleshchev, line 6793: This search seems pointless.

If this is always reseting own skeleton, why not just call gAgentAvatarp->resetSkeleton(false) imediately?

Issue #2482, opened 2024-09-03
Second Life reports incorrect amount of available video memory on some discrete GPUs on Windows


RunitaiLinden:
### Description

Testing on NVIDIA GT 1030 2GB, the viewer reports 4GB of vram are available and exhausts available video memory, leading to poor performance.

Test Plan:
1. Open task manager, click performance, click GPU, and note "Dedicated GPU Memory"
2. In Second Life, click Help->About Second Life and note "Texture Memory"
EXPECTED: Texture Memory should agree with "Dedicated GPU Memory" on discrete GPUs
3. Open Texture Console (Ctrl+Shift+3) and note "Est. Free"
EXPECTED: Est. Free should agree with task manager

RunitaiLinden:
Fixed in DeltaFPS and did a low-end NVIDIA performance pass:

- Use GL_NVX_gpu_memory_info when available
- Disable transparent water on Mid+ and lower
- Adjust GPU benchmark to better tell the truth
- Texture bias tune up

RunitaiLinden:
GT 1050 looks like a pass, but the OS is giving some other app extra VRAM.  Est. Free is calculated by subtracting Total from Texture Memory, then taking off an additonal 512MB plus some percent of Total, which usually gets in the ballpark of what Task Manager reports, but sometimes some other app is eating texture memory.  When this is the case, the OS should free vram and give it to the app that has focus when vram is exhausted, so we pretend we have access to most of the available vram even if other apps are taking vram, otherwise we end up never using available vram and always ceding to backgrounded tasks.

The Quadro is definitely a fail... looking into that.

RunitaiLinden:
Potentially fixed in DeltaFPS

Don't have a quadro to test with, but from the logs it looks like we were getting the right VRAM number from OpenGL, then overriding it with the wrong number from DirectX.  Fix was to only ask DirectX if OpenGL was unable to provide a sensible VRAM number.

DmitrySProductengine:
Passed QA.
Verified on the Second Life Release 7.1.10.10800445603 (64bit) on Win10/Win11 in the scope of [IQA #332](https://github.com/secondlife/iqa/issues/332).

During the test, video cards with different memory sizes were tested.
The parameter 'Est. Free' almost corresponds to the value of free dedicated video memory.

Video cards:
1. ATI Radeon HD 2600 XT - 512MB video memory  
[...]

PR #2486, opened 2024-09-03
#2482 Low end NVIDIA compatibility pass

RunitaiLinden:
- Use GL_NVX_gpu_memory_info when available
- Disable transparent water on Mid+ and lower
- Adjust GPU benchmark to better tell the truth
- Texture bias tune up

### DIFF
--- a/indra/newview/llviewertexture.cpp
+++ b/indra/newview/llviewertexture.cpp
@@ -511,33 +511,20 @@ void LLViewerTexture::updateClass()
 
     F32 over_pct = (used - target) / target;
 
-    bool is_low = over_pct > 0.f;
+    bool is_sys_low = isSystemMemoryLow();
+    bool is_low = is_sys_low || over_pct > 0.f;
 
     if (isSystemMemoryLow())
     {
-        is_low = true;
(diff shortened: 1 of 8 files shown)


indra/newview/llviewertexture.cpp
```diff
@@ -511,33 +511,20 @@ void LLViewerTexture::updateClass()
 
     F32 over_pct = (used - target) / target;
 
-    bool is_low = over_pct > 0.f;
+    bool is_sys_low = isSystemMemoryLow();
+    bool is_low = is_sys_low || over_pct > 0.f;
 
     if (isSystemMemoryLow())
```
RyeMutt, line 517: Redundant with `bool is_sys_low = isSystemMemoryLow();` above?

RyeMutt (approved):
Small comment about possible cleanup, but otherwise looks good!

indra/newview/llviewertexture.cpp
RunitaiLinden: Whoops... I had changes that were using "is_sys_low" to make it not slam update all the virtual sizes at once unless we were low on system memory, guess I fat fingered "undo" at some point.  Good catch.

PR #2489, opened 2024-09-03
Fix visualizing luminance buffer and small cleanup

RyeMutt:
Fix visualizing luminance buffer

### DIFF
--- a/indra/newview/pipeline.cpp
+++ b/indra/newview/pipeline.cpp
@@ -7871,7 +7871,6 @@ void LLPipeline::renderFinalize()
     renderDoF(&mPostMap, &mRT->screen);
 
     LLRenderTarget* finalBuffer = &mRT->screen;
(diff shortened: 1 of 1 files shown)


indra/newview/pipeline.cpp
```diff
@@ -7880,7 +7880,6 @@ void LLPipeline::renderFinalize()
 
     if (RenderBufferVisualization > -1)
     {
-        finalBuffer = &mPostMap;
```
Ansariel, line 7883: So is this correct that you overwrite the buffer you have been written to last? Or am I missing something in visualizeBuffers()?

Also could remove the unused `aa_type` LLCachedControl.

RyeMutt: Yes that's correct. `visualizeBuffers()` overwrites the output buffer with whatever buffer is being requested to be visualized.

RunitaiLinden (approved):

PR #2497, opened 2024-09-04
Added missing comma in a list. Fixed loop with too small a loop variable.

Hecklezz:
The build indices loop in LLImagePreviewSculpted::setPreviewTarget had the loop variable set as a U16 when it should be a U32.
The blacklist_untrusted list in LLFloaterOpenHandler::canHandleUntrusted is missing a comma for one of the entries.


akleshchev (approved):

marchcat (approved):
Thank you!

Issue #2500, opened 2024-09-04
Possible error in llmd5 finalize function leaving sensitive information not zeroized?


Hecklezz:
### Description

Unless I'm mistaken, is [line 219 in llmd5.cpp](https://github.com/secondlife/viewer/blob/c2724537adab9d31c23d33e36002772a9c56f4c3/indra/llcommon/llmd5.cpp#L219) not supposed to be:
`memset (buffer, 0, sizeof(buffer));`
or more preferably in modern c++:
`buffer = {};`

Or is it intentional that only the first byte of the buffer is getting zeroized?

marchcat:
For QA: there were code quality improvements made in #2507, which shouldn't affect any visible behavior.

Dan-Linden:
Nothing to test. Moving this to Pending Release.

vladyslavasproductengine:
Passed QA according to the Dan-Linden comment.
The issue was investigated in the scope of https://github.com/secondlife/iqa/issues/356.
The ticket is in Pending Release status and has not been changed.

PR #2507, opened 2024-09-05
Fixed md5 buffer in LLMD5::finalize not being correctly zero-initialized

Hecklezz:
Fixes the error mentioned in my issue posted here-
https://github.com/secondlife/viewer/issues/2500

Unless I am mistaken and it is intentional, the md5 buffer in the LLMD5::finalize function is being incorrectly zeroized and instead only the first byte is being zeroized.

Also went ahead and refactored the code so it more closely matches the styling of the rest of the codebase.

### DIFF
--- a/indra/llcommon/llmd5.cpp
+++ b/indra/llcommon/llmd5.cpp
@@ -68,10 +68,6 @@ documentation and/or software.
 
  */
 
-
-
-
-
 #include "linden_common.h"
 
 #include "llmd5.h"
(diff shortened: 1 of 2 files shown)


marchcat (approved):
Nice! Thank you!

marchcat:
Mac build fails with
```
/Users/runner/work/viewer/viewer/indra/llcommon/llmd5.cpp:219:10: error: array type 'uint8_t[64]' is not assignable
  buffer = {};
  ~~~~~~ ^
1 error generated.
```
so the buffer should be either initialized as 
```
std::array<uint8_t, 64> buffer = {};
```
or, alternatively, we can use std::fill:
```
std::fill(std::begin(buffer), std::end(buffer), 0);
```

Hecklezz:
Whoops my bad, I really should have tested compiling first before blindly making that change and trusting intellisense that it said nothing was wrong and not remembering that C arrays are not assignable like that after they have already been defined. 

Should we stick with memset since a large portion of the rest of the codebase is still using memset, or should the code be changed to be more in line with modern C++ standards?

Nicky-D:
While at it, can we maybe change the finalized member to a bool?

Regarding memset. If this is down deemed security related and the buffer must be cleared, then memset_s or something like SecureZeroMemory needs to be employed. 
Otherwise the compiler can happily optimize that call to memset away (dunno if it would do here, probably not?)

Hecklezz:
I went back to using regular memset. Doing some googling, supposedly MSVC does not even have memset_s implemented even if you set the define it wants, and SecureZeroMemory is windows specific so that isn't helpful either.

I also went ahead and did a bit of a tidy-up of the llmd5 files so they weren't so outdated interms of code styling.

marchcat (approved):
It looks much better now.

Thank you for your efforts!

PR #2515, opened 2024-09-05
Fix remaining occurrences of fake BOOL

Ansariel:
@nat-goodspeed you missed 2 occurrences during the merge


akleshchev (approved):

marchcat (approved):
Good catch, thanks!

Ansariel:
@marchcat can you give an update on the different physicsextensions libs? Are they updated already so we can remove the BOOL-related definitions and remaining methods?

Ansariel:
@nat-goodspeed can we merge this before this branch is getting merched into `develop`, please?

maxim-productengine (approved):

PR #2518, opened 2024-09-06
Feature: Added toggles to avatar dropdown for hear media and sound or voice from avatar.

Hecklezz:
Added two toggles to the Avatar top dropdown menu:
1. Hear Media and Sound from Avatar
2. Hear Voice from Avatar

The first one simply toggles the "MediaSoundsEarLocation" control and the second toggles the "VoiceEarLocation" control and fires the EnableVoiceChat callback.

This feature is just a small QOL so people who want to be able to change these options on the fly for certain scenarios, don't need to go digging through preferences.

Possible additions required/wanted:
1. Translations for other languages.
2. Toggleable toolbar buttons that users can add to their toolbars (disabled by default).

secondlife-bin_2024-09-06_14-34-18


marchcat (approved):
Thank you very much!

PR #2520, opened 2024-09-06
Changes towards C++20 compatibility

Ansariel:
Since there are still issues on macOS side preventing the actual switch to C++20 (see #2330), I broke out some changes that also work on C++17 and that will be needed in the future. I left out some changes I am not sure of if they would cause issues on macOS currently.

### DIFF
--- a/indra/newview/llviewertexteditor.cpp
+++ b/indra/newview/llviewertexteditor.cpp
@@ -503,7 +503,7 @@ S32 LLEmbeddedItems::getIndexFromEmbeddedChar(llwchar wch)
     }
     else
     {
-        LL_WARNS() << "Embedded char " << wch << " not found, using 0" << LL_ENDL;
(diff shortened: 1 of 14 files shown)


marchcat (approved):
Thank you, @Ansariel!

Mac builds are green for this one, merging.

indra/newview/llviewertexteditor.cpp
```diff
@@ -503,7 +503,7 @@ S32 LLEmbeddedItems::getIndexFromEmbeddedChar(llwchar wch)
     }
     else
     {
-        LL_WARNS() << "Embedded char " << wch << " not found, using 0" << LL_ENDL;
+        LL_WARNS() << "Embedded char " << (int)wch << " not found, using 0" << LL_ENDL;
```
nat-goodspeed, line 506: See above remarks about `ll_convert_to<std::string>(wch)`.

PR #2523, opened 2024-09-06
Put viewer's Lua functionality behind a feature flag, default off.

nat-goodspeed:
When disabled, the Lua machinery simply returns an error message.

Remove Lua floaters from the default menu structure, but add a Lua autorun script to dynamically re-add them if Lua is enabled.

### DIFF
--- a/indra/newview/lluilistener.cpp
+++ b/indra/newview/lluilistener.cpp
@@ -90,12 +90,14 @@ LLUIListener::LLUIListener():
     add("addMenuItem",
         "Add new menu item [\"name\"] with displayed [\"label\"]\n"
         "and call-on-click UI function [\"func\"] with optional [\"param\"]\n"
-        "to the [\"parent_menu\"] within the Top menu.",
+        "to the [\"parent_menu\"] within the Top menu.\n"
+        "If [\"pos\"] is present, insert at specified 0-relative position.",
         &LLUIListener::addMenuItem,
         required_args.with("func", LLSD()));
 
     add("addMenuSeparator",
-        "Add menu separator to the [\"parent_menu\"] within the Top menu.",
+        "Add menu separator to the [\"parent_menu\"] within the Top menu.\n"
+        "If [\"pos\"] is present, insert at specified 0-relative position.",
         &LLUIListener::addMenuSeparator,
         llsd::map("parent_menu", LLSD(), "reply", LLSD()));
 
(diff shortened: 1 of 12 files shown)


indra/newview/lluilistener.cpp
```diff
@@ -302,9 +311,19 @@ void LLUIListener::addMenuItem(const LLSD&event) const
     item_params.on_click = item_func;
     if(LLMenuGL* parent_menu = get_parent_menu(response, event))
     {
-        if(!parent_menu->append(LLUICtrlFactory::create<LLMenuItemCallGL>(item_params)))
+        auto item = LLUICtrlFactory::create<LLMenuItemCallGL>(item_params);
+        // Clamp pos to getItemCount(), meaning append. If pos exceeds that,
+        // insert() will silently ignore the request.
+        auto pos = get_pos(event, parent_menu->getItemCount());
```
maxim-productengine, line 317: Correct me if I'm wrong, but after clamping to getItemCount() here we'll still to insert() and it will fail.

indra/newview/lluilistener.cpp
```diff
@@ -264,6 +266,13 @@ LLMenuGL* get_parent_menu(LLEventAPI::Response& response, const LLSD&event)
     return parent_menu;
 }
 
+// Return event["pos"].asInteger() if passed, but clamp (0 <= pos <= size).
+// Reserve -1 return to mean event has no "pos" key.
+S32 get_pos(const LLSD& event, U32 size)
+{
+    return event["pos"].isInteger()? llclamp(event["pos"].asInteger(), 0, size) : -1;
```
maxim-productengine, line 273: 0 menu position is used for 'double line' in the top menu (which allows to undock that menu). Should we allow inserting new items before it, basically pushing 'double line' down? It's ok for me if you think we should (though it's somewhat inconsistent), Otherwise, let's clamp to 1.

maxim-productengine (approved):

indra/newview/lluilistener.cpp
nat-goodspeed: Hmm! I based this `llclamp()` call on the test in `LLMenuGL::insert()`. Are you saying that if we allow inserting at 0, tearing off the menu would leave that first (inserted) item attached to the top menu?

If we change this to 1, should we also change the test in `LLMenuGL::insert()`?

maxim-productengine: Using test_top_menu.lua: current script vs inserting "Debug console" at 0.
image

I'm not sure about `LLMenuGL::insert()`. As far as I remember, only top menus have these 'tear-off' items(LLMenuItemTearOffGL). So, it makes sense to allow inserting at 0 in general for other menus, the exception is the top menu.

PR #2525, opened 2024-09-07
Update calcDataSizeJ2C to stop undersized blocks

TommyTheTerrible:
The initial block area for the pyramid walk should not be smaller than the max_block_size area so there needs to be an llmax to not allow multiplication below 1.

This was causing decode errors for complex small images (128x128 or smaller) on discards 1 and 2.

Tested on OpenJPEG 2.5. 

Should improve quality of decodes for affected cases in KDU but probably should be tested just in case.


marchcat (approved):
Thank you!

PR #2527, opened 2024-09-09
Reduce 'World map' floater draw time

maxim-productengine:
'World map' floater draw time is reduced by approx. 750μs when observing a cluster of 4 PE regions.

### DIFF
--- a/indra/newview/llfloaterworldmap.cpp
+++ b/indra/newview/llfloaterworldmap.cpp
@@ -1773,8 +1773,8 @@ void LLFloaterWorldMap::onCommitSearchResult()
     }
     LLStringUtil::toLower(sim_name);
 
-    std::map<U64, LLSimInfo*>::const_iterator it;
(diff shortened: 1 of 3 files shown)


akleshchev (approved):

marchcat (approved):

indra/newview/llfloaterworldmap.cpp
```diff
@@ -1773,8 +1773,8 @@ void LLFloaterWorldMap::onCommitSearchResult()
     }
     LLStringUtil::toLower(sim_name);
 
-    std::map<U64, LLSimInfo*>::const_iterator it;
-    for (it = LLWorldMap::getInstance()->getRegionMap().begin(); it != LLWorldMap::getInstance()->getRegionMap().end(); ++it)
+    LLWorldMap::sim_info_map_t::const_iterator end = LLWorldMap::instance().getRegionMap().end();
+    for (LLWorldMap::sim_info_map_t::const_iterator it = LLWorldMap::getInstance()->getRegionMap().begin(); it != end; ++it)
```
nat-goodspeed, line 1777: I realize this is already merged, but while touching an iterator-based `for` is a pretty good time to change to range `for`.

PR #2528, opened 2024-09-09
[Maint B] Fix viewer channel selection based on a branch

marchcat:


marchcat:
The build log now shows
```
    AUTOBUILD_VCS_BRANCH: marchcat/channel-fix
```
for the PR branch, so this should work. It was empty before the fix.

akleshchev (approved):

.github/workflows/build.yaml
```diff
@@ -152,7 +152,7 @@ jobs:
         id: build
         shell: bash
         env:
-          AUTOBUILD_VCS_BRANCH: ${{ needs.setvar.outputs.branch }}
+          AUTOBUILD_VCS_BRANCH: ${{ steps.which-branch.outputs.branch }}
```
akleshchev, line 155: Isn't it supposed to take channel from setvar?

marchcat: It's an old piece of code that somehow sneaked in during the merge. The current code in the develop branch uses `which-branch`.

Issue #2532, opened 2024-09-09
Voice: After calling and hanging up on a group, the call button remains grey.


roxanneskelly:
### Environment

Second Life Release 7.1.9.10515727195 (64bit)
Release Notes

You are at 128.0, 128.0, 23.0 in webRTC1 located at simhost-07c23d76c83837d55.aditi
SLURL: secondlife://Aditi/secondlife/webRTC1/128/128/23
(global coordinates 259,968.0, 245,376.0, 23.0)
WebRTC-Voice 2024-08-30.10639499260
Release Notes

CPU: Intel(R) Core(TM) Ultra 9 185H (3072 MHz)
Memory: 32213 MB
OS Version: Microsoft Windows 11 64-bit (Build 22631.4037)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 4060 Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.5612
OpenGL Version: 4.6.0 NVIDIA 556.12

Window size: 1694x989
Font Size Adjustment: 96pt
UI Scaling: 1.05
Draw distance: 128m
Bandwidth: 10000kbit/s
LOD factor: 1.75
Render quality: 5
Texture memory: 7192MB
Disk cache: Max size 3993.6 MB (100.0% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 0/211 (0.0%)
September 09 2024 10:58:53

### Description

After calling and hanging up on a group, without closing chat, the call button remains disabled.

### Reproduction steps

1. User logs in.
2. Bring up a group via 'people' and start a chat.
3. Click the call button.
4. hang up the call.
5. Observe the call button.

Expected: The call button re-enables after the call is hung up.
Actual: The call button remains disabled.

Workaround: close the chat session and re-open it.

katerynamproductengine:
Passed QA.

This issue is not reproduced on Agni and Aditi grids on Vivox/WebRTC regions.
(Simulator WebRTC:
[...]

PR #2533, opened 2024-09-09
Callback button for groups was not being reenabled after hanging up on group call.

roxanneskelly:
For #2532, callback button was remaining disabled after hanging up on a group call.

Also, fix https://github.com/secondlife/viewer-private/issues/283 where user was automatically added to a group call after the initiator of the group call hangs up and re-calls the group.

See those issues for repro steps.

### DIFF
--- a/indra/newview/llfloaterimsessiontab.cpp
+++ b/indra/newview/llfloaterimsessiontab.cpp
@@ -450,7 +450,8 @@ void LLFloaterIMSessionTab::enableDisableCallBtn()
         else
         {
             // We allow to start call from this state only
-            if (mSession->mVoiceChannel->getState() == LLVoiceChannel::STATE_NO_CHANNEL_INFO &&
+            if (mSession->mVoiceChannel  &&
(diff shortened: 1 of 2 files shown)


indra/newview/llfloaterimsessiontab.cpp
```diff
@@ -494,10 +495,7 @@ void LLFloaterIMSessionTab::onCallButtonClicked()
     }
     else
     {
-        LLVoiceChannel::EState channel_state = mSession && mSession->mVoiceChannel ?
-            mSession->mVoiceChannel->getState() : LLVoiceChannel::STATE_NO_CHANNEL_INFO;
-        // We allow to start call from this state only
-        if (channel_state == LLVoiceChannel::STATE_NO_CHANNEL_INFO)
+        if (mSession->mVoiceChannel && !mSession->mVoiceChannel->callStarted())
```
roxanneskelly, line 498: states of hung up and error should also allow calls.

brad-linden (approved):

PR #2538, opened 2024-09-10
fix for alpha-gamma-bypass prototype

AndrewMeadows:
This PR fixes viewer-side problems in the "alpha-gamma-prototype".

### DIFF
--- a/indra/llprimitive/llprimitive.cpp
+++ b/indra/llprimitive/llprimitive.cpp
@@ -125,6 +125,8 @@ const LLUUID SCULPT_DEFAULT_TEXTURE("be293869-d0d9-0a69-5989-ad27f1946fd4"); // 
 // can't be divided by 2.   See DEV-19108
 const F32   TEXTURE_ROTATION_PACK_FACTOR = ((F32) 0x08000);
 
+constexpr U8 EXTRA_PROPERTY_ALPHA_GAMMA = 0x01;
+
 struct material_id_type // originally from llrendermaterialtable
 {
     material_id_type()
@@ -1262,6 +1264,7 @@ bool LLPrimitive::packTEMessage(LLMessageSystem *mesgsys) const
             memcpy(&material_data[face_index*16],getTE(face_index)->getMaterialID().get(),16);  /* Flawfinder: ignore */
         }
 
+        // pack required properties
         cur_ptr += packTEField(cur_ptr, (U8 *)image_ids, sizeof(LLUUID),last_face_index, MVT_LLUUID);
         *cur_ptr++ = 0;
         cur_ptr += packTEField(cur_ptr, (U8 *)colors, 4 ,last_face_index, MVT_U8);
@@ -1284,8 +1287,12 @@ bool LLPrimitive::packTEMessage(LLMessageSystem *mesgsys) const
         *cur_ptr++ = 0;
         cur_ptr += packTEField(cur_ptr, (U8 *) material_data, 16, last_face_index, MVT_LLUUID);
         *cur_ptr++ = 0;
-        *cur_ptr++ = 0x01;
+        // end of required properties
+
+        // pack extra properties
+        *cur_ptr++ = EXTRA_PROPERTY_ALPHA_GAMMA; // indicator alpha_gamma is present
         cur_ptr += packTEField(cur_ptr, (U8 *) alpha_gamma, 1, last_face_index, MVT_U8);
+        *cur_ptr++ = 0; // always null-terminate each propery
     }
     mesgsys->addBinaryDataFast(_PREHASH_TextureEntry, packed_buffer, (S32)(cur_ptr - packed_buffer));
 
@@ -1349,6 +1356,7 @@ bool LLPrimitive::packTEMessage(LLDataPacker &dp) const
             memcpy(&material_data[face_index*16],getTE(face_index)->getMaterialID().get(),16);  /* Flawfinder: ignore */
         }
 
+        // pack required properties
         cur_ptr += packTEField(cur_ptr, (U8 *)image_ids, sizeof(LLUUID),last_face_index, MVT_LLUUID);
         *cur_ptr++ = 0;
         cur_ptr += packTEField(cur_ptr, (U8 *)colors, 4 ,last_face_index, MVT_U8);
@@ -1371,8 +1379,12 @@ bool LLPrimitive::packTEMessage(LLDataPacker &dp) const
         *cur_ptr++ = 0;
         cur_ptr += packTEField(cur_ptr, (U8 *) material_data, 16, last_face_index, MVT_LLUUID);
         *cur_ptr++ = 0;
-        *cur_ptr++ = 0x01;
+        // end of required properties
+
+        // pack extra properties
+        *cur_ptr++ = EXTRA_PROPERTY_ALPHA_GAMMA; // indicator alpha_gamma is present
         cur_ptr += packTEField(cur_ptr, (U8 *) alpha_gamma, 1, last_face_index, MVT_U8);
+        *cur_ptr++ = 0; // always null-terminate each propery
     }
 
     dp.packBinaryData(packed_buffer, (S32)(cur_ptr - packed_buffer), "TextureEntry");
@@ -1445,17 +1457,18 @@ S32 LLPrimitive::parseTEMessage(LLMessageSystem* mesgsys, char const* block_name
         tec.material_ids[i].set(&(material_data[i]));
     }
 
-    if ((cur_ptr < buffer_end) && (*cur_ptr == 0x01))
+    if ((cur_ptr < buffer_end) && (*cur_ptr == EXTRA_PROPERTY_ALPHA_GAMMA))
     {
+        ++cur_ptr; // skip the indicator
         if (!unpack_TEField<U8>(tec.alpha_gamma, tec.face_count, cur_ptr, buffer_end, MVT_U8))
         {
-            LL_WARNS("TEXTUREENTRY") << "Baddly formed alphagamma unpacking texture entry." << LL_ENDL;
+            LL_WARNS("TEXTUREENTRY") << "Malformed alphagamma unpacking texture entry." << LL_ENDL;
         }
     }
 
     retval = 1;
     return retval;
-    }
+}
 
 S32 LLPrimitive::applyParsedTEMessage(LLTEContents& tec)
 {
@@ -1521,6 +1534,7 @@ S32 LLPrimitive::unpackTEMessage(LLDataPacker &dp)
     U8          bump[MAX_TES];
     U8          media_flags[MAX_TES];
     U8          glow[MAX_TES];
+    U8          alpha_gamma[MAX_TES];
     material_id_type material_data[MAX_TES];
 
     memset((void*)scale_s, 0, sizeof(scale_s));
@@ -1531,6 +1545,9 @@ S32 LLPrimitive::unpackTEMessage(LLDataPacker &dp)
     memset((void*)bump, 0, sizeof(bump));
     memset((void*)media_flags, 0, sizeof(media_flags));
     memset((void*)glow, 0, sizeof(glow));
+
+    constexpr U8 DEFAULT_ALPHA_GAMMA = 100;
+    memset((void*)alpha_gamma, DEFAULT_ALPHA_GAMMA, sizeof(alpha_gamma));
 
     S32 size;
     U32 face_count = 0;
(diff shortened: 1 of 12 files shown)


indra/llprimitive/llprimitive.cpp
```diff
@@ -125,6 +125,8 @@ const LLUUID SCULPT_DEFAULT_TEXTURE("be293869-d0d9-0a69-5989-ad27f1946fd4"); //
 // can't be divided by 2.   See DEV-19108
 const F32   TEXTURE_ROTATION_PACK_FACTOR = ((F32) 0x08000);
 
+constexpr U8 EXTRA_PROPERTY_ALPHA_GAMMA = 0x01;
+
```
AndrewMeadows, line 129: Use `EXTRA_PROPERTY_ALPHA_GAMMA` for less magic.

indra/llprimitive/llprimitive.cpp
```diff
@@ -1588,6 +1605,15 @@ S32 LLPrimitive::unpackTEMessage(LLDataPacker &dp)
...
 
+    if ((cur_ptr < buffer_end) && (*cur_ptr == EXTRA_PROPERTY_ALPHA_GAMMA))
+    {
+        ++cur_ptr; // skip the indicator
+        if (!unpack_TEField<U8>(alpha_gamma, face_count, cur_ptr, buffer_end, MVT_U8))
+        {
+            LL_WARNS("TEXTUREENTRY") << "Malformed alphagamma unpacking texture entry." << LL_ENDL;
+        }
+    }
+
```
AndrewMeadows, line 1616: Whoops, the version of `LLPrimitive::unpackTEMessage()` being used for the data from the server had not been instrumented to handle the new "AlphaGamma" data!  This is a good example of how bugs creep into code that is not **DRY**'ed.  Unfortunately, de-duplicating the code paths here would be a little harder.  It should be done, however I'm willing to be convinced we shouldn't block this PR for it.

indra/newview/llselectmgr.cpp
```diff
@@ -2338,21 +2350,20 @@ void LLSelectMgr::selectionSetAlphaGamma(U8 gamma)
...
         }
     } setfunc(gamma);
     getSelection()->applyToTEs(&setfunc);
 
-    LLSelectMgrSendFunctor sendfunc;
+    // TODO: remove LLSelectMgrAlphaGammaBypassFunctor and instead
+    // use the normal LLSelectMgrSendFunctor and enhance the server to know when to allow bypass
+    LLSelectMgrAlphaGammaBypassFunctor sendfunc;
     getSelection()->applyToObjects(&sendfunc);
 }
```
AndrewMeadows, line 2368: That was the wrong place to call `packAlphaGammaOverride()`.  By putting it there the viewer would send one message for every object face!

The correct way to do it is to first make ALL of the changes to the local object (if permission rules allow) and only when all changes made: send an update to the server.  The server validates the changes and will always send a confirmation or denied update back to the viewer that tried to make the change.

indra/newview/llselectmgr.h
```diff
@@ -901,12 +901,13 @@ class LLSelectMgr : public LLEditMenuHandler, public LLSimpleton<LLSelectMgr>
...
     static void logDetachRequest(LLSelectNode* node, void *user_data);
     static bool confirmDelete(const LLSD& notification, const LLSD& response, LLObjectSelectionHandle handle);
-    static void packAlphaGammaOverride(LLViewerObject* object);
 
     // Get the first ID that matches test and whether or not all ids are identical in selected objects.
     void getFirst(LLSelectGetFirstTest* test);
 
 public:
+    static void packAlphaGammaBypass(LLViewerObject* object);
+
```
AndrewMeadows, line 910: This method must be `public `because it is called by `LLSelectMgrAlphaGammaBypassFunctor`.

indra/llprimitive/llprimitive.cpp
```diff
@@ -159,6 +162,71 @@ const U8 material_id_type::s_null_id[] = { 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0 };
...
+    glow = reinterpret_cast<U8*>(data + offset);
+    offset += num_textures * sizeof(U8);
+
+    alpha_gamma = reinterpret_cast<U8*>(data + offset);
+    offset += num_textures * sizeof(U8);
+
+    // verify we correctly computed data size
+    llassert(offset == num_bytes);
+}
```
AndrewMeadows, line 222: `LLTEContents` ctor now takes an argument for the number of TEs to expect.  This allows it to only allocate as much memory as it actually needs (e.g. 6 for a box) instead of the default (45, which is how many the avatar uses).

indra/llprimitive/tests/llprimitive_test.cpp
```diff
@@ -266,6 +231,170 @@ namespace tut
...
+            ensure_equals(te_A->getOffsetS(), te_B->getOffsetS());
+            ensure_equals(te_A->getOffsetT(), te_B->getOffsetT());
+            ensure_equals(te_A->getRotation(), te_B->getRotation());
+
+            ensure_equals(te_A->getBumpShinyFullbright(), te_B->getBumpShinyFullbright());
+            ensure_equals(te_A->getMediaFlags(), te_B->getMediaFlags());
+            ensure_equals(te_A->getGlow(), te_B->getGlow());
+            ensure_equals(te_A->getAlphaGamma(), te_B->getAlphaGamma());
+        }
+    }
```
AndrewMeadows, line 397: I'm adding this unit-test for the **TextureEntry** pack/unpack logic.

indra/llprimitive/tests/lltextureentry_stub.cpp
```diff
@@ -0,0 +1,316 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Linden Research, Inc., 945 Battery Street, San Francisco, CA  94111  USA
+ * $/LicenseInfo$
+ */
+
+#pragma once
+
```
AndrewMeadows, line 27: In order for the unit-test to work we need a lot of **functional** stubbery for `LLTextureEntry`.

indra/newview/llpanelface.cpp
```diff
@@ -1325,14 +1324,6 @@ void LLPanelFace::updateUI(bool force_set_values /*false*/)
...
         }
 
-        // Alpha gamma.  This should always be enabled, no matter if the object is mod or not.
-        {
-            U8   alpha_gamma           = 100;
-            bool identical_alpha_gamma = false;
-            LLSelectedTE::getAlphaGamma(alpha_gamma, identical_alpha_gamma);
-            mComboAlphaGamma->getSelectionInterface()->selectByValue(alpha_gamma);
-            mComboAlphaGamma->setEnabled(true);
-        }
```
AndrewMeadows, line 1335: This moved down below so the UI could enabled even when permissions are **no-mod**.

indra/newview/llviewerobject.cpp
```diff
@@ -3117,8 +3117,8 @@ struct LLFilenameAndTask
         LL_DEBUGS() << "Destroying LLFilenameAndTask: " << sCount << LL_ENDL;
     }
 private:
-    LLFilenameAndTask(const LLFilenameAndTask& rhs);
-    const LLFilenameAndTask& operator=(const LLFilenameAndTask& rhs) const;
+    LLFilenameAndTask(const LLFilenameAndTask& rhs) = delete;
+    const LLFilenameAndTask& operator=(const LLFilenameAndTask& rhs) = delete;
```
AndrewMeadows, line 3121: VisualStudio was warning about these unimplemented methods.  It looks like someone wanted to restrict the usage of these normally-default methods so I did it correctly using the `delete `keyword and the compiler did not complain.

indra/newview/llvoavatar.cpp
```diff
@@ -243,7 +243,8 @@ struct LLTextureMaskData
 
 struct LLAppearanceMessageContents: public LLRefCount
 {
-    LLAppearanceMessageContents():
+    LLAppearanceMessageContents(U8 num_tes):
+        mTEContents(num_tes),
```
AndrewMeadows, line 247: `LLTEContents `now requires "num_textures" argument in its ctor, so `LLAppearanceMessage `(which has a `LLTEContents `data member) also requires the same argument so it can pass it over.

indra/llprimitive/llprimitive.cpp
```diff
@@ -1507,111 +1502,35 @@ S32 LLPrimitive::unpackTEMessage(LLDataPacker &dp)
...
-        // Note:  This is an optimization to send common colors (1.f, 1.f, 1.f, 1.f)
-        // as all zeros.  However, the subtraction and addition must be done in unsigned
-        // byte space, not in float space, otherwise off-by-one errors occur. JC
-        color.mV[VRED]      = F32(255 - colors[i].mV[VRED])   / 255.f;
-        color.mV[VGREEN]    = F32(255 - colors[i].mV[VGREEN]) / 255.f;
-        color.mV[VBLUE]     = F32(255 - colors[i].mV[VBLUE])  / 255.f;
-        color.mV[VALPHA]    = F32(255 - colors[i].mV[VALPHA]) / 255.f;
-
-        retval |= setTEColor(i, color);
-    }
```
AndrewMeadows, line 1612: The motivation for the DRY pass was to eliminate duplicate unpacking logic.  We now have a single method where this is done.
```
// static
S32 LLPrimitive::parseTEMessage(U8* packed_buffer, U32 data_size, LLTEContents& tec)
```

indra/llprimitive/llprimitive.h
```diff
@@ -498,12 +512,15 @@ class LLPrimitive : public LLXform
 
     void copyTEs(const LLPrimitive *primitive);
     S32 packTEField(U8 *cur_ptr, U8 *data_ptr, U8 data_size, U8 last_face_index, EMsgVariableType type) const;
+
+    S32  packTEMessageBuffer(U8* packed_buffer) const;
     bool packTEMessage(LLMessageSystem *mesgsys) const;
-    bool packTEMessage(LLDataPacker &dp) const;
```
AndrewMeadows, line 502: The `packTEMessage()` that expected an `LLDataPacker&` argument has been removed: it was never called anywhere.

indra/llprimitive/llprimitive.cpp
Geenz: Disclaimer: popping in as I drive across the country for my move.

I see you've largely taken on de-crustifying things - though I don't feel like I'm the best person to really review that given my inexperience with the message system (for now - probably gonna be taking a few stabs with the upcoming VR work).  It seems like this comment is obsolete?

indra/newview/llselectmgr.cpp
Geenz: Noted!  Thank you!

indra/newview/llselectmgr.cpp
```diff
@@ -1736,6 +1736,18 @@ struct LLSelectMgrSendFunctor : public LLSelectedObjectFunctor
...
+{
+    virtual bool apply(LLViewerObject* object)
+    {
+        if (object->permYouOwner())
+        {
+            LLSelectMgr::packAlphaGammaBypass(object);
+        }
+        return true;
+    }
+};
```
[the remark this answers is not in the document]
Geenz: To be clear - we're sticking with the new message, yes?

Geenz:
This generally looks good to me - I'm willing to accept the de-crustifying of LLPrimitive's TE packing at face value unless someone would like to take up that review.  It generally looks good to me on that front.  If there's no objections, I'll merge into `legacy-alpha-params`

Geenz (approved):

indra/newview/llselectmgr.cpp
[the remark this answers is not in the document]
AndrewMeadows: Yes, we're sticking with the new message.

indra/llprimitive/llprimitive.cpp
AndrewMeadows: Yes, this comment is obsolete.  I will resolve the conversation.

indra/llprimitive/llprimitive.cpp
```diff
@@ -1284,8 +1287,12 @@ bool LLPrimitive::packTEMessage(LLMessageSystem *mesgsys) const
         *cur_ptr++ = 0;
         cur_ptr += packTEField(cur_ptr, (U8 *) material_data, 16, last_face_index, MVT_LLUUID);
         *cur_ptr++ = 0;
-        *cur_ptr++ = 0x01;
+        // end of required properties
+
+        // pack extra properties
+        *cur_ptr++ = EXTRA_PROPERTY_ALPHA_GAMMA; // indicator alpha_gamma is present
         cur_ptr += packTEField(cur_ptr, (U8 *) alpha_gamma, 1, last_face_index, MVT_U8);
+        *cur_ptr++ = 0; // always null-terminate each propery
```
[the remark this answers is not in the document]
AndrewMeadows: This comment is also obsolete.  My understanding has evolved: the sender does NOT null-terminate the last **TEField**, however the receiving side will explicitly add a null-termination to make it easier to unpack.

indra/llprimitive/llprimitive.cpp
```diff
@@ -1371,8 +1379,12 @@ bool LLPrimitive::packTEMessage(LLDataPacker &dp) const
...
```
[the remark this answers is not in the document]
AndrewMeadows: The **DRY**ing is complete.  Well... things are **DRY**er.  This conversation can be resolved.

PR #2544, opened 2024-09-10
Profile guided optimizations

RunitaiLinden:
* Eliminate superfluous copy of texture coordinate array
* Prevent redundant calls to LLVertexBuffer::unmapBuffer
* Increase bytes sent per call for glBufferSubData and glTexSubImage2D
* Remove CPU downscaling of textures when desired discard is lowered between texture decode and GPU texture creation
* Reduce time spent in analyzeAlpha by doing a 1/4 res sample instead of a 1/2 res sample


RyeMutt (approved):
LGTM

cosmic-linden (approved):

brad-linden (approved):
👍

PR #2547, opened 2024-09-11
Fix names in group chat sometimes showing as "Loading..." and never resolve

Ansariel:
Finally available in the official viewer thanks to a special request from @RyeMutt : Fix the 10+ year old bug where names in group chat sometimes/often get stuck as "Loading..." and never resolve (BUG-3829 / MAINT-3118). Also see https://jira.firestormviewer.org/browse/FIRE-11330 for a description of the bug. 😄


marchcat (approved):
Thank you!

PR #2548, opened 2024-09-11
Make Develop->Render Tests->Frame Profile dump JSON to a file too (#2412)

### LINKED ISSUE #2412: Capture develop->render tests->frame profile output as CSV
Currently frame profiles are dumped to the log file in a fairly verbose form - various numbers listed for each shader. It would be more useful to have this data in a csv file to support analysis.

Let's leave the log file output, but also dump each frame profile to a new csv file in the log dir.

TBD: how should these be named and how many should we keep? We will want to support the common use case of running the same test on multiple viewers and then comparing the values, so we don't want every dump file to have the same name and overwrite the previous one.

## Test Plan

Use any existing tests for frame profile, and verify that the results are also dumped to a csv file.

nat-goodspeed:
Make `LLGLSLShader::finishProfile()` accept a string pathname instead of a
bool and, in addition to logging statistics to the viewer log, output
statistics to that file as JSON. The calls that used to pass
`emit_report=false` now pass `report_name=std::string()`.

Make llviewerdisplay.cpp's `display()` function synthesize a profile filename
in the viewer's logs directory, and pass that filename to
`LLGLSLShader::finishProfile()`.

Add a Lua script to be run during a session that TPs to a known location, looks at a particular point and captures a frame profile.

Add a bash script + Lua script to be run on each of several viewers that automatically logs into a known location, looks at a particular point and captures a frame profile.

Add a Python script to pretty-print a frame profile JSON file. The file written by the viewer is a dense single line of JSON.

Add a Python script to convert a frame profile JSON file to CSV.

Fixes #2412.

### DIFF
--- a/indra/llcommon/hexdump.h
+++ b/indra/llcommon/hexdump.h
@@ -25,7 +25,7 @@ namespace LL
 class hexdump
 {
 public:
-    hexdump(const std::string_view& data):
+    hexdump(std::string_view data):
         hexdump(data.data(), data.length())
     {}
 
@@ -66,7 +66,7 @@ private:
 class hexmix
 {
 public:
-    hexmix(const std::string_view& data):
+    hexmix(std::string_view data):
         mData(data)
     {}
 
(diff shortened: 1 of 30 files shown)


nat-goodspeed:
[Testable build](https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%231b7fdac) of this branch

RunitaiLinden (approved):

vir-linden (approved):
Looks good! Curious how we wound up needing so much work on file names and string conversions

PR #2552, opened 2024-09-12
#2542 Buoyant Nametags

LLGuru:


marchcat (approved):

akleshchev (changes requested):
This code was originallt added in 'SL-19887 Water cuts avatar's name bubble' https://github.com/secondlife/viewer/commit/2a87b64d26fba502e6b8514084d91b1f79965fd1
Purpose seems to be to not cut bubble in two due to visibility issues which does not conflic with current issue of being always above water.

Intead of simply removing it, code should be updated to snap to either be under water or above water instead of strictly above water. Or visibility issues should be fixed in some other way.

marchcat:
> snap to either be under water or above water instead of strictly above water.

This makes sense,

indra/newview/llvoavatar.cpp
```diff
@@ -3696,21 +3696,22 @@ LLVector3 LLVOAvatar::idleCalcNameTagPosition(const LLVector3 &root_pos_last)
...
+    if (mNameText)
     {
-        if (LLViewerCamera::getInstance()->getOrigin()[VZ] >= water_height)
-        {
-            name_position[VZ] = water_height;
-        }
-        else if (mNameText) // both camera and HUD are below watermark
+        F32 water_height = getRegion()->getWaterHeight();
+        static const F32 WATER_HEIGHT_ABOVE_DELTA = 0.25;
+        if (name_position[VZ] < water_height + WATER_HEIGHT_ABOVE_DELTA)
```
akleshchev, line 3704: I'm a bit confused about this condition. Shuldn't it be something like this?

if (name_position_bottom[VZ] < water_height && name_position_bottom[VZ] > water_height )

akleshchev (approved):

PR #2556, opened 2024-09-12
Fix warning spam from missing call to unmapBuffer introduced in #2544

RyeMutt:
Fixes log spam from `Missing call to unmapBuffer or flushBuffers` introduced from changes in #2544 by changing to LL_WARNS_ONCE

### DIFF
--- a/indra/llrender/llvertexbuffer.cpp
+++ b/indra/llrender/llvertexbuffer.cpp
@@ -1600,7 +1600,7 @@ void LLVertexBuffer::setBuffer()
 
     if (mMapped)
     {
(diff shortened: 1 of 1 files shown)


RunitaiLinden (changes requested):
Don't do this any more.  We get a performance boost by flushing all the buffers at once instead of doing some work, flushing a buffer, doing some work, flushing a buffer...

There's new machinery in LLVertexBuffer that queues up the buffers to be flushed and then flushes them all at once when you call LLVertexBuffer::flushBuffers.  If you want to suppress the warning spam while I track down the cases that aren't quite aligned with the new approach, make it an LL_WARNS_ONCE.

RunitaiLinden (approved):

PR #2557, opened 2024-09-12
OS X - Fix black screen and Tracy build

RunitaiLinden:


indra/llrender/llvertexbuffer.cpp
```diff
@@ -1301,6 +1301,9 @@ U8* LLVertexBuffer::mapIndexBuffer(U32 index, S32 count)
 void LLVertexBuffer::flush_vbo(GLenum target, U32 start, U32 end, void* data, U8* dst)
 {
 #if LL_DARWIN
+    // on OS X, flush_vbo doesn't actually write to the GL buffer, so be sure to call
+    // _mapBuffer to tag the buffer for flushing to GL
+    _mapBuffer();
```
brad-linden, line 1306: is it intentional that this is outside the profile zone?

brad-linden (approved):

indra/llrender/llvertexbuffer.cpp
RunitaiLinden: No, but it's inconsequential

Issue #2553, opened 2024-09-12
[OSX] The 'CoreFoundation!__NSArrayM_new.cold.1' crash occurs after navigating to Help > About Second Life


sl-service-account:
**BugSplat** [Key Crash 94456](https://app.bugsplat.com/v2/keycrash?stackKeyId=94456&database=SecondLife_Viewer_2018)

**Sample Version:**
7.1.11.10800977053 (7.1.11.10800977053)

**Sample Callstack:**
<table><tr><td>CoreFoundation!__NSArrayM_new.cold.1</td><td></td></tr><tr><td>CoreFoundation!__NSArrayM_new</td><td></td></tr><tr><td>CoreFoundation!+[NSMutableArray arrayWithCapacity:]</td><td></td></tr><tr><td>BugsplatMac!0x107201000</td><td></td></tr><tr><td>BugsplatMac!0x107201000</td><td></td></tr><tr><td>BugsplatMac!0x107201000</td><td></td></tr><tr><td>BugsplatMac!0x107201000</td><td></td></tr><tr><td>BugsplatMac!0x107201000</td><td></td></tr><tr><td>BugsplatMac!0x107201000</td><td></td></tr><tr><td>libc++abi.dylib!_ZSt11__terminatePFvvE</td><td></td></tr><tr><td>libc++abi.dylib!__cxa_rethrow</td><td></td></tr><tr><td>libobjc.A.dylib!objc_exception_rethrow</td><td></td></tr><tr><td>AppKit!NSApplication _handleEvent</td><td></td></tr><tr><td>AppKit!NSApplication run</td><td></td></tr><tr><td>AppKit!NSApplicationMain</td><td></td></tr><tr><td>???!0x0</td><td></td></tr></table>

**Steps to reproduce:**
1. Download and login to the [Second Life Test 7.1.11.10800977053 viewer](https://github.com/secondlife/viewer/actions/runs/10800977053).
2. Go to Help > About Second Life.

**Actual result:** Crash occurs after trying to open About Second Life.

**Expected result:** No crash should occur.

**Additional info:**
The issue is reproduced only on OSX.
The issue is not reproduced on current production (7.1.9.10515727195).
The issue was found during https://github.com/secondlife/iqa/issues/335 testing.

maestrolinden:
[...]

PR #2561, opened 2024-09-12
#2553 about sl crash locale init

brad-linden:
Attempted fix for mac crash when showing About Second Life.
secondlife/viewer#2553

The app is crashing in the debugger with an unhandled exception issued from LLStringUtil::formatNumber() when attempting to imbue the locale.  This makes no sense since the call to construct the locale is directly wrapped in a try/catch block that should be handling it.  The bugsplat call stack looks slightly different and fails in some Cocoa array allocation handling when terminating due to the uncaught exception.

Nonetheless this whole situation can be avoided by not attempting to set invalid locales such as `MissingString(macosLocale)`.  I have not yet discovered why we are attempting to look up locale names in strings.xml which have not been there for some time.

### DIFF
--- a/indra/llcommon/llerror.cpp
+++ b/indra/llcommon/llerror.cpp
@@ -110,7 +110,7 @@ namespace {
         virtual void recordMessage(LLError::ELevel level,
                                     const std::string& message) override
         {
-            LL_PROFILE_ZONE_SCOPED_CATEGORY_LOGGING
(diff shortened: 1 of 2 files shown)


marchcat (approved):

brad-linden:
here is the line that is throwing the exception:
https://github.com/secondlife/viewer/blob/develop/indra/llcommon/llstring.cpp#L1436

PR #2562, opened 2024-09-13
Clean up Windows build

Ansariel:
* APR_DECLARE_STATIC and APU_DECLARE_STATIC gets already defined in APR.cmake
* Move both _CRT_SECURE_NO_WARNINGS and _WINSOCK_DEPRECATED_NO_WARNINGS definitions to 00-Common.cmake
* Always define WIN32_LEAN_AND_MEAN and include subset of Windows API by default
* Remove llwin32headerslean.h and remove unnecessary WIN32_LEAN_AND_MEAN definition handling in llwin32headers.h
* Clean up includes of Windows API headers
* Get rid of workaround to link against IPHLPAPI.lib in lluuid.cpp - this seems to have been an issue in the past that has been fixed


marchcat (approved):
TYVM!

marchcat:
You were faster than me at fixing that again 😸 
Let's check how it builds on github now.

PR #2563, opened 2024-09-13
Fix macOS locale not being set due to missing locale strings

RyeMutt:
This fixes the macOS locale not being set correctly due to OS string changes that were not propagated to UI xml


akleshchev (approved):

brad-linden (commented):
👍 

aha!  that's why I couldn't find it, thanks!  for future work we should also probably examine upgrading the english locale settings to en_US.UTF-8 across the board

PR #2569, opened 2024-09-13
send AgentUpdate ASAP when control bits change

AndrewMeadows:
This is a port of changes in the [GameControl ](https://wiki.secondlife.com/wiki/LSL_Game_Control_Beta) viewer relating to sending **AgentUpdate** messages ASAP whenever the avatar's movement control bits change.  This affects the timing and rate of [LSL **control**](https://wiki.secondlife.com/wiki/Control) events for in-world scripts (e.g. vehicles).

### DIFF
--- a/indra/newview/llagent.cpp
+++ b/indra/newview/llagent.cpp
@@ -439,8 +439,6 @@ LLAgent::LLAgent() :
     mIsDoNotDisturb(false),
 
     mControlFlags(0x00000000),
-    mbFlagsDirty(false),
-    mbFlagsNeedReset(false),
 
     mAutoPilot(false),
     mAutoPilotFlyOnStop(false),
@@ -936,8 +934,6 @@ void LLAgent::setFlying(bool fly, bool fail_sound)
 
     // Update Movement Controls according to Fly mode
     LLFloaterMove::setFlyingMode(fly);
-
-    mbFlagsDirty = true;
 }
 
 
@@ -1068,7 +1064,6 @@ void LLAgent::setRegion(LLViewerRegion *regionp)
             {
                 regionp->setCapabilitiesReceivedCallback(LLAgent::capabilityReceivedCallback);
             }
-
         }
         else
         {
@@ -1556,7 +1551,6 @@ U32 LLAgent::getControlFlags()
 void LLAgent::setControlFlags(U32 mask)
 {
     mControlFlags |= mask;
-    mbFlagsDirty = true;
 }
 
 
@@ -1565,28 +1559,7 @@ void LLAgent::setControlFlags(U32 mask)
 //-----------------------------------------------------------------------------
 void LLAgent::clearControlFlags(U32 mask)
 {
-    U32 old_flags = mControlFlags;
     mControlFlags &= ~mask;
-    if (old_flags != mControlFlags)
-    {
-        mbFlagsDirty = true;
-    }
-}
-
-//-----------------------------------------------------------------------------
-// controlFlagsDirty()
-//-----------------------------------------------------------------------------
-bool LLAgent::controlFlagsDirty() const
-{
-    return mbFlagsDirty;
-}
-
-//-----------------------------------------------------------------------------
-// enableControlFlagReset()
-//-----------------------------------------------------------------------------
-void LLAgent::enableControlFlagReset()
-{
-    mbFlagsNeedReset = true;
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


indra/newview/llagent.cpp
```diff
@@ -439,8 +439,6 @@ LLAgent::LLAgent() :
     mIsDoNotDisturb(false),
 
     mControlFlags(0x00000000),
-    mbFlagsDirty(false),
-    mbFlagsNeedReset(false),
```
AndrewMeadows, line 443: This "FlagsDirty" and "FlagsNeedReset" stuff is not helpful, so it is being removed.

indra/newview/llagent.cpp
```diff
@@ -2085,11 +2053,19 @@ void LLAgent::propagate(const F32 dt)
...
+    }
 
-    const F32 PITCH_RATE = 90.f * DEG_TO_RAD;           // radians per second
-    pitch(PITCH_RATE * gAgentCamera.getPitchKey() * dt);
+    constexpr F32 PITCH_RATE = 90.f * DEG_TO_RAD;            // radians per second
+    angle = PITCH_RATE * gAgentCamera.getPitchKey() * dt;
+    if (fabs(angle) > 0.0f)
+    {
+        pitch(angle);
+    }
```
AndrewMeadows, line 2068: Minor unrelated (to the bug being fixed) cleanup: don't bother applying yaw/pitch when the angle change is zero.

indra/newview/llappviewer.cpp
```diff
@@ -4711,30 +4709,13 @@ void LLAppViewer::idle()
...
-        // just that it will be checked.
-        if (force_send || timed_out)
-        {
-            LL_PROFILE_ZONE_SCOPED_CATEGORY_NETWORK;
-            // Send avatar and camera info
-            mLastAgentControlFlags = gAgent.getControlFlags();
-            mLastAgentForceUpdate = force_send ? 0 : agent_force_update_time;
-            send_agent_update(force_send);
-            agent_update_timer.reset();
-        }
```
AndrewMeadows, line 4737: All of the logic (above) about whether an update needs to be sent has been moved into `send_agent_update()`.

indra/newview/llappviewer.h
```diff
@@ -341,10 +339,6 @@ class LLAppViewer : public LLApp
     bool mIsFirstRun;
 };
 
-// consts from viewer.h
-const S32 AGENT_UPDATES_PER_SECOND  = 125; // Value derived experimentally to avoid Input Delays with latest PBR-Capable Viewers when viewer FPS is highly volatile.
-const S32 AGENT_FORCE_UPDATES_PER_SECOND  = 1;
-
```
AndrewMeadows, line 347: These constants have been moved to where they are actually used.

indra/newview/llviewermessage.cpp
```diff
@@ -3139,87 +3139,56 @@ void process_crossed_region(LLMessageSystem* msg, void**)
...
 
 
-
-// Sends avatar and camera information to simulator.
-// Sent roughly once per frame, or 20 times per second, whichever is less often
-
-const F32 THRESHOLD_HEAD_ROT_QDOT = 0.9997f;    // ~= 2.5 degrees -- if its less than this we need to update head_rot
-const F32 MAX_HEAD_ROT_QDOT = 0.99999f;         // ~= 0.5 degrees -- if its greater than this then no need to update head_rot
-                                                // between these values we delay the updates (but no more than one second)
-
```
AndrewMeadows, line 3149: The constants above have been moved closer to where they are actually used.

indra/newview/llviewermessage.cpp
```diff
@@ -3228,7 +3197,22 @@ void send_agent_update(bool force_send, bool send_reliable)
                             AGENT_CONTROL_ML_LBUTTON_UP ;
     }
 
-    control_flag_change = last_control_flags ^ control_flags;
+    // any change in control_flags should be sent ASAP, so we fold that into force_send
+    force_send = force_send || (control_flags != last_control_flags);
```
AndrewMeadows, line 3201: This is the main fix right here: when `control_flags` have changed we send an update ASAP regardless of send rate limits.  The rest of the changes are cleanup and reorganizing how all of the other conditions are checked.

indra/newview/llviewermessage.cpp
```diff
@@ -3240,159 +3224,140 @@ void send_agent_update(bool force_send, bool send_reliable)
...
-                LL_INFOS("Messaging") << "AgentUpdate left button up" << LL_ENDL;
+                // close enough
+                return;
+            }
+            else if (head_rot_qdot < MIN_HEAD_ROT_QDOT)
+            {
+                // way off
+                send_update = true;
+                break;
             }
```
brad-linden, line 3288: am I reading this wrong?  these conditions appear backwards

AndrewMeadows: For two quaternions that are very close to each other their dot-product will be very close to 1.0.  So:

(1) When the `dot_product` is above the `MAX `threshold then they are close --> no need to send an update at all.

(2) When the `dot_product `is less than the `MIN `threshold then definitely need to send an update.

(3) When the `dot_product` is between the two then maybe we need to send an update (to help mouselook aim) so we compute a "dynamic" threshold that moved from `MIN `to `MAX `over `THRESHOLD_GROWTH_PERIOD`.

I added the `fabs()` operation on the dot product because the quaternion hypersphere has a discrepancy of 2 when mapped to rotation space, so that handles the rare case when the two quaternions happen to be on opposite hemispheres.

brad-linden (approved):

indra/newview/llviewermessage.cpp
brad-linden: aha, ok thanks for the explanation

PR #2571, opened 2024-09-13
Restore LUA debug consoles in viewer menu and tie visibility to feature flag

Ansariel:
@nat-goodspeed Instead of removing the LUA consoles from to viewer menu so they might get forgotten and lost forever, I added them back and instead tied their visibility to the corresponding feature flag.

### DIFF
--- a/indra/newview/skins/default/xui/en/menu_viewer.xml
+++ b/indra/newview/skins/default/xui/en/menu_viewer.xml
@@ -2583,6 +2583,37 @@ function="World.EnvPreset"
                parameter="scene monitor" />
             </menu_item_check>
             <menu_item_separator/>
+            <menu_item_check
+               label="LUA Debug Console"
+               name="LUA Debug Console">
+              <menu_item_check.on_check
+               function="Floater.Visible"
+               parameter="lua_debug" />
+              <menu_item_check.on_click
+               function="Floater.Toggle"
+               parameter="lua_debug" />
+              <menu_item_check.on_visible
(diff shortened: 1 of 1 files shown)


nat-goodspeed (commented):
Thank you, I didn't know about that `"CheckControl"` feature.

But did you see [the menus.lua autorun script](https://github.com/secondlife/viewer/blob/release/luau-scripting/indra/newview/scripts/lua/auto/menus.lua) that injects the Lua consoles into the menu tree when Lua is in fact enabled?

My intention is to minimize the number of places that have to check the `LuaFeature` flag. Presumably at some point we'll remove that setting entirely. Once the setting is deleted, would `"CheckControl" "LuaFeature"` prevent those menu items from being displayed?

Ansariel:
In fact I did not actually see that script. However, it uses a fixed position where the menu items get inserted. What if somebody changes the menu and the position needs to be changed, but whoever changes it is not aware of this script? And aren't you running risk of creating an obtrusive way of inserting things into the menu? I have a feeling in the future there might be questions where the menu items are magically coming from since they are not in the XUI file.

Also: Does injecting the menu items via script not make localizations impossible?

IMHO it's easier to tie the menu items directly to the setting. If the setting gets removed in the future, a simple search over all files should reveal all locations.

nat-goodspeed (approved):
I find your argument about localization compelling.

nat-goodspeed:
https://github.com/secondlife/viewer/pull/2574

PR #2573, opened 2024-09-13
Add script to compare a Frame Profile JSON stats file vs. a baseline file.

nat-goodspeed:
Enrich stats files with context to provide more meaningful information to a human reader.


nat-goodspeed:
Dave, I'm curious if this new `profile_cmp.py` script addresses what you have in mind.

maxim-productengine (approved):

PR #2585, opened 2024-09-17
Xcode16 build fix

marchcat:


Ansariel:
I see you are trying to enable C++20 on macOS at least. Are you going to fix the broken hash generation in llgltfmaterial and llmaterial we discovered in #2330?

marchcat:
That should be the next step.
For now, it builds successfully (with the test temporarily disabled) on my mac with xcode 16.0 and fails on github, so it's a bit harder to fix with all this commit-juggling. I'm going to return to fixing the test after some success with the actual build.

Ansariel:
Ok, but you should check `LLMaterial::getHash()` as well - it does the same kind of hash generation.

indra/newview/llviewerwindow.cpp
```diff
@@ -1331,7 +1331,7 @@ LLWindowCallbacks::DragNDropResult LLViewerWindow::handleDragNDrop( LLWindow *wi
                                 // Check the whitelist, if there's media (otherwise just show it)
                                 if (te->getMediaData() == NULL || te->getMediaData()->checkCandidateUrl(url))
                                 {
-                                    if ( obj != mDragHoveredObject)
+                                    if (obj != mDragHoveredObject.get())
```
nat-goodspeed, line 1334: Let's pull in Ansariel's free functions to compare `(T* == LLPointer<T>)` and `!=`.

nat-goodspeed (approved):

indra/newview/llviewerwindow.cpp
Ansariel: They are already in the `develop` branch (d91d39fa0f7f4f204d6fb7ff66b9817e498dbd61). The problem here is that obj is a LLVOVolume pointer and mDragHoveredObject is an LLPointer<LLViewerObject>, so the overloaded operator doesn't work.

nat-goodspeed: If we're comparing pointers to different types in the same class hierarchy, shouldn't we formally be downcasting one of them to match the other? We don't use a lot of multiple inheritance, but a pointer to object A's base class doesn't necessarily have the same bit pattern as a pointer to A.

Ansariel: That would probably be a smart thing to do.

nat-goodspeed: I should have realized [the language takes care of that](https://en.cppreference.com/w/cpp/language/implicit_conversion#Pointer_conversions).

That being the case, I'll try tweaking the `(T* == LLPointer<T>)` comparisons to accept a raw pointer to arbitrary type, letting the compiler produce an error if the two pointer types can't be compared.

nat-goodspeed: [#2681](https://github.com/secondlife/viewer/pull/2681)

Ansariel:
Excuse me while I'm capturing this PR for an unrelated, but kinda similar question: Is the Linux runner using Ubuntu 22.04? And if so, can we get a Ubuntu 24.04 / GCC13 compatibility pass? (Buggy) GCC13 gets really angry about some templating stuff that was added with the Lua feature merge.

nat-goodspeed:
Sigh, Windows dislikes the lerp() change. On it.

marchcat:
It still fails. Probably just renaming our `lerp()` implementation to something like `ll_lerp()` will work better, like in 1c30ce503bce36bea360a86e52351d3f243e2e86.

nat-goodspeed:
Yes, I see. In this case it's a tinygltf header that #includes <math.h>, so we'd have to patch the autobuild package to correct that. I'll pursue the tactic of renaming our lerp() function, but I'll change the references only where MSVC complains: where it does not, I see no reason to use ours instead of std::lerp().

.github/workflows/build.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
     needs: setup
     strategy:
       matrix:
-        runner: [windows-large, macos-12-xl]
+        runner: [windows-large, macos-14-xlarge]
         configuration: ${{ fromJSON(needs.setup.outputs.configurations) }}
         include:
         - runner: linux-large
```
marchcat, line 52:
```suggestion
        - runner: linux-large
          continue-on-error: true 
```
While we're at it, let's ensure failed Linux builds don't cancel others.

marchcat:
> Is the Linux runner using Ubuntu 22.04? And if so, can we get a Ubuntu 24.04 / GCC13 compatibility pass?

linux-large is currently Ubuntu 22.04.5. I tried to build on ubuntu-24.04 runner and got the same set of errors, but slower.

brad-linden:
> > Is the Linux runner using Ubuntu 22.04? And if so, can we get a Ubuntu 24.04 / GCC13 compatibility pass?
> 
> linux-large is currently Ubuntu 22.04.5. I tried to build on ubuntu-24.04 runner and got the same set of errors, but slower.
we will certainly want to keep our distributed builds on 22.04 for some time.  we can do work for compatibility with newer compilers, but moving to a newer glibc based system will limit our distribution compatibility even further

.github/workflows/build.yaml
```diff
@@ -46,10 +46,11 @@ jobs:
...
+          continue-on-error: true
```
marchcat, line 53:
```suggestion
          continue-on-error: true
```

nat-goodspeed:
This PR actually does turn on C++20. Once it succeeds, do we need further discussion with anyone before merging to develop?

Ansariel:
> This PR actually does turn on C++20. Once it succeeds, do we need further discussion with anyone before merging to develop?

What about the hash calculation for BP and GLTF materials? Is that fixed or does it need more attention?

nat-goodspeed:
Dave was running some head-to-head performance tests with various hash tactics; I don't know where he committed the results of his work.

Ansariel:
But does the test succeed now? Previously xcode seem to have changed something to the memory alignment of the said material classes, causing the test to fail.

vir-linden (approved):

Ansariel:
> Sigh, Windows dislikes the lerp() change. On it.

Looking at the cased that got changed to flerp, could it be VS got mad because in all those cases you are mixing parameter types (int/float)?

Ansariel:
FYI: C++23 builds fine with VS 🫡

nat-goodspeed:
A new [branch tip build](https://github.com/secondlife/viewer/actions/runs/11076889958) for evaluation

nat-goodspeed:
build = https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%2371e89d6

akleshchev:
```
  Unit test group_started name=LLPrimitive
   INFO #TestRunner# test/test.cpp(274) LLTestCallback::test_completed : [LLPrimitive, 4: Test Static llprimitive constructor and initer.] ok
  [LLPrimitive, 5: unknown] exception: Windows_SEH_exception: 'test threw c0000005 (Access Violation)'
  Unit test group_completed name=LLPrimitive
  	Total Tests:	5
  	Passed Tests:	4
  *********************************
  Failed Tests:	1
  Please report or fix the problem.
  *********************************
```

marchcat:
Thanks, that should be fixed now.

nat-goodspeed:
THANK YOU! I didn't know how to debug that, given the peculiar way we invoke these tests as a side effect of library creation.

marchcat:
It basically took me to check it line by line, and then I saw `NULL` being passed as a string argument.

nat-goodspeed:
Second Life Release build with yesterday's merge from develop:

build = https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%230f7aadf

nat-goodspeed:
build = https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%238796361

nat-goodspeed:
Atlas says:

>Thumbs up for "light testing". Nothing exploded during testing with this build on a couple different machines.
The following 2 issues do pop up but these appear to be Develop branch specific
https://github.com/secondlife/viewer/issues/2899
https://github.com/secondlife/viewer/issues/2870

marchcat:
#2899 should be fixed in #2929.

nat-goodspeed:
So we should keep an eye on #2929 and, once it's merged, update this branch again from develop? -- unless this PR merges first, of course.

marchcat:
Yes, I'll update this PR once ready.

marchcat:
#2870 has a workaround (#2890), should we get it in here as well?

marchcat:
Yay! Thank you, @nat-goodspeed!

Issue #2408, opened 2024-08-23
The long covenant with emojis significantly slows down some operations in the viewer


DmitrySProductengine:
### Environment
<details><summary>Environment</summary>
<p>

Second Life Test 7.1.9.10164740217 (64bit)
Release Notes

You are at 105.8, 109.8, 22.2 in Product Engine 3 located at simhost-0cd70bedfc4667751.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine%203/106/110/22
(global coordinates 336746.0, 307054.0, 22.2)
Second Life Server 2024-07-22.10048683488
Release Notes

CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (2600 MHz)
Memory: 16384 MB
OS Version: macOS 14.5.0 Darwin 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:09:52 PDT 2024; root:xnu-10063.121.3~5/RELEASE_X86_64 x86_64
Graphics Card Vendor: ATI Technologies Inc.
Graphics Card: AMD Radeon Pro 555X OpenGL Engine

OpenGL Version: 4.1 ATI-5.5.17

Window size: 1058x836
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.25
Render quality: 2
Texture memory: 4096MB
Disk cache: Max size 1638.4 MB (19.0% used)
HiDPI display mode: 

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202310131309
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/834 (0.0%)
August 23 2024 07:43:07

</p>
</details> 

### Reproduction steps

1. Log in to the Viewer as an estate owner or estate manager.
2. Create any notecard with long text and emoji (the example file is attached below - its contents can be copied to the notecard).
3. Drag the notecard into the Covenant field in the Region/Estate floater.
4. Observe the viewer's behavior during various operations (opening tabs in region/estate or about land floaters; general periodic freezing of the viewer).

[...]

PR #2592, opened 2024-09-18
#2408 The long covenant with emojis significantly slows down FPS

LLGuru:
When the covenant is loaded, it is being parsed 4 times
This operation takes much time for splitting the text on segments
Each line of the text is at least a segment, and each line break is a segment

The provided example contains almost 6000 lines of text with emoji
On my PC it takes 20+ seconds to show the covenant (4 times of 5+ sec.)

The proposed change decreases this delay 4 times, because the parsing is being done only once,
and then the contents of the 1st text editor is being copied to other text editors as is

### DIFF
--- a/indra/llui/lltextbase.cpp
+++ b/indra/llui/lltextbase.cpp
@@ -2368,6 +2368,34 @@ S32 LLTextBase::removeFirstLine()
     return 0;
 }
 
+// virtual
+void LLTextBase::copyContents(const LLTextBase* source)
+{
+    llassert(source);
+    if (!source)
+        return;
+
+    beforeValueChange();
+    deselect();
+
+    mSegments.clear();
+    for (const LLTextSegmentPtr& segp : source->mSegments)
+    {
+        mSegments.emplace(segp->clone(*this));
+    }
+
+    mLineInfoList.clear();
+    for (const line_info& li : mLineInfoList)
+    {
+        mLineInfoList.push_back(line_info(li));
+    }
+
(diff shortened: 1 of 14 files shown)


indra/llui/lltextbase.cpp
```diff
@@ -3233,6 +3261,24 @@ boost::signals2::connection LLTextBase::setIsObjectBlockedCallback(const is_bloc
...
+    LLStyle* style = new LLStyle(params);
+
+    // Take some params from source
+    style->setLinkHREF(source->getLinkHREF());
+    if (source->isImage())
+    {
+        style->setImage(source->getImage()->getName());
+    }
+
+    return style;
```
marchcat, line 3278: This whole idea of using `new` like this:
```
LLSomething* methodName()
{
    return new LLSomething();
}
```
without a corresponding `delete` looks like a huge potential for a memory leak. It would be much safer to use `LLPointer` instead.

LLGuru:
> Why does a covenant even need to contain emojis?

The issue is being observed not only with emojis but with any long text with many lines
Emojis just make the delay even longer

Why does a covenant even need to contain 6000 lines of text?
I don't know, but there is a bug issue, and we can make the code more effective in this point

indra/llui/lltextbase.cpp
LLGuru: > It would be much safer to use `LLPointer` instead.

Done

akleshchev (approved):

marchcat (approved):

marchcat (approved):
Let's test it.

PR #2594, opened 2024-09-18
Tiny argument names correction (tittle -> title)

PanteraPolnocy:
This is by no means a change that would be visible to the user, but for the sake of consistency of the "titles"...


marchcat (approved):
Thank you!

PR #2593, opened 2024-09-18
Remove quads rendering mode entirely

Ansariel:
Quads had to go, so I guided them out. 👋 

This is taken from my [changes made to Firestorm](https://github.com/FirestormViewer/phoenix-firestorm/commit/0c4c1b59685ff99b93e220a03e79ba2b9df0bbfc) over 7 years ago and unless I made some copy&paste error, this should run just fine.

PR #2596, opened 2024-09-18
Fix copy&paste error in llfloaterimagepreview.cpp

Ansariel:
Follow-up to #2593 : Restore image preview rendering based on original values - there were some changes in Firestorm that superseeded the original change, which I missed while adding the changes to the LL viewer.


akleshchev (approved):

marchcat (approved):
Thank you!

PR #2612, opened 2024-09-19
#1519 ObjectGrab message includes invalid SurfaceInfo data when in mouselook mode

LLGuru:


marchcat (approved):

indra/newview/lltoolgrab.cpp
```diff
@@ -147,7 +147,7 @@ bool LLToolGrabBase::handleMouseDown(S32 x, S32 y, MASK mask)
     if (!gAgent.leftButtonGrabbed() || ((mask & DEFAULT_GRAB_MASK) != 0 && !gAgentCamera.cameraMouselook()))
     {
         // can grab transparent objects (how touch event propagates, scripters rely on this)
-        gViewerWindow->pickAsync(x, y, mask, pickCallback, /*bool pick_transparent*/ true);
+        gViewerWindow->pickAsync(x, y, mask, pickCallback, true, false, true);
```
akleshchev, line 150: So, if left button not grabbed by script or not in mouselook. Isn't the issues specifically about mouselook? How changing this pick help with mouselook?

LLGuru: This really helps :-)

The 7th argument of the pickAsync() method is `pick_unselectable`
Its value goes to the constructor of the class `LLPickInfo` as the argument `pick_surface_info` AKA `pick_uv_coords`
and is stored in the member field `mWantSurfaceInfo`

Probably this is the wrong order of the values sent to the constructor because the 8th argument is named `pick_unselectable` but it always receives `true`

Anyway, to fix the issue we should sent `true` to `pick_surface_info` to have true in `mWantSurfaceInfo`

akleshchev: And why this isn't an issues outside of mouselook? Does it launch a different pickAsync?

LLGuru: See the next variant

akleshchev (approved):

indra/newview/lltoolgrab.cpp
akleshchev: Looks good.
Suggestion, cache gSavedSettings.getBOOL("SelectInvisibleObjects") while you are at it.

LLGuru: > P.S. Still confused why it's mouselook specific, but fix makes sense.

In Mouselook mode we use `pickAsync`, but with no Mouselook mode we use `pickImmediate`
That's why this mistake in `pickAsync` doesn't take effect with no Mouselook mode

PR #2622, opened 2024-09-19
Restore option to change location of existing pick

Ansariel:
Here, have a quick fix for #2614 : Restore functionality to update the location of an existing pick by returning the original code that was removed in a2d75995d17c596ca57d9c9a092cbf45d335d90c because the button went missing.

### DIFF
--- a/indra/newview/llpanelprofilepicks.cpp
+++ b/indra/newview/llpanelprofilepicks.cpp
@@ -801,7 +801,7 @@ void LLPanelProfilePick::onClickSave()
     {
         mParcelCallbackConnection.disconnect();
     }
-    if (mLocationChanged) 
+    if (mLocationChanged)
     {
         onClickSetLocation();
     }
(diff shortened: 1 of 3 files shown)


akleshchev (approved):

marchcat (approved):
Thank you!

indra/newview/llpanelprofilepicks.cpp
```diff
@@ -769,6 +801,10 @@ void LLPanelProfilePick::onClickSave()
     {
         mParcelCallbackConnection.disconnect();
     }
+    if (mLocationChanged) 
+    {
```
marchcat, line 805:
```suggestion
    if (mLocationChanged)
    {
```
to pass pre-commit

marchcat: Wow, it allowed me to fix it right away this time.

Issue #2462, opened 2024-08-29
Optimize unloading of prims


cosmic-linden:
Optimize `LLMeshRepository::unregisterMesh`.

`LLViewerObjectList::cleanDeadObjects` takes a long time to run, especially during shutdown, but also when moving around the world, leading to intermittent slow frames. `LLMeshRepository::unregisterMesh` was found to be a bottleneck for this function.

## Test plan

Due to the nature of the performance fix, regressions are not expected. There is a small chance of increased crashes with `LLMeshRepository` in the stack trace, most noticably on shutdown.

Dan-Linden:
Passed QA.
Verified on Second Life Release 26.2.0.21655710006 on Win11/OSX in the scope of https://github.com/secondlife/iqa/issues/606

PR #2623, opened 2024-09-20
secondlife/viewer#2462: Optimize unloading of prims

cosmic-linden:


indra/newview/llvovolume.cpp
```diff
@@ -998,6 +1001,28 @@ LLDrawable *LLVOVolume::createDrawable(LLPipeline *pipeline)
     return mDrawable;
 }
 
+// Inverse of gMeshRepo.loadMesh and gMeshRepo.getSkinInfo, combined into one function
+// Assume a Collada mesh never changes after being set.
```
cosmic-linden, line 1005: This is a big assumption. If the sculpt ID were to ever change, we would also need to call `unregisterOldMeshAndSkin` from `LLPrimitive::setVolume`.

It's my understanding that features to change sculpt IDs were disabled long ago.

marchcat (approved):

cosmic-linden:
Thank you!

PR #2640, opened 2024-09-20
Improve inventory finder floater performance: Don't parse XUI tree each frame

Ansariel:
...and fix some more issues found by Visual Studio


marchcat (approved):
Awesome, thank you very much!

Issue #2618, opened 2024-09-19
Roles and Members tab is a mess


akleshchev:
### Environment

Second Life Release 7.1.9.10515727195 (64bit)

### Description

Lots of unused space, weird jumping, spacing. banned members is the worst off.

### Reproduction steps

1. Create or edit a group
2. Add a large list of banned residents.
3. Switch tabs away and to banned (view will jump oddly, hiding tabs)
4. Try scrolling ban list (scroll goes off screen)
5. Extra free space below tabs list (either whole list should fit on screen or there should be no free space)
6. Free, potentially useles space in other tabs.

Example: space after the ban list
<img width="479" height="771" alt="Image" src="https://github.com/user-attachments/assets/697b5bdc-b016-429e-bbf7-e761ff72a225" />

Dan-Linden:
Reopened.
Tested on Second Life Release 7.2.3.19079415635 on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/566

There is longer a space below the ban residents button as shown in the description. That's fixed.

Each time you switch to the Banned Residents tab, it is scrolled down quite a ways. Video: https://github.com/user-attachments/assets/b8143190-50cb-4613-864e-f03e5fcd744c

Dan-Linden:
This fix came in after build 7.2.3.19118091373.
Leaving as "In QA" in case we cut another 25.08 build.

Geenz:
@Dan-Linden I’ll review what we have today that’s queued up- might do a build later

Dan-Linden:
Passed QA.
Verified on Second Life Release 7.2.3.19241056504 on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/568

PR #2642, opened 2024-09-21
#2618 Roles and Members tab is a mess (show ctrl pos in debug views)

LLGuru:
The "Debug Views" mode will work as follows:
image


marchcat (approved):
This may be useful.

Issue #2655, opened 2024-09-23
UI/UX: Make the new alpha gamma drop down in the build floater not break the build floater


Geenz:
In the new alpha gamma feature, we've added a new field that allows a resident to select what gamma curve to apply to content.  We need this field to not cause the build floater to overflow.  Attached is a screenshot of its current state.

![image](https://github.com/user-attachments/assets/e51dde81-b661-4c04-a049-435e48c70bf3)

The alpha gamma changes can be found in https://github.com/secondlife/viewer/tree/geenz/feat/legacy-alpha-params

PR #2663, opened 2024-09-24
#2655 update alpha gamma drop down position

maxim-productengine:
Also minor changes: disable drop down when the object is not editable and enable correct text label otherwise.
image


akleshchev (approved):

Geenz:
One caveat to this- this UI should always be enabled. We have some special handling for alpha gamma that allows it to function even on no-mod objects through a special message that was introduced in the alpha params branch.

maxim-productengine:
Oh, I see. Reverted that change, so this control is always enabled.

akleshchev (approved):

PR #2680, opened 2024-09-25
Restore currently entered text in chat entry textbox after going through history with Ctrl-PgUp/PgDown

Ansariel:
This changes a rather inconvenient annoyance in the chat entry textbox. Say you entered several messages and started typing a new message into the chat entry textbox, you then decide to scroll through the history with Ctrl-PgUp/PgDown and then return to the actual input, everything you entered is gone and you have to enter it again. This change will restore the current input from before scrolling through the history.

Sounds complicated, therefor I made videos showing the difference.

Old behavior: https://gyazo.com/f6222fb2ec2a2885e000f8f5f7f71286
New behavior: https://gyazo.com/2069706990534b771e859573f8bb7840


marchcat (approved):
Thank you, @Ansariel!

PR #2681, opened 2024-09-25
Generalize LLPointer's comparison operators to allow comparable LLPointer types.

nat-goodspeed:
Also remove hugely redundant restatement of the `LLPointer` class for `LLConstPointer`, which can be effected with an alias template.

Use canonical copy-and-swap idiom for `LLPointer`'s assignment operators.


marchcat (approved):
Awesome, thank you!

PR #2688, opened 2024-09-25
Fix various issues with snapshot upload fee calculation

Ansariel:
* Removed orphaned LLPanelOutfitSnapshotInventory
* Calculation of upload in snapshot to inventory panel for insufficient funds was based on preview image size instead of encoded image size, resulting in upload fees of L$50 for textures <= 1024x1024 pixels
* Calculation on the options button had the same issue plus additionally was using referring to the previously used snapshot type which isn't necessarily the same as snapshot to inventory. Removed the upload fee from the button.
* Added back display of calculated upload fee on the snapshot to inventory panel itself. The fee is updated if the image size changes to always display the correct upload fee.

### DIFF
--- a/indra/newview/llpanelsnapshot.cpp
+++ b/indra/newview/llpanelsnapshot.cpp
@@ -37,6 +37,7 @@
 
 // newview
 #include "llsidetraypanelcontainer.h"
+#include "llsnapshotlivepreview.h"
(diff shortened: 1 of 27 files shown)


marchcat (approved):
Thank you!

PR #2695, opened 2024-09-26
fix xcode-16 build

AiraYumi:


AiraYumi:
Can you check if you can build it?

AiraYumi:
Thank you @nat-goodspeed.
You can probably build it with a new option, but it's probably best not to use it - if you can get it to work this PR might not be necessary.

nat-goodspeed:
The `lerp()` conflict is addressed in the target marchcat/xcode-16 branch.

I think I'll also try to resolve the ambiguous-reversed-operator conflict in source code on that branch. Thank you for pointing out the issue though!

AiraYumi:
I'm glad I could help you fix it.

AiraYumi:
This PR is no longer in use so we are closing it.

PR #2712, opened 2024-09-26
Fix lerp issues and eliminate flerp in favor of std::lerp

Ansariel:


marchcat (approved):
Oh, that's nice. TYVM!

nat-goodspeed (commented):
I like it! Thank you!

Looks like I just put the `f` in the wrong place. :-)

PR #2713, opened 2024-09-27
Don't clear big render target if all pixels will be replaced

cosmic-linden:


Geenz (commented):
LGTM - I like the new API addition.

Geenz (approved):

Geenz:
You may want to re-merge with develop and let CI re-run. Looks like Linux build isn't happy.

indra/newview/pipeline.cpp
```diff
@@ -7476,7 +7478,7 @@ void LLPipeline::generateSMAABuffers(LLRenderTarget* src)
             LLGLSLShader& blend_weights_shader = gSMAABlendWeightsProgram[fsaa_quality];
 
             dest.bindTarget();
-            dest.clear(GL_COLOR_BUFFER_BIT);
+            dest.invalidate(GL_COLOR_BUFFER_BIT);
```
RyeMutt, line 7481: This could have side effects on SMAA as the edge detection shaders utilize discard.

Geenz: That's actually a pretty good point - discards get weird when you don't have something clear.  Do you have any good scenes to test with?

cosmic-linden:
@Geenz The Linux build is complaining about an autobuild issue with conflicting dependencies, so I've merged the code into develop for now.

PR #2722, opened 2024-09-27
Add UI callback to invoke specified script via menu

maxim-productengine:


indra/newview/llviewermenu.cpp
```diff
@@ -9461,6 +9463,23 @@ void LLUploadCostCalculator::calculateCost(const std::string& asset_type_str)
...
+    {
+        LL_WARNS() << "Script name is not specified" << LL_ENDL;
+        return;
+    }
+    fsyspath fs_script_path(script_path);
+    script_path = fs_script_path.lexically_normal().u8string();
+    if (!fs_script_path.is_absolute())
+    {
+        script_path = (fsyspath(gDirUtilp->getExpandedFilename(LL_PATH_SCRIPTS, "lua")) / script_path).u8string();
+    }
```
nat-goodspeed, line 9479: At this point, `LLLUAmanager::runScriptFile()` [uses `ScriptCommand` to look for any relative script path on `"LuaCommandPath"`](https://github.com/secondlife/viewer/blob/maxim/lua-uicallback-invoke/indra/newview/llluamanager.cpp#L195-L198) from settings.xml. So maybe we don't need lines 9474-9479?

nat-goodspeed (approved):

indra/newview/llviewermenu.cpp
maxim-productengine: You are right, thanks.

PR #2729, opened 2024-09-27
Linux/GCC 13 build fix for template specialization of pasteTextWithLinebreaks

Nicky-D:
GCC 13 did not like having a template<> inside the class body at all.
Thus turn the specialization of pasteTextWithLinebreaks into a class methode pasteTextWithLinebreaksImpl.

### DIFF
--- a/indra/llui/lltexteditor.h
+++ b/indra/llui/lltexteditor.h
@@ -313,10 +313,9 @@ public:
     template <typename STRINGTYPE>
     void            pasteTextWithLinebreaks(const STRINGTYPE& clean_string)
     {
-        pasteTextWithLinebreaks<LLWString>(ll_convert(clean_string));
+        pasteTextWithLinebreaksImpl(ll_convert(clean_string));
     }
-    template <>
-    void            pasteTextWithLinebreaks<LLWString>(const LLWString & clean_string);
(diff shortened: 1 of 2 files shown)


Ansariel:
Another GCC bug. This construct is allowed as per C++17. GCC bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282

indra/llui/lltexteditor.h
```diff
@@ -313,10 +313,9 @@ class LLTextEditor :
     template <typename STRINGTYPE>
     void            pasteTextWithLinebreaks(const STRINGTYPE& clean_string)
     {
-        pasteTextWithLinebreaks<LLWString>(ll_convert(clean_string));
+        pasteTextWithLinebreaksImpl(ll_convert(clean_string));
```
Ansariel, line 316: What about
```
    {
        if constexpr (std::is_same_v<LLWString, STRINGTYPE>())
            pasteTextWithLinebreaksImpl(clean_string);
        else
            pasteTextWithLinebreaks<LLWString>(ll_convert(clean_string));
    }
```
to prevent an unnecessary ll_convert if not needed?

marchcat (approved):
Thank you!

PR #2731, opened 2024-09-28
Fix GLTF Load Crash

TommyTheTerrible:
For some reason, shader is null at some point in the Develop -> GLTF -> Load process, so this needs to be qualified to avoid assigning a nullptr to sCurBoundShaderPtr.

Probably a bad idea to allow this, anyway.

### DIFF
--- a/indra/llrender/llglslshader.cpp
+++ b/indra/llrender/llglslshader.cpp
@@ -1050,7 +1050,7 @@ void LLGLSLShader::bind()
 
     gGL.flush();
 
-    if (sCurBoundShader != mProgramObject)  // Don't re-bind current shader
+    if (this && sCurBoundShader != mProgramObject)  // Don't re-bind current shader
     {
         if (sCurBoundShaderPtr)
         {
@@ -1064,7 +1064,7 @@ void LLGLSLShader::bind()
         LLVertexBuffer::setupClientArrays(mAttributeMask);
     }
 
-    if (mUniformsDirty)
+    if (this && mUniformsDirty)
     {
         LLShaderMgr::instance()->updateShaderUniforms(this);
         mUniformsDirty = false;
--- a/indra/llrender/llglslshader.cpp
+++ b/indra/llrender/llglslshader.cpp
@@ -1050,7 +1050,7 @@ void LLGLSLShader::bind()
 
     gGL.flush();
 
-    if (this && sCurBoundShader != mProgramObject)  // Don't re-bind current shader
+    if (sCurBoundShader != mProgramObject)  // Don't re-bind current shader
     {
         if (sCurBoundShaderPtr)
         {
@@ -1064,7 +1064,7 @@ void LLGLSLShader::bind()
         LLVertexBuffer::setupClientArrays(mAttributeMask);
     }
 
-    if (this && mUniformsDirty)
+    if (mUniformsDirty)
     {
         LLShaderMgr::instance()->updateShaderUniforms(this);
         mUniformsDirty = false;
--- a/indra/newview/gltfscenemanager.cpp
+++ b/indra/newview/gltfscenemanager.cpp
@@ -655,14 +655,14 @@ void GLTFSceneManager::render(Asset& asset, U8 variant)
 
             if (!shader_bound)
             { // don't bind the shader until we know we have somthing to render
-                if (opaque)
-                {
-                    gGLTFPBRMetallicRoughnessProgram.bind(variant);
-                }
-                else
-                { // alpha shaders need all the shadow map setup etc
-                    gPipeline.bindDeferredShader(gGLTFPBRMetallicRoughnessProgram.mGLTFVariants[variant]);
-                }
+                //if (opaque)
+                //{
+                //    gGLTFPBRMetallicRoughnessProgram.bind(variant);
+                //}
+                //else
+                //{ // alpha shaders need all the shadow map setup etc
+                //    gPipeline.bindDeferredShader(gGLTFPBRMetallicRoughnessProgram.mGLTFVariants[variant]);
+                //}
 
                 if (!rigged)
                 {


marchcat (approved):
Thank you!

marchcat:
Well, this was unexpected..
```
/home/runner/work/viewer/viewer/indra/llrender/llglslshader.cpp:1053:9: error: 'this' pointer cannot be null in well-defined C++ code; pointer may be assumed to always convert to true [-Werror,-Wundefined-bool-conversion]
    if (this && sCurBoundShader != mProgramObject)  // Don't re-bind current shader
        ^~~~ ~~
/home/runner/work/viewer/viewer/indra/llrender/llglslshader.cpp:1067:9: error: 'this' pointer cannot be null in well-defined C++ code; pointer may be assumed to always convert to true [-Werror,-Wundefined-bool-conversion]
    if (this && mUniformsDirty)
        ^~~~ ~~
2 errors generated.
```

TommyTheTerrible:
Well, I'm glad Clang and I agree, but the pointer is definitely null at some point which causes the GLTF load to crash the viewer on Windows. Visual Studio told me so!

Does the GLTF Load crash on Mac or Linux, or is this only a Windows issue?

I will try hunting through the process more to see why it has a nullptr.

RyeMutt:
The LLGLSLShader object would be invalid in the case of the shader failing to load. The correct fix would be to figure out what shader is failing to load and correcting it in the shader source.

Nicky-D:
```
Well, I'm glad Clang and I agree, but the pointer is definitely null at some point which causes the GLTF load to crash the viewer on Windows. Visual Studio told me so!

Does the GLTF Load crash on Mac or Linux, or is this only a Windows issue?
```
If you send me the file I'm happy to test on Linux.

As Rye pointed out, is checking the shader before the call the proper fix.
clang more or less low key tells you that. `this` being nullptr would be undefined behavior and thus `this` can be assumed to always be valid. Hence often compilers can also optimize that check away (and render the fix potentially useless).

There's another problem regardless, parts of the object gets access without that bogus `this` check. That yields UB/crashes as well.

RunitaiLinden (changes requested):
If you're null checking this, you're just burying a bug that exists elsewhere.

Pretty sure the real issue is that we unload the GLTF shaders if GLTFEnabled gets set to false, and GLTFEnabled gets set to false when you connect to a non-GLTF enabled region.  We're not loading the GLTF shaders by default right now because they fail to compile on some systems we support, which was making SL unusable on those systems because of how shader compilation errors are handled.  Develop->GLTF->Open should be disabled when GLTFEnabled is false.  If power users want to use it on non-GLTF enabled regions, they can manually set that setting to true.

RunitaiLinden:
> ```
> Well, I'm glad Clang and I agree, but the pointer is definitely null at some point which causes the GLTF load to crash the viewer on Windows. Visual Studio told me so!
> 
> Does the GLTF Load crash on Mac or Linux, or is this only a Windows issue?
> ```
> 
> If you send me the file I'm happy to test on Linux.
> 
> As Rye pointed out, is checking the shader before the call the proper fix. clang more or less low key tells you that. `this` being nullptr would be undefined behavior and thus `this` can be assumed to always be valid. Hence often compilers can also optimize that check away (and render the fix potentially useless).
> 
> There's another problem regardless, parts of the object gets access without that bogus `this` check. That yields UB/crashes as well.

Null checking the shader pointer at the draw call level is wrong, too.

I'd add an llassert at the top of LLGLTFSceneManager::render that sCurBoundShaderPtr is not null, and document that that is a precondition of that function.

Asserting on preconditions is something we do a lot as way of having the code validate itself in developer builds, but folks seem to have some confusion around what the term "precondition" _means_.

A precondition MUST be satisfied before the function is called.  The function MUST NOT contain code to handle the case where the precondition is not met.  This bubbles the root cause of the bug up to a place where it makes sense to handle the condition and generally makes for cleaner, faster code.  In the example here, adding null checks everywhere sCurBoundShaderPtr is accessed would mask the true bug, clutter up many inner loops with null checks, and would result in the viewer getting into a state where it's pumping the GLTF render pipe every frame while drawing nothing at all.

TommyTheTerrible:
I'm sorry to have submitted such a simple and inappropriate change. Clang and others are definitely right, this was not the right way to solve but Microsoft's compiler let me so I ran with it.

I tracked through the callstack and the shader seems to be the gGLTFPBRMetallicRoughnessProgram. It is being passed a variant in GLTFSceneManager::render(Asset& asset, U8 variant), but the object shows no variants available. I removed the passing of a variant to it and the crashes stopped but the loaded GTLF loaded asset did not draw.

I commented gGLTFPBRMetallicRoughnessProgram out from the GLTFSceneManager::render function and the GLTF asset loads and draws. I do not see anything else missing or looking different (PBR assets seem to have normal, emissive, etc.), so I assume this is an old shader that was meant to be removed? I don't see it being used anywhere else.

Updating PR with the nullptr check removed and gGLTFPBRMetallicRoughnessProgram as well.

TommyTheTerrible:
> If you're null checking this, you're just burying a bug that exists elsewhere.
> 
> Pretty sure the real issue is that we unload the GLTF shaders if GLTFEnabled gets set to false, and GLTFEnabled gets set to false when you connect to a non-GLTF enabled region. We're not loading the GLTF shaders by default right now because they fail to compile on some systems we support, which was making SL unusable on those systems because of how shader compilation errors are handled. Develop->GLTF->Open should be disabled when GLTFEnabled is false. If power users want to use it on non-GLTF enabled regions, they can manually set that setting to true.

I'm sorry, I just read this after replying. It was minimized on the website for some reason.

You are right, GLTFEnabled was False in testing and this is probably the reason. I had thought GTLF was on because PBR materials are rendering in the test region.

 I will close the PR if this is truly the case.

PR #2733, opened 2024-09-29
SDL and Linux improvements

RyeMutt:
This is a collection of changes to improve Linux and SDL
* Implements many stubbed out portions of LLWindowSDL
* Converts legacy LLWindowSDL code to use modern SDL functions
* Fixes finding graphics memory on linux
* Cleans up old hacks that are no longer relevent for linux
* Updates Linux feature table to be largely equivalent to windows


Nicky-D:
Very nice, only thing I'm a little worried about is the keyboard handling.

I recall I did originally start with just the SDL functions rather than that whole X11 dance, but hit a snag when pasting large amounts of data (people sometimes do that when writing scripts or notecard 'offline').
I dunno if this is finally fixed, it has been 2 years after all, but https://github.com/libsdl-org/SDL/issues/8871 seems to suggest not. 
That's definitively something to look out for.

marchcat (approved):

PR #2737, opened 2024-09-30
set CURLOPT_ENCODING to nullptr in case of curl7

Nicky-D:
The intention for this is for a discussions about the implications and ramifications rather then getting this in as is (there would be a few more lines for Curl7)

Curl 7 introduced more strictness how it handles CURLOPT_ENCODING.

With earlier versions it as apparently okay when the server sent an encoding curl did not implement.

Newer versions are much more stricter and refuse to complete downloads under such circumstances. This leads to almost nothing loading. The culrpit seemingly being either the simulator or asset servers sending headers curls does not accept.

The workaround is to set CURLOPT_ENCODING to nullptr rather than "". This does come with the implication that curl will not handle decompression transparently, rather this burden gets shifted to the application. It will also not send Accept-Encoding headers automatically.


marchcat (approved):
Thank you!

marchcat:
We should probably adjust the related unit tests as well.

Nicky-D:
Good call, I replaced a few more occurrences of CURLOPT_ENCODING. Though I saw no tests related to this.
Still a bit wary though how safe this change is, then again I came to the conclusion that it will only affect Curl 7 for now, so maybe it is a good test balloon. Find possible edge cases with extended curl 7 usage first, see if something barfs

PR #2743, opened 2024-09-30
Linux: Remove APR from Linux-Symbolgrabber

Nicky-D:
dlopen and dlsym are well…suited for and let us avoid the need to deal with the pecularities of APR.

This fixes a problem after the latest  updates of APR and WebRTC where APR sometimes would crash when trying to load the DSOs and resolve symbols (or sometimes even when creating the apr pool).


marchcat (approved):
Thank you!

PR #2746, opened 2024-09-30
Clean up LLUI and fix/add suggestions from VS

Ansariel:
@nat-goodspeed contains the changes from #2732 and a lot of other fixes suggested by VS


Ansariel:
@marchcat are we waiting for something before this PR can be merged?

marchcat:
@Ansariel, it looks good to me. I'm just keeping it open since you mentioned @nat-goodspeed.

Ansariel:
> @Ansariel, it looks good to me. I'm just keeping it open since you mentioned @nat-goodspeed.

I just mentioned Nat because it contains the cleanup changes from #2732 which otherwise would have gotten lost.

nat-goodspeed (commented):
Thanks, Ansariel. I confirm that I observe the changes from #2732 here.

I appreciate all the method annotations! Especially those that affect a whole inheritance tree.

I'm rather surprised at all the unused/deleted methods. One that caught my eye is `LLNotificationForm::fromLLSD()`. I gather that method definition was previously deleted, which hasn't mattered since there are no existing callers.

Since we have a conversion from Lua data to `LLSD`, if that `fromLLSD()` method still existed, a Lua script could conceivably create a new notification. That's an interesting idea; no idea if it has any actual use cases. I guess it's moot though.

nat-goodspeed:
To my surprise, I can't find a changeset that deletes that method - maybe it was never written?! Or it was deleted before we cut over to Subversion or something.

PR #2758, opened 2024-10-01
Port Fontcache from Firestorm.

Nicky-D:
The cache will make sure a font only gets loaded once in memory and then reused, saving a lot of unnecessary I/O and memory..

Test case: Open viewer on Mac and Windows (I did Linux). Verify program does display menus etc,

### DIFF
--- a/indra/llrender/llfontfreetype.h
+++ b/indra/llrender/llfontfreetype.h
@@ -43,15 +43,28 @@ typedef struct FT_FaceRec_* LLFT_Face;
 struct FT_StreamRec_;
 typedef struct FT_StreamRec_ LLFT_Stream;
 
+namespace nd
+{
+    namespace fonts
+    {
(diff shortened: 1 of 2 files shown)


indra/llrender/llfontfreetype.h
```diff
@@ -43,15 +43,28 @@ typedef struct FT_FaceRec_* LLFT_Face;
 struct FT_StreamRec_;
 typedef struct FT_StreamRec_ LLFT_Stream;
 
+namespace nd
```
Ansariel, line 46: Maybe change the namespace to "ll"? 😉

Nicky-D: Whoops 😬 
You're right, will do that tomorrow

akleshchev (approved):

marchcat (approved):
Thank you!

PR #2767, opened 2024-10-02
Restore Linux Release config builds

RyeMutt:
* Restores support for Release build config on Linux
* Updates openal package with linux pipewire support
* Default OpenAL audio backend on so all build configurations have functional audio
* Switch windows autobuild build command to utilize `cmake --build` to shave 4 minutes off windows GHA time

### DIFF
--- a/autobuild.xml
+++ b/autobuild.xml
@@ -3234,7 +3234,7 @@ Copyright (c) 2012, 2014, 2015, 2016 nghttp2 contributors</string>
                 <key>options</key>
                 <array>
                   <string>-G</string>
-                  <string>Ninja</string>
+                  <string>Unix Makefiles</string>
                 </array>
               </map>
               <key>build</key>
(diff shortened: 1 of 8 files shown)


brad-linden (approved):

marchcat (approved):

autobuild.xml
```diff
@@ -3203,8 +3201,7 @@ Copyright (c) 2012, 2014, 2015, 2016 nghttp2 contributors</string>
                 <key>options</key>
                 <array>
                   <string>-G</string>
-                  <string>Ninja</string>
-                  <string>-DUSE_OPENAL:BOOL=ON</string>
+                  <string>Unix Makefiles</string>
```
Nicky-D, line 3204: We should stick to one build system, either ninja (as it is still in the ReleaseOS config) or Makefiles.

ninja used to be faster, but I haven't tested that in a while. Generally I do not mind either, just that we use *one* standard and not multiple

RyeMutt: I've unified both configurations to the `Unix Makefiles` generator.

PR #2771, opened 2024-10-02
secondlife/viewer#2472: Debug tools for labeling textures

cosmic-linden:


indra/newview/llviewertexturelist.cpp
```diff
@@ -1113,6 +1147,62 @@ F32 LLViewerTextureList::updateImagesCreateTextures(F32 max_time)
...
+    if (!debug_texture_label()) { return; }
+
+    static const std::string local_prefix = "lltexlocal";
+    static const std::string other_prefix = "lltexother";
+
+    std::string label;
+    bool error;
+    for (image_list_t::iterator it = mImageList.begin(); it != mImageList.end(); ++it)
+    {
+        LLViewerFetchedTexture* image = *it;
```
brad-linden, line 1195: probably would prefer
```suggestion
    for (LLViewerFetchedTexture * image : mImageList)
    {
```

brad-linden (approved):

indra/newview/app_settings/settings.xml
```diff
@@ -7932,6 +7932,17 @@
       <key>Value</key>
       <integer>0</integer>
     </map>
+    <key>RenderDebugTextureLabel</key>
+    <map>
+      <key>Comment</key>
+      <string>Enable texture labels via glObjectLabel. Requires restart for some features.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
```
RunitaiLinden, line 7941: Make this not persist.  It will reduce performance forever when enabled, and since glObjectLabel only exists in GL 4.3 and later, enabling this on previous OpenGL versions will make the viewer crash forever on many machines if settings get saved with this enabled.  Also maybe protect calls to glObjectLabel with a GL version check.

RunitaiLinden (changes requested):

indra/newview/llviewertexturelist.cpp
cosmic-linden: Yea that wasn't necessary for this loop. Always good to simplify.

indra/newview/app_settings/settings.xml
cosmic-linden: OK, this no longer persists. For convenience reasons, I've split the feature into two configs so that some functionality no longer requires a restart.

GL version checks are also put in place.

PR #2782, opened 2024-10-03
viewer_private#300 Picker fixes

akleshchev:


marchcat (approved):

maxim-productengine (approved):

akleshchev:
Not sure if it's an entirely correct thing to do, but will see how it goes, alternatively can cause a 'cancel' instead.

Issue #2465, opened 2024-08-29
Add controls for dynamic exposure


RunitaiLinden:
### Description

RenderDynamicExposureEnabled - dynamic exposure on or off
RenderDynamicExposureEVOffset - Exposure Value offset to apply to graphics preferences “exposure” setting
RenderDynamicExposureMinimum -  Minimum EV offset
RenderDynamicExposureMaximum - Maximum EV offset
NOTE: EV Min/Max offset is relative to EV Offset applied to exposure slider
RenderDynamicExposureSpeed - number of seconds to take to reach desired exposure

Also increase maximum Exposure value in preferences to 2.0

PR #2786, opened 2024-10-03
#2465 Setup the ev offset/min/max and add some sane default values.

Geenz:


indra/newview/pipeline.cpp
```diff
@@ -7035,13 +7035,8 @@ void LLPipeline::generateExposure(LLRenderTarget* src, LLRenderTarget* dst, bool
...
         {
-            F32 hdr_scale = sqrtf(LLEnvironment::instance().getCurrentSky()->getGamma()) * 2.f;
-
-            if (hdr_scale > 1.f)
-            {
-                exp_min = 1.f / hdr_scale;
-                exp_max = hdr_scale;
-            }
+            exp_min = dynamic_exposure_ev_offset - dynamic_exposure_ev_min;
+            exp_max = dynamic_exposure_ev_offset + dynamic_exposure_ev_max;
```
cosmic-linden, line 7039: I learned the hard way that you need to actually call the cached controls for them to work reliably. Maybe it's only an issue for bool values? Or maybe there's something wrong with the `LLCachedControl` API...

brad-linden: I'm curious to hear more about this.  if the implicit casting `operator const T&()` for LLCachedControl is not functioning properly, we should figure out how to fix it

cosmic-linden: To clarify, I would try to read the LLCachedControl bool for the first time before opening a floater to decide what UI to display, and it wouldn't work the first time.

Relevant PR:

https://github.com/secondlife/viewer/pull/2492

Geenz:
This was backported and shipped with ExtraFPS.

PR #2749, opened 2024-10-01
[DEV] Add Native Support for Wayland

AiraYumi:
Enables native support for the Wayland desktop environment.

FLTK 1.4 is still in the prerelease stage, but the API changes are said to be stable, so I would like to start working on it at this stage.

Related Issues : https://github.com/secondlife/viewer/issues/1442 [Linux] Add Native Support for Wayland

- [x] Compile viewer
- [x] Use libdecor for GNOME
- [x] Allows you to launch the viewer.
- [x] Add Wayland Native Dialogs
- [x] Enable IME

PR #2803, opened 2024-10-04
Implement the groundwork for wayland support.

Nicky-D:
This contributes the ground work needed for #2749

During compile time try to detect libwayland-client-dev is installed,
  - If yes, compile a viewer which can run and start under wayland. libwayland-client is needed to detect if the viewer is hidden/minimized.
  - if no, disable wayland support and refuse to start the viewer under wayland (xwayland is okay).

To not introduce a hard link dependency on libwayland-client.so the object will be loaded dynamically if needed.


marchcat:

marchcat:
pre-commit looks unhappy with
```
indra/llwindow/llwindowsdl.cpp: 1 line starting with tabs found
```
however, it doesn't specify the line number.

marchcat (approved):

PR #2804, opened 2024-10-04
#2542 Buoyant Nametags (hide the name tag for deep underwater avatar)

LLGuru:


akleshchev:
I'm not sure this is expected behavior. If you see avatar you should be able to see avatars nametag. Plus QA failed the ticket because nametag was not visible, why make it even more invisible?

akleshchev (approved):
I tested build 7.0.0.582000 and there nametag stops rendering underwater if viewer from over water. Approving.

PR #2810, opened 2024-10-04
More cleanup of LLMath and the color classes in particular

Ansariel:


marchcat (approved):
Yay, cleanup! 
Thank you, @Ansariel!

PR #2811, opened 2024-10-04
Test matrix configurations GHA change

brad-linden:


AiraYumi:
Does this PR re-enable Release builds for Linux?

brad-linden:
unfortunately no, this branch did not work out the way we wanted for that.  but we are working towards that elsewhere

PR #2817, opened 2024-10-05
Fix issues with game control preferences

Ansariel:
* Add missing names for controls to make content localizable
* Fix labels for controls that are obviously wrong/make no sense


marchcat (approved):
Thank you!

AndrewMeadows:
Thank you for this PR.  There was a reason why I restricted the language of that UI to be just "Forward" instead of "Forward/Back" etc, however that was before the per-device settings allowed for an axis inversion, and I think my reasoning would no longer be convincing, especially if this UI choice was poking you in the eye.

AndrewMeadows (approved):

Issue #1883, opened 2024-06-28
Terrain Painting: Client test


cosmic-linden:

cosmic-linden:
Doing a WIP code check-in as this has been sitting for a while:

https://github.com/secondlife/viewer/pull/2823

Potential terrain brush shader implementation for future reference:

![Image](https://github.com/user-attachments/assets/c1188b6e-dba1-46ff-8554-27395b960dcc)

PR #2823, opened 2024-10-07
secondlife/viewer#1883: Terrain paint client prototype WIP code check-in

cosmic-linden:
This is a non-functioning, partial prototype that I'm checking in now, as it's been sitting on the back burner for a bit and was starting to accumulate conflicts.

Currently, a lot of paint data is stored in CPU RAM in the interest of modularity. More data could be kept exclusively GPU-side, and the alpha stamp queue is likely not needed except maybe possibly for implementing undo (which might not be a thing because it's unclear if that's a valuable feature to have in the long run).

## Test plan

The terrain paint feature is not functional yet, so there's no new features to test here.

Look out for system-specific startup crashes associated with shader compilation with the new terrain bake/terrain stamp shaders.

### DIFF
--- a/indra/newview/llterrainpaintmap.cpp
+++ b/indra/newview/llterrainpaintmap.cpp
@@ -573,7 +573,7 @@ LLTerrainPaintQueue LLTerrainPaintMap::convertPaintQueueRGBAToRGB(LLViewerTextur
             paint_out->mStartY = paint_in->mStartY;
             paint_out->mWidthX = paint_in->mWidthX;
             paint_out->mWidthY = paint_in->mWidthY;
-            paint_out->mBitDepth = 8; // Will be reduced to 5 bits later
+            paint_out->mBitDepth = 8; // Will be reduced to TerrainPaintBitDepth bits later
             paint_out->mComponents = LLTerrainPaint::RGB;
 #ifdef SHOW_ASSERT
             paint_out->assert_confined_to(tex);
@@ -602,7 +602,8 @@ LLTerrainPaintQueue LLTerrainPaintMap::convertPaintQueueRGBAToRGB(LLViewerTextur
         // round-trip which will reduce the bit depth, making the
         // pre-conversion step not necessary.
         queue_out.enqueue(paint_out);
-        queue_out.convertBitDepths(queue_out.size()-1, 5);
+        LLCachedControl<U32> bit_depth(gSavedSettings, "TerrainPaintBitDepth");
+        queue_out.convertBitDepths(queue_out.size()-1, bit_depth);
     }
 
     queue_in.clear();
@@ -775,7 +776,7 @@ LLTerrainPaintQueue LLTerrainPaintMap::convertBrushQueueToPaintRGB(const LLViewe
             const F32 dY = brush_start_y - F32(paint_out->mStartY);
             paint_out->mWidthX = U16(ceil(brush_width_x + dX));
             paint_out->mWidthY = U16(ceil(brush_width_y + dY));
-            paint_out->mBitDepth = 8; // Will be reduced to 5 bits later
+            paint_out->mBitDepth = 8; // Will be reduced to TerrainPaintBitDepth bits later
             paint_out->mComponents = LLTerrainPaint::RGB;
             // The brush strokes are expected to sometimes partially venture
             // outside of the paintmap bounds.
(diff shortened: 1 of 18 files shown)


indra/newview/llterrainpaintmap.cpp
```diff
@@ -283,3 +301,690 @@ bool LLTerrainPaintMap::bakeHeightNoiseIntoPBRPaintMapRGB(const LLViewerRegion&
...
+            const LLGLenum format = paint_in->mComponents == 4 ? GL_RGBA : GL_RGB;
+            constexpr LLGLenum type = GL_UNSIGNED_BYTE;
+            stamp_image = new LLImageGL(stamp_tex_name, components, target, internal_format, format, type, LLTexUnit::TAM_WRAP);
+            // Nearest-neighbor filtering to reduce surprises
+            stamp_image->setFilteringOption(LLTexUnit::TFO_POINT);
+
+            // Initialize the image dimensions in GL
+            constexpr U8* undefined_data_for_now = nullptr;
+            gGL.getTexUnit(0)->bind(stamp_image, false, true);
+            glTexImage2D(GL_TEXTURE_2D, 0, internal_format, width_rounded, height_rounded, 0, format, type, undefined_data_for_now);
```
brad-linden, line 544: TIL `data` is allowed to be `nullptr`:
https://registry.khronos.org/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml
>data may be a null pointer. In this case, texture memory is allocated to accommodate a texture of width width and height height. You can then download subtextures to initialize this texture memory. The image is undefined if the user tries to apply an uninitialized portion of the texture image to a primitive.

indra/newview/llterrainpaintmap.cpp
```diff
@@ -283,3 +301,690 @@ bool LLTerrainPaintMap::bakeHeightNoiseIntoPBRPaintMapRGB(const LLViewerRegion&
...
+        // Apply ("stamp") the paint to the render target
+        {
+            shader.bindTextureImageGL(LLShaderMgr::DIFFUSE_MAP, stamp_image);
+            const F32 width_fraction = F32(paint_in->mWidthX) / F32(width_rounded);
+            const F32 height_fraction = F32(paint_in->mWidthY) / F32(height_rounded);
+            shader.uniform2f(LLShaderMgr::TERRAIN_STAMP_SCALE, width_fraction, height_fraction);
+            buf->setBuffer();
+            buf->draw(LLRender::TRIANGLES, buf->getIndicesSize(), 0);
+        }
+
```
brad-linden, line 568: I realize this is a preliminary version of this, but is here where we would want to possibly return and enqueue a job to later check the results via `glReadPixels()` to avoid stalling?

indra/newview/llterrainpaintmap.cpp
```diff
@@ -283,3 +301,690 @@ bool LLTerrainPaintMap::bakeHeightNoiseIntoPBRPaintMapRGB(const LLViewerRegion&
...
+        // Apply the "brush" to the render target
+        {
+            // TODO: Use different shader for this - currently this is using the stamp shader. The white image is just a placeholder for now
+            shader.bindTexture(LLShaderMgr::DIFFUSE_MAP, LLViewerFetchedTexture::sWhiteImagep);
+            shader.uniform2f(LLShaderMgr::TERRAIN_STAMP_SCALE, 1.0f, 1.0f);
+            buf->setBuffer();
+            buf->draw(LLRender::TRIANGLES, buf->getIndicesSize(), 0);
+        }
+
```
brad-linden, line 769: similar question here about expected `glReadPixels()` stall

brad-linden (changes requested):

indra/newview/llterrainpaintmap.h
```diff
@@ -26,9 +26,16 @@
 
 #pragma once
 
+#include "llviewerprecompiledheaders.h"
```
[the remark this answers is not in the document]
cosmic-linden: Huh, how did that get there? Nice catch.

indra/newview/llterrainpaintmap.cpp
cosmic-linden: The intent of the queues is to allow them to be called somewhere more sensible than directly in llviewermenu.cpp . I'm not 100% sure where they would land, but I would consider the multithreading limitations of OpenGL, plus the desire for real-time painting.

brad-linden (approved):

Issue #2696, opened 2024-09-26
[Win10,11]Crash the viewer after pressing on different types of gestures quickly and simultaneously


vladyslavasproductengine:
### Environment

Second Life Release 7.1.11.10945332304 (64bit)
Release Notes

You are at 189.6, 93.2, 21.4 in Product Engine located at simhost-04fc9fd708f788a32.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine/190/93/21
(global coordinates 336,830.0, 307,293.0, 21.4)
Second Life Server 2024-07-22.10048683488
Release Notes

CPU: Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz (2712 MHz)
Memory: 8067 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4894)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: GeForce 940MX/PCIe/SSE2

Windows Graphics Driver Version: 25.21.14.1735
OpenGL Version: 4.6.0 NVIDIA 417.35

Window size: 1920x1001
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 96m
Bandwidth: 3000kbit/s
LOD factor: 1.125
Render quality: 1
Texture memory: 2048MB
Disk cache: Max size 1638.4 MB (6.3% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/1,236 (0.0%)
September 26 2024 03:27:07

### Description

see reproduction steps.

### Reproduction steps

Steps to reproduce:

1. Login in the Viewer.
2. Click "Communicate"→"Gestures".
3. Try quickly and simultaneously pressing different types of gestures.

Actual behavior: 
Crash the viewer after pressing on different types of gestures quickly and simultaneously.
Expected behavior:
The viewer should not crash.

Additional information:
This issue was found according to the test plan:3.5.11 Gestures and Sounds
[...]

PR #2827, opened 2024-10-09
#2696 The viewer crashes on gestures floater (fix navigation)

LLGuru:


marchcat (approved):

indra/llrender/llgltexture.cpp
```diff
@@ -80,6 +80,7 @@ void LLGLTexture::cleanup()
     if(mGLTexturep)
     {
         mGLTexturep->cleanup();
+        mGLTexturep = nullptr;
```
akleshchev, line 83: For me crash reproes despite this change.

indra/llui/llscrolllistctrl.cpp
```diff
@@ -1105,38 +1105,32 @@ S32 LLScrollListCtrl::getItemIndex( const LLUUID& target_id ) const
...
+        // don't allow navigation to disabled elements
+        if (item->getEnabled())
+        {
+            prev_item = item;
         }
     }
 
+    if (!prev_item)
+    {
+        reportInvalidInput();
```
akleshchev, line 1128: This seem like a change in behavior.
Previously if nothing was selected it would select last item, which is expected, but now if nothing was selected it would reportInvalidInput();?

akleshchev (approved):

PR #2589, opened 2024-09-17
Minimal RLVa contribution

KittyBarnett:
As discussed this PR contributes (close to) the minimal possible amount of code to have the RLVa command parsing/processing base in the official viewer.

BTW - If the PR is still too big the individual commits can be applied incrementally (they were actually intended to be reviewed that way to begin with).

I tried to isolate all the RLVa code into its own 'Rlv' namespace to avoid polluting the regular intellisense.

Some things might unfortunately seem 'weird' (such as RlvActions being empty except for an 'isEnabled' function but that's because many things were ripped out and will be filled in as more and more functionality is added).

I've (only just) started some documentation which can - for the moment - be found at https://kittybarnett.github.io/rlva-docs/development/layout/ and also includes some test scenarios. I'll keep adding to this over time.

~~Some of the SFINAE code for command processors and handlers in rlvhelper.h used to have a fix for the Mac compiler but I've removed it hoping that recent updates made that go away (I can't however test this myself).~~ The template fix on Mac was still needed and has been reapplied.

---

For TPV devs: the contributed code was refactored and moved around/renamed in many places. I do maintain  a sync'ed branch that modifies the existing TPV code to match the contributed code and should significantly reduce/remove merge hassles so don't freak out 😸 .

@vir-linden

PR #2830, opened 2024-10-10
Project: RLVa

marchcat:
Based on #2589, for further testing and integration.


vir-linden (approved):
Good to go. Thanks Kitty!

PR #2833, opened 2024-10-11
Add 'Review Validity Check' GHA workflow

marchcat:


akleshchev (dismissed):

marchcat:
So it works. Awesome! 😸

akleshchev (approved):

Geenz:
@marchcat Can you rebase this onto develop?

marchcat:
@Geenz Done! However, this check doesn't start due to the CLA. In any other case, it should have already dismissed the current approval.

bennettgoble:
Hey, @marchcat this functionality is available in Github's repository rule settings. If we want to enable it (which I'm not sure we do) it could be done so through repository settings.

dismiss-2025-03-04 132137

marchcat:
@bennettgoble Thank you!

That other option, "Require approval of the most recent push," seems even closer to what I've been trying to achieve. Dismissing the reviews just felt like the most obvious solution to me at the time.

PR #2834, opened 2024-10-11
Remove traces of FLTK

Ansariel:


marchcat (approved):
Thank you, @Ansariel!

PR #2837, opened 2024-10-11
#2408 The long covenant with emojis (no double requesting)

LLGuru:


akleshchev (approved):

marchcat (approved):

indra/newview/llfloaterregioninfo.h
```diff
@@ -144,7 +145,7 @@ class LLPanelRegionInfo : public LLPanel
     void onChangeAnything();
     static void onChangeText(LLLineEditor* caller, void* user_data);
 
-    virtual bool refreshFromRegion(LLViewerRegion* region);
+    virtual bool refreshFromRegion(LLViewerRegion* region, bool no_request);
```
akleshchev, line 148: Please add a comment of what no_request does and when to use/not use it.

refreshFromRegion(region) is intuitive enough, but no_request feels out of place.

P.S. Personaly I would have done either some kind of state machine or throttle 'request was sent this frame, don't re-request' mechanics.

LLGuru: Replaced bool with enum

PR #2839, opened 2024-10-11
Fix GHA matrix config linux build

RyeMutt:
This works around a parse order issue/behavior in GHA config files that prevented the build from generating correctly.


brad-linden (approved):

marchcat (approved):

nat-goodspeed (commented):
I'm less enthused. I have an experiment in [branch nat/matrix](https://github.com/secondlife/viewer/compare/nat/matrix?expand=1) where we don't even put the configuration in the matrix at all (since there's only one for all platforms), but I can't see GH even attempting to run it.

PR #2856, opened 2024-10-15
Develop → Maint C sync

marchcat:
Known regressions:
- [x] #2857 (fixed at 35edd296a6f850b811c000c8b337abaa1fde9835);
- [x] Inventory: create notecard, add it to favorites, delete it there, go back, create a folder -> assert/crash (fixed at 4d8d0d85f159d95edde7692dee23d04db315d66e);
- [x] Upload to folder: conflict/regression (fixed at d3d349ae0f17a72481f30b9354b9367b1cd3b639);
- [x] #2862

PR #2857, opened 2024-10-15
Fix the filter in Inventory-Favorites (#2856 regression)

marchcat:
This filter didn't work after the merge for some reason; it seems that setFilterFavorites() should be called there, so I added it.


akleshchev (changes requested):
This shouldn't be needed. This panel uses special favorites_inventory_panel (due to hierarhy changes) and it displays both favorite items and non-favorite content of favorite folders, so filter is likely to break intended behavior for this panel.

Edit: I think I found the problem (initRootContent() is missing), building and checking.

marchcat:
Thank you!

marchcat:
Right, the code is present, but not for LLInventoryFavoritesItemsPanel.

marchcat:
It works like a charm now 😸

marchcat:
Closing since 35edd296a6f850b811c000c8b337abaa1fde9835 is already integrated.

PR #2858, opened 2024-10-15
Reduce MAX_AGENT_UPDATES_PER_SECOND

TommyTheTerrible:
This number has historically been 10.

Back in June, Beq raised it to 125 in this commit https://github.com/secondlife/viewer/commit/2e836ab06f934038394b46e9c861312c0441b657.

Their commit to the Firestorm repository makes a reference to a 25 number that I assume is the sim's tick rate. https://github.com/FirestormViewer/phoenix-firestorm/commit/beea955388333103972609f2809c221156ced499

I believe this change lead to several networking issues, including the increase in avatar outfit issues reported by users as the change was rolled out in updates to the community.

Regardless of this being the case, this update rate is absurdly high. In popular fast-paced shooter games, 60 updates a second is typical and 120 in highly competitive environments. Meanwhile, more leisurely-paced gaming platforms like Roblox is 30 and Minecraft is 20.

Pinging stakeholders or those in past discussions on the issue: @beqjanus @vldevel

### DIFF
--- a/indra/newview/llviewermessage.cpp
+++ b/indra/newview/llviewermessage.cpp
@@ -3211,7 +3211,7 @@ void send_agent_update(bool force_send, bool send_reliable)
                         last_camera_at;
 
     // compute sec_since_last_send
-    constexpr F64 MAX_AGENT_UPDATES_PER_SECOND = 125.0; // Value derived experimentally to avoid Input Delays with latest PBR-Capable Viewers when viewer FPS is highly volatile.
+    constexpr F64 MAX_AGENT_UPDATES_PER_SECOND = 20.0; // Value derived experimentally to avoid Input Delays with latest PBR-Capable Viewers when viewer FPS is highly volatile.
     constexpr F64 MIN_AGENT_UPDATES_PER_SECOND = 1.0; // keep-alive rate
     constexpr F64 MIN_AGENT_UPDATE_PERIOD = 1.0 / MAX_AGENT_UPDATES_PER_SECOND;
     constexpr F64 MAX_AGENT_UPDATE_PERIOD = 1.0 / MIN_AGENT_UPDATES_PER_SECOND;
--- a/indra/newview/llviewermessage.cpp
+++ b/indra/newview/llviewermessage.cpp
@@ -3211,7 +3211,7 @@ void send_agent_update(bool force_send, bool send_reliable)
                         last_camera_at;
 
     // compute sec_since_last_send
-    constexpr F64 MAX_AGENT_UPDATES_PER_SECOND = 20.0; // Value derived experimentally to avoid Input Delays with latest PBR-Capable Viewers when viewer FPS is highly volatile.
+    constexpr F64 MAX_AGENT_UPDATES_PER_SECOND = 45.0; // Match server tick rate
     constexpr F64 MIN_AGENT_UPDATES_PER_SECOND = 1.0; // keep-alive rate
     constexpr F64 MIN_AGENT_UPDATE_PERIOD = 1.0 / MAX_AGENT_UPDATES_PER_SECOND;
     constexpr F64 MAX_AGENT_UPDATE_PERIOD = 1.0 / MIN_AGENT_UPDATES_PER_SECOND;


TommyTheTerrible:
Possible related tracked issues:

https://feedback.secondlife.com/server-bugs/p/attachment-loss-on-rc-channel-2024-08-2910619830788
https://feedback.secondlife.com/bug-reports/p/voice-crashing-on-teleport

KrsityKu:
A few things:

1. SL Simulator tick rate is 45/s
2. The attachment issue is purely server side, and affects all viewers, and is supposedly fixed in 2024-10-15.11356152186
3. This limit should not be lowered, unless it can be verified that lowering does not cause these issues to return:
https://feedback.secondlife.com/bug-reports/p/directional-input-delays-with-latest-pbr-capable-viewers 
https://jira.firestormviewer.org/browse/FIRE-34171
4. Maximum is not the minimum, I doubt a viewer running at 20 fps is sending 125 updates.

TommyTheTerrible:
Where are people getting all these numbers? Beq says 25, KrsityKu says 45. Both are far lower than 125.

The issue is not what happens when the viewer is running 20 FPS, the issue is when you teleport to a new region the viewer tends to be going faster than on average because there is nothing to render yet. You will generally get your highest FPS in the region during this time.

So, in the first several seconds of arrival with a value this high, most users are sending more updates than the "server tick rate" while they are also trying to send other information to the sim, like outfit. If the sim is running a Last In First Out (LIFO) methodology which is typical of network services trying to keep near real-time, then information sent early will go stale and purged before being processed if the queue is receiving updates faster than the rate of processing.

Meanwhile, users able to handle FPS higher than the tick rate once the scene is loaded will be constantly sending more updates than the tick rate and possibly causing the same thing for themselves or others depending on how the sim queuing is setup.

In any case, we should not be trying to send more updates to the server than the processing rate. If it's 45, then the position updates should be below this number. 

I suggested the 20 number because I thought the tick rate was Beq's 25 number and a 20% margin seems ample, but to be completely honest these numbers should only be determined by someone familiar with the back-end code.

As far as #2, there's been no detailed explanation of the issue on the sim that could have caused it and if they are making code adjustments trying to handle the increase in position updates while not increasing the tick rate, then we're just going to be back here again later when all these FPS optimizations are done and more people are able to handle sustained updates faster than the processing rate.

KrsityKu:
https://wiki.secondlife.com/w/index.php?title=Viewerhelp:Statistics

> Sim FPS - The simulator frame rate. This should now always be the same as the physics frame rate -- 45.0 when things are running well

TommyTheTerrible:
> https://wiki.secondlife.com/w/index.php?title=Viewerhelp:Statistics
> 
> > Sim FPS - The simulator frame rate. This should now always be the same as the physics frame rate -- 45.0 when things are running well

This seems to suggest 45 is the maximum ("when things are running well") but not the guaranteed value. I wonder if Beq's 25 is the guaranteed minimum?

TommyTheTerrible:
I went through git blame to triage better and I think Beq's selection of 125 was to combat changes to the position input logic made in this commit: https://github.com/secondlife/viewer/commit/a67dde1f16c144247cff0b765201e64f8f8e75c2

The increase to 125 forced a bypass of the new slowdown logic in the commit.

This number was copied over to new logic in this commit: https://github.com/secondlife/viewer/commit/da87e8bd370ea079576f8b412a4ddb80c0715bd1

So, it seems that there was an attempt at reducing the position updates to the sim by @akleshchev, which caused complaints and @beqjanus increased the value to bypass the reduction. Without understanding how that value was derived, @AndrewMeadows  copied the value over to new logic that uses it for a latency bound and allowing position updates to be send at 125/45 (2.77) times more than the server tick.

This value definitely needs to be reduced. I will update the suggested change to 45 to match the maximum server tick.

TommyTheTerrible:
To appease the concerns that @KrsityKu brought up, I compiled 7.1.11 with various values and SLV Head as well to test Alistar Snook's test script (in the Firestorm Jira link provided) to see how each value performed.

Firestorm 7.1.9  @ 10 	= 275 +/- 1
Firestorm 7.1.11 @ 20 	= 326 +/- 4
Firestorm 7.1.11 @ 45 	= 419 +/- 8
Firestorm 7.1.11 @ 125 	= 450 +/- 0 (max)

No camera movement:

Second Life Develop @ 20 = 230 +/- 0
Second Life Develop @ 45 = 230 +/- 0
Second Life Develop @ 90 = 230 +/- 0
Second Life Develop @ 125 = 230 +/- 0

With camera movement:

Second Life Develop @ 20 = 317 +/- 1
Second Life Develop @ 45 = 428 +/- 3
Second Life Develop @ 90 = 422 +/- 12
Second Life Develop @ 125 = 406 +/- 6

Because @AndrewMeadows latest code is limiting input position updates to camera changes, I kept getting 230 until I started moving forward with the left click to test. Please ignore the variance on the 90 and 125 with camera movment, I started there with the tests and I gradually got better pathing to keep the camera moving so those values are a skill issue (avoiding obstacles in my skybox) and not a code issue.

The latest logic seems to work fine at 45 but 20 would not have been enough, so I appreciate KrsityKu challenging.

PR #2859, opened 2024-10-15
Reintroduce headless osmesa support to llapputil

RyeMutt:
* Reintroduces support for the llwindowmesaheadless backend to avoid the need for an X server in utilities
* Restores RelWithDebInfo linux build configs
* Fixes build-variables not applying flags in correct order resulting in some being discarded
* Fixes llapputil build on non-linux platforms


rohvani (approved):
LGTM, I've gone ahead and cherry-picked these commits into https://github.com/secondlife/viewer/pull/2353 due to the branch divergence.

PR #2864, opened 2024-10-15
Provide affordances to disable automatic downrezing of textures when SL is in background

brad-linden:
secondlife/viewer#2549


brad-linden:
Proposed notification text:
> To improve system performance, [SECOND_LIFE] has reduced texture memory usage after being in the background for [DELAY] seconds.  It may take some time for texture image quality to return to normal.

image

brad-linden:
additionally I presume most people will not be interested in seeing this beyond the first time, is there a way to default that `Don't show me this again` checkbox to on?

Ansariel:
Shouldn't this setting be exposed in preferences?

Also: It would probably be great to have an option to downres textures when in background or alternatively when the window is minimized. Latter option actually makes more sense because switching to the webbrowser for a moment shouldn't be a common situation where a downres is required. But if you need the resources, you could minimize the viewer and downres textures to free VRAM.

brad-linden:
> Shouldn't this setting be exposed in preferences?

we could possibly also do that, but for most people that would still be difficult to find, and the preferences floater is already extremely cluttered.  I am hesitant to make that situation worse.  I wanted to get a partial solution started while that design conversation can continue.

> Also: It would probably be great to have an option to downres textures when in background or alternatively when the window is minimized. Latter option actually makes more sense because switching to the webbrowser for a moment shouldn't be a common situation where a downres is required. But if you need the resources, you could minimize the viewer and downres textures to free VRAM.

we are currently handling both of those cases together here. are you suggesting it would be better to add separate configuration for these?  I am really trying to avoid an unnecessary proliferation of configuration complexity here.  The goal is to get a pretty good solution that will generally work well for most people, so now it will default to keeping full res for 60 seconds.  People who want even more control can go straight to the debug setting `TextureDiscardBackgroundedTime`

Ansariel:
> we are currently handling both of those cases together here. are you suggesting it would be better to add separate configuration for these? I am really trying to avoid an unnecessary proliferation of configuration complexity here. The goal is to get a pretty good solution that will generally work well for most people, so now it will default to keeping full res for 60 seconds. People who want even more control can go straight to the debug setting `TextureDiscardBackgroundedTime`

I personally - and what I am most likely doing in Firestorm - would make this super flexible by making two settings:
* The first setting which allows downres of textures if the viewer is in the background after a configurable amount of time
* The second setting would enable an immediate downres if the viewer gets minimized

I like the second setting because it gives direct control of when a downres happens and I can immediately trigger it if needed.

Re preferences: If you expose the setting in preferences, it can be saved in a preset. For example if you have a preset "Gaming" which turns down all kind of settings to free up resources while you want to play a different game and keep the viewer running in the background, you can enable texture downres basically on demand - and keep it disabled for normal cases where you switch to the webbrowser or do other things while logged into SL.

indra/newview/llviewertexture.cpp
```diff
@@ -560,16 +573,21 @@ void LLViewerTexture::updateClass()
...
+    static F32 last_desired_discard_bias = 1.f;
+    static LLCachedControl<F32> backgrounded_discard_time(gSavedSettings, "TextureDiscardBackgroundedTime", 60.f);
 
     bool in_background = (gViewerWindow && !gViewerWindow->getWindow()->getVisible()) || !gFocusMgr.getAppHasFocus();
 
     if (in_background)
     {
-        if (backgrounded_timer.getElapsedTimeF32() > 10.f)
+        F32 background_elapsed = backgrounded_timer.getElapsedTimeF32();
+        if (backgrounded_discard_time > 0.f && background_elapsed > backgrounded_discard_time)
```
akleshchev, line 584: Wouldn't this show message after each return from background?

akleshchev (approved):

Quinn-Elara:
> * The first setting which allows downres of textures if the viewer is in the background after a configurable amount of time
> 
> * The second setting would enable an immediate downres if the viewer gets minimized

Seconding this idea - sending the viewer into the background (for example, clicking an SLURL) vs. fully minimizing the viewer (when needing to do something else for a while) should be handled separately.

maxim-productengine (approved):

kylelinden:
This is a good starting point, we can improve on it later or accept a contribution.

brad-linden (commented):
included improvements from rye

marchcat (approved):

PR #2865, opened 2024-10-16
Remove dependency on libwayland-dev

AiraYumi:
Fixes the issue that booting with wayland will not work unless libwayland-dev is installed.


akleshchev (approved):

marchcat (approved):

marchcat:
Thanks!

Issue #2572, opened 2024-09-13
Voice dot appears next to avatar's name in Conversations floater when Voice Chat was disabled prior to login


maestrolinden:
### Environment

Second Life Test 7.1.11.10800977053 (64bit)
Release Notes

CPU: Apple M1 Pro (2400 MHz)
Memory: 16384 MB
OS Version: macOS 14.6.1 Darwin 23.6.0 Darwin Kernel Version 23.6.0: Mon Jul 29 21:14:30 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6000 x86_64
Graphics Card Vendor: Apple
Graphics Card: Apple M1 Pro

OpenGL Version: 4.1 Metal - 88.1

Window size: 1225x867
Font Size Adjustment: 96pt
UI Scaling: 0.75
Draw distance: 96m
Bandwidth: 10000kbit/s
LOD factor: 1.125
Render quality: 1
Texture memory: 10922MB
Disk cache: Max size 1638.4 MB (100.0% used)
HiDPI display mode: 

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091638
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version:  

September 13 2024 11:25:24

### Description

If "Voice Chat" is disabled prior to logging in, the Conversations panel will incorrectly show a voice dot next to the user's own name, even though voice is not connected.
![image (1)](https://github.com/user-attachments/assets/a31e56df-3f31-4ecf-a349-07be232b0bba)

I don't see this bug if "Voice Chat" is disabled after login time - all voice dots and indicators disappear as expected in that case.

### Reproduction steps

1. Launch the viewer
2. Disable voice chat: Open Preferences -> Sound & Media, and uncheck "Enabled" next to "Voice Chat"
3. Login to any region (reproduces with both webRTC and Vivox-mode regions)
4. Open the Conversations floater, and look for voice dots

Expected results:
No voice dots should appear anywhere, since Voice Chat is disabled.

Actual results:
A voice dot appears next to your avatar's name in the Conversations floater.  No voice dot appears above the avatar's head, however.

LLGuru:
[...]

PR #2871, opened 2024-10-16
#2572 Voice dot appears next to avatar's name in Conversations floater

LLGuru:


marchcat (approved):

akleshchev (approved):

indra/newview/llconversationview.cpp
```diff
@@ -578,7 +578,12 @@ participant_id(),
 avatar_icon("avatar_icon"),
 info_button("info_button"),
 output_monitor("output_monitor")
-{}
+{
+    if (!output_monitor.visible.isProvided())
+    {
+        output_monitor.visible = false;
+    }
+}
```
akleshchev, line 586: This is a wrong way to set default value for params

akleshchev: Apparently output_monitor("output_monitor", LLOutputMonitorCtrl::Params().with(...)), but since it's apparently internal element and not a simple value, it should be defined in xml.

LLGuru: Reworked

PR #2882, opened 2024-10-17
Add the latest viewer stats packet into Frame Profile dump file.

nat-goodspeed:


indra/newview/llviewerstats.cpp
```diff
@@ -488,6 +493,11 @@ void update_statistics()
     }
 }
 
+LLSD get_viewer_stats()
```
vir-linden, line 496: Do we want to return this by value?

vir-linden (approved):
Looks good!

PR #2885, opened 2024-10-17
viewer#2884 Convert webrtc and experience coroutines into idle callbacks

akleshchev:
This doesn't fix the problem, just makes it ten times rarer (vivox is still a looping coroutine), but I hope it will be good enough untill we get some tracy coverage for coroutines. It also brings these functions back to main thread making them properly observable.

Covers:
- WebRTC
- Experience cache

### DIFF
--- a/indra/newview/llvoicewebrtc.cpp
+++ b/indra/newview/llvoicewebrtc.cpp
@@ -302,6 +302,10 @@ void LLWebRTCVoiceClient::stopTimer()
 {
     if (mIsTimerActive)
     {
+        LLMuteList::instanceExists();
+        {
+            LLMuteList::getInstance()->removeObserver(this);
+        }
         mIsTimerActive = false;
         LL::Timers::instance().cancel(mVoiceTimerHandle);
     }
@@ -470,13 +474,6 @@ void LLWebRTCVoiceClient::connectionTimer()
     LL_PROFILE_ZONE_SCOPED_CATEGORY_VOICE;
     try
     {
-        // TODO: Doing some measurement and calculation here,
-        // we could reduce the timeout to take into account the
-        // time spent on the previous loop to have the loop
-        // cycle at exactly 100ms, instead of 100ms + loop
-        // execution time.
-        // Could help with voice updates making for smoother
-        // voice when we're busy.
         bool voiceEnabled = mVoiceEnabled;
 
         if (!isAgentAvatarValid())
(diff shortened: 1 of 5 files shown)


indra/newview/llvoicewebrtc.cpp
```diff
@@ -456,118 +465,109 @@ void LLWebRTCVoiceClient::removeObserver(LLFriendObserver *observer)
...
-            // cycle at exactly 100ms, instead of 100ms + loop
-            // execution time.
-            // Could help with voice updates making for smoother
-            // voice when we're busy.
-            llcoro::suspendUntilTimeout(UPDATE_THROTTLE_SECONDS);
-            if (sShuttingDown) return; // 'this' migh already be invalid
-            bool voiceEnabled = mVoiceEnabled;
-
-            if (!isAgentAvatarValid())
+        // TODO: Doing some measurement and calculation here,
```
roxanneskelly, line 473: I don't think this TODO comment applies anymore, as the timer updates every 100ms.

indra/newview/llvoicewebrtc.cpp
```diff
@@ -1569,10 +1572,13 @@ void LLWebRTCVoiceClient::setVoiceEnabled(bool enabled)
...
             updatePosition();
-            if (!mIsCoroutineActive)
+            if (!mIsTimerActive)
             {
-                LLCoros::instance().launch("LLWebRTCVoiceClient::voiceConnectionCoro",
-                    boost::bind(&LLWebRTCVoiceClient::voiceConnectionCoro, LLWebRTCVoiceClient::getInstance()));
+                LL_DEBUGS("Voice") << "Starting" << LL_ENDL;
+                mIsTimerActive = true;
+                LLMuteList::getInstance()->addObserver(this);
+                const F32 SECS_BETWEEN_REQUESTS = 0.5f;
```
roxanneskelly, line 1580: where is this used?

roxanneskelly (approved):
This is great stuff.  Thanks.

indra/newview/llvoicewebrtc.cpp
akleshchev: Nowhere, I'm surprised build didn't fail. removing. Also forgot to add LLMuteList::getInstance()->removeObserver(). No idea why there was no such call before.

indra/newview/llvoicewebrtc.cpp
akleshchev: Cleaned up.

Issue #2870, opened 2024-10-16
Keypresses to text input fields are duplicated


maestrolinden:
### Environment

Second Life Test 7.1.12.11370694217 (64bit)
Release Notes

You are at 36.7, 134.1, 21.3 in Tustin located at simhost-0d8e20bef03d618f9.agni
SLURL: http://maps.secondlife.com/secondlife/Tustin/37/134/21
(global coordinates 244773.0, 328326.0, 21.3)
Second Life RC LeTigre 2024-09-13.10853867644
Release Notes

CPU: Apple M1 Pro (2400 MHz)
Memory: 16384 MB
OS Version: macOS 14.7.0 Darwin 23.6.0 Darwin Kernel Version 23.6.0: Wed Jul 31 20:49:39 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T6000 x86_64
Graphics Card Vendor: Apple
Graphics Card: Apple M1 Pro

OpenGL Version: 4.1 Metal - 88.1

Window size: 1072x831
Font Size Adjustment: 96pt
UI Scaling: 0.8
Draw distance: 128m
Bandwidth: 10000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 10922MB
Disk cache: Max size 1638.4 MB (87.9% used)
HiDPI display mode: 

RestrainedLove API: (disabled)
J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091638
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Not Connected
Packets Lost: 6/6862 (0.1%)
October 16 2024 13:24:55

### Description

When I type text into any input field in a [recent build](https://github.com/secondlife/viewer/actions/runs/11370694217?pr=2668) of the geenz/feat/legacy-alpha-params branch, which @Geenz has just pulled develop from, characters are randomly duplicated, like so:

> HHii evveerybbooddyy!   I'mm  tyyping  eexxttrra  tooddaayy!

[...]

PR #2890, opened 2024-10-18
temporary workaround for #2870: disable SDL2 on mac

AndrewMeadows:
**SDL2** has a builtin pipeline that uses Cocoa’s NSTextInputClient protocol, and the input submitted during the `SDL_PollEvent()` call.  Meanwhile we have our own hookup to NSTextInputClient, hence the Mac viewer now gets two events for almost every keystroke.

The proper solution probably requires us to hack our SDL2 on Mac to remove the automatic text input of that library, however in the meantime we offer this **workaround**: disable SDL2 for DARWIN builds.  This means we won't have access to game controllers through SDL2, however that isn't necessarily terrible because AFAICT most game controllers don't work on the Mac operating system: only a small handful of game controllers are supported because "Apple knows best".

### DIFF
--- a/indra/llwindow/llgamecontrol.cpp
+++ b/indra/llwindow/llgamecontrol.cpp
@@ -1530,6 +1530,9 @@ void LLGameControl::init(const std::string& gamecontrollerdb_path,
     llassert(saveObject);
     llassert(updateUI);
 
+#ifndef LL_DARWIN
+    // SDL2 is temporarily disabled on Mac, so this needs to be a no-op on that platform
+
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

marchcat (approved):

brad-linden:
I'll repeat my comment here from secondlife/viewer#2891:  we probably in the long run will want a proper solution of using SDL for all input and windowing on mac rather than hacking in some custom functionality into SDL

PR #2897, opened 2024-10-19
viewer#2741 Don't reallocate buffer all the time

akleshchev:
It doesn't look like scratch needs to be reallocated all the time, it gets deleted immediately after use and glTexImage2D copies content to target.

I have no idea how high fonts can go, but I assume 4096x4096 is more than enough.


aiaustin (commented):
Spelling: CONVERTION -> CONVERSION

RyeMutt (approved):

Issue #2245, opened 2024-08-09
Allow Lua scripts to be run from the Conversations floater


AtlasLinden:
Allow a Lua script to be executed via the Conversations floater.

For instance, typing `/test.lua` will execute the Lua script with that name.

(Something to consider is what if a user has named a gesture test.lua? It's possible!)

maxim-productengine:
Nearby chat field (Conversation floater) can be used now to run scripts, for example `/test_luafloater_speedometer.lua`
By default, it run the script from `scripts/lua` (which is specified by debug setting `LuaCommandPath`).

PR #2909, opened 2024-10-22
#2245 Allow Lua scripts to be run from the Conversations floater

maxim-productengine:


indra/newview/llfloaterimnearbychat.cpp
```diff
@@ -606,6 +608,13 @@ void LLFloaterIMNearbyChat::sendChat( EChatType type )
                 if(!LLGestureMgr::instance().triggerAndReviseString(utf8text, &utf8_revised_text))
                 {
                     utf8_revised_text = utf8text;
+                    // check if the message is /filename.lua and execute the Lua script
+                    static const boost::regex is_lua_script("^/.*\\.luau?$");
```
nat-goodspeed, line 612: I recommend using `ll_regex_match()` from llregex.h: `boost::regex_match()` can throw. `ll_regex_match()` defends against that case.

Also, now that `runScriptFile()` can pass command-line arguments to the subject Lua script, I would match even if there's more text beyond the `.lua` extension.

nat-goodspeed (changes requested):

indra/newview/llfloaterimnearbychat.cpp
maxim-productengine: Thanks. You're right about passing arguments.

indra/newview/llfloaterimnearbychat.cpp
```diff
@@ -606,6 +608,13 @@ void LLFloaterIMNearbyChat::sendChat( EChatType type )
                 if(!LLGestureMgr::instance().triggerAndReviseString(utf8text, &utf8_revised_text))
                 {
                     utf8_revised_text = utf8text;
+                    // check if the message is /filename.lua and execute the Lua script
+                    static const boost::regex is_lua_script("^/.*\\.luau?(?:\\s+\\S+)*$");
```
nat-goodspeed, line 612: Wouldn't it be equivalent (and more readable) just to end the pattern string at `"...luau?"`

nat-goodspeed (approved):

indra/newview/llfloaterimnearbychat.cpp
maxim-productengine: Currently it matches strings like `/filename.lua` or `/filename.lua arg1 arg2`. Changing pattern to "^/.*\\.luau?" will allow matches like `/filename.luaqwerty123'.

maxim-productengine: Changing pattern to `^/.*\\.luau?` won't allow matches like  `/filename.lua arg1 arg2`.

nat-goodspeed: Even if you omit the final `$`? I thought that would match any string prefix.

You're right about trailing characters though. You could use `"^/.*\\.luau?\\b"` ?

Issue #2910, opened 2024-10-22
Allow terminating multiple scripts at once


maxim-productengine:
Currently you can only highlight one script at a time in 'Lua scripts' floater.
So, allow selecting and terminating multiple scripts at once.

PR #2919, opened 2024-10-23
#2910 Allow terminating multiple scripts at once

maxim-productengine:


akleshchev (approved):

maxim-productengine:
Trimmed trailing whitespaces.

indra/llui/llscrolllistctrl.cpp
```diff
@@ -423,6 +423,19 @@ std::vector<LLScrollListItem*> LLScrollListCtrl::getAllSelected() const
     return ret;
 }
 
+std::vector<LLSD> LLScrollListCtrl::getAllSelectedValues() const
```
nat-goodspeed, line 426: I like introducing `getAllSelectedValues()`; I'm a little surprised that didn't already exist. But I do wonder why you return a `std::vector<LLSD>` instead of an LLSD array?

PR #2946, opened 2024-10-25
Small fix in LLTeleportHandler::handle()

LLGuru:


akleshchev:
Does this have a ticket, what exactly does this fix and how to repro this problem?

marchcat:
An "Issue" type ticket may be optional if the PR is sufficiently documented.
However, I can't determine what this change does based on the code alone either.

marchcat (approved):
I see now; the new condition ensures that the code won't attempt to access tokens that may not exist, potentially preventing runtime errors like out-of-bounds access.

akleshchev:
> the new condition ensures that the code won't attempt to access tokens that may not exist

No, it does not, it will still attempt to access token 3 even if it does not exist and there are only 3 tokens (0, 1 and 2). But given that this is LLSD it probably won't crash.

marchcat:
So something like `size > 2 && size < 4` should be correct then?

akleshchev:
Probably, but in cases like this, probably best to check QA's SLURL test plan

PR #2959, opened 2024-10-25
Fix for IM session name and icon.

Rider-Linden:
Adds optional parameters to addMessage which allow the function to override the icon and session name used for the chat while correctly naming and constructing the chat session.

### DIFF
--- a/indra/newview/llimview.cpp
+++ b/indra/newview/llimview.cpp
@@ -3143,9 +3143,16 @@ void LLIMMgr::addMessage(
     const LLUUID& region_id,
     const LLVector3& position,
     bool is_region_msg,
-    U32 timestamp) // May be zero
+    U32 timestamp,  // May be zero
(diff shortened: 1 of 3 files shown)


AndrewMeadows (approved):

indra/newview/llimview.cpp
```diff
@@ -3241,7 +3248,7 @@ void LLIMMgr::addMessage(
             }
 
             //Play sound for new conversations
-            if (!skip_message & !gAgent.isDoNotDisturb() && (gSavedSettings.getBOOL("PlaySoundNewConversation")))
+            if (!skip_message && !gAgent.isDoNotDisturb() && (gSavedSettings.getBOOL("PlaySoundNewConversation")))
```
akleshchev, line 3251: IMs aren't particularly frequent, but probably better to use LLCachedControl here.

PR #2971, opened 2024-10-28
secondlife/viewer#2768: Partial fix for PBR texture animations stopping. May also fix some Blinn-Phong texture animations.

cosmic-linden:
This is a partial fix to https://github.com/secondlife/viewer/issues/2768 . The fix appears to prevent PBR texture animations from sometimes stopping when looking away and then looking back. However, there is still an issue with PBR texture animations not starting.

### DIFF
--- a/indra/newview/llvovolume.cpp
+++ b/indra/newview/llvovolume.cpp
@@ -786,7 +786,24 @@ void LLVOVolume::updateTextureVirtualSize(bool forced)
         LLFace* face = mDrawable->getFace(i);
         if (!face) continue;
         const LLTextureEntry *te = face->getTextureEntry();
-        LLViewerTexture *imagep = face->getTexture();
+        LLViewerTexture *imagep = nullptr;
+        U32 ch_min;
+        U32 ch_max;
+        if (!te->getGLTFRenderMaterial())
+        {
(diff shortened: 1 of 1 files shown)


indra/newview/llvovolume.cpp
```diff
@@ -786,7 +786,24 @@ void LLVOVolume::updateTextureVirtualSize(bool forced)
...
+        else
+        {
+            ch_min = LLRender::BASECOLOR_MAP;
+            ch_max = LLRender::EMISSIVE_MAP;
+        }
+        for (U32 ch = ch_min; (!imagep && ch <= ch_max); ++ch)
+        {
+            // Get _a_ non-null texture if possible (usually diffuse/basecolor, but could be something else)
+            imagep = face->getTexture(ch);
+        }
```
cosmic-linden, line 806: The virtual size was calculated to be 0, which caused an early exit in `LLVOVolume::animateTextures`.

cosmic-linden, line 806: The virtual size was stuck at 0 in some cases due to the `continue` below, which caused an early exit in `LLVOVolume::animateTextures`.

RunitaiLinden (approved):

PR #2982, opened 2024-10-31
Give control over thread queue's automatic shutdown

akleshchev:
queue can shutdown before parent thread is done

### DIFF
--- a/indra/llcommon/threadpool.h
+++ b/indra/llcommon/threadpool.h
@@ -124,7 +124,7 @@ namespace LL
                         size_t threads=1,
                         size_t capacity=1024*1024,
                         bool auto_shutdown = true):
-            ThreadPoolBase(name, threads, new queue_t(name, capacity), auto_shutdown)
+            ThreadPoolBase(name, threads, new queue_t(name, capacity, false), auto_shutdown)
         {}
         ~ThreadPoolUsing() override {}
 
--- a/indra/llcommon/workqueue.cpp
+++ b/indra/llcommon/workqueue.cpp
@@ -29,14 +29,17 @@ using Lock  = LLCoros::LockType;
 /*****************************************************************************
 *   WorkQueueBase
 *****************************************************************************/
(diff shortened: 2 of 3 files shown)


marchcat (approved):

nat-goodspeed (commented):
My question is: now that `WorkQueueBase` has `auto_shutdown`, why don't we remove that feature from `ThreadPoolBase`? Put differently, it seems to me the `auto_shutdown` flag passed to `ThreadPoolUsing` and its alias `ThreadPool` could _just_ pass through to `queue_t`, instead of engaging redundant support in `ThreadPoolBase`'s constructor.

akleshchev:
>My question is: now that WorkQueueBase has auto_shutdown, why don't we remove that feature from ThreadPoolBase?

@nat-goodspeed Are you suggesting to remove mStopListener from ThreadPoolBase?

nat-goodspeed:
Hmm, now that I'm looking at `ThreadPoolBase::close()`, I see that it does more than call `WorkQueueBase::close()`: it also joins the `ThreadPoolBase` threads. Sorry, my previous suggestion was ill-considered.

So this PR is about the case when we have a `WorkQueue` or `WorkSchedule`, _not_ associated with a `ThreadPool`, that shouldn't `close()` when `LLCoros` signals the start of viewer shutdown? Maybe it should instead listen for a later value of `"status"` ?

akleshchev:
> Maybe it should instead listen for a later value of "status" ?

This is mostly for the sake of window thread which under normal circumstances needs to keep living long after event system dies.

nat-goodspeed:
Okay, I see. In that case I agree with your earlier suggestion.

> Arguably I should be calling new queue_t(name, capacity, false) to let thread decide when it's ready to die.

If both `ThreadPoolBase` and `WorkQueueBase` see `auto_shutdown=true`, they register redundant listeners.

akleshchev:
Updated with `new queue_t(name, capacity, false)`.

nat-goodspeed (approved):

PR #2990, opened 2024-10-31
More detailed frame time logging in ViewerStats

vir-linden:
Capture more details about frame times/fps while in foreground and background, including standard deviation.


indra/newview/llviewerstats.cpp
```diff
@@ -274,6 +274,16 @@ void LLViewerStats::updateFrameStats(const F64Seconds time_diff)
         // old stats that were never really used
         F64Seconds jit = (F64Seconds)std::fabs((mLastTimeDiff - time_diff));
         sample(LLStatViewer::FRAMETIME_JITTER, jit);
+
+        if (gFocusMgr.getAppHasFocus())
+        {
+            mForegroundFrameStats.push(F32(F64(time_diff)));
```
nat-goodspeed, line 280: The double casting looks odd, but I assume you tried it with simpler code.

nat-goodspeed (approved):

indra/llcommon/llstatsaccumulator.h
```diff
@@ -100,21 +113,47 @@ class LLStatsAccumulator
...
-    F32 mSumOfSquares;
+    F64 mSum;
+    F64 mSumOfSquares;
     F32 mMinValue;
     F32 mMaxValue;
     U32 mCountOfNextUpdatesToIgnore;
 };
 
+// Assumes the samples are frame times
+inline F32 fps(LLStatsAccumulator& accum)
```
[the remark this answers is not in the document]
vir-linden: These functions only make sense if the quantity being measured is frame times, so they don't quite fit in a generic accumulator. I could derive a new class from LLStatsAccumulator and add the methods there, but hate to promulgate a whole class hierarchy just to fit in a couple of small functions. If I add histogram support at some point might add them then.

indra/newview/llviewerstats.cpp
vir-linden: Thought I had but it turns out the simpler formulation still works.

PR #2995, opened 2024-11-01
viewer#2985 Odd crash when notifying voice observers

akleshchev:
Crash points at enableDisableCallBtn, but parent class (LLFloaterIMSession) was cleaned earlier and has a removeObserver call so an observer record should be long gone. Likely something else is going on.

Went over various callbacks and made sure they are cleaned. But in case LLFloaterIMSession somehow remained, added mSession = NULL on a session removal callback.

### DIFF
--- a/indra/newview/llfloaterimsession.cpp
+++ b/indra/newview/llfloaterimsession.cpp
@@ -638,6 +638,7 @@ void LLFloaterIMSession::onClose(bool app_quitting)
     // Last change:
     // EXT-3516 X Button should end IM session, _ button should hide
     gIMMgr->leaveSession(mSessionID);
+    mSession = nullptr; // leaveSession should have deleted it.
     // *TODO: Study why we need to restore the floater before we close it.
     // Might be because we want to save some state data in some clean open state.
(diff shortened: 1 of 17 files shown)


indra/newview/llfloaterimsession.cpp
```diff
@@ -638,6 +638,7 @@ void LLFloaterIMSession::onClose(bool app_quitting)
     // Last change:
     // EXT-3516 X Button should end IM session, _ button should hide
     gIMMgr->leaveSession(mSessionID);
+    mSession = NULL; // leaveSession should have deleted it.
```
marchcat, line 641: `nullptr`?

marchcat (approved):

indra/newview/llfloaterimsession.cpp
akleshchev: Fixed, thanks. Did NULL out of habit.

roxanneskelly (approved):

maxim-productengine (approved):

akleshchev:
Moved bugsplat and shader logging into separate commit.

PR #2944, opened 2024-10-24
Post C++20 improvements

Ansariel:
* Make eligible inline functions in llmath.h constexpr (held back this change to avoid conflict with std::lerp change)
* Replace boost reverse foreach with range view adoptor

@nat-goodspeed C++20 now contains several mathematical constants that are currently manually defined in the viewer in llmath.h. What is your thought about replacing the manual specified values with the ones from https://en.cppreference.com/w/cpp/numeric/constants ?

PR #2996, opened 2024-11-01
Check if it can be built with the ubuntu 24.04

AiraYumi:


marchcat:
Started the build, let's see.

AiraYumi:
Oops. It seems ubuntu-latest is still 22.04.
If you want to build it on 24.04, you need to specify it.

AiraYumi:
It still doesn't seem to build on 24.04.
Let's hope #2944 improves the situation.

By the way, Check CLA is probably broken right now.

PR #3006, opened 2024-11-04
[mac] remove unused LSRequiresCarbon key

larsnaesbye:
This key has been pointless since support for pre-macOS X was dropped.


marchcat (approved):
Thank you!

PR #3016, opened 2024-11-05
Tracy Improvements

RyeMutt:
* Bump GHA macos build agent to 15 to fix invariant timer issues
* Fix tracy memory profiling integration on non-windows
* Fix up tracy gpu profiling integration
* Update Tracy to 0.11.1 with fixed macos and linux tools

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -47,7 +47,7 @@ jobs:
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        runner: [windows-large, macos-14-xlarge]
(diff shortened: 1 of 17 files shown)


.github/workflows/build.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        runner: [windows-large, macos-14-xlarge]
+        runner: [windows-large, macos-15-xlarge]
```
marchcat, line 50: Thanks! That was the initial plan for the xcode16 update, but there were no available runners at the time.

marchcat (approved):

PR #3026, opened 2024-11-06
secondlife/viewer#2768: Fix for PBR texture animations sometimes not starting 

cosmic-linden:
Follow-up to https://github.com/secondlife/viewer/pull/2971 .

This fixes another case where PBR texture animations did not play, if the PBR base material does not contain textures but the GLTF material overrides do. The GLTF material overrides were arriving sufficiently late such that the texture virtual size was calculated to be 0, which prevented the texture matrix from being copied over to the `LLDrawInfo`. Since the contents of the overrides cannot be predicted in advance, a rebuild is required.

### DIFF
--- a/indra/newview/llviewermenu.cpp
+++ b/indra/newview/llviewermenu.cpp
@@ -9897,7 +9897,6 @@ void initialize_menus()
     view_listener_t::addMenu(new LLAdvancedClickGLTFEdit(), "Advanced.ClickGLTFEdit", cb_info::UNTRUSTED_BLOCK);
     view_listener_t::addMenu(new LLAdvancedClickResizeWindow(), "Advanced.ClickResizeWindow", cb_info::UNTRUSTED_BLOCK);
     view_listener_t::addMenu(new LLAdvancedPurgeShaderCache(), "Advanced.ClearShaderCache", cb_info::UNTRUSTED_BLOCK);
-    view_listener_t::addMenu(new LLAdvancedRebuildTerrain(), "Advanced.RebuildTerrain", cb_info::UNTRUSTED_BLOCK);
 
     #ifdef TOGGLE_HACKED_GODLIKE_VIEWER
     view_listener_t::addMenu(new LLAdvancedHandleToggleHackedGodmode(), "Advanced.HandleToggleHackedGodmode");
@@ -9915,10 +9914,10 @@ void initialize_menus()
     view_listener_t::addMenu(new LLAdvancedResetInterestLists(), "Advanced.ResetInterestLists");
 
     // Develop > Terrain
-    view_listener_t::addMenu(new LLAdvancedRebuildTerrain(), "Advanced.RebuildTerrain");
-    view_listener_t::addMenu(new LLAdvancedTerrainCreateLocalPaintMap(), "Advanced.TerrainCreateLocalPaintMap");
-    view_listener_t::addMenu(new LLAdvancedTerrainEditLocalPaintMap(), "Advanced.TerrainEditLocalPaintMap");
-    view_listener_t::addMenu(new LLAdvancedTerrainDeleteLocalPaintMap(), "Advanced.TerrainDeleteLocalPaintMap");
+    view_listener_t::addMenu(new LLAdvancedRebuildTerrain(), "Advanced.RebuildTerrain", cb_info::UNTRUSTED_BLOCK);
+    view_listener_t::addMenu(new LLAdvancedTerrainCreateLocalPaintMap(), "Advanced.TerrainCreateLocalPaintMap", cb_info::UNTRUSTED_BLOCK);
+    view_listener_t::addMenu(new LLAdvancedTerrainEditLocalPaintMap(), "Advanced.TerrainEditLocalPaintMap", cb_info::UNTRUSTED_BLOCK);
+    view_listener_t::addMenu(new LLAdvancedTerrainDeleteLocalPaintMap(), "Advanced.TerrainDeleteLocalPaintMap", cb_info::UNTRUSTED_BLOCK);
 
     // Advanced > UI
     registrar.add("Advanced.WebBrowserTest", boost::bind(&handle_web_browser_test, _2));  // sigh! this one opens the MEDIA browser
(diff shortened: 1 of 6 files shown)


indra/newview/llvovolume.cpp
```diff
@@ -832,12 +841,25 @@ void LLVOVolume::updateTextureVirtualSize(bool forced)
...
+                // dirtyGeom+markRebuild tells the engine to call
+                // LLVolumeGeometryManager::rebuildGeom, which rebuilds the
+                // LLDrawInfo for the spatial group containing this LLFace,
+                // safely copying the mTextureMatrix from the LLFace the the
+                // LLDrawInfo. While it's not ideal to call it here, prims with
+                // animated faces get moved to a smaller partition to reduce
+                // side-effects of their updates (see shrinkWrap in
+                // LLVOVolume::animateTextures).
+                mDrawable->getSpatialGroup()->dirtyGeom();
+                gPipeline.markRebuild(mDrawable->getSpatialGroup());
```
cosmic-linden, line 862: Main fix is here

indra/newview/llvovolume.cpp
```diff
@@ -641,6 +641,15 @@ void LLVOVolume::animateTextures()
...
                     facep->mTextureMatrix = new LLMatrix4();
+                    if (facep->getVirtualSize() > MIN_TEX_ANIM_SIZE)
+                    {
+                        // Fix the one edge case missed in
+                        // LLVOVolume::updateTextureVirtualSize when the
+                        // mTextureMatrix is not yet present
+                        gPipeline.markRebuild(mDrawable, LLDrawable::REBUILD_TCOORD);
+                        mDrawable->getSpatialGroup()->dirtyGeom();
+                        gPipeline.markRebuild(mDrawable->getSpatialGroup());
+                    }
```
cosmic-linden, line 652: This addresses a rare theoretical edge case not handled by the fix in `updateTextureVirtualSize`

indra/newview/llviewermenu.cpp
```diff
@@ -9915,10 +9914,10 @@ void initialize_menus()
...
 
     // Develop > Terrain
-    view_listener_t::addMenu(new LLAdvancedRebuildTerrain(), "Advanced.RebuildTerrain");
-    view_listener_t::addMenu(new LLAdvancedTerrainCreateLocalPaintMap(), "Advanced.TerrainCreateLocalPaintMap");
-    view_listener_t::addMenu(new LLAdvancedTerrainEditLocalPaintMap(), "Advanced.TerrainEditLocalPaintMap");
-    view_listener_t::addMenu(new LLAdvancedTerrainDeleteLocalPaintMap(), "Advanced.TerrainDeleteLocalPaintMap");
+    view_listener_t::addMenu(new LLAdvancedRebuildTerrain(), "Advanced.RebuildTerrain", cb_info::UNTRUSTED_BLOCK);
+    view_listener_t::addMenu(new LLAdvancedTerrainCreateLocalPaintMap(), "Advanced.TerrainCreateLocalPaintMap", cb_info::UNTRUSTED_BLOCK);
+    view_listener_t::addMenu(new LLAdvancedTerrainEditLocalPaintMap(), "Advanced.TerrainEditLocalPaintMap", cb_info::UNTRUSTED_BLOCK);
+    view_listener_t::addMenu(new LLAdvancedTerrainDeleteLocalPaintMap(), "Advanced.TerrainDeleteLocalPaintMap", cb_info::UNTRUSTED_BLOCK);
```
cosmic-linden, line 9920: There seemed to be a merge error here, so this is my best guess of what the code should look like. In any case, these are developer-only testing options, plus terrain painting is on the shelf, so this functionality is not critical.

indra/newview/llvovolume.cpp
TommyTheTerrible: Is this logic going to work? If virtual size is below that threshold, then the matrix will be filled and it might be possible the rebuild won't happen later since the matrix would not be null.

What's the harm in removing the virtual size qualification?

cosmic-linden: > What's the harm in removing the virtual size qualification?

The virtual size comparison logic has not changed, I've just cleaned it up.

For the rest of your questions, see also `LLVolumeGeometryManager::registerFace` for the logic that ultimately determines whether the texture matrix gets copied. The `dirtyGeom` and `markRebuild` lines I've added force that function to get recalled for faces impacted by this bug.

TommyTheTerrible: My repository is behind (downstream Firestorm), so this is the first time I have seen the added virtual size qualification and I apologize if it was added by someone else.

But I think the combination of the two qualifications (null check on the matrix, virtual size threshold) narrows the chances of a rebuild down too far. Which is why I was asking if the second was even necessary.

If the matrix is null and virtual size is outside the threshold, then the rebuild will not happen until some other outside logic forces a rebuild because the first qualification will never pass again.

LLVolumeGeometryManager::registerFace seems to only be called when the face is added to the spatial group, so when will the face be rebuilt once the virtual size threshold is met? When the LOD on the drawable changes? What if it has only one LOD level?

I think the bug will persist if the face is distant or small and enters into frustum from a previous cull. In this situation and if the second qualification is used, I think it will not start animating as the camera get closer and the virtual size grows larger than the threshold until a LOD level change initiates a rebuild.

I know this is possibly a narrow set of circumstances but I think it will be noticeable when it does happen and I do not see any harm in removing the virtual size qualification.

cosmic-linden: > If the matrix is null

This is explained in the comments, as well as the edge case fix that I also included in `LLVOVolume::animateTextures`

> LOD

I didn't change the LOD code

> I do not see any harm in removing the virtual size qualification

I already told you that code has not changed. Can you please confirm exactly which subset of the code diff you are referring to by copying and pasting it in your reply, and explain exactly how the code change maps to the issue you are describing?

RunitaiLinden (approved):

cosmic-linden:
Based on my testing, this is a net improvement to texture animation behavior which fixes an observed bug.

cosmic-linden:
Per our GitHub policies, the merging of this pull request locks the conversation. However, feedback is still welcome on these changes, provided that they are specific and actionable.

Issue #3029, opened 2024-11-06
Log spam when dragging a folder in Inventory


maxim-productengine:
### Environment

Current develop build on top of `6b448af1bbb80ee663d9cb9587b529a7e777074a` commit

### Description

Log spam when dragging a folder inside Inventory floater.

### Reproduction steps

1, Open Develop>Consoles>Debug console
2. Open Inventory and create a new folder
3. Select this folder and start dragging it to the closest folder, and then to itself
4. Observe log spam
`WARNING   LLControlGroup::get : Control DragAndDropToolTipDelay not found.`

katerynamproductengine:
Passed QA.

Verified on the Second Life Release 7.1.13.13816156139 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/464.

PR #3031, opened 2024-11-06
#3029 restore tooltip delay const variable

maxim-productengine:
Probably it was lost in merges, so restoring DRAG_N_DROP_TOOLTIP_DELAY.


marchcat (approved):
Thank you!

PR #3056, opened 2024-11-08
viewer#3055 Fix texture fetch being stuck

akleshchev:
destroyRawImage() is the only function that calls saveRawImage(), not calling it results in fetcher thinking that it still needs data to do the saving and looping back to trying to fetch.


marchcat (approved):

akleshchev:
Not sure if mNeedsAux even needs to be here (the threadless creation doesn't reset it) or be before or after saving. But 'before' the saving should be the safest.

Issue #3025, opened 2024-11-05
on_main_coro developer assert when clicking resident name on minimap


cosmic-linden:
### Environment

Developer-built version of Second Life viewer
RelWithDebInfo
Built on top of commit `e37996269d8d874c73b4de22f965a1db4a4c75aa` from the tip of origin/develop from 2024-11-05

<details>
<summary><strong>About Second Life</strong></summary>
SpaceDebris2024_tracy 7.1.12.243102349 (64bit)
Release Notes
Build Configuration RelWithDebInfo

You are at 137.0, 130.0, 28.1 in Sunborne located at simhost-0ad6efcb1f7c17b88.agni
SLURL: http://maps.secondlife.com/secondlife/Sunborne/137/130/28
(global coordinates 222,089.0, 305,538.0, 28.1)
Second Life Server 2024-10-15.11356152186
Release Notes

CPU: AMD Ryzen 9 7945HX3D with Radeon Graphics       (2295.63 MHz)
Memory: 64731 MB
OS Version: Microsoft Windows 11 64-bit (Build 22631.4391)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 4090 Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6603
OpenGL Version: 4.6.0 NVIDIA 566.03

Window size: 2560x1334
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 16376MB
Disk cache: Max size 1638.4 MB (100.0% used)

RestrainedLove API: (disabled)
J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/1,187 (0.0%)
November 05 2024 16:03:31
</details>

### Description

I am triggering a reproducible assert on `llassert(LLCoros::on_main_coro());` in my dev build of the viewer when clicking a resident name on the minimap.

### Reproduction steps

[...]

PR #3060, opened 2024-11-08
Fix assert when retrieving data while mousing over minimap avatar dots

RyeMutt:
Callback was happening in network thread instead of on main thread

secondlife/viewer#3025


Geenz (approved):
LGTM

Issue #2637, opened 2024-09-20
Projected texture is visible on an alpha blending surface


Dan-Linden:
### Environment

Second Life Release 7.1.11.10947335521 (64bit)
Release Notes

You are at 67.9, 39.5, 38.8 in Danger! Danger! located at simhost-0dd7e33bcf6d85949.agni
SLURL: http://maps.secondlife.com/secondlife/Danger%21%20Danger%21/68/40/39
(global coordinates 261,188.0, 249,640.0, 38.8)
Second Life Server 2024-07-22.10048683488
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.92 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4894)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.3623
OpenGL Version: 4.6.0 NVIDIA 536.23

Window size: 1556x1196
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (11.9% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 2/6,467 (0.0%)
September 20 2024 14:02:10

### Description

Projected texture is visible on an alpha blending surface in the ExtraFPS viewer.

### Reproduction steps

Repro:
1. Upload the attached "254 of 255 opacity.png". (It's the big white gap after this sentence.) ![254 of 255 opacity](https://github.com/user-attachments/assets/f545045f-cac9-479b-b32a-b9d062257a83)
2. Apply the texture to a box.
3. Point a projector with a texture at the box.
4. Verify the projector texture is not visible on the box.

[...]

PR #3061, opened 2024-11-08
#2637 Revert to the previous alpha analysis to maintain legacy content parity.

Geenz:
Fixes projectors not functioning as expected with certain alpha textures.


RyeMutt (approved):
LGTM

Issue #1493, opened 2024-05-16
Avatars that are nearby, but outside camera frustum, stop updating in mirrors.


Dan-Linden:
### Environment

Second Life Release 7.1.5.8472515256 (64bit)
Release Notes

You are at 60.7, 245.1, 22.9 in Rumpus Room 2050 located at simhost-01129fe6e38889eb3.aditi
SLURL: secondlife://Aditi/secondlife/Rumpus%20Room%202050/61/245/23
(global coordinates 36,412.7, 14,581.1, 22.9)
PMFP 2024-04-13.8669470587
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.91 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4291)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.3623
OpenGL Version: 4.6.0 NVIDIA 536.23

Window size: 1560x1184
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 7304MB
Disk cache: Max size 1638.4 MB (100.0% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: FMOD Studio 2.02.13
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Not Connected
Packets Lost: 0/1,073 (0.0%)
May 16 2024 11:13:29

### Description

Agents outside the camera frustum only show positional updates in mirrors. Updates for animation and attachments are not reflected in mirrors.

### Reproduction steps

1. Make a mirror that reflects everything.
2. Watch the mirror while someone wearing attachments walks around behind you.

Observed: The other avatar's attachments stop updating and their avatar moves away from the attachments. 
https://github.com/secondlife/viewer/assets/119471733/821dedb9-5fde-4d34-8faa-3310a5a25539

Expected: When looking in a mirror, the reflection of the scene behind you should look accurate.

PR #3064, opened 2024-11-09
Server Issue #1493: New notification message for llTransferOwnership.

Rider-Linden:
This is a new notification message that may be sent from the simulator after a call to llTransferOwnership

See server pull request here: https://github.com/secondlife/server/pull/1564

### DIFF
--- a/indra/newview/skins/default/xui/en/notifications.xml
+++ b/indra/newview/skins/default/xui/en/notifications.xml
@@ -10348,6 +10348,14 @@ You are now the owner of object [OBJECT_NAME]
 
   <notification
    icon="alertmodal.tga"
+   name="NowOwnObjectInv"
+   type="notify">
+   <tag>fail</tag>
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

indra/newview/skins/default/xui/en/notifications.xml
```diff
@@ -10348,6 +10348,14 @@ You are now the owner of object [OBJECT_NAME]
 
   <notification
    icon="alertmodal.tga"
+   name="NowOwnObjectInv"
+   type="notify">
+   <tag>fail</tag>
```
akleshchev, line 10353: probably shouldn't be 'fail'

Rider-Linden: All the others around use fail... (It's supposed to behave like `NowOwnObject` immediately above it and that is marked as fail also.

PR #3066, opened 2024-11-10
Fix calling LLFloaterPreference::setPanelVisibility resulting in selecting the 1st tab

Ansariel:
I'm not quite sure if there currently is a way to reproduce this in the LL viewer, but if changing any graphics settings will trigger this method being called and changing the selected tab to the first one in the list.


marchcat (approved):
Thank you!

marchcat (approved):

Issue #3034, opened 2024-11-07
Emoji temporarily breaks when chat scrolls up.


Dan-Linden:
### Environment

Second Life Release 7.1.11.11565212741 (64bit)
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.92 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5011)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6590
OpenGL Version: 4.6.0 NVIDIA 565.90

Window size: 1732x1234
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (42.7% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version:  

November 06 2024 18:14:41

### Description

Emoji temporarily breaks when chat scrolls up.

### Reproduction steps

1. Clear your chat.txt file, or at least don't have a copy of the hushed emoji in chat history.
2. Type some words in chat so the chat history window is full and ready to scroll up.
3. Type `:hushed` and press Enter to chat that emoji.
4: Chat something else so the chat history scrolls up.

Observed: The emoji is replaced with some other glyph.
https://github.com/user-attachments/assets/bfcb993f-619d-4b15-b236-b65634a289a0

Expected: The emoji should look the same.

Workaround: After chat scrolls one more time, it looks correct.

Note 1: I was not able to reproduce this on DeltaFPS, so this appears to be a regression in ExtraFPS.

AtlasLinden:
Same repro but different symbol: https://gyazo.com/e3b221657dee880772d4a11aff45aeb0

akleshchev:
Might be fixed if those work through LLUIString
#3222 Emoji tabs losing symbols

[...]

PR #3077, opened 2024-11-12
#3034 fix for Emoji temporarily breaks when chat scrolls up

maxim-productengine:


marchcat (approved):

akleshchev (approved):

indra/llui/lltextbase.cpp
```diff
@@ -3801,6 +3801,21 @@ LLEmojiTextSegment::LLEmojiTextSegment(const LLUIColor& color, S32 start, S32 en
...
+F32 LLEmojiTextSegment::draw(S32 start, S32 end, S32 selection_start, S32 selection_end, const LLRectf& draw_rect)
+{
+    bool reset_font_buffers = (mLastGeneration != mEditor.getTextGeneration()) && (mLastGeneration != -1);
+
+    F32 result = LLNormalTextSegment::draw(start, end, selection_start, selection_end, draw_rect);
+
+    //reset font buffers one more time next iteration, after all other segments are actually drawn
+    if (reset_font_buffers)
+    {
+        mLastGeneration = -1;
```
akleshchev, line 3814: Why not add some kind of LLNormalTextSegment::resetBuffer?

indra/llui/lltextbase.cpp
```diff
@@ -3801,6 +3801,21 @@ LLEmojiTextSegment::LLEmojiTextSegment(const LLUIColor& color, S32 start, S32 en
 {
 }
 
+
+F32 LLEmojiTextSegment::draw(S32 start, S32 end, S32 selection_start, S32 selection_end, const LLRectf& draw_rect)
+{
+    bool reset_font_buffers = (mLastGeneration != mEditor.getTextGeneration()) && (mLastGeneration != -1);
+
+    F32 result = LLNormalTextSegment::draw(start, end, selection_start, selection_end, draw_rect);
```
akleshchev, line 3809: Approved, but this seems odd. Why does it need to wait for other segments? Buffer is a collection of points, it shouldn't need to reset unless position changes because points remain unchanged.

akleshchev: Might be fixed in foreverFPS without the need to double-generate.

Issue #3068, opened 2024-11-11
When first opening the Experience tab in the Region/Estate floater, experience names with emojis are displayed incompletely or misspelled


DmitrySProductengine:
### Environment
<details><summary>Environment</summary>
<p>

Second Life Release 7.1.12.11386494836 (64bit)
Release Notes

You are at 134.4, 117.0, 22.9 in Product Engine 3 located at simhost-06b407db4579a48da.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine%203/134/117/23
(global coordinates 336774.0, 307061.0, 22.9)
Second Life Server 2024-10-15.11356152186
Release Notes

CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (2600 MHz)
Memory: 16384 MB
OS Version: macOS 15.0.1 Darwin 24.0.0 Darwin Kernel Version 24.0.0: Tue Sep 24 23:36:30 PDT 2024; root:xnu-11215.1.12~1/RELEASE_X86_64 x86_64
Graphics Card Vendor: ATI Technologies Inc.
Graphics Card: AMD Radeon Pro 555X OpenGL Engine

OpenGL Version: 4.1 ATI-6.1.13

Window size: 1181x1053
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.25
Render quality: 2
Texture memory: 4096MB
Disk cache: Max size 1638.4 MB (31.7% used)
HiDPI display mode: 

RestrainedLove API: (disabled)
J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091638
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/850 (0.0%)
November 11 2024 01:51:38

</p>
</details> 

### Reproduction steps

1. Download and install the Second Life 7.1.12.11386494836 viewer (https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%230ef7a9b-maint-c-2024-10-17).
2. Log in as an estate owner or estate manager.
[...]

PR #3090, opened 2024-11-13
#3068 Experience names are displayed incompletely (fix batch rendering)

LLGuru:


indra/llrender/llfontgl.cpp
```diff
@@ -341,24 +349,35 @@ S32 LLFontGL::render(const LLWString &wstr, S32 begin_offset, F32 x, F32 y, cons
...
             bitmap_entry.first == EFontGlyphType::Grayscale ? text_color
                                                             : emoji_color;
+        // This function emplaces data to arrays vertices, uvs, colors
+        // and increases glyph_count on 1, 2 or 6 depending on shadow
         drawGlyph(glyph_count, vertices, uvs, colors, screen_rect, uv_rect,
                   col, style_to_add, shadow, drop_shadow_strength);
+        llassert(glyph_count <= GLYPH_BATCH_SIZE);
+        if (!glyph_per_char)
+        {
+            glyph_per_char = glyph_count;
```
akleshchev, line 365: Shouldn't this be drawGlyph(glyph_per_char, ... and glyph_count += glyph_per_char;?

LLGuru: Definitely No

Each call of `drawGlyph` increases `glyph_count` on the same value (1, 2 or 6, depending on `shadow`)
Initially `glyph_per_char` is set to 0, and only after the first call of `drawGlyph` it is reinitialized again
Using this value we can prognose the next size of `glyph_count` before the next call of `drawGlyph`

akleshchev: Aren't you missing a reset in such case?

LLGuru: Yes, the parameter `shadow` is the same for all characters in the provided string,
so the value of `glyph_per_char` is absolutely the same for the whole string fragment

indra/llrender/llfontgl.cpp
```diff
@@ -295,30 +300,33 @@ S32 LLFontGL::render(const LLWString &wstr, S32 begin_offset, F32 x, F32 y, cons
...
         {
             LL_ERRS() << "Missing Glyph Info" << LL_ENDL;
             break;
         }
+
         // Per-glyph bitmap texture.
-        std::pair<EFontGlyphType, S32> next_bitmap_entry = fgi->mBitmapEntry;
-        if (next_bitmap_entry != bitmap_entry)
+        const std::pair<EFontGlyphType, S32>& bitmap_entry = fgi->mBitmapEntry;
+        if (wch != last_wch || i == 0)
```
akleshchev, line 324: last_wch doesn't get set anywhere?

PR #3101, opened 2024-11-16
Add Support HTTP Multiplexing

AiraYumi:
Enables HTTP/2 Multiplexing.
If this doesn't work, it will use HTTP pipelining.

### DIFF
--- a/indra/llcorehttp/_httplibcurl.cpp
+++ b/indra/llcorehttp/_httplibcurl.cpp
@@ -487,7 +487,7 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
         policy.stallPolicy(policy_class, false);
         mDirtyPolicy[policy_class] = false;
 
-        if (options.mPipelining > 1)
+        if (options.mPipelining)
         {
             // We'll try to do pipelining on this multihandle
             check_curl_multi_setopt(multi_handle,
--- a/indra/llcorehttp/_httplibcurl.cpp
+++ b/indra/llcorehttp/_httplibcurl.cpp
@@ -493,7 +493,7 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
             LL_INFOS("HttpLibcurl") << "HTTP Pipelining Enable." << LL_ENDL;
             check_curl_multi_setopt(multi_handle,
                                      CURLMOPT_PIPELINING,
-                                     1L);
+                                     CURLPIPE_MULTIPLEX | CURLPIPE_HTTP1);
             check_curl_multi_setopt(multi_handle,
                                      CURLMOPT_MAX_PIPELINE_LENGTH,
                                      long(options.mPipelining));
@@ -509,7 +509,7 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
             LL_INFOS("HttpLibcurl") << "HTTP Pipelining Disable." << LL_ENDL;
             check_curl_multi_setopt(multi_handle,
                                      CURLMOPT_PIPELINING,
-                                     0L);
+                                     CURLPIPE_NOTHING);
             check_curl_multi_setopt(multi_handle,
                                      CURLMOPT_MAX_HOST_CONNECTIONS,
                                      0L);
--- a/indra/llcorehttp/_httplibcurl.cpp
+++ b/indra/llcorehttp/_httplibcurl.cpp
@@ -490,6 +490,7 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
         if (options.mPipelining)
         {
             // We'll try to do pipelining on this multihandle
+            LL_INFOS("HttpLibcurl") << "HTTP Pipelining Enable." << LL_ENDL;
             check_curl_multi_setopt(multi_handle,
                                      CURLMOPT_PIPELINING,
                                      1L);
@@ -505,6 +506,7 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
         }
         else
         {
+            LL_INFOS("HttpLibcurl") << "HTTP Pipelining Disable." << LL_ENDL;
             check_curl_multi_setopt(multi_handle,
                                      CURLMOPT_PIPELINING,
                                      0L);


AiraYumi:
I'm currently working on it, and at least I added gSavedSettings to "indra/llcorehttp/_httplibcurl.cpp", but it didn't build.
I'll try adding it to HttpOptions.

akleshchev:
You can also remake mPipelined into a enum with options like no|pipelining|multiplexing. It already loads from gSavedSettings.getBOOL(http_pipelining), so it will just have to become getS32(http_pipelining).

monty-linden:
Per-request and per-multi-handle options are kept well-apart in llcorehttp.  So `curl_multi_setopt` values are controlled by options sent in via httppolicyclass.  Per-request options in HttpOptions find their way into _httpoprequest where global, per-policy-class, and per-request options are baked into `curl_easy_setopt` calls in prepareRequest() (mostly).

It sounds like you want to extend the policy class definition.  But maybe describe what libcurl options you want to change and that will shape an answer.  There's also the README.Linden file which hasn't been maintained but has useful tutorial information.

AiraYumi:
It is now possible to switch between CURLPIPE_MULTIPLEX or CURLPIPE_HTTP1 depending on the value registered in httppolicyclass.
The default is 0, so it works the same as before the change.

--- Here's what I think ---
Setting UseHTTP2Multiplexing to 1 will use HTTP/2 Multiplexing, and setting it to 0 will use HTTP Pileline.
Since they operate mutually exclusive, enabling HTTP/2 will disable HTTP Pileline.

AiraYumi:
I think it should work, let me know if you have any problems.

indra/newview/llappcorehttp.cpp
```diff
@@ -281,6 +283,15 @@ void LLAppCoreHttp::init()
         LL_INFOS("Init") << "HTTP Pipelining " << (mPipelined ? "enabled" : "disabled") << "!" << LL_ENDL;
     }
 
+    // Global multiplexing setting
+    static const std::string http_multiplexing("UseHTTP2Multiplexing");
```
akleshchev, line 287: Personally I would have merged "HttpPipelining" and "UseHTTP2Multiplexing" into some kind of 'mode switch' since they are sort of exclusive.

akleshchev (approved):

indra/llcorehttp/_httplibcurl.cpp
```diff
@@ -487,12 +487,22 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
...
         mDirtyPolicy[policy_class] = false;
 
-        if (options.mPipelining > 1)
+        if (options.mPipelining)
         {
             // We'll try to do pipelining on this multihandle
-            check_curl_multi_setopt(multi_handle,
-                                     CURLMOPT_PIPELINING,
-                                     1L);
+            LL_INFOS("HttpLibcurl") << "HTTP Pipelining Enable." << LL_ENDL;
```
monty-linden, line 493: Is this debug logging left in?  Should it not use the supplied LOG_* tag values above?

indra/llcorehttp/_httplibcurl.cpp
```diff
@@ -487,12 +487,22 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
...
+            if (options.mMultiplexing)
```
monty-linden, line 494: The nesting conditionals are probably wrong, in practice.  If mMultiplexing is on but mPipelining is off, multiplexing ends up disallowed.  Probably not intended.  One of those areas we'll have to sort out.  If the 'mPipelining' concept is changed to concurrency, this is more nearly correct.

indra/llcorehttp/httprequest.h
```diff
@@ -217,6 +217,10 @@ class HttpRequest
         /// Per-class only
         PO_PIPELINING_DEPTH,
 
+        /// If it is 1, enable HTTP Multiplexing.
```
monty-linden, line 220: More explanation about this API.  The 'public' API presented here is *not* meant to be a facade for libcurl.  At the time this was written and also possibly for future purposes, libcurl might be replaced by another library.  (For Sansar, we had our own low-level libcurl-like library which could have slotted in.)  The 'public' API avoided using libcurl definitions.

Instead, this API is meant to describe how the viewer wants to use HTTP requests and how HTTP traffic should be managed and shaped.  The intent is to have this description be as minimal as possible.  If a concept isn't used by the viewer, it won't appear here.

Internal APIs, such as _httplibcurl.*, are the translation layer between the public API and whatever is going on below.  An adapter layer.  There's only one concrete adapter at this point but that is okay.

And so we have the comment above:
> When PIPELINING_DEPTH is 2 or more, libcurl performs

That is still correct.  The public API describes an abstract usage and a pipelining depth of 2 or more enables pipelining.  And so the code above in _httplibcurl is no longer correct.

What *should* this API be in the http/2 world?  Well, we don't know yet.  We haven't stood up the infrastructure to support it.  But we would probably want to modify llcorehttp differently.  PO_PIPELINING_DEPTH might instead be changed to PO_REQUEST_CONCURRENCY covering both the http/1 (pipelining) and http/2 (multiplexing) cases with added explicit enables/disables for pipelining and http/2.  However, we may find that we need independent http/2 parameters for the policy classes.

(There's another weak reason to keep existing API definitions as-is:  we use a copy of this code in server-side binaries and subtle changes may not be caught.  This isn't meant to stop changes and progress in the viewer.  But when all other considerations are equal, keeping an API spec has a benefit to us.)

monty-linden (changes requested):
One very large problem I have with these http/2 PRs is that this code, while disabled by default, cannot even be tested experimentally anywhere.  Linden has a lot of infrastructure changes to make.  (https: support, http/2 support, origin server changes, multiple grids, CDN configuration, new certs and cipher negotiation)  So this is code that has never been tested and cannot simply be picked up and developed without that work.  I approve of the spirit but this may not be the way to start.

Given the amount of infrastructure that must change and the need for testing around the world, this is work that should probably be packaged in a project viewer and not targeted at 'develop'.

I was surprised when these PRs appeared.  Is there some particular need driving the interest in HTTP/2?  We do have a project outlined but other work must happen first.  And now I am curious if there is an important need for http/2 now....

indra/llcorehttp/_httplibcurl.cpp
```diff
@@ -487,12 +487,22 @@ void HttpLibcurl::policyUpdated(unsigned int policy_class)
         policy.stallPolicy(policy_class, false);
         mDirtyPolicy[policy_class] = false;
 
-        if (options.mPipelining > 1)
```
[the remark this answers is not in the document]
AiraYumi: fix it.

AiraYumi:
The reason we decided to introduce the concept of HTTP/2 to the viewer was the end of support for OpenSSL.

One of the obstacles to updating OpenSSL is updating Curl.
One of the obstacles to updating Curl is the HTTP Pipeline.
When Curl is updated, the HTTP Pipeline becomes unavailable, which means performance is reduced.
As a result, the HTTP Pipeline technology is a major obstacle to updating Curl and OpenSSL.

We believe that if HTTP/2 is successfully introduced, one of the obstacles to updating OpenSSL will be removed.
More than 10 years have passed since the concept of the HTTP Pipeline was introduced to the viewer.

During this time, HTTP has been updated to HTTP/3, but the viewer continued to rely on HTTP/1.1...

AiraYumi:
The issues pointed out that did not work as intended have been fixed.

I now understand that this content is likely to be something that should be put into the project viewer.
I certainly understand the argument that things that can't be tested shouldn't be put into develop.
If some of the CDNs in the beta SIM could support HTTP/2, we'd be able to test it...

monty-linden:
Yes, there is quite a bit of work that must happen on our side.  For now, I recommend making this a 'draft pull request' so that it isn't lost.  It may be useful to start here in the future.

I do have an http/2 project outline.  I will put this in a better place later.  But I can share it here for reference:

### Milestone 1 - Testbed

- Build and run viewer with special instrumentation to recover the asset fetch profile for interesting sessions.  These are the mix of asset fetch and HTTP range options one will expect on entry to dense regions.
- PoC changes to llcorehttp to support HTTP/2 in the crudest fashion.  Used to get an initial benchmark with the above profile.  Identifies:
  - Performance gap introduced by https:
  - API changes needed in the library to support dual configuration (both HTTP/1 and HTTP/2).
  - Excessive connection concurrency and head-of-line issues as connections transition to /2.

### Milestone 2 - Infrastructure

- Initial asset system and CDN work:
  - Cert creation and distribution.
  - HTTP/2 configuration as needed.
  - CDN origin server modifications.
- Simulator:
  - New and/or modified Capabilities to support HTTP/2 endpoints without impacting existing legacy client code.

### Milestone 3 - Internal Viewer

- Llcorehttp modified according to design fallout from above.
- Additional library support updated as needed.
- Viewer that can be used internally by others with some inconveniences.
- TPVs:  begin communications

### Milestone 4 - Project Viewer

- Llcorehttp and libraries take final form.
- Viewer uses new infrastructure, presents new configuration options, etc.
- Viewer gets any final polishing needed including additional diagnostics for connection problems, etc.
- Publish and share.

### Milestone 5 - Final Pass

- Expect additional tuning and control needs after this gets some exposure in the field.  Users with particularly bad connectivity or wonky networking equipment may need special consideration.

monty-linden:
@akleshchev pulled me back onto this PR and my previous comments still apply.  But some significant changes are coming to the 'develop' branch.  Expect some updated information about processes in the next days.  It will likely apply to this work.

akleshchev:
Sorry, should have stated reason why I pulled you back, This now has a toggle and is disabled by default. I'm not sure if there is a point holding it except for a lack of 'home' (develop in current form is going away).

bennettgoble:
@AiraYumi thank you very much for your work here. Lower level and major changes like this need to happen in tight concert with our backend development efforts.

I'm closing this PR for now, but what we will need is to prepare the backend changes so that this can be tested while development is ongoing. It's not possible to accept a PR for HTTP/2 without HTTP/2 support on our servers.

PR #3105, opened 2024-11-18
brad/2744 handle shader errors

brad-linden:


indra/newview/llfeaturemanager.cpp
```diff
@@ -655,6 +655,21 @@ void LLFeatureManager::applyBaseMasks()
...
+        {
+            // if we don't have OpenGL 4.6 on intel, set it to OpenGL 3.3
+            // we also want to trigger the GL3 fallbacks on these chipsets
+            // this is expected to be mainly pre-Haswell Intel HD Graphics 4X00 and 5X00.
+            // A lot of these chips claim 4.3 or 4.4 support, but don't seem to work.
+            // https://code.blender.org/2019/04/supported-gpus-in-blender-2-80/
+            // https://docs.blender.org/manual/en/latest/troubleshooting/gpu/windows/intel.html#legacy-intel-hd-4000-5000
+            // https://www.intel.com/content/www/us/en/support/articles/000005524/graphics.html
+            // this will disable things like reflection probes, HDR, FXAA and SMAA
+            gGLManager.mGLVersion = llmax(gGLManager.mGLVersion, 3.33f);
```
RunitaiLinden, line 668: do you mean llmin?

brad-linden: yes I do, drat

brad-linden: one other problem here is that it is kind of too late to set the GL Version, since the GL context has already been created at this point.  I'm not sure what the right solution for this would be.   I would like to force recreate the context, but not sure that would actually work

PR #3110, opened 2024-11-19
Drop emissive on old Intel GPUs

Geenz:


RunitaiLinden (dismissed):
Definitely too risky to go into ExtraFPS right away, might need to wait for a hotfix release. 
 Let's dogfood it for a day tomorrow and loop in kyle/atlas on if we want some QA on it before it merges down.

brad-linden (approved):

PR #3112, opened 2024-11-20
Update to openssl 3.4.0 and test if the viewer works.

AiraYumi:


AiraYumi:
If the version of curl was newer than 7.54.1 it would not build with the newer openssl.

AiraYumi:
~~~
Error -11: terminated by signal SIGSEGV
Command PhaseScriptExecution failed with a nonzero exit code
~~~

This is the first time I've encountered this error, so I don't know what to do.

AiraYumi:
3p-curl was being built on macos-12.
Changed to build on macos-15.

bennettgoble:
@AiraYumi thank you for the work on dependencies. As a matter of policy we can't/shouldn't run builds if it is referencing code built on your account. In this case, it would be best to test these changes locally before submitting a PR (Both to 3p-curl and viewer.)

We also really love to get issues describing work, so we can use it as a conversation piece. What new features does this version of curl bring in? (Bug fixes? Etc.) A short summary would be appreciated.

PR #3114, opened 2024-11-20
Merge ExtraFPS into develop

akleshchev:
to pick up two priority fixes


marchcat (approved):

maxim-productengine (approved):

akleshchev:
@RyeMutt linux build fails on an SDL package, due to libpng mismatch. I wanted to update SDL, but looks like you are already doing something there? Can I just merge this? Or should I pick newer SDL?

PR #3120, opened 2024-11-21
#2696 Viewer crashes on gestures : add checkbox 'Play all infinitely'

LLGuru:
This checkbox is displayed at the bottom right corner of the GESTURES floater
It is visible only when the "Show Advanced Menu" option is ON
When this checkbox is checked, all gestures are being started in a cycle
This allows to reproduce the bug now and will allow to check for a regression in future (when the bug will be fixed)
imageimage

### DIFF
--- a/indra/newview/llfloatergesture.cpp
+++ b/indra/newview/llfloatergesture.cpp
@@ -36,6 +36,7 @@
 
 #include "llagent.h"
 #include "llappearancemgr.h"
+#include "llcheckboxctrl.h"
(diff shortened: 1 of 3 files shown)


akleshchev (approved):

akleshchev:
Show Advanced Menu is still a user menu, I suggest using develop (which are still in use by users) or a debug setting without UI.

LLGuru:
> I suggest using develop

Replaced with "Develop" menu

PR #3129, opened 2024-11-22
prevent erroneous edit of wrong parcel

AndrewMeadows:
This PR fixes jira-archive-internal/issues/70771

https://github.com/secondlife/jira-archive-internal/issues/70771

The problem was the update for the selected parcel was being sent to the wrong region, because the viewer logic for finding the corresponding region was flawed.

### DIFF
--- a/indra/newview/llviewerparcelmgr.cpp
+++ b/indra/newview/llviewerparcelmgr.cpp
@@ -1327,7 +1327,7 @@ const S32 LLViewerParcelMgr::getAgentParcelId() const
     return INVALID_PARCEL_ID;
 }
 
-void LLViewerParcelMgr::sendParcelPropertiesUpdate(LLParcel* parcel, bool use_agent_region)
+void LLViewerParcelMgr::sendParcelPropertiesUpdate(LLParcel* parcel)
 {
     if(!parcel)
         return;
--- a/indra/newview/llviewerparcelmgr.cpp
+++ b/indra/newview/llviewerparcelmgr.cpp
@@ -1332,7 +1332,7 @@ void LLViewerParcelMgr::sendParcelPropertiesUpdate(LLParcel* parcel, bool use_ag
     if(!parcel)
         return;
 
-    LLViewerRegion *region = use_agent_region ? gAgent.getRegion() : LLWorld::getInstance()->getRegionFromPosGlobal( mWestSouth );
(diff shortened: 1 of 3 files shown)


indra/newview/llviewerparcelmgr.cpp
```diff
@@ -1332,7 +1332,7 @@ void LLViewerParcelMgr::sendParcelPropertiesUpdate(LLParcel* parcel, bool use_ag
     if(!parcel)
         return;
 
-    LLViewerRegion *region = use_agent_region ? gAgent.getRegion() : LLWorld::getInstance()->getRegionFromPosGlobal( mWestSouth );
+    LLViewerRegion *region = LLWorld::getInstance()->getRegionFromID(parcel->getRegionID());
```
AndrewMeadows, line 1335: This is where the bug lived.  The region pointer was being found based on (a) whether the agent was in the region or (b) a cached value of `mWestSouth` which was supposed to be the global-position of the region's origin corner, but could sometimes be from the wrong region.

The fix is to cache the region_id in each parcel so it can be used to find the region in this context.

AndrewMeadows:
I  **think** the repro recipe goes something like this:

**(1)**  Find two adjacent regions **A** and **B**, each with only one parcel (to make things easy)
**(2)** While on **regionA** sidle up to the boundary with **regionB**, but don't cross over.
**(3)** Select **parcelA** on **regionA** and open **About Land** floater (just to make it visible)
**(4)** On **regionA** terrain right-click the terrain and select **Edit Terrain**
**(5)** Left-click **parcelB** on **regionB** --> notice the parcel info populates the previously openened **About Land** floater
**(6)** Change the Name of **parcelB**
**(7)** Select **parcelA** in **regionA** --> notice its name has changed to what you tried to change on **parcelB**

Why does this work?  Because there is a way to cache the `LLViewerParcelMgr::mWestSouth` corner of **regionA** and then change to a parcel in **regionB** without updating that cached value.  When the old viewer would try to apply a parcel properties change it would use the cached value of `mWestSouth` to figure out which region to send it to.  I'm not 100% sure that is the exact repro recipe above, but it is  something like that.  The operation of opening the **About Land** floater up will, for most cases, correctly cache `mWestSouth` for the appropriate region, so you need to find that path that caches the corner and then later updates the **About Land** parcel selection without recomputing it.

marchcat (approved):

Issue #3104, opened 2024-11-18
Outfit Gallery images do not load


Dan-Linden:
<details>
<summary>Environment</summary>
Second Life Release 7.1.11.11750364439 (64bit)

Release Notes

You are at 118.0, 95.7, 21.8 in By Design located at simhost-01d3500f306966b31.agni

SLURL: http://maps.secondlife.com/secondlife/By%20Design/118/96/22

(global coordinates 261494.0, 246880.0, 21.8)

Second Life Preflight 2024-10-19.11413863173

Release Notes

CPU: Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz (2800 MHz)

Memory: 16384 MB

OS Version: macOS 11.7.10 Darwin 20.6.0 Darwin Kernel Version 20.6.0: Thu Jul  6 22:12:47 PDT 2023; root:xnu-7195.141.49.702.12~1/RELEASE_X86_64 x86_64

Graphics Card Vendor: Intel Inc.

Graphics Card: Intel Iris Pro OpenGL Engine

OpenGL Version: 4.1 INTEL-16.5.14

Window size: 1280x800

Font Size Adjustment: 96pt

UI Scaling: 1

Draw distance: 96m

Bandwidth: 3000kbit/s

LOD factor: 1.125

Render quality: 1

Texture memory: 1536MB

Disk cache: Max size 819.2 MB (100.0% used)

HiDPI display mode:

J2C Decoder Version: KDU v7.10.4

Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft

Dullahan: 1.14.0.202408091638

  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54

  Chromium: 118.0.5993.54

LibVLC Version: 3.0.21

Voice Server Version: Secondlife WebRTC Gateway

Packets Lost: 87/65023 (0.1%)

November 18 2024 11:13:56
</details>

### Description

Outfit Gallery images do not load on Kyle Linden's ole Mac.

### Reproduction steps

1. Install https://releasenotes.secondlife.com/viewer/7.1.11.11750364439.html on an old Mac
2. Log in and open Avatar > My Outfits, then Outfit Gallery.

Observed: 
![Image](https://github.com/user-attachments/assets/cfbd209b-b49d-4dd4-93cb-da4079e724ad)

Expected: The images should load.

[...]

PR #3136, opened 2024-11-22
secondlife/viewer#3104: Fix outfit gallery pictures sometimes not loading. Remove texture resolution restriction.

cosmic-linden:
I have decided against the more common pattern of using texture boosting in this case. As mentioned in the ticket, outfits floater and texture loading need further follow-up. However, that work can be deferred.

### DIFF
--- a/indra/newview/lloutfitgallery.cpp
+++ b/indra/newview/lloutfitgallery.cpp
@@ -55,8 +55,15 @@
 
 static LLPanelInjector<LLOutfitGallery> t_outfit_gallery("outfit_gallery");
 
-#define MAX_OUTFIT_PHOTO_WIDTH 256
-#define MAX_OUTFIT_PHOTO_HEIGHT 256
+// The maximum resolution at which to load the outfit photo. If the given
+// texture has a higher resolution, tell the texture streaming system to
+// only load the resolution needed. An in-world object may request to load
+// the texture at a higher resolution, but that won't affect textures
+// loaded with LLViewerTexture::FETCHED_TEXTURE. (see
+// LLOutfitGalleryItem::setImageAssetId and also
+// LLViewerTexture::LOD_TEXTURE)
+#define MAX_OUTFIT_PHOTO_LOAD_WIDTH 256
+#define MAX_OUTFIT_PHOTO_LOAD_HEIGHT 256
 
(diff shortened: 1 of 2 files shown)


indra/newview/lloutfitgallery.cpp
```diff
@@ -979,28 +986,18 @@ void LLOutfitGalleryItem::draw()
...
 
-            gl_draw_scaled_image(interior.mLeft - 1, interior.mBottom, interior.getWidth(), interior.getHeight(), mTexturep, UI_VERTEX_COLOR % alpha);
+        // Pump the priority
+        const F32 stats = (F32)llmin(interior.getWidth() * interior.getHeight(), MAX_OUTFIT_PHOTO_LOAD_WIDTH * MAX_OUTFIT_PHOTO_LOAD_HEIGHT);
+        mTexturep->addTextureStats(stats);
 
-            // Pump the priority
-            mTexturep->addTextureStats((F32)(interior.getWidth() * interior.getHeight()));
-        }
+        // If the floater is focused, don't apply its alpha to the texture (STORM-677).
```
akleshchev, line 998: Probably no point referencing STORM-677, it's should be not accessible by now.

akleshchev (approved):

indra/newview/lloutfitgallery.cpp
```diff
@@ -979,28 +986,18 @@ void LLOutfitGalleryItem::draw()
...
-        {
-            LLRect interior = border;
-            interior.stretch(-1);
+        LLRect interior = border;
+        interior.stretch(-1);
 
-            gl_draw_scaled_image(interior.mLeft - 1, interior.mBottom, interior.getWidth(), interior.getHeight(), mTexturep, UI_VERTEX_COLOR % alpha);
+        // Pump the priority
+        const F32 stats = (F32)llmin(interior.getWidth() * interior.getHeight(), MAX_OUTFIT_PHOTO_LOAD_WIDTH * MAX_OUTFIT_PHOTO_LOAD_HEIGHT);
+        mTexturep->addTextureStats(stats);
```
akleshchev, line 996: Does this happen only when in view? Most LLView elements only draw when their rect is in view, but some lists are extempt to that.

Issue #2848, opened 2024-10-14
[ExtraFPS] Scale of animated textures is changing


Dan-Linden:
### Environment

Second Life Release 7.1.11.11074622243 (64bit)
Release Notes

You are at 140.9, 225.5, 133.7 in Linden Lab HQ 4 located at simhost-0ed1af4f6c09b0a7a.agni
SLURL: http://maps.secondlife.com/secondlife/Linden%20Lab%20HQ%204/141/226/134
(global coordinates 238,989.0, 244,194.0, 133.7)
Second Life Preflight 2024-09-30.11111372425
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.91 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.4894)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6590
OpenGL Version: 4.6.0 NVIDIA 565.90

Window size: 1712x1217
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.75
Render quality: 5
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (100.0% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 0/2,157 (0.0%)
October 14 2024 11:56:49

### Description

The scale of animated textures changes in the 7.1.11.11074622243 - ExtraFPS viewer
This does not reproduce in the 7.1.10.10800445603 - DeltaFPS viewer.

### Reproduction steps

1. Go to http://maps.secondlife.com/secondlife/Linden%20Lab%20HQ%204/141/226/134
2. Make the Second Life window relatively small (I used 1512 x 1257)
3. Watch the animated leaves on the tree.
Observed: The texture of the leaves occasionally grows larger and smaller.

[...]

PR #3143, opened 2024-11-23
secondlife/viewer#2768, secondlife/viewer#2848: Partial cherry pick of "Animated pbr textures don't animate" from develop

cosmic-linden:


akleshchev (approved):
No longer can repro with these commits applied. Note: leaves aren't aniamting.

marchcat (approved):

PR #3144, opened 2024-11-24
Fix memory leak in LLVertexBuffer

Ansariel:


marchcat (approved):
Thank you, @Ansariel!

marchcat:
Wow, I even approved this twice! 😹

PR #3168, opened 2024-11-27
Kill explicit vintage mode settings for quality modes. Only explicitly set it in GL3.

Geenz:


brad-linden (changes requested):

brad-linden:
this did correct the log warnings I was seeing about reenabling disabled features though, so mostly this is good

brad-linden (approved):

PR #3173, opened 2024-11-27
Add diagnostic output for LLProcess event history failure.

nat-goodspeed:


brad-linden:
rebased since the upstream branch got squashed when PR #3168 merged

brad-linden (approved):

PR #3175, opened 2024-11-27
viewer#3171 Exposure setting enabled in Vintage mode

akleshchev:


brad-linden (approved):

Ansariel:
Why not simply use `enabled_control` attribute for the control in the XML file?

akleshchev:
Because that requires the setting to be in reverse, so i havent even considered it, but now that you saoid that, I double checked and there apparently is disabled_control. Will try with that, thanks,

Ansariel:
Yeah, so just use `disabled_control` instead:

```
  <slider
    control_name="RenderExposure"
    disabled_control="RenderVintage"
    decimal_digits="1"
    follows="left|top"
    height="16"
    increment="0.1"
    initial_value="160"
    label="Brightness (exposure)"
    label_width="145"
    layout="topleft"
    left="30"
    min_val="0.5"
    max_val="4.0"
    name="RenderExposure"
    show_text="true"
    top_pad="14"
    width="260">
  </slider>
```

Geenz (approved):

PR #3177, opened 2024-11-27
ExtraFPS -> develop merge

brad-linden:


marchcat (approved):

marchcat:
Oh, I didn't notice it had auto-merge enabled. But that is convenient.

brad-linden:
> Oh, I didn't notice it had auto-merge enabled. But that is convenient.

Yeah I turned on the option for enabling this,  I wasn't sure how to experiment with it without surprising people.  My hope is that this will reduce some of the motivation to skip checks

marchcat:
> My hope is that this will reduce some of the motivation to skip checks

In this case, should we make the green build checks required as well?

brad-linden:
> In this case, should we make the green build checks required as well?

I would definitely like to get that done.  I don't know for sure yet what impediments may still be in the way of doing that, but until we do know, I want to keep removing any possible impediments to that.

PR #3179, opened 2024-11-28
viewer#3170 Fix tonemaping slider

akleshchev:
At the moment slider is in general settings, not per environment


maxim-productengine (approved):

Geenz:
This was not the correct fix for that.  The correct fix for it is here: https://github.com/secondlife/viewer/pull/3174

PR #3182, opened 2024-11-29
Check whether the viewer built on Ubuntu 24.04 can be run on Ubuntu 22.04.

AiraYumi:


marchcat:
Let's see how it works, starting a build.

vldevel:
Under Linux, the backward compatibility (from a newer distro version to an older one) is usually not possible. However, forward compatibility is a given, i.e. you can always use a viewer built on an older distro on a newer distro.

You need the libraries used by your build to be older or equal to the target Linux system. glibc and libstdc++ are two such libraries that **absolutely** need to be older or equal on your build system than on any target system.

That's why I am always using a build system that is 4-5 years older than currently supported Linux distros for the builds of the Cool VL Viewer.

FYI, I am currently using a Devuan Beowulf build VM.

.github/workflows/build.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
         runner: [windows-large, macos-15-xlarge]
         experimental: [false]
         include:
-        - runner: linux-large
+        - runner: ubuntu-24.04
```
brad-linden, line 53: vldevel is right about how the versioning of glibc and libstdc++ need to be managed. We will likely never want to upgrade our linux runners unless we are intentionally dropping support for the viewer running on older linux distros.  even if you could get an example built this way working on your ubuntu 22.04 machine, it is extremely unlikely that it would work more broadly and we couldn't distribute it.

brad-linden (changes requested):

.github/workflows/build.yaml
vldevel: I would personally recommend to use a 2020-ish distro for the build system (and not just for the viewer, but *also* for all the 3p libraries).

Ubuntu 20.04 LTS is still maintained (an will still receive security fixes until 2030), and would be a good candidate for a Linux build system in 2024/2025.

brad-linden: I think we ended up where we are on 22.04 as something of a compromise between having a new enough GCC for c++20 and an old enough glibc relatively broad support.  I'm open to trying to build back on ubuntu 20.04 if we have a story for how to run a good c++20 compiler there that won't hold back Windows/MSVC and Mac/clang development with compatibility problems

AiraYumi: There are two ways to make an old ubuntu compatible with the latest gcc: installing a PPA for "Toolchain test builds" or including gcc in a package.
Installing a PPA is the easiest, but it will install things that are not standard ubuntu.
Packaging gcc requires several hours of build time and regular performance and security evaluations.

brad-linden: right.  I am speaking specifically of the runner-images that github has prepared:

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md#installed-software
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#installed-software
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#installed-software

Ideally we should be able to use those off the shelf

vldevel: > I think we ended up where we are on 22.04 as something of a compromise between having a new enough GCC for c++20 and an old enough glibc for relatively broad support. I'm open to trying to build back on ubuntu 20.04 if we have a story for how to run a good c++20 compiler there that won't hold back Windows/MSVC and Mac/clang development with compatibility problems

Well, if you really *need* C++20 (you won't, should llwebrtc be moved out of the viewer sources tree and into a 3p-library, like what happens for Dullahan/CEF), then you still can opt for building the viewer with clang. Even "old" clang versions got C++20 support.

brad-linden: I didn't do the experiment myself, but I have heard reports that the clang versions available on the stock runner-images are even worse for c++20 support.  There is work in develop now that uses c++20 constructs beyond just llwebrtc, so this is gonna become a bigger deal.

AiraYumi: As of gcc 11, almost all of C++20 is implemented,
but clang seems to still lack some implementations even in 18.

You need to consider how that might be a problem.

gcc
https://gcc.gnu.org/projects/cxx-status.html#cxx20

clang
https://clang.llvm.org/cxx_status.html#cxx20

AiraYumi:
Cancelled as it is too early to attempt migration

PR #3183, opened 2024-11-29
Fix merge error and restore crash fix in LLMaterialEditor

Ansariel:
@marchcat 😃


marchcat (approved):
I agree, it should be a smoother merge back into develop. Thank you, @Ansariel!

PR #3192, opened 2024-12-02
Rename "Vintage Mode" to "HDR and Emissive" in advanced settings.

Geenz:


akleshchev (approved):

indra/newview/app_settings/settings.xml
```diff
@@ -16071,7 +16071,7 @@
     <key>Value</key>
     <integer>1</integer>
   </map>
-    <key>RenderVintageMode</key>
+    <key>RenderDisableVintageMode</key>
     <map>
         <key>Comment</key>
         <string>Disable different rendering pipeline features to improve performance on older machines that makes the world look closer to how it used to prior to V7.</string>
```
brad-linden, line 16077: this doc comment string probably should be updated to reflect the fact that things are reversed now too

Geenz: Updated.

brad-linden (approved):

indra/newview/app_settings/settings.xml
Ansariel: If it is better graphics, why is it not enabled by default and if I have old hardware, I can get back to the old/more performant mode by enabling it? This makes literally no sense: "Here's your new car. It's the same as the old one. If you want it to actually be new. you have to press this button first..."

akleshchev: It is enabled by default for everything above Low+

indra/newview/app_settings/settings.xml
```diff
@@ -16071,16 +16071,16 @@
     <key>Value</key>
     <integer>1</integer>
   </map>
-    <key>RenderVintageMode</key>
+    <key>RenderDisableVintageMode</key>
```
Ansariel, line 16074: Name it`RenderEnableAdvancedPipelineFeatures` instead of inverse logic of disabling something to enable new features.

Geenz: We did this to make the feature table happy.  We may rename it later, but for now it's fine.

Issue #3119, opened 2024-11-20
Lua: Printing LLDebugSetting crashes viewer


AtlasLinden:
### Environment

Second Life Develop build: [7.1.12.11839252315](https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%23a0e5e23)

### Description

The below script to edit and print the value of a debug setting is crashing the viewer upon execution.

No change if a different debug setting is used or if the debug setting is manually changed.

### Reproduction steps

```
local LLDebugSettings = require 'LLDebugSettings'

LLDebugSettings.set("360CaptureCameraFOV", 100)

debug = LLDebugSettings.get("360CaptureCameraFOV")
LL.print_info(debug)
```

PR #3197, opened 2024-12-03
#3119 crash when lua requests debug settings value

maxim-productengine:
'Quick fix' to eliminate the crash #3119
(later we should investigate and don't allow overwriting the global table)

### DIFF
--- a/indra/llcommon/lua_function.cpp
+++ b/indra/llcommon/lua_function.cpp
@@ -764,7 +764,7 @@ int lua_metaipair(lua_State* L)
 
 } // anonymous namespace
 
-bool LuaState::push_debug_traceback()
+int LuaState::push_debug_traceback()
 {
     // Push debug.traceback() onto the stack as lua_pcall()'s error
     // handler function. On error, lua_pcall() calls the specified error
@@ -778,18 +778,18 @@ bool LuaState::push_debug_traceback()
     {
         lua_pop(mState, 1);
         LL_WARNS("Lua") << "'debug' table not found" << LL_ENDL;
-        return false;
+        return 0;
     }
     lua_getfield(mState, -1, "traceback");
     if (!lua_isfunction(mState, -1))
     {
         lua_pop(mState, 2);
         LL_WARNS("Lua") << "'traceback' func not found" << LL_ENDL;
-        return false;
+        return 0;
     }
     // ditch "debug"
     lua_remove(mState, -2);
-    return true;
+    return lua_gettop(mState);
 }
 
 LuaState::~LuaState()
@@ -812,6 +812,8 @@ LuaState::~LuaState()
     // stack contains Registry.atexit
     if (lua_istable(mState, -1))
     {
+        int atexit = lua_gettop(mState);
+
         // We happen to know that Registry.atexit is built by appending array
         // entries using table.insert(). That's important because it means
         // there are no holes, and therefore lua_objlen() should be correct.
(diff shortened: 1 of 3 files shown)


akleshchev (approved):

nat-goodspeed (changes requested):
I think I see: the bug is that the original code assumes that `debug` should be the Luau library module, and its `traceback` should be the documented function. If there's no `debug`, or no `debug.traceback`, or if either of them aren't as expected, then we run into trouble. You test whether they're correct (or as correct as can be readily verified).

The trouble with this fix is that if the subject script inadvertently overwrote the global `debug` variable, `~LuaState()` skips calling functions registered with `LL.atexit()`. That's potentially problematic, since we rely on `atexit()` functions to clean up viewer `ResultSet` objects, e.g. for inventory.

It would be better if, when `push_debug_traceback()` succeeds, you set an `int` variable to `lua_gettop(L)`, and pass that `int` variable to `lua_pcall()` as the error function stack index. When it fails, that variable is 0, meaning no error function.

You could conditionally pop `debug.traceback` without retesting the `push_debug_traceback()` return `bool` by instantiating a `LuaPopper` whose initial count is 0, but which is set to 1 on success. Up to you, though.

nat-goodspeed:
> don't allow overwriting the global table

Can we set a metatable for the global table? If so, we might consider raising an error if a Lua script (inadvertently or otherwise) attempts to assign to [the Luau library module names](https://luau.org/library): `math`, `table`, `string`, `coroutine`, `bit32`, `utf8`, `os`, `debug`, `buffer` and `vector`.

maxim-productengine:
Thanks, Nat. You're right; we shouldn't skip LL.atexit() functions. I've also added a simple demo script since there wasn't one with debug settings api. And it can be used to verify that the crash no longer occurs and atexit() functions are firing properly.

Regarding your question about the global table, I agree it sounds like a good idea. However, I'm not sure what the typical approach is in other projects, so I'll do some research.

indra/llcommon/lua_function.cpp
```diff
@@ -764,6 +764,34 @@ int lua_metaipair(lua_State* L)
...
+    lua_getfield(mState, -1, "traceback");
+    if (!lua_isfunction(mState, -1))
+    {
+        lua_pop(mState, 2);
+        LL_WARNS("Lua") << "'traceback' func not found" << LL_ENDL;
+        return false;
+    }
+    // ditch "debug"
+    lua_remove(mState, -2);
+    return true;
```
nat-goodspeed, line 792: I just realized that this function should return an `int`: 0 on failure, `lua_gettop(mState)` on success. If the return value is used in an `if` condition, it will still have the same sense. But more specifically, that allows the caller to directly capture the error function stack index to be passed to `lua_pcall()`.

indra/llcommon/lua_function.cpp
```diff
@@ -793,36 +821,25 @@ LuaState::~LuaState()
...
+        // if debug_traceback is true, stack now contains atexit, /debug.traceback()/
+        // otherwise just atexit
         for (int i(len); i >= 1; --i)
         {
             lua_pushinteger(mState, i);
-            // stack contains Registry.atexit, debug.traceback(), i
-            lua_gettable(mState, -3);
-            // stack contains Registry.atexit, debug.traceback(), atexit[i]
+            // stack contains Registry.atexit, /debug.traceback()/, i
+            lua_gettable(mState, debug_traceback ? -3 : -2);
```
nat-goodspeed, line 832: No - up around line 815, right after `if (lua_istable(mState, -1)) {`, capture `int atexit = lua_gettop(mState);`

Then you can use `atexit` here.

indra/llcommon/lua_function.cpp
```diff
@@ -793,36 +821,25 @@ LuaState::~LuaState()
...
+            // stack contains Registry.atexit, /debug.traceback()/, atexit[i]
             // Call atexit[i](), no args, no return values.
             // Use lua_pcall() because errors in any one atexit() function
             // shouldn't cancel the rest of them. Pass debug.traceback() as
             // the error handler function.
-            LL_DEBUGS("Lua") << LLCoros::getName()
-                             << ": calling atexit(" << i << ")" << LL_ENDL;
-            if (lua_pcall(mState, 0, 0, -2) != LUA_OK)
+            LL_DEBUGS("Lua") << LLCoros::getName() << ": calling atexit(" << i << ")" << LL_ENDL;
+            if (lua_pcall(mState, 0, 0, debug_traceback ? -2 : 0) != LUA_OK)
```
nat-goodspeed, line 839: Pass the error function index `traceback` returned by `push_debug_traceback()`.

indra/llcommon/lua_function.cpp
```diff
@@ -831,7 +848,8 @@ LuaState::~LuaState()
             // stack contains atexit, debug.traceback()
         }
         // pop debug.traceback()
-        lua_pop(mState, 1);
+        if(debug_traceback)
+            lua_pop(mState, 1);
```
nat-goodspeed, line 852: And this can test for nonzero `debug.traceback()` stack index, and call `lua_remove(mState, traceback)`.

indra/newview/scripts/lua/test_debugsettings.lua
```diff
@@ -0,0 +1,10 @@
+local LLDebugSettings = require 'LLDebugSettings'
+local SETTING_NAME = '360CaptureCameraFOV'
+LL.atexit(function() LL.print_info('Finished') end)
+
+LL.print_info('Updating ' .. SETTING_NAME)
+LLDebugSettings.set(SETTING_NAME, 100)
+
+debug = LLDebugSettings.get(SETTING_NAME)
```
nat-goodspeed, line 8: Let's add a comment here explaining that the point of this assignment is to "inadvertently" hide the Luau library module `debug`.

nat-goodspeed (changes requested):
Sorry it took me so long to figure out that `push_debug_traceback()` should return the stack index, but that really seems better to me.

Issue #3106, opened 2024-11-19
The viewer does not fully load all textures on the Intel HD Graphics 610 video card


DmitrySProductengine:
### Environment
<details><summary>Environment</summary>
<p>

Second Life Release 7.1.11.11864501455 (64bit)
Release Notes

You are at 3.7, 127.9, 85.3 in LagLand located at simhost-07a82ae0a96565d36.agni
SLURL: http://maps.secondlife.com/secondlife/LagLand/4/128/85
(global coordinates 260,100.0, 248,192.0, 85.3)
Second Life Server 2024-10-15.11356152186
Release Notes

CPU: Intel(R) Celeron(R) CPU G3930 @ 2.90GHz (2904.01 MHz)
Memory: 3522 MB
OS Version: Microsoft Windows 11 64-bit (Build 22631.4460)
Graphics Card Vendor: Intel
Graphics Card: Intel(R) HD Graphics 610

Windows Graphics Driver Version: 31.0.101.2128
OpenGL Version: 4.6.0 - Build 31.0.101.2128

Window size: 1920x1009
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 96m
Bandwidth: 3000kbit/s
LOD factor: 1.125
Render quality: 1
Texture memory: 1584MB
Disk cache: Max size 1638.4 MB (9.2% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: Undefined
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/1,357 (0.0%)
November 19 2024 02:47:03

</p>
</details> 

### Reproduction steps

1. Download and install the Second Life 7.1.11.11864501455 viewer (https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%23c654dac-ExtraFPS).
2. Log in to the viewer on a computer with an Intel HD Graphics 610 video card.
3. Teleport to a region with many textures (for example: LagLand - http://maps.secondlife.com/secondlife/LagLand/4/128/85).
4. Wait some time for the textures to load.
5. Observe the loaded textures.

**Actual result:**
The viewer does not fully load all textures on the Intel HD Graphics 610 video card.

**Expected result:**
[...]

PR #3207, opened 2024-12-04
#3106 Lower maximum texture bias back to 4, reintroduce camera importance texture scaling

RunitaiLinden:


brad-linden (approved):
looks good

PR #3208, opened 2024-12-04
viewer-private#330 Fix Vintage mode in feature table

akleshchev:
Windows only issue.
These parameters are related to RenderDisableVintageMode, either make them follow Vintage in every mode or remove entirely and let vintage set them.


akleshchev:
Added fix for couple cached controls, they weren't saving.

marchcat (approved):

Issue #3210, opened 2024-12-04
Don't load higher resolution textures than requested


RunitaiLinden:
### Description

The log file is full of "LLViewerTextureList::updateImagesCreateTextures : ONCE (3300th time seen): Texture will be downscaled immediately after loading."

This indicates a higher resolution image was loaded than was requested.  Find the cause and eliminate it.

RunitaiLinden:
Fixed in ExtraFPS

There was an "optimization" in the texture fetcher that would load a texture at full resolution if all the bytes were available.  The viewer would also sometimes request a higher resolution image, then determine a lower resolution was desired before the higher resolution decode completes. 

Fix was to adjust the logic so that we never decode a higher resolution than requested, and if a higher resolution decode than currently desired completes, don't use the high res version. 

A future optimization might attempt to abort the request in progress if the viewer detects it doesn't want the higher resolution image before the decode is complete.

maxim-productengine:
We won't try anymore to downscale texture below level 5, so there should be less log messages.

DmitrySProductengine:
Passed QA.
Verified on the Second Life Release 7.1.12.13509103967 (64bit) on Win11/OSX in the scope of [IQA #451](https://github.com/secondlife/iqa/issues/451).
Tested by visiting two locations with a large number of people for one login:
1. http://maps.secondlife.com/secondlife/Blue%20Moon%20Bay/183/64/27
2. http://maps.secondlife.com/secondlife/London%20City/71/194/23

As described - the number of entries in the log decreased:
On OSX - 10 entries (versus 37 in Sergey's log).
On Win11 - 11 entries.
See attached logs:
1. [7.1.12.13509103967_OSX_SecondLife.log](https://github.com/user-attachments/files/18962857/7.1.12.13509103967_OSX_SecondLife.log)
2. [7.1.12.13509103967_Win11_SecondLife.log](https://github.com/user-attachments/files/18962864/7.1.12.13509103967_Win11_SecondLife.log)

PR #3212, opened 2024-12-05
#3210 Fix for "Texture will be downscaled" happening too often.

RunitaiLinden:


indra/newview/llviewertexture.cpp
```diff
@@ -559,7 +559,10 @@ void LLViewerTexture::updateClass()
         // lower discard bias over time when free memory is available
         if (sDesiredDiscardBias > 1.f && over_pct < 0.f)
         {
-            sDesiredDiscardBias -= gFrameIntervalSeconds * 0.01f;
+            static LLCachedControl<F32> high_mem_discard_increment(gSavedSettings, "RenderHighMemMinDiscardDecrement", .1f);
```
akleshchev, line 562: Typo:
high_mem_discard_**increment**

Also won't it be better to make this dynamic? The more memory is free the faster bias falls?

RunitaiLinden: See line below, it is faster when more memory is free (over_pct is negative), fixing typo...

indra/newview/llviewertexturelist.cpp
```diff
@@ -1066,13 +1066,26 @@ F32 LLViewerTextureList::updateImagesCreateTextures(F32 max_time)
...
+
+        // desired discard may change while an image is being decoded. If the texture in VRAM is sufficient
+        // for the current desired discard level, skip the texture creation.  This happens more often than it probably
+        // should
+        bool redundant_load = imagep->hasGLTexture() && imagep->getDiscardLevel() <= imagep->getDesiredDiscardLevel();
+
+        if (!redundant_load)
+        {
+           imagep->createTexture();
+        }
```
akleshchev, line 1078: May be add else LL_DEBUGS() here?

indra/newview/llviewertexture.cpp
akleshchev: Sorry, somehow missed that.

indra/newview/llviewertexturelist.cpp
```diff
@@ -1066,13 +1066,26 @@ F32 LLViewerTextureList::updateImagesCreateTextures(F32 max_time)
...
+        
         imagep->postCreateTexture();
         imagep->mCreatePending = false;
         mCreateTextureList.pop();
 
         if (imagep->hasGLTexture() && imagep->getDiscardLevel() < imagep->getDesiredDiscardLevel())
         {
+            // NOTE: this may happen if the desired discard reduces while a decode is in progress and does not
+            // necessarily indicate a problem, but if log occurrences excede that of dsiplay_stats: FPS,
+            // something has probably gone wrong.
```
akleshchev, line 1088: Scaling is new, so might be a good idea to add some kind of debug logging here, that shows Id and by how much image gets scaled.
Or compare to mDebugId then log details.

And texture in a 'scaled' state probably should be represented in texture console somehow.

Issue #2922, opened 2024-10-23
Add 'Create Landmark' above 'Copy SLurl...' in the chat context menu


LLGuru:
* Context menu for the SLURL in the chat should contain an item for creating a landmark

* New menu item should be placed above the "Copy SLurl to clipboard" menu item
![Image](https://github.com/user-attachments/assets/fc669868-bbf6-4e99-9f67-1ed2e925eb41)

* It should open the same floater as we see when clicking the star in the viewer address bar:
![Image](https://github.com/user-attachments/assets/e711369a-d6cb-4250-ac83-05274affd723)

LLGuru:
- Implemented in https://github.com/secondlife/viewer/pull/3217

jellyfish-linden:
I didn't see this in 26.2.0.24254827122. Is this in that version?

PR #3217, opened 2024-12-05
#2922 Add 'Create Landmark' to the chat context menu

LLGuru:
Using new capabily "CreateLandmarkForPosition" introduced in https://github.com/secondlife/server/pull/1553

### DIFF
--- a/indra/newview/lllandmarkactions.cpp
+++ b/indra/newview/lllandmarkactions.cpp
@@ -39,6 +39,7 @@
 
 #include "llagent.h"
 #include "llagentui.h"
+#include "llfloaterreg.h"
 #include "llinventorymodel.h"
 #include "lllandmarklist.h"
 #include "llslurl.h"
@@ -46,6 +47,7 @@
 #include "llviewerinventory.h"
 #include "llviewerparcelmgr.h"
 #include "llworldmapmessage.h"
+#include "llviewernetwork.h"
 #include "llviewerwindow.h"
 #include "llwindow.h"
 #include "llworldmap.h"
@@ -233,13 +235,13 @@ void LLLandmarkActions::createLandmarkHere(
     const std::string& desc,
     const LLUUID& folder_id)
 {
-    if(!gAgent.getRegion())
+    if (!gAgent.getRegion())
     {
         LL_WARNS() << "No agent region" << LL_ENDL;
         return;
     }
-    LLParcel* agent_parcel = LLViewerParcelMgr::getInstance()->getAgentParcel();
-    if (!agent_parcel)
+
+    if (!LLViewerParcelMgr::getInstance()->getAgentParcel())
     {
         LL_WARNS() << "No agent parcel" << LL_ENDL;
         return;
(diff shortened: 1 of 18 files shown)


marchcat (approved):

akleshchev (changes requested):
Hide or at least disable new UI items if CreateLandmarkForPosition is not available.

Edit, nevermind, I see it, the process is a bit colvoluted.

indra/newview/llviewerinventory.cpp
```diff
@@ -1230,6 +1230,61 @@ void create_inventory_settings(const LLUUID& agent_id, const LLUUID& session_id,
         static_cast<U8>(settype), next_owner_perm, cb);
 }
 
+void create_inventory_landmark(const LLUUID& folder_id, const std::string& name, const std::string& desc,
+    const std::string& region_name, S32 x, S32 y, S32 z)
+{
+    if (LLApp::isExiting())
+        return;
+
+    LLViewerRegion* viewer_region = gAgent.getRegion();
```
akleshchev, line 1239: you can use gAgent.getRegionCapability()

indra/newview/skins/default/xui/en/menu_url_slurl.xml
```diff
@@ -27,6 +27,15 @@
     </menu_item_call>
     <menu_item_separator
      layout="topleft" />
+    <menu_item_call
+     label="Create Landmark for SLurl"
+     layout="topleft"
+     name="landmark">
+        <menu_item_call.on_click
+         function="Url.CreateLandmark" />
+        <menu_item_call.on_enable
```
akleshchev, line 36: It's better to hide new element instead of simply disabling it. Users will be confused why element is disabled and will look for a way to enable it, better for it to not be seen untill feature is deployed server side.

indra/newview/llviewerinventory.cpp
```diff
@@ -1230,6 +1230,61 @@ void create_inventory_settings(const LLUUID& agent_id, const LLUUID& session_id,
...
+            LLCore::HttpStatus status = LLCoreHttpUtil::HttpCoroutineAdapter::getStatusFromLLSD(httpResults);
+
+            if (!status)
+            {
+                LL_WARNS() << "Error " << status.getType() << ": '" << status.toString() << "'" << LL_ENDL;
+                LLNotificationsUtil::add("CantCreateLandmark");
+            }
+            else if (result.has("item_id"))
+            {
+                LL_INFOS() << "Created item id: '" << result["item_id"] << "'" << LL_ENDL;
```
akleshchev, line 1284: Might be a good idea to pop inventory or landmark panek open and highligh new landmark, like we do when we upload something.

indra/newview/lllandmarkactions.cpp
```diff
@@ -265,6 +267,63 @@ void LLLandmarkActions::createLandmarkHere()
...
+        LL_INFOS() << "Unsupported URL: '" << url << "'" << LL_ENDL;
+        return;
+    }
+
+    LLSD data;
+    data["region"] = slurl.getRegion();
+    data["x"] = (S32)std::round(slurl.getPosition()[VX]);
+    data["y"] = (S32)std::round(slurl.getPosition()[VY]);
+    data["z"] = (S32)std::round(slurl.getPosition()[VZ]);
+    data["title"] = title.empty() ? slurl.getRegion() : title;
```
akleshchev, line 285: May be use coordinates as a part of a tittle if you are using region's name? I suggest discussing this behavior wit vir. I understad that title might not be unavailable yet, but may be server can fill it in? create_inventory_landmark will be done anyway so might be a good chance to grab a tittle.

akleshchev (approved):

indra/newview/skins/default/xui/en/menu_url_slurl.xml
LLGuru: I think the same way, and I started from the attribute is_visible, but it is not supported here

indra/newview/llviewerinventory.cpp
LLGuru: Yes, but this function hides the actual problem: no region or no cap support by the region

indra/newview/llviewerinventory.cpp
LLGuru: Maybe. This is a question for UX experts (like @Steeltoe-Linden)

indra/newview/lllandmarkactions.cpp
```diff
@@ -265,6 +267,63 @@ void LLLandmarkActions::createLandmarkHere()
...
+        return false; // Cap is not supported by the region
+
+    LLSLURL slurl(url);
+    if (slurl.getType() == LLSLURL::LOCATION) // LLUrlEntryPlace
+    {
+        if (LLGridManager* gm = LLGridManager::getInstance())
+        {
+            std::string current_grid = gm->getGrid();
+            std::string grid_from_url = gm->getGrid(slurl.getGrid());
+            if (grid_from_url == current_grid) // The same grid location
```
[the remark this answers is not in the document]
LLGuru: Maybe. This is a question for UX experts (like @Steeltoe-Linden)

indra/newview/skins/default/xui/en/menu_url_slurl.xml
akleshchev: menu_item_call.on_visible is used just fine in other places, how can it be not supported?

PR #3235, opened 2024-12-09
Adding C++23 compatibility

AiraYumi:
Test that it builds with gcc-13 and C++23.
Fix any errors that occur.

### DIFF
--- a/indra/newview/llpanelface.h
+++ b/indra/newview/llpanelface.h
@@ -57,6 +57,18 @@ class LLRadioGroup;
 class PBRPickerAgentListener;
 class PBRPickerObjectListener;
 
+// This is specializations are needed to support std::unique_ptr<T> where T is an incomplete type
+namespace std {
+    template<> struct default_delete<PBRPickerAgentListener> {
+        constexpr default_delete() noexcept = default;
+        void operator()(PBRPickerAgentListener* ptr) const noexcept;
+    };
+    template<> struct default_delete<PBRPickerObjectListener> {
+        constexpr default_delete() noexcept = default;
+        void operator()(PBRPickerObjectListener* ptr) const noexcept;
+    };
+}
+
 // Represents an edit for use in replicating the op across one or more materials in the selection set.
 //
 // The apply function optionally performs the edit which it implements
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -72,7 +72,7 @@ jobs:
       # autobuild-package.xml.
       AUTOBUILD_VCS_INFO: "true"
       AUTOBUILD_VSVER: "170"
-      DEVELOPER_DIR: "/Applications/Xcode_16.0.app/Contents/Developer"
+      DEVELOPER_DIR: "/Applications/Xcode_16.1.app/Contents/Developer"
       # Ensure that Linden viewer builds engage Bugsplat.
       BUGSPLAT_DB: ${{ needs.setup.outputs.bugsplat_db }}
       build_coverity: false
--- a/indra/CMakeLists.txt
+++ b/indra/CMakeLists.txt
@@ -29,7 +29,7 @@ else()
   set( USE_AUTOBUILD_3P ON )
 endif()
 
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include(Variables)
--- a/indra/newview/llpanelface.cpp
+++ b/indra/newview/llpanelface.cpp
@@ -1838,6 +1838,7 @@ public:
         mChangePending = false;
     }
 };
+void std::default_delete<PBRPickerAgentListener>::operator()(PBRPickerAgentListener * ptr) const noexcept { delete ptr; }
 
 // One-off listener that updates the build floater UI when the prim inventory updates
 class PBRPickerObjectListener : public LLVOInventoryListener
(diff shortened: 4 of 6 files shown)


nat-goodspeed (approved):

nat-goodspeed:
...but it looks as though there's more yet to fix.

AiraYumi:
It seems that Xcode 16 is needed to fix char8_t compatibility in c++23...
https://developer.apple.com/xcode/cpp/

The cause is not immediately clear. I will try disabling C++23 when building llcommon on my mac.

nat-goodspeed:
Oh yes! Definitely use Xcode 16 on Mac for C++20 and beyond.

AiraYumi:
I gave up on enabling C++23 by default on macOS because I couldn’t build it even with the latest beta version of Xcode 16.1.

AiraYumi:
maybe fixed.

indra/newview/llpanelface.h
```diff
@@ -57,6 +57,18 @@ class LLRadioGroup;
 class PBRPickerAgentListener;
 class PBRPickerObjectListener;
 
+// Need default_delete Template for C++23
```
AiraYumi, line 60: This part.
Is there a better way?

indra/newview/llpanelface.h
```diff
@@ -57,6 +57,18 @@ class LLRadioGroup;
...
+namespace std {
+    template<> struct default_delete<PBRPickerAgentListener> {
+        constexpr default_delete() noexcept = default;
+        void operator()(PBRPickerAgentListener* ptr) const noexcept;
+    };
+    template<> struct default_delete<PBRPickerObjectListener> {
+        constexpr default_delete() noexcept = default;
+        void operator()(PBRPickerObjectListener* ptr) const noexcept;
+    };
+}
```
AiraYumi, line 70: This part.
Is there a better way?

indra/newview/llpanelface.h
```diff
@@ -57,6 +57,18 @@ class LLRadioGroup;
...
+
```
AiraYumi, line 71: This part.
Is there a better way?

nat-goodspeed: Sorry for intermittent attention; I'm on Christmas holidays.

If I understand the problem and your solution, it's that `PBRPickerAgentListener` and `PBRPickerObjectListener` are incomplete types, so the default `std::default_delete`, which tries to `delete` the referents of those `std::unique_ptr`s, can't be instantiated in a translation unit that only sees the header file.

Your specializations of `std::default_delete` work around that by putting their `operator()` methods in the .cpp file, so the referents of the `std::unique_ptr`s can be destroyed without publishing their classes in the .h file.

The solution that comes to my mind would be to ensure that `~LLPanelFace()` is in the .cpp file - but it already is. Then I recalled that any assignment to either of those `unique_ptr`s would need to engage `default_delete` - but I saw no such assignment in the header file.

Maybe the issue arises from the memberwise assignments in the default copy constructor, copy assignment, move constructor and move assignment methods? I don't see explicit mention of any of those in the declaration of `LLPanelFace`, so presumably the compiler generates the default ones.

Have you tried declaring those four methods as `= delete` (without your `std::default_delete` specializations)?

AiraYumi: I haven't tried it yet. I will.

AiraYumi: I tried to define "operator()" within LLPanelFace, but a compilation error occurred.
Also, it seems that neither default nor delete can be defined for operator.

nat-goodspeed: I'm sorry, I seem to have confused things.

What I meant to suggest was that you try adding to the `LLPanelFace` class declaration:
```
LLPanelFace(const LLPanelFace&) = delete;  // copy constructor
LLPanelFace(LLPanelFace&&) = delete;       // move constructor
LLPanelFace& operator=(const LLPanelFace&) = delete;  // copy assignment operator
LLPanelFace& operator=(LLPanelFace&&) = delete;       // move assignment operator
```
Without explicit declarations, the compiler will generate default inline copy and move constructors and assignment operators, which operate on individual members. `LLPanelFace` contains a couple of `std::unique_ptr` members. Assigning to a `std::unique_ptr` requires deleting its previous referent, if any. So it could be that the generated assignment operators are responsible for the original incomplete-type errors.

With those declarations in place, you may discover other errors from some code (inadvertently?) trying to copy or move an `LLPanelFace` instance. If those actions are erroneous, you can fix them. If they're legit, you might have to code an actual copy or move assignment operator, placing the method definition in the .cpp file.

AiraYumi: i wil try.

AiraYumi: Simply adding it to the original code did not fix the error.

nat-goodspeed: Thank you for testing that.

(And Happy New Year!)

AiraYumi: I was able to build after moving the definitions of PBR PickerAgentListener and PBR PickerObjectListener to .h.
Which is better, this or the current fix?

nat-goodspeed: I think your original solution, specializing `std::default_delete`, is better than moving the `PBRPicker*Listener` classes into the header file.

Please expand the comment at line 60 to explain that these specializations are needed to support `std::unique_ptr<T>` where `T` is an incomplete type: deleting, or assigning to, either of those `std::unique_ptr` members requires `unique_ptr` to delete its referent. Moving `std::default_delete<T>::operator()` to the .cpp file, where each `T` is defined, addresses the problem.

I suppose another alternative would be to declare each of those `std::unique_ptr` instances with a custom deleter whose implementation is likewise in the .cpp file. That might require a few less lines of code, but would still call for explanatory comments.

nat-goodspeed (approved):

indra/newview/llpanelface.h
AiraYumi: added

nat-goodspeed:
Thank you! At this point we're trying to decide how to establish to our satisfaction that there are no negative performance impacts across the range of hardware we support.

brad-linden:
I think we may be able to take the c++ changes in the interim while we are figuring out the testing needs

Issue #3251, opened 2024-12-10
Remove obsolete intel driver notification


RunitaiLinden:
### Description

When launching the viewer with an Intel HD 4000 on Windows 10, a notification may be popped up that asks you to update your drivers.  This notification doesn't display properly on all systems and can block users from logging in, and no driver update is necessary (or even available in many cases).  

Remove the notification.

RunitaiLinden:
Not sure this is testable without a Dell Latitude E6530 or similar.

igorlproductengine:
Passed QA.
Verified on the Second Life Release 7.1.11.12306234817 (64bit) in the scope of https://github.com/secondlife/iqa/issues/411 on Win10.
This issue was verified on the Graphics Card: Intel(R) HD Graphics 4000.

PR #3252, opened 2024-12-10
#3251 Remove obsolete intel driver notification

RunitaiLinden:


akleshchev (approved):

akleshchev:
I'm surprised this even triggers, aren't most driver version 30+ nowadays?

I think originally this was added because old drivers couldn't run 64bit viewer. It worked with 32bit viewer but it was better to encourage transition to 64 bit than to install 32 bit.

PR #3262, opened 2024-12-11
Fix mesa failing to link shaders due to missing vertex shader outputs

RyeMutt:
This ensures the vertex shader outputs for `atmosphericsVarsF.glsl` from `atmosphericsVarsV.glsl` are available at shader link time to avoid error state in mesa 24.3.x

https://jira.firestormviewer.org/browse/FIRE-34877

### DIFF
--- a/indra/llrender/llshadermgr.cpp
+++ b/indra/llrender/llshadermgr.cpp
@@ -79,7 +79,7 @@ bool LLShaderMgr::attachShaderFeatures(LLGLSLShader * shader)
     //////////////////////////////////////
 
     // NOTE order of shader object attaching is VERY IMPORTANT!!!
-    if (features->calculatesAtmospherics)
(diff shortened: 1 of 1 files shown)


indra/llrender/llshadermgr.cpp
```diff
@@ -79,7 +79,7 @@ bool LLShaderMgr::attachShaderFeatures(LLGLSLShader * shader)
     //////////////////////////////////////
 
     // NOTE order of shader object attaching is VERY IMPORTANT!!!
-    if (features->calculatesAtmospherics)
+    if (features->calculatesAtmospherics || features->hasGamma || features->isDeferred)
```
brad-linden, line 82: shouldn't isDeferred always be on now?

RyeMutt: It's a separate shader feature flag controlling deferredUtil shader helpers being linked or not.

brad-linden (approved):

PR #3264, opened 2024-12-11
more commands in test_LLChatListener.lua

vir-linden:
Demo prep

### DIFF
--- a/indra/newview/scripts/lua/test_LLChatListener.lua
+++ b/indra/newview/scripts/lua/test_LLChatListener.lua
@@ -1,25 +1,49 @@
 local LLAgent = require 'LLAgent'
+local LLAppearance = require 'LLAppearance'
 local LLChat = require 'LLChat'
 local LLListener = require 'LLListener'
 local UI = require 'UI'
 
 -- Chat listener script allows to use the following commands in Nearby chat:
--- open inventory   -- open defined floater by name
--- close inventory  -- close defined floater by name
--- closeall         -- close all floaters
--- stop             -- close the script
--- any other messages will be echoed.
-function openOrEcho(message)
-    local open_floater_name = string.match(message, "^open%s+(%w+)")
-    local close_floater_name = string.match(message, "^close%s+(%w+)")
-    if open_floater_name then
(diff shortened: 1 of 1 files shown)


indra/newview/scripts/lua/test_LLChatListener.lua
```diff
@@ -1,25 +1,49 @@
...
+    print("handling message " .. message)
+    local cmd, arg = string.match(message, '^/(%w+)%s+(.*)$')
+    if arg then
+      -- remove quotes from arg
+      arg = string.gsub(arg, '"', '')
     else
-      LLChat.sendNearby('Echo: ' .. message)
+      cmd = string.match(message, '^/(%w+)')
+    end
+    print("cmd: " .. (cmd or '') .. " arg: " .. (arg or ''))
```
nat-goodspeed, line 23: This would be a good use for Luau's string interpolation (using backticks):
```
print(`cmd: {cmd} arg: {arg}`)
```
In Lua, when you're not trying to concatenate `nil` with a string, `nil` stringizes as "nil".

indra/newview/scripts/lua/test_LLChatListener.lua
```diff
@@ -32,11 +56,14 @@ function listener:handleMessages(event_data)
       return true
     elseif string.sub(event_data.message, 1, 5) == '[LUA]' then
       return true
-    elseif event_data.message == 'stop' then
+    elseif event_data.message == '/stop' then
       LLChat.sendNearby('Closing echo script.')
       return false
     else
-      openOrEcho(event_data.message)
+      ok, ret = pcall(processCommands, event_data.message)
```
nat-goodspeed, line 63: yes, exactly

nat-goodspeed (approved):
This is good. It's tempting to reorganize it as a table of functions:
```
commands = {
    open=function(arg),
        -- ...
    end,
    close=function(arg),
        -- ...
    end,
    ...
}
```
and then generate the `/help` text by running through the keys, but this works.

fwiw Lua indicates table lookup failure only by returning `nil`.

PR #3268, opened 2024-12-11
Add photo-demo.lua to take snapshots at various graphics quality settings

nat-goodspeed:
The other commits are support features used by photo-demo.lua.

### DIFF
--- a/indra/newview/lluilistener.cpp
+++ b/indra/newview/lluilistener.cpp
@@ -147,6 +147,11 @@ LLUIListener::LLUIListener():
         &LLUIListener::uploadLocalTexture,
         llsd::map("filename", LLSD::String(), "reply", LLSD()));
 
+    add("getGraphicsQuality",
+        "Return graphics quality \"level\": from 0 (Low) to 6 (Ultra)",
+        &LLUIListener::getGraphicsQuality,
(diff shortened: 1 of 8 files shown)


indra/newview/lluilistener.cpp
```diff
@@ -443,6 +448,11 @@ void LLUIListener::uploadLocalTexture(const LLSD& event) const
     });
 }
 
+LLSD LLUIListener::getGraphicsQuality(LLSD const& request)
```
maxim-productengine, line 451: We can use `LLDebugSettings.get('RenderQualityPerformance')` in Lua

maxim-productengine (approved):

indra/newview/lluilistener.cpp
nat-goodspeed: You're right. It would be cleaner to revert the change I added.

nat-goodspeed: https://github.com/secondlife/viewer/pull/3268/commits/ca5b1e4ea7aa8cb012356c44133fa5037d202449

maxim-productengine: ship it

Issue #3272, opened 2024-12-12
[MAC] The 'RenderReflectionsEnabled' setting is set to True on Low graphics setting


irinabproductengine:
### Environment

<details><summary>Environment</summary>
<p>

Second Life Release 7.1.11.12263287958 (64bit)
Release Notes

You are at 32.3, 240.3, 38.2 in Danger! Danger! located at simhost-04bcfd96634d43e62.agni
SLURL: http://maps.secondlife.com/secondlife/Danger%21%20Danger%21/32/240/38
(global coordinates 261152.0, 249840.0, 38.2)
Second Life Server 2024-11-25.12013542687
Release Notes

CPU: Apple M1 Pro (2400 MHz)
Memory: 32768 MB
OS Version: macOS 15.1.1 Darwin 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:15 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6000 x86_64
Graphics Card Vendor: Apple
Graphics Card: Apple M1 Pro

OpenGL Version: 4.1 Metal - 89.3

Window size: 1712x968
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 64m
Bandwidth: 3000kbit/s
LOD factor: 1.125
Render quality: 0
Texture memory: 21845MB
Disk cache: Max size 1638.4 MB (100.0% used)
HiDPI display mode: 

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091638
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 39/29658 (0.1%)
December 12 2024 03:49:27

</p>
</details> 

### Reproduction steps

1. Login to the viewer with recommended settings.
2. Set Graphics to Low.
3. Open Debug settings (Advanced menu -> Show Debug Settings).
4. Input 'RenderReflectionsEnabled'.

Actual result: The 'RenderReflectionsEnabled' setting is set to True.
Expected result: The 'RenderReflectionsEnabled' setting should be set to False on the Low graphics setting.

[...]

PR #3274, opened 2024-12-12
#3272 [MAC] Fix wrong RenderReflectionsEnabled on Low

akleshchev:


marchcat (approved):

marchcat:
So that's the one responsible for lower performance on Low setting compared to higher ones?

akleshchev:
It's also On on higher settings, unlikely to be the cause.

maxim-productengine (approved):

PR #3277, opened 2024-12-12
#2590 Radeon mac optimization pass

RunitaiLinden:
This should:
- Skip updating of reflection probes that are not the default probe when probe coverage is set to "None"
- enable RenderAppleUseMultGL and disable occlusion culling on Macs with AMD GPUs
- Reduce the number of texture decode threads on Macs with intel cpus.
- Move texture deletion to LLImageGL::updateClass and prevent textures from staying resident in vram longer than 3 frames
- Disable SSAO by default on Macs with intel CPUs

### DIFF
--- a/indra/newview/llappviewer.cpp
+++ b/indra/newview/llappviewer.cpp
@@ -2171,7 +2171,12 @@ bool LLAppViewer::initThreads()
 
     // get the number of concurrent threads that can run
     S32 cores = std::thread::hardware_concurrency();
-
+#if LL_DARWIN
+    if (!gGLManager.mIsApple)
+    {
+        cores /= 2;
+    }
+#endif
     U32 max_cores = gSavedSettings.getU32("EmulateCoreCount");
     if (max_cores != 0)
     {
--- a/indra/newview/llviewerwindow.cpp
+++ b/indra/newview/llviewerwindow.cpp
@@ -5391,6 +5391,8 @@ bool LLViewerWindow::cubeSnapshot(const LLVector3& origin, LLCubeMapArray* cubea
         camera->setUserClipPlane(clipPlane);
     }
 
+    gPipeline.pushRenderTypeMask();
+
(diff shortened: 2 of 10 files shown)


indra/newview/llappviewer.cpp
```diff
@@ -2171,7 +2171,12 @@ bool LLAppViewer::initThreads()
 
     // get the number of concurrent threads that can run
     S32 cores = std::thread::hardware_concurrency();
-
+#if LL_DARWIN
+    if (!gGLManager.mIsApple)
+    {
+        cores /= 2;
```
brad-linden, line 2177: is this an SMT/hyperthreading workaround?  is there a way to detect that?

brad-linden (approved):
looks good, testing now...

RyeMutt (approved):

indra/newview/llappviewer.cpp
RunitaiLinden: Pretty much, yeah... I don't know of a way to detect that... on Windows we don't seem to get thread starvation as much when treating virtual cores as real cores, and all seems fine on Apple silicon.  Note that the viewer process still gets blocked pretty hard sometimes (all threads pause), but with that change I don't see what looks like thread starvation when decoding textures.

indra/newview/llviewerwindow.cpp
```diff
@@ -5479,16 +5481,7 @@ bool LLViewerWindow::cubeSnapshot(const LLVector3& origin, LLCubeMapArray* cubea
...
-    if (!dynamic_render)
-    {
-        for (int i = 0; i < dynamic_render_type_count; ++i)
-        {
-            if (prev_dynamic_render_type[i])
-            {
-                gPipeline.toggleRenderType(dynamic_render_types[i]);
-            }
-        }
-    }
```
RunitaiLinden, line 5491: whoops... this shouldn't have bee nremoved

Issue #3288, opened 2024-12-13
BugSplat Crash: SecondLifeViewer!LLGLTFMaterial::operator=(117)


sl-service-account:
### Test plan 1
1. Create a box. 
2. While the box is still being edited, drag a texture from Inventory onto the top face of the box.
3. Verify the viewer does not crash.

### Test plan 2
1. Create a box.
2. Select the Texture tab in the Edit Tools
3. Under Material, select PBR Metallic Roughness
4. Click Choose from inventory
5. In the PICK:TEXTURE floater, click a material in the Inventory, then click the Cancel button.
6. Verify the viewer does not crash.

**BugSplat** [Key Crash 97662](https://app.bugsplat.com/v2/keycrash?stackKeyId=97662&database=SecondLife_Viewer_2018)

**Sample Version:**
7.1.11.12306234817

**Sample Callstack:**
[...]

PR #3289, opened 2024-12-13
#3288 null overrides crashes

brad-linden:
Fixes for secondlife/viewer#3288 override copy related crashes when overrides can be nullptr


brad-linden:
recommend hiding whitespace when viewing this

akleshchev (approved):

marchcat (approved):

PR #3297, opened 2024-12-16
Roll back to Xcode 14.0.1 to fix secondlife/viewer-private#347

brad-linden:
secondlife/viewer-private#347

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,7 +42,7 @@ jobs:
     needs: setup
     strategy:
       matrix:
-        runner: [windows-large, macos-13-xlarge]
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

akleshchev:
How is mac build related to inventory? Is there some code that is ambigious and needs updating?

brad-linden:
there is definitely a deeper problem here.  we will need to get a better fix going forward.

marchcat:
Wow, that was unexpected. Can it be something related to the Apple silicon runner vs. Intel?

brad-linden:
the macos 13 runners were still intel I think.  I am continuing to investigate with a focus on libc++ changes and container usage

PR #3313, opened 2024-12-19
Merge ExtraFPS develop into maxim/potato

nat-goodspeed:
This PR is to build and verify functionality before completing the merge.


nat-goodspeed:
On successful build, I ran the `photo-demo-compare.lua` and `liz-demo.lua` scripts, which behaved as they ought.

PR #3320, opened 2024-12-24
Remove obsolete cmake_minimum_required that is lower than the required version in the main CMakeLists.txt

Ansariel:
Not only is the additional minimum required version statement useless, but it also triggers a deprecation warning in CMake 3.31.3:
`Compatibility with CMake < 3.10 will be removed from a future version of CMake.`


marchcat (approved):
Right, thank you!

PR #3321, opened 2024-12-27
Add Analyze Compile Flags

AiraYumi:
Adds compilation options for debugging.

### DIFF
--- a/indra/cmake/00-Common.cmake
+++ b/indra/cmake/00-Common.cmake
@@ -176,6 +176,14 @@ if (LINUX)
             -lm
     )
   endif()
+
+  if (CMAKE_BUILD_TYPE MATCHES "RelWithDebInfo")
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+      add_compile_options(--analyze)
+    elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+      add_compile_options(-fanalyzer)
+    endif ()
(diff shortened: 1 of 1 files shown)


indra/cmake/00-Common.cmake
```diff
@@ -199,6 +207,10 @@ if (DARWIN)
   # required for clang-15/xcode-15 since our boost package still uses deprecated std::unary_function/binary_function
   # see https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#C++-Standard-Library
   add_compile_definitions(_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION)
+
+  if (CMAKE_BUILD_TYPE REGEX "RelWithDebInfo")
```
marchcat, line 211: Mac build fails here. Should it be `if( CMAKE_BUILD_TYPE MATCHES "RelWithDebInfo" )`?

RyeMutt:
#3071 Also introduces support for sanitizers which I hope to have ready to merge soon.

indra/cmake/00-Common.cmake
AiraYumi: fix it.

marchcat: Thank you!

marchcat (approved):
@RyeMutt, since your change is not ready yet, can we merge this now?

Issue #3322, opened 2025-01-02
No confirmation is shown when gifting all your money


LLGuru:
### Environment

Second Life Release 7.1.11.12363455226 (64bit)
And I suppose any other release contains this bug

### Description

There is a no-confirmation pay-amount limit (L$ 200)
When the gifting sum exceeds this limit, the confirmation is shown
![Image](https://github.com/user-attachments/assets/993b41f2-678b-43a7-abe8-782f4d230883)

But when the gifting sum is equal to the total balance, the money go away without any confirmation,

**The payment confirmation should be shown when the amount exceeds L$ 200 even if this amount is equal to the total balance**

### Reproduction steps

1. Try to pay L$ to the resident (amount > 200 and amount < total)
2. Observe the confirmation
3. Press the "Cancel" button
4. Try to pay L$ to the resident (amount > 200 and amount == total)

Expected behavior: The confirmation is shown
Actual behavior: No confirmation is shown

LLGuru:
- Fixed in https://github.com/secondlife/viewer/pull/3323

Dan-Linden:
Passed QA.
Verified on Second Life Release 26.2.0.21655710006 on Win11/OSX in the scope of https://github.com/secondlife/iqa/issues/606

PR #3323, opened 2025-01-02
#3322 No confirmation is shown when gifting all your money

LLGuru:


marchcat (approved):
This looks reasonable, but shouldn't it still be like `gStatusBar->getBalance() >= amount`?

LLGuru:
> This looks reasonable, but shouldn't it still be like `gStatusBar->getBalance() >= amount`?

I believe we should ask for a confirmation even if the balance is not enough
(in this case the user will be prompted to top up his balance)

First, when the amount is big, it makes sense to always double-check the target and sum
Second, just the same balance and amount comparison is being done when processing the operation

akleshchev (approved):

PR #3338, opened 2025-01-03
Server#1036 A blocked user should not be able to gift L$

LLGuru:
Viewer-part  of the solution for https://github.com/secondlife/server/issues/1036

### DIFF
--- a/indra/newview/skins/default/xui/en/notifications.xml
+++ b/indra/newview/skins/default/xui/en/notifications.xml
@@ -12030,6 +12030,14 @@ Money transfers to objects are currently disabled in this region.
 
   <notification
    icon="alertmodal.tga"
+   name="CantTransferMoneyAgentIsBlocked"
+   type="notify">
+   <tag>fail</tag>
+Unable to make payment due to the agent being blocked by the recipient.
+  </notification>
+
+  <notification
+   icon="alertmodal.tga"
    name="DroppedMoneyTransferRequest"
    type="notify">
    <tag>fail</tag>
--- a/indra/newview/llmutelist.cpp
+++ b/indra/newview/llmutelist.cpp
@@ -367,8 +367,8 @@ void LLMuteList::updateAdd(const LLMute& mute)
     msg->nextBlockFast(_PREHASH_MuteData);
     msg->addUUIDFast(_PREHASH_MuteID, mute.mID);
     msg->addStringFast(_PREHASH_MuteName, mute.mName);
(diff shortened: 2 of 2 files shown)


indra/newview/skins/default/xui/en/notifications.xml
```diff
@@ -12028,6 +12028,14 @@ Unable to find object.
 Money transfers to objects are currently disabled in this region.
   </notification>
 
+  <notification
+   icon="alertmodal.tga"
+   name="CantTransfterMoneyAgentIsBlocked"
+   type="notify">
+   <tag>fail</tag>
+Unable to make payment due to agent is blocked by the recipient.
```
marchcat, line 12036: "Unable to make payment due to the agent being blocked by the recipient." would be smoother.

marchcat (approved):

indra/newview/skins/default/xui/en/notifications.xml
akleshchev: Agree. But might be better to remake it more
"Unable to make payment due recipient blocking the transaction."
Also not sure if it's a good idea to expose the fact that you are blocked, people might start 'probing' block list that way even if it costs them.

akleshchev: @vir-linden please take a look.

akleshchev (approved):

indra/newview/skins/default/xui/en/notifications.xml
LLGuru: > Also not sure if it's a good idea to expose the fact that you are blocked

This idea might be not so good, but it is already implemented in the IM floater
If you're blocked by the resident and open IM chat with him, you see the following:
image

So the fact that you're blocked by the resident is not a secret :-)
And it makes sense to follow the same logics in payments

LLGuru: It seems to be an UI/UX question: I bet @vir-linden will send us to @Steeltoe-Linden :-)

akleshchev:
You haven't fixed broken wording.

indra/newview/skins/default/xui/en/notifications.xml
vir-linden: I would say ask @kylelinden about wording, particularly whether we want to explain the reason.
There is a typo in the name, "Transfter" should be "Transfer"

PR #3363, opened 2025-01-07
Internal PR corresponding to AiraYumi's #3235

nat-goodspeed:
This PR is to run CI internal builds.


indra/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ else()
   set( USE_AUTOBUILD_3P ON )
 endif()
 
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 23)
```
brad-linden, line 32: I think maybe we could take this PR if we made this optional and allow it to be set from the cmake/autobuild command line.  For example:

```suggestion
if (NOT DEFINED CMAKE_CXX_STANDARD)
  set(CMAKE_CXX_STANDARD 20)
endif()
```

nat-goodspeed: https://github.com/secondlife/viewer/pull/3363/commits/0c4b3d4d7c4cb3b979a9e4f0a581fc19deeda4fd

brad-linden (approved):

AiraYumi:
Thank you @nat-goodspeed

I'm currently fixing the build error that occurs when using gcc 15 (Alpha version) in #3389, so I'm okay with you merging it in once that's done.

PR #3378, opened 2025-01-09
VRAM Budget adjustments

akleshchev:
1. Instead of reducing bias when not over budget, start resducing it only once there is 10% free memory to use, to avoid memory immediately going over budget
2. On a 8gb PC intel was using 3.5Gb of vram which is way too generous given viewer's ram needs. For intels caped vram to 25% ram.


marchcat (approved):

indra/llwindow/llwindowwin32.cpp
```diff
@@ -4681,6 +4682,23 @@ void LLWindowWin32::LLWindowWin32Thread::checkDXMem()
 
                     // Alternatively use GetDesc from below to get adapter's memory
                     UINT64 budget_mb = info.Budget / (1024 * 1024);
+                    if (gGLManager.mIsIntel)
```
Ansariel, line 4685: What about Intel's discrete GPU cards? Isn't there a more general distinction required?

Ansariel:
What about slowing down bias decrease? Currently it is evaluating and decreasing bias every frame AFAIK.

indra/llwindow/llwindowwin32.cpp
akleshchev: I suspect those would have accurate gGLManager.mVRAM already, but without having a physical one, hard to be sure. worst case 25% is still a reasonable limit.

Potentially we can substract desc.SharedSystemMemory to get real memory in case of descrete GPU, but I'm not comfortable doing that without having hardware to test it.

akleshchev:
> What about slowing down bias decrease?
That is indeed too fast. I will check it, but might leave it for later.

indra/llwindow/llwindowwin32.cpp
Ansariel: Or `mIsIntel` should only be true for integrated GPUs - if there is a way to detect that.

akleshchev: That's not the purpose of mIsIntel. AMD has built in GPUs and those need same treatment just in case OS/bios is overeager with shared memory. The problem is detection.

PR #3366, opened 2025-01-08
Fix develop nightly builds

brad-linden:
attempt to get the tag-release.yaml workflow to actually succeed at triggering a release build

PR #3384, opened 2025-01-09
PR #3366 didn't work, revert it and fix the right way

brad-linden:
Revert "attempt to get the tag-release.yaml workflow to actually succeed at triggering a release build (#3366)"

This reverts commit 3b79d2ca2405cac4bc7ea7892a45767e3f2b5538.

will be proceeding with provisioning a new token to try to fix this the right way:
https://stackoverflow.com/a/71372524

### DIFF
--- a/.github/workflows/tag-release.yaml
+++ b/.github/workflows/tag-release.yaml
@@ -37,7 +37,12 @@ jobs:
       - name: Update Tag
         uses: actions/github-script@v7.0.1
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # use a real access token instead of GITHUB_TOKEN default.
+          # required so that the results of this tag creation can trigger the build workflow
+          # https://stackoverflow.com/a/71372524
+          # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow
+          # this token will need to be renewed anually in January
+          github-token: ${{ secrets.LL_TAG_RELEASE_TOKEN }}
           script: |
             github.rest.git.createRef({
               owner: context.repo.owner,
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,10 +5,6 @@ on:
   push:
     branches: ["main", "release/*", "project/*"]
     tags: ["Second_Life*"]
-  workflow_run:
-    workflows: ["Tag a Build"]
-    types:
-      - completed
 
 jobs:
   # The whole point of the setup job is that we want to set variables once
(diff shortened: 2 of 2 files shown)


nat-goodspeed (approved):
Since it doesn't work, I approve this revert.

But I'm surprised that GITHUB_TOKEN doesn't do everything a Personal Access Token would.

brad-linden:
> Since it doesn't work, I approve this revert.
> 
> But I'm surprised that GITHUB_TOKEN doesn't do everything a Personal Access Token would.

yeah the devil is in the details here in the GH docs, apparently it is intended as a protection against accidentally creating mutually recursive workflows
https://stackoverflow.com/a/71372524
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow

.github/workflows/tag-release.yaml
```diff
@@ -37,7 +37,12 @@ jobs:
       - name: Update Tag
         uses: actions/github-script@v7.0.1
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # use a real access token instead of GITHUB_TOKEN default.
+          # required so that the results of this tag creation can trigger the build workflow
+          # https://stackoverflow.com/a/71372524
+          # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow
+          # this token will need to be renewed anually in January
```
nat-goodspeed, line 44: I suggest that you store a companion secret encoding the upcoming expiration date of `LL_TAG_RELEASE_TOKEN` and, once it gets close, start emitting a [warning message](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message) on the GH workflow page. You don't need to use the extra `::warning ...` syntax; it works to code:
`echo "::warning::LL_TAG_RELEASE_TOKEN must be recreated before $expiration"`

brad-linden: capturead as #3386

brad-linden:
it looks like this worked, although since it is my access token being used, all the builds look like they were triggered by a push by me: 

https://github.com/secondlife/viewer/releases/tag/Second_Life_Develop%2395afff07-2025-01-11
https://github.com/secondlife/viewer/releases/tag/Second_Life_Develop%2395afff07-2025-01-14

nat-goodspeed:
I bet those with organization authority could create a "buildbot" (or some such) PAT for use in this secret.

PR #3387, opened 2025-01-09
Add Analyze Compile Flags

AiraYumi:
Fixed version of #3321


marchcat:
Just checked, and it throws the same errors on my mac builds as before. Let's keep this one open until we fix the builds.

AiraYumi:
I can build it with the GHA build, but I don't know why an error occurs on the actual device.
I don't own a Mac, so I can't check the error.

marchcat:
Right, because GHA builds use "Release" configuration.

marchcat:
@AiraYumi, are your local linux builds fine with these changes and "ReleaseWithDebInfo" configuration? I don't have a linux dev setup right now, so can't verify that.

AiraYumi:
oops, Relwithdebinfoes options did not exist in the Linux version and an error occurred. You need to fix it.

AiraYumi:
I think this is necessary.
https://github.com/secondlife/viewer/pull/2353/commits/e84a180d39ea79576d5f53fa7c43a4def969ebfb

marchcat:
Right, this will allow you to test this change as well.

AiraYumi:
I checked. There were a lot of build errors caused by adding the debug option.
This will probably take a fair amount of time to fix. maybe.

AiraYumi:
Can you merge this into a branch other than develop?
We expect a large number of CVE-related errors to occur, so we cannot determine the scale of the fix in a short time...

AiraYumi:
I built it with GCC_DISABLE_FATAL_WARNINGS enabled.
[analyzer.log](https://github.com/user-attachments/files/18383875/analyzer.log)

AiraYumi:
Since the cause of the build error is unknown, we will change it to Draft for now.
We need to determine whether it is a problem with boost or not.

AiraYumi:
I contacted the boost team about the issue but got no response.
Upgrading from 1.86.0 will prevent 3p-boost from building.
I'm going to close this PR until it's resolved, but I'd like to hear your thoughts.

bennettgoble:
@AiraYumi Yes, unfortunately we can't merge until builds are fixed. If you want to revive this change you could re-open a fresh PR against develop. Thank you.

PR #3389, opened 2025-01-09
fix build for gcc 15

AiraYumi:
Fix build error for gcc 15

### DIFF
--- a/indra/newview/llvoiceclient.cpp
+++ b/indra/newview/llvoiceclient.cpp
@@ -310,7 +310,7 @@ const LLVoiceVersionInfo LLVoiceClient::getVersion()
     }
     else
     {
-        LLVoiceVersionInfo result;
+        LLVoiceVersionInfo result = {};
         result.serverVersion = std::string();
         result.voiceServerType = std::string();
         result.mBuildVersion = std::string();


indra/newview/llvoiceclient.cpp
```diff
@@ -310,7 +310,7 @@ const LLVoiceVersionInfo LLVoiceClient::getVersion()
     }
     else
     {
-        LLVoiceVersionInfo result;
+        LLVoiceVersionInfo result = {};
         result.serverVersion = std::string();
         result.voiceServerType = std::string();
         result.mBuildVersion = std::string();
```
nat-goodspeed, line 316: If `serverVersion`, `voiceServerType` and `mBuildVersion` are `std::string` members of `LLVoiceVersionInfo`, wouldn't the new initialization at line 313 make lines 314-316 redundant?

nat-goodspeed (approved):

indra/newview/llvoiceclient.cpp
AiraYumi: I'll check.
There's a good chance that lines 314 to 316 can be deleted.

AiraYumi: I found that lines 314 to 316 could be deleted.

nat-goodspeed: Designated initialization would be great if any one of those members had a non-default value. But to be honest, at least from my perspective,
```
LLVoiceVersionInfo result{};
```
is clearer than the designated initialization you wrote above. I think we all understand that value-initializing `std::string` produces an empty string, and value-initializing an `int` produces 0. If, in reading the code, I hit that 8-line initialization expression, I must consider it to understand why it's written it out that way. Is any one of those members being initialized to a non-default value? No? Are we intentionally omitting mention of any members, maybe because they're declared with their own initial values in the struct definition? That sends me looking for the struct definition.

If I see a one-line initialization to `{}`, I can nod and understand that all members are initialized to their default values, and go on.

AiraYumi:
all fixed

nat-goodspeed:
fwiw https://github.com/secondlife/viewer/commit/d9aeb91ace9579fee8c7e1c8478a5cbaa250dfcd

Issue #3408, opened 2025-01-15
Develop Branch Build Error (Linux)


AiraYumi:
### Description

~~~
================ Starting base copy
Processing ../../scripts/messages/message_template.msg => app_settings/message_template.msg ... 1 files
Processing ../../etc/message.xml => app_settings/message.xml ... 1 files
Processing licenses-linux.txt => licenses.txt ... 1 files
Processing client-readme.txt => README-linux.txt ... 1 files
Processing client-readme-voice.txt => README-linux-voice.txt ... 1 files
Processing client-readme-joystick.txt => README-linux-joystick.txt ... 1 files
Processing wrapper.sh => secondlife ... 1 files
Processing handle_secondlifeprotocol.sh => etc/handle_secondlifeprotocol.sh ... 1 files
Processing register_secondlifeprotocol.sh => etc/register_secondlifeprotocol.sh ... 1 files
Processing refresh_desktop_app_entry.sh => etc/refresh_desktop_app_entry.sh ... 1 files
Processing launch_url.sh => etc/launch_url.sh ... 1 files
Processing install.sh => install.sh ... 1 files
Processing secondlife-bin => bin/do-not-directly-run-secondlife-bin ... 1 files
Processing ../llplugin/slplugin/SLPlugin => bin/SLPlugin ... 1 files
Processing appearance-utility-bin => appearance-utility-bin ... 1 files
Processing appearance-utility-headless-bin => appearance-utility-headless-bin ... 
unable to find 'appearance-utility-headless-bin'; looked in:
  /usr/local/secondlife/viewer/build-linux-x86_64/newview/../llappearanceutility
  /usr/local/secondlife/viewer/build-linux-x86_64/newview/../llappearanceutility
  /usr/local/secondlife/viewer/build-linux-x86_64/newview/
Processing res-sdl => res-sdl ... 48 files
Processing secondlife_256.png => secondlife_icon.png ... 1 files
Processing secondlife_256.BMP => res-sdl/ll_icon.BMP ... 1 files
Processing libllwebrtc.so => lib/libllwebrtc.so ... 1 files
Processing gstreamer10/libmedia_plugin_gstreamer10.so => bin/llplugin/libmedia_plugin_gstreamer.so ... 1 files
Processing cef/libmedia_plugin_cef.so => bin/llplugin/libmedia_plugin_cef.so ... 1 files
Processing libcef.so => lib/libcef.so ... 1 files
Processing libEGL* => lib/libEGL* ... 1 files
[...]

PR #3434, opened 2025-01-20
#3408 Fix linux 3p builds

akleshchev:


marchcat (approved):

maxim-productengine (approved):

akleshchev:
No idea why linux-large fails, it doesn't reach my fix so should be something else.

PR #3443, opened 2025-01-21
check runner os version

AiraYumi:
Allows you to check the OS of the runner that was executed (Linux only).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -131,6 +131,9 @@ jobs:
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: |
+          # Check Runner Version
+          sudo cat /etc/os-release
+          # Upgrade
           sudo apt update
           sudo apt install -y \
             libpulse-dev libunwind-dev \


.github/workflows/build.yaml
```diff
@@ -131,6 +131,9 @@ jobs:
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
         run: |
+          # Check Runner Version
+          sudo cat /etc/os-release
```
marchcat, line 135: Does it need to be called with `sudo`?

marchcat (changes requested):
Thank you!

I see the following output now:
```
PRETTY_NAME="Ubuntu 24.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.1 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
```

.github/workflows/build.yaml
AiraYumi: I don't think sudo is necessary, but I used it just in case.
My thinking was that since the linux-large build is failing right now, I wanted this information to help solve the problem, so I had no intention of merging it.
However, if merging it would be better for the future, I will make the change.

AiraYumi:
I don't think the fact that ubuntu has become 24.04 is the intended behavior.

.github/workflows/build.yaml
marchcat: These versions are also present at the very beginning of the log, so there is indeed no need to merge this one.

image

marchcat:
Since the required info is already in the logs, this can be closed.

Issue #3400, opened 2025-01-13
Consider watchdog process for bugsplat to ensure all crash reports are received


kylelinden:
_What does this feature/enhancement do? Why is it needed?_

Consider watchdog process for bugsplat to ensure all crash reports are received.

We know we are not receiving all crash reports given our current design.

## Test Plan

_Include information, such as guidance and test scenarios, QA will need to test this feature._

To be completed

nat-goodspeed:
I don't know of any way a watchdog process could improve the reliability of BugSplat's own crash report upload.

What I believe a watchdog process could do for us would be to notify a Second Life service (new path on the login server?) that the viewer has crashed. We can send information about the viewer version, host ID etc., to try to correlate such notifications with BugSplat crash reports that _do_ succeed. But most of the utility will be to close the statistical gap from BugSplat crash report uploads that do _not_ succeed.

PhilipRosedale:
First priority is simply to understand the following question:  DID A VIEWER SESSION END IN A CRASH?  

Presently, it seems we do not have this data.   Please explain to me what we have in place right now.  In particular:  What % of first-time sessions end in a crash?

nat-goodspeed:
What's in place right now is that we trust BugSplat to send viewer crash reports.

On Windows, the embedded BugSplat library tries to send the report as the last act of the crashing viewer. Empirically, it seems that sometimes things are too badly awry for that to succeed.

On Mac, the (different, acquired?) BugSplat library stores crash information on disc, and sends the report the next time the viewer is launched. This is more reliable, _if_ and when the same viewer is launched again.

Presumably this ticket is about creating an additional mechanism by which we can get a definite YES THIS VIEWER SESSION ENDED IN A CRASH. Before BugSplat, before its predecessor, we used to have our own crash-reporting infrastructure, with public endpoints to which the viewer could submit such a report. I believe those endpoints are no more. So this ticket requires not only viewer code to try to send a report, but server work to receive and "do something" with it.

vir-linden:
Disclaimer: we're working on reviewing the viewer crash infrastructure right now, so some of this may get updated as we learn more. Best understanding right now:

[...]

PR #3482, opened 2025-01-28
#3400 Reimplement LLError marker for crash statistics

akleshchev:
Since there are now more error states, decided to reuse error code from error marker and remove what remained of old llerror marker.

I'm not sure if this should be under #3400 or needs it's own home.

### DIFF
--- a/indra/llcommon/llerror.h
+++ b/indra/llcommon/llerror.h
@@ -308,6 +308,7 @@ namespace LLError
     class LLUserWarningMsg
     {
     public:
+        // error codes, tranlates to last_exec states like LAST_EXEC_OTHER_CRASH
         typedef enum
         {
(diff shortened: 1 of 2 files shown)


nat-goodspeed (approved):
I like it, it's simpler. I still need a more thorough understanding of `processMarkerFiles()`, but I'll dive in as part of the watchdog project.

maxim-productengine (approved):

akleshchev:
> I still need a more thorough understanding of processMarkerFiles(), but I'll dive in as part of the watchdog project.

It's really primitive. Viewer creates a mMarkerFileName on startup and removes on shutdown. If marker exists on next launch: previous session crashes or froze. Other markers sans exec one are there to narrow down the cause.

PR #3492, opened 2025-01-31
Appearance fixes

KittyBarnett:
This PR resolves a number of bugs related to avatar attachments and COF item tracking. These aren't new, but have existed in their current form dating back to at least when I filed VWR-18512.

Attached are a number of screengrabs that illustrate the various issues. It's not all of them, but they are the ones that are easy to reproduce on demand to verify that the PR actually does fix the underlying issues.

This PR is a prerequisite for the RLVa wardrobe support since it makes use of the LLAppearanceMgr/LLAttachmentMgr functionalities.

---

I've tried to split the commits across domain responsibilities, but at least some of them are interdependent in order to fix all the repros above.

* Clean up unused code (n
	-> This code hasn't been used for >10 years

* LLViewerObject::getAttachmentItemID() sometimes returns the NULL UUID for the avatar's own attachments
	-> If the object is already attached then LLViewerJointAttachment::removeObject() ends up getting called, which sets the object's item to the NULL UUID so we need to extract it the item id after the block, not before
	-> Not easy to reproduce, but tracing execution makes it obvious that extractAttachmentItemID should be called after removeObject()

* LLAppearanceMgr::updateAppearanceFromCOF() doesn't properly filter items collected from folder links
    -> When a folder or outfit uses folder links you can end up picking the same inventory item multiple times, so deduping is needed

* Resolve race condition when removing/add attachments inside of a LLWearableHoldingPattern
    -> there's no guarantee that wearable holding patterns will complete in order, e.g. the first wear could be require an asset fetch while the second wear could be cached locally, the second will then complete first and when the first completes it will remove the attachments added by the second
    -> if the asset fetch takes longer then one or more attachments could already be added/remove since its creation causing user frustration when removed things reattach or added things detach

* Resolve issues tracking worn attachments that aren't in COF yet (see PR for more details)
    -> Improves the tracking of "worn but not in COF" attachment and makes sure that they're included in each subsequent appearance change (until their COF link finally gets created)
	
* If multiple LLWearableHoldingPattern instances are waiting for completion, only the most recently created one should progress
   -> Smallest surface level fix
   
* Accurately reflect the worn status of inventory items (the reasons for why this was needed have been resolved in previous commits)
   -> Treating COF links as 'worn' was added as a way to address some of the issues in the screengrabs above, but it causes some issues of its own (and the underlying issues should be resolved by the code above)
   
* Attaching a rezzed object incorrectly delays attachment link creation by 5 seconds
  -> LLSelectNode::mItemID is NULL for a newly created in-world object or the old inventory item UUID if it was rezzed from inventory
  -> a new inventory UUID will be assigned so there's no use in waiting for an item that will never come (and we should never wait on the NULL UUID)

---

Repros:
* AttachFromWorld
1. Rez a prim
2. Open inventory with Current Outfits visible
3. Right-click the prim and Wear/Add
4. Not that there's no COF item being created
5. Wear any clothing layer

Observed: the box detaches
Post fix: the box remains attached

* Attach+CofDesync
1. Rez a prim and take it into inventory
2. Duplicate it three times + rename so you can keep them apart
3. *Quickly* double-click each one in succession (this may take 2-3 tries)
4. Observe that only one cube is attached (they all attach to the same pointo so that's correct)

Observed: incorrect COF (one or more of the objects are linked in COF but aren't worn)

5. Detach the one you are wearing

Observed: the detach happens, but another (unexpected to an end user) attachment pops into existence
Post-fix: COF and appearance are sync'ed (the code in llattachmentmgr largely caused this issue)

---

* DetachAfterWear
1. Create some wearables (can be default)
2. Create some attachments
3. Select both wearables and attachments and then pick 'Wear'

Observed: the attachments come on, then the wearables, then the attachments detach
Post-fix: everything is worn as you'd expect

---

* DetachAfterWear-2
1. Find a wearable that isn't cached locally (you'll have a bit more time for the third double click if the asset needs to fetch - see video)
2. Create two attachments (not attaching to the same point)
2. Double-click the first attachment
3. Double-click the wearble
4. Double-click the second attachment

Observed: perfect example of why syncing attachment inside of a wearable holding pattern is wrong
Post-fix: everything is correctly worn and attached

---

* DetachOnOutfit
1. Have a prim with the script in it ready to go (accept the permission prompt)
2. Have an outfit ready to go (that isn't local wearable and simulator attachment asset fetched)
3. Replace wear the outfit
4. Immediately click the prim to attach it

Observed: the prim gets lots in the outfit change
Post-fix: the prim remains attached

This has the same cause as the others above but illustrates it's not just limited to single inventory wears. The worst case of this one is when you quickly switch between two outfits and end up wearing some of either and some linked in COF but unworn.

https://github.com/user-attachments/assets/af1fce43-8c86-49c1-8537-3d6a8636230e
https://github.com/user-attachments/assets/fc37dd4d-a8da-4608-b808-27d6ff2a390c
https://github.com/user-attachments/assets/08cdaa70-ca1f-4727-a7d0-a36943ddbd17
https://github.com/user-attachments/assets/43d53a2e-fdb4-41fc-ac46-71715f920509
https://github.com/user-attachments/assets/7744bb89-05cc-42dc-897c-8bb3f73f51f5


Ansariel:
This is 95% the code I removed from Firestorm because it will randomly caused stuck attachments when changing outfits a lot. Also unclear if the same problems have been fixed in the meantime for RLVa itself: https://jira.firestormviewer.org/browse/FIRE-16497

The actual problem revolves around the way attachments get attached/detached via simulator messages and the delay until the correc status is reflected in COF. However, this usually isn't that much of an issue unless you add several attachments really fast - which is apparently what the RLVa wardrobe does. A different solution would probably that RLVa would collect the attachment changes, calculate the changes to COF and then use the AIS function to slam the new COF contents to COF in one go instead of sending attach/detach simulator messages each time such a RLVa command is getting processed.

KittyBarnett:
> This is 95% the code I removed from Firestorm because it randomly caused stuck attachments when changing outfits a lot. Also unclear if the same problems have been fixed in the meantime for RLVa itself: https://jira.firestormviewer.org/browse/FIRE-16497

Repro or it didn't happen. Sorry, but you've been banging on this drum and refusing to actually provide anyone with any reason. It just feels spiteful at this point.

I provided repros and explanations of what the changes do, please have the courtesy to counter with tangible facts.

Ansariel:
I already explained it can happen when you change a lot of outfits - I mean a lot of outfit changes!

> These fixes also have absolutely nothing to do with RLVa; all repros involve performing actions residents do on a continuous basis.

Quote from the PR description:
> This PR is a prerequisite for the RLVa wardrobe support since it makes use of the LLAppearanceMgr/LLAttachmentMgr functionalities.

The RLVa wardrobe relies on adding attachments/wearables in quick succession which then exhibit a lot of the issues about. However, I already explained an alternate solution which doesn't involve spamming attach object simulator messages and rather use the same method that is used when changing outfits and that uses AIS, circumventing the race conditions and COF desync that happens when quickly adding or detaching attachments!

I'm not saying the existing bugs shouldn't be fixed, but I am saying these changes can result in stuck attachments. Some issues could be solved even better with a combination of server and viewer side changes, e.g. the whole process of sending a simulator message to attach an object, then wait for the attachment to arrive before the viewer itself updates COF to include the attachment which is the primary reason quickly attach/detaching attachments results in an indeterministic outcome. In fact, most of the issues in the description stem from this fact.

KittyBarnett:
> I'm not saying the existing bugs shouldn't be fixed, but I am saying these changes can result in stuck attachments. Some issues could be solved even better with a combination of server and viewer side changes, e.g. the whole process of sending a simulator message to attach an object, then wait for the attachment to arrive before the viewer itself updates COF to include the attachment which is the primary reason quickly attach/detaching attachments results in an indeterministic outcome. In fact, most of the issues in the description stem from this fact.

Agreed that COF has always been problematic, but I would definitely challenge that any of these changes would make a practical difference to the problem of switching back and forth between outfit folders in quick succession. Also keep in mind that Firestorm has more changes than just these, it's a subset of, not the same.

And we don't get to make simulator PRs so this is the best we can do 😋 .

I also never said that none of the RLVa code won't change either either as it gets refactored, but the bugs above are a problem that needs addressing first, and they've been there for an eternity.

Ansariel:
The reason I mentioned server side changes here is because unlike trying to make things somehow work solely in the viewer to sail around the broken process itself, contributing changes might have the chance of a combined effort together with LL to eliminate the actual source of the problem, namely the old legacy way how attachments are handled outside of a full outfit change - looking at @vir-linden right now...

marchcat (approved):
Code cleanup is always good, thanks!
However, we still have to decide how to address the issues mentioned above.

Geenz:
@KittyBarnett Just checking in - any updates on the rebase work so we can keep this moving forward?

KittyBarnett:
> @KittyBarnett Just checking in - any updates on the rebase work so we can keep this moving forward?

Sorry, it's rebased back onto old develop now & (re)tested.

Geenz:
Thanks @KittyBarnett! Reviewing it I don't see anything _obvious_ - but this isn't an area that I've had the time to really familiarize myself with.  So I'll be requesting additional input from other developers prior to merging it in.

In the meantime, can you provide some tests that our QA folks can work through to validate that this works as intended?  Just a regular step by step test plan is fine.

KittyBarnett:
> In the meantime, can you provide some tests that our QA folks can work through to validate that this works as intended? Just a regular step by step test plan is fine.

There's a repro with an accompanying video of the problem for each of the individual issues. Do you need them in a different format?

Ansariel:
@Geenz Better also do a lot of outfit changes - and I mean a lot - to check if attachments get stuck at some point.

And I would still prefer getting an actual fix of the way attachments are handled: Directly via COF and not by sending a simulator message to attach something, then wait for it to arrive, then update COF via AIS - and then run into race conditions requiring some cruft in the viewer to work around that problem.

Geenz:
@Ansariel Define a lot.  Are we talking just a lot in a single session?  Or a lot of rapid changes?  Or rapid changes and a lot of changes in a session?

Ansariel:
> @Ansariel Define a lot. Are we talking just a lot in a single session? Or a lot of rapid changes? Or rapid changes and a lot of changes in a session?

@Geenz I went through wearing my outfits one by one when at some point attachments got stuck - no wild clicking, no rapid changes, just normal changing from one outfit to another and sometimes making changes to them. Of course I didn't keep track of an outfit change counter - could have been 50, 100 or even more outfit changes.

Geenz:
@Ansariel Noted.  I'll make sure QA spends some extra time cycling through as many outfits they can to try and trigger it - at minimum it's worth understanding the characteristics of the bug if we can get some solid repros.

KittyBarnett:
@Ansariel  - can you clarify what you mean with 'stuck attachment' because to me that could mean different things:
* local-only stuck: simulator detached it, but the viewer never got the kill so didn't destroy the attachment (in this case scripts or edits don't work, and others don't see it on you, and you can't detach it)
* stuck COF link: an inventory link in COF doesn't end up getting delete for some reason so the attachment gets reattached between outfit changes (in this case it would be good to know what happens after a relog to know if the inventory item is real or a phantom one that didn't exist on the inventory server)
* worn but not in COF: attachment is real, not linked in COF, sticks around between outfit changes but can be manually detached

Or something else entirely?

KittyBarnett:
> Stuck as in "the attachment is worn, linked in CoF, but it can't be detached"

Which seems to point to "detached on the simulator side but viewer didn't receive an object kill message"? Which is an issue outside of the appearance code.

AtlasLinden:
Tests were conducted on the above mentioned issues using the build generated from the latest [commit](https://github.com/secondlife/viewer/pull/3492/commits/d327ee29181db75f9c72084a265b10e0e197a7b4). 
Thanks KittyBarnett for providing the repro steps and also the very helpful videos!

Build: Second Life Test 7.1.14.15620674661 (64bit)
- AttachFromWorld
This was repeated against every type of clothing asset. No issues spotted upon attaching every other type of clothing or body parts on both Win11 and MacOS.
- Attach+CofDesync
Unable to reproduce this on Win11/MacOS. Although it is interesting to note that if all 3 attachments are selected followed by pressing Enter/Return it takes slightly longer for the object to appear in COF: https://gyazo.com/15c976c8fd94f57abdb7fc5309bfd0ad. This may be expected. Nevertheless, when attempting to add other attachments during this time, no errors are encountered.
- DetachAfterWear
Unable to reproduce this on Win11/MacOS
Potential bug when selecting Add on multiple clothing and objects followed by selecting Take Off is sometimes leaving clothing attached: https://gyazo.com/f7ce994beb20410338b53915aa74df34  (Naked Ruth Warning XD). This could be fixed by adding another piece of clothing but after testing this many times I actually ended up in a state where I could not remove the gloves or underpants: https://gyazo.com/af06f1d262f204a96f6f0935947e49ae. Relogging shows these clothes still attached and can then be removed.
A new ticket will be made for this unless one already exists in the viewer repository. 
- DetachAfterWear-2
Unable to reproduce this on Win11/MacOS
- DetachOnOutfit
Unable to reproduce this on Win11/MacOS

In regards to changing many outfits, no “stuck” attachments were observed. However, the following issue regarding high complexity outfits was observed:
- If the action to replace the current outfit is selected before the previous outfit has finished attaching, there are times where an attachment on the previous outfit will appear onto the newly selected outfit. These attachments can be detached and reselecting to replace the current outfit can resolve this. https://gyazo.com/26e772b9db0d7412f3c413f5f27e947b 
Note that this reproduces on the latest Beta 7.2.0.16729091892.  A separate issue will be made for this if one does not already exist.

Ansariel, if you do continue to spot attachments being “stuck” and you have a clear repro, please create an issue for it and we can investigate it further (including forwarding it onto the server team if necessary.)

raspopov:
In my opinion, adding a small delay to network packets, such as those that exist on VPNs, helps identify the problems described here.

Geenz:
As an update here.

For 2025.08 we've pivoted into a maint release, so this is tentatively rescheduled into 2025.09.  I'll be trying to get the planning train moving on that ASAP.

indra/newview/llappearancemgr.cpp
```diff
@@ -2636,6 +2577,11 @@ void LLAppearanceMgr::updateAppearanceFromCOF(bool enforce_item_restrictions,
     remove_non_link_items(wear_items);
     remove_non_link_items(obj_items);
     remove_non_link_items(gest_items);
+    // Since we're following folder links we might have picked up new duplicates, or exceeded MAX_CLOTHING_LAYERS
+    removeDuplicateItems(wear_items);
+    removeDuplicateItems(obj_items);
+    removeDuplicateItems(gest_items);
+    filterWearableItems(wear_items, 0, LLAgentWearables::MAX_CLOTHING_LAYERS);
```
Hecklezz: Calling filterWearableItems(wear_items, 0, LLAgentWearables::MAX_CLOTHING_LAYERS) here will result in a bug that I finally got around to investigating in Firestorm that has existed for years (this very same line has been in Firestorm since even before 6.6.14 and I tested our 6.6.17 and the bug was in it).

Currently the viewer and simulator allows you to have 60 clothing wearables added, NOT including body parts (shape, hair, eyes, and skin, etc). However, "wear_items" at this point in the code includes those body parts, and calling filterWearableItems() with LLAgentWearables::MAX_CLOTHING_LAYERS, will result in clothing that exceeds (MAX_CLOTHING_LAYERS - BodyPartCount) being removed from many wearable containers in the code while still being attached, and becoming completely bugged and introduces a number of weird inventory issues, and from testing sometimes bake fails upon login.

I brought this PR into my local build of SLViewer and it now exhibits the same bug that currently exists in Firestorm and did not exist in SLViewer before this PR.
<details><summary>Screenshots showing the inventory and edit outfit side breaking as a result of this line being added-</summary>
Working and current develop:
Image

After merging this PR:
Image

</details>
As you can see in the screenshot after this PR is merged, clothing items after too many are equipped (while still less than the 60 clothing limit), no longer show as worn in the inventory, and right clicking them in edit outfit no longer allows you to take them off or replace them.

Additionally, once you do actually reach 60 clothing items, right clicking another clothing item still says you can Add it because COF only thinks you have say 56 clothing items attached (if you have 4 body parts), so it tries and then gets rejected by the server and undoes the Add.

___

The solution? Unless I'm mistaken, body parts count is ALWAYS 4, never less and never more. So the check should probably be MAX_CLOTHING_LAYERS(60) + MAX_BODY_PARTS(4).
If I am mistaken, and body parts count can somehow be < or > 4, then a different solution will be required (possibly an edit to filterWearableItems to allow you to specify if you want to ignore body parts?).

This is the commit I current pushed to Firestorm to fix this, but feel free to do a better solution (especially if body part count can be not 4)-
https://github.com/FirestormViewer/phoenix-firestorm/commit/3d2fb677dcebab3ebb45e6b9c758f23cd1c05fc0

KittyBarnett: Thanks for the detailed feedback, @Hecklezz ! I'm going to have a look at it right now, your suggestion makes sense 🙂 .

KittyBarnett: @Hecklezz I went a slightly different way @ https://github.com/secondlife/viewer/pull/3492/changes/73eab608d11437b76f086f733cc775dd389bd047 , only because it will always work regardless of the number of body parts SL will have.

@Geenz

KittyBarnett: @Hecklezz Thanks for letting me know btw and for the detailed explanation!

indra/newview/llappearancemgr.cpp
```diff
@@ -1000,6 +948,10 @@ bool LLWearableHoldingPattern::pollFetchCompletion()
     {
         // runway skip here?
         LL_WARNS() << self_av_string() << "skipping because LLWearableHolding pattern is invalid (superceded by later outfit request)" << LL_ENDL;
+
+        // If we were signalled to stop then we shouldn't do anything else except poll for when it's safe to delete ourselves
+        doOnIdleRepeating(boost::bind(&LLWearableHoldingPattern::pollStopped, this));
+        return true;
```
akleshchev, line 954: Can LLWearableHoldingPattern instance cease to be before the idle call? Like pollMissingWearables calls `delete this` then pollStopped gets checked? Or do they never run together?

Nitpick: please use lambda, not boost.

indra/newview/llappearancemgr.cpp
akleshchev: Was this resolved?

> will result in clothing that exceeds (MAX_CLOTHING_LAYERS - BodyPartCount)

@Hecklezz please, clarify, an example scenario would be to attempt to wear ~60 tattos?

Hecklezz: Any sort of system clothing, alpha, tattoos or a mixture of them will work (I just copied tattoos as part of quickly testing it), anything that falls under the "Clothing" category. It should be fixed in commit https://github.com/secondlife/viewer/pull/3492/commits/73eab608d11437b76f086f733cc775dd389bd047 by Kitty, however I have not tested it personally yet and can do later today.

Hecklezz: Sorry for the delay, I confirmed the few bugs that were introduced are no longer present after https://github.com/secondlife/viewer/commit/73eab608d11437b76f086f733cc775dd389bd047

akleshchev: Thank you!

PR #3495, opened 2025-02-01
Downgrade gcc13 to gcc11.

AiraYumi:
This will restore the Linux build to a working state.

However, it will no longer work on versions older than Ubuntu-24.04, so that will need to be resolved separately.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -268,8 +268,8 @@ jobs:
           fi
           export PYTHON_COMMAND_NATIVE="$(native_path "$PYTHON_COMMAND")"
 
-          # This is necessary because the runner is now ubuntu24.04 and cannot be built.
-          # TODO: It needs to be removed once the runner is back to ubuntu 22.04 base.
+          # Temporary workaround: explicitly set gcc-11 and g++-11 on Linux.
+          # TODO: This should be rolled back once gcc-13 issues are resolved.
           if [[ "$RUNNER_OS" == "Linux" ]]; then
             export CC=gcc-11
             export CXX=g++-11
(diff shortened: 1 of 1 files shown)


marchcat:
`linux-large` runner is currently `ubuntu-24.04` so there shouldn't be any problems related to that. Let's try building.

AiraYumi:
Will there be a rollback of linux-large to 22.04 base in the future?

marchcat:
Since GitHub regularly updates these aliases to the latest stable runners, we definitely can just specify the version directly (like we do that with macos). If I remember correctly, Ubuntu LTS releases are supported for five years, so 22.04 will be available for at least two more years.

Quinn-Elara:
IMO ideally there would not be a rollback of the runner unless it's required for some other reason prior to Linux getting an official release copy of the SL viewer, so (personally) I'd be in favor of keeping the runner on 24.04 and making adjustments to get the viewer to build correctly rather than rolling back the runner version to then have to change it again in future (and potentially causing incompatibilities with release copies)

marchcat (approved):
Let's accept it as a temporary workaround.

.github/workflows/build.yaml
```diff
@@ -267,6 +268,13 @@ jobs:
           fi
           export PYTHON_COMMAND_NATIVE="$(native_path "$PYTHON_COMMAND")"
 
+          # This is necessary because the runner is now ubuntu24.04 and cannot be built.
+          # TODO: It needs to be removed once the runner is back to ubuntu 22.04 base.
```
marchcat, line 272:
```suggestion
          # Temporary workaround: explicitly set gcc-11 and g++-11 on Linux.
          # TODO: This should be rolled back once gcc-13 issues are resolved.
```

PR #3524, opened 2025-02-08
Fix for negative fresnel as reported by Whirly

Geenz:


aiaustin:
A bit of feedback on a recent Water fixes build by Geenz... the distant water is now not looking as good as before these changes. There are waves up to a distinct line, then the distant water turns to dark little lines/dots. 
image

Geenz:
@aiaustin There were numerous other environments where the previous fixes would royally break creating a difficult situation where we were likely to break _even more_ content.  I'm still tweaking and trying to find the best middle ground, but horizon brightening is looking increasingly problematic when combined with an environment's haze.  This is increasingly looking like more options for water (which right now we're not in a great spot to add right now).

aiaustin:
The little dancing dashes/dots in the distant water though do draw the eye, and not in a good way. They remind me of the bad SSR black dots if SSR is enabled.  Below is a comparison shot of a SL LL dev version about 7.1.9 that I happen to still have installed.  The distant water looks much more like gentle waves and doe snot draw the eye... 
image

RyeMutt (approved):
LGTM

aiaustin:
Feedback 2025-02-11: Second Life Test 7.1.12.13254202909  looking better to my eye. Interesting one though.. maybe always there? When  you zoom right out for a wide shot on regions that border voids/no region seas, the w3ave effect suddenly becomes smooth water. Should the void be rendered exactly as any distant water? 
image

aiaustin:
image

aiaustin:
As sundown approached I noticed a distinct line in distant water where reflections made a wider line, and water beyond was quite still. On 2025-02-11 in Second Life Test 7.1.12.13254202909. 
Snapshot_001

aiaustin:
Second Life Test 7.1.12.13266555840 ( looks much better to my eye Geenz, and the patchy waves areas above void flat water at a distance is gone.  
image

PR #3548, opened 2025-02-13
viewer-private#394 Ensure correct item id is used on mac

maxim-productengine:
Big thanks to Caladbolg for assisting and helping to find the cause!

### DIFF
--- a/indra/newview/llinventorypanel.cpp
+++ b/indra/newview/llinventorypanel.cpp
@@ -579,7 +579,7 @@ void LLInventoryPanel::itemChanged(const LLUUID& item_id, U32 mask, const LLInve
     {
         if (model_item && view_item && viewmodel_item)
         {
-            const LLUUID& idp = viewmodel_item->getUUID();
+            const LLUUID idp = viewmodel_item->getUUID();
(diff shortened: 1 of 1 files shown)


marchcat (approved):

indra/newview/llinventorypanel.cpp
```diff
@@ -579,7 +579,7 @@ void LLInventoryPanel::itemChanged(const LLUUID& item_id, U32 mask, const LLInve
     {
         if (model_item && view_item && viewmodel_item)
         {
-            const LLUUID& idp = viewmodel_item->getUUID();
+            const LLUUID idp = viewmodel_item->getUUID();
```
akleshchev, line 582: other palces do it like this:

```
// Remove the item ID before destroying the view because the view-model-item gets
// destroyed when the view is destroyed
removeItemID(viewmodel_item->getUUID());

// Item is to be moved outside the panel's directory (e.g. moved to trash for a panel that
// doesn't include trash).  Just remove the item's UI.
view_item->destroyView();
```

akleshchev (approved):

PR #3565, opened 2025-02-14
drain UDP socket to avoid dropped packets

AndrewMeadows:
This PR changes `LLPacketRing` to be a proper ring and uses it to buffer packets that could otherwise be dropped from the UDP socket during long frames. 

This work to help address issue #3566 .

### DIFF
--- a/indra/llappearance/llavatarappearance.cpp
+++ b/indra/llappearance/llavatarappearance.cpp
@@ -799,7 +799,6 @@ void LLAvatarAppearance::buildCharacter()
     bool status = loadAvatar();
     stop_glerror();
 
-//  gPrintMessagesThisFrame = true;
     LL_DEBUGS() << "Avatar load took " << timer.getElapsedTimeF32() << " seconds." << LL_ENDL;
 
     if (!status)
--- a/indra/llmessage/llpacketbuffer.cpp
+++ b/indra/llmessage/llpacketbuffer.cpp
@@ -32,8 +32,6 @@
 #include "lltimer.h"
 #include "llhost.h"
 
-///////////////////////////////////////////////////////////
-
 LLPacketBuffer::LLPacketBuffer(const LLHost &host, const char *datap, const S32 size) : mHost(host)
 {
     mSize = 0;
@@ -51,7 +49,6 @@ LLPacketBuffer::LLPacketBuffer(const LLHost &host, const char *datap, const S32 
             mSize = size;
         }
     }
-
 }
 
 LLPacketBuffer::LLPacketBuffer (S32 hSocket)
@@ -59,13 +56,9 @@ LLPacketBuffer::LLPacketBuffer (S32 hSocket)
     init(hSocket);
 }
 
-///////////////////////////////////////////////////////////
-
 LLPacketBuffer::~LLPacketBuffer ()
 {
 }
-
-///////////////////////////////////////////////////////////
 
 void LLPacketBuffer::init(S32 hSocket)
 {
@@ -74,7 +67,7 @@ void LLPacketBuffer::init(S32 hSocket)
     mReceivingIF = ::get_receiving_interface();
 }
 
-void LLPacketBuffer::init(char* buffer, S32 data_size, const LLHost& host)
+void LLPacketBuffer::init(const char* buffer, S32 data_size, const LLHost& host)
 {
     if (data_size > NET_BUFFER_SIZE)
     {
--- a/indra/llmessage/llpacketbuffer.cpp
+++ b/indra/llmessage/llpacketbuffer.cpp
@@ -41,7 +41,7 @@ LLPacketBuffer::LLPacketBuffer(const LLHost &host, const char *datap, const S32 
 
     if (size > NET_BUFFER_SIZE)
     {
-        LL_ERRS() << "Sending packet > " << NET_BUFFER_SIZE << " of size " << size << LL_ENDL;
+        LL_ERRS() << "Constructing packet with size=" << size << " > " << NET_BUFFER_SIZE << LL_ENDL;
     }
     else
     {
@@ -67,10 +67,25 @@ LLPacketBuffer::~LLPacketBuffer ()
 
 ///////////////////////////////////////////////////////////
 
-void LLPacketBuffer::init (S32 hSocket)
+void LLPacketBuffer::init(S32 hSocket)
 {
     mSize = receive_packet(hSocket, mData);
     mHost = ::get_sender();
     mReceivingIF = ::get_receiving_interface();
+}
+
+void LLPacketBuffer::init(char* buffer, S32 data_size, const LLHost& host)
+{
+    if (data_size > NET_BUFFER_SIZE)
+    {
+        LL_ERRS() << "Initializing packet with size=" << data_size << " > " << NET_BUFFER_SIZE << LL_ENDL;
+    }
+    else
+    {
+        memcpy(mData, buffer, data_size);
+        mSize = data_size;
+        mHost = host;
+        mReceivingIF = ::get_receiving_interface();
+    }
 }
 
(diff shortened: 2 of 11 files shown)


indra/llappearance/llavatarappearance.cpp
```diff
@@ -799,7 +799,6 @@ void LLAvatarAppearance::buildCharacter()
     bool status = loadAvatar();
     stop_glerror();
 
-//  gPrintMessagesThisFrame = true;
```
AndrewMeadows, line 802: This gPringMessagesThisFrame global variable is legacy debug instrumentation that is not being used anymore.  I'm purging it to help keep the code more readable.  Fellow developers: please purge your personal debugging frameworks unless they are broadly useful and and come with some comments or other documentation.

indra/llmessage/llpacketbuffer.cpp
```diff
@@ -67,10 +67,25 @@ LLPacketBuffer::~LLPacketBuffer ()
...
+        LL_ERRS() << "Initializing packet with size=" << data_size << " > " << NET_BUFFER_SIZE << LL_ENDL;
+    }
+    else
+    {
+        memcpy(mData, buffer, data_size);
+        mSize = data_size;
+        mHost = host;
+        mReceivingIF = ::get_receiving_interface();
+    }
+}
```
AndrewMeadows, line 90: I'm adding this `LLPacketBuffer::init()` method to allow the `LLPacketRing` to recycle existing instances for new packets.

indra/llmessage/llpacketring.cpp
```diff
@@ -43,329 +43,301 @@
...
+
+LLPacketRing::LLPacketRing ()
+    : mPacketRing(DEFAULT_BUFFER_RING_SIZE, nullptr)
 {
+    LLHost invalid_host;
+    for (size_t i = 0; i < mPacketRing.size(); ++i)
+    {
+        mPacketRing[i] = new LLPacketBuffer(invalid_host, nullptr, 0);
+    }
 }
```
AndrewMeadows, line 57: This is almost a complete rewrite of LLPacketRing.

Previously it wasn't actually a ring although it had a seldom-used API that made it kinda look like a ring.  I'm purging that seldom-used stuff since it was only ever used for simulating slow network connections and these days there are better (external) tools available to force the viewer into low-bandwidth environments.

Going forward `LLPacketRing` has a real ring of pre-allocated `LLPacketBuffers` that can be used to buffer inbound UDP packets that we don't yet have the per-frame time to process.

indra/llmessage/llpacketring.cpp
```diff
@@ -43,329 +43,301 @@
...
 
-void LLPacketRing::setUseInThrottle(const bool use_throttle)
-{
-    mUseInThrottle = use_throttle;
+    return send_packet( socket,
+                        headered_send_buffer,
+                        data_size + SOCKS_HEADER_SIZE,
+                        LLProxy::getInstance()->getUDPProxy().getAddress(),
+                        LLProxy::getInstance()->getUDPProxy().getPort());
 }
```
AndrewMeadows, line 102: Why do I bother to implement send_packet_helper() instead of just putting the dirty work in `LLPacketRing::sendPacket()`?   Because when I first moved it here we called it twice: `LLPacketRing` had a "delay and throttle outgoing packets" feature but I ripped that out.  Just in case we ever want to put that back in I'm leaving this logic abstracted into its own helper method.

indra/newview/llstartup.cpp
```diff
@@ -623,21 +623,6 @@ bool idle_startup()
...
-                LL_DEBUGS("AppInit") << "Setting packetring incoming bandwidth to " << inBandwidth << LL_ENDL;
-                msg->mPacketRing.setUseInThrottle(true);
-                msg->mPacketRing.setInBandwidth(inBandwidth);
-            }
-            if (outBandwidth != 0.f)
-            {
-                LL_DEBUGS("AppInit") << "Setting packetring outgoing bandwidth to " << outBandwidth << LL_ENDL;
-                msg->mPacketRing.setUseOutThrottle(true);
-                msg->mPacketRing.setOutBandwidth(outBandwidth);
-            }
```
AndrewMeadows, line 640: The in/out bandwidth throttle features are being removed.  These were purely for debug purposes for simulating low bandwidth scenarios.   Perhaps we'll add them back in if people rely on this feature, but there are other ways to simulate low bandwidth these days so hopefully we won't need to.

indra/newview/llappviewer.cpp
```diff
@@ -5251,50 +5247,41 @@ void LLAppViewer::idleNetwork()
...
+                {
+                    // Increase CheckMessagesMaxTime so that we will eventually catch up
+                    CheckMessagesMaxTime *= 1.035f; // 3.5% ~= 2x in 20 frames, ~8x in 60 frames
+                }
+            }
+            else
+            {
+                // Reset CheckMessagesMaxTime to default value
+                CheckMessagesMaxTime = CHECK_MESSAGES_DEFAULT_MAX_TIME;
             }
```
AndrewMeadows, line 5280: These are the relevant changes above; the rest of the changes were just to make this possible.

We now drain the UDP socket whenever we exit the loop without handling ALL of the packets.  In one sense this just moves them from one buffer (maintained by the OS) to another buffer (maintained bu us) but the benefit here is we now know that we have buffered packets and can do something about them.

indra/llmessage/net.cpp
```diff
@@ -76,14 +76,8 @@ static U32 gsnReceivingIFAddr = INVALID_HOST_IP_ADDRESS; // Address to which dat
...
-#if LL_DARWIN
-    // macOS returns an error when trying to set these to 400000.  Smaller values succeed.
-    const int   SEND_BUFFER_SIZE    = 200000;
-    const int   RECEIVE_BUFFER_SIZE = 200000;
-#else // LL_DARWIN
-    const int   SEND_BUFFER_SIZE    = 400000;
-    const int   RECEIVE_BUFFER_SIZE = 400000;
-#endif // LL_DARWIN
+const int   SEND_BUFFER_SIZE    = 200000;
+const int   RECEIVE_BUFFER_SIZE = 800000;
```
AndrewMeadows, line 81: The code above that would use a smaller `RECEIVE_BUFFER_SIZE` for Mac was Really Old (pre 2007).  I checked on a modern Mac build and was able to successfully configure larger buffer sizes.

Also, the viewer doesn't send as many UDP packets as it receives, so unbalanced buffer sizes between the two directions here is ok.

indra/llmessage/llpacketbuffer.cpp
akleshchev: May be add that comment before the memcpy?

indra/llmessage/llpacketring.cpp
```diff
@@ -43,329 +43,301 @@
...
-    // If using the throttle, simulate a limited size input buffer.
-    if (mUseInThrottle)
+    // pull straight from socket
+    if (LLProxy::isSOCKSProxyEnabled())
     {
-        bool done = false;
-
-        // push any current net packet (if any) onto delay ring
-        while (!done)
+        char buffer[NET_BUFFER_SIZE + SOCKS_HEADER_SIZE];   /* Flawfinder ignore */
```
akleshchev, line 142: Pehaps make this static or thread static?

akleshchev (approved):

indra/llmessage/llpacketring.cpp
```diff
@@ -43,329 +43,301 @@
...
+        packet_size = receive_packet(socket, static_cast<char*>(static_cast<void*>(buffer)));
```
akleshchev, line 143: is this double cast nessesary?

AndrewMeadows: Thanks for pointing this out.  I inherited  that line from elsewhere when buffer had been initialized to `U8[]`.  I will fix it.

PR #3586, opened 2025-02-19
Update allow list for CLA.

Geenz:


brad-linden (approved):
yamllint.com says valid yaml!

PR #3589, opened 2025-02-20
theoretical earlier start to asset downloads during startup

AndrewMeadows:
I seemed like maybe we could start processing UDP messages a little earlier during the startup sequence, however I wasn't able to measure much improvement: there was just too much noise during login.  Some logins go faster than others both with or without these changes.

### DIFF
--- a/indra/newview/llappviewer.cpp
+++ b/indra/newview/llappviewer.cpp
@@ -348,8 +348,6 @@ std::string gLastVersionChannel;
 LLVector3 gWindVec(3.0, 3.0, 0.0);
 LLVector3 gRelativeWindVec(0.0, 0.0, 0.0);
 
-U32 gPacketsIn = 0;
-
 bool gRandomizeFramerate = false;
 bool gPeriodicSlowFrame = false;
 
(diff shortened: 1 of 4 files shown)


indra/newview/llappviewer.cpp
```diff
@@ -348,8 +348,6 @@ std::string gLastVersionChannel;
 LLVector3 gWindVec(3.0, 3.0, 0.0);
 LLVector3 gRelativeWindVec(0.0, 0.0, 0.0);
 
-U32 gPacketsIn = 0;
```
AndrewMeadows, line 351: This was never actually used.  It looks like old debug cruft that someone forgot to remove.

indra/newview/llstartup.cpp
```diff
@@ -297,16 +297,59 @@ void callback_cache_name(const LLUUID& id, const std::string& full_name, bool is
...
+        }
+        if (needs_drain || gMessageSystem->mPacketRing.getNumBufferedPackets() > 0)
+        {
+             gMessageSystem->drainUdpSocket();
+        }
+        lmc.processAcks();
+    }
+    // ...then call display_startup()
+    display_startup();
+}
```
AndrewMeadows, line 336: A wrapper for display_startup() that tries to process UDP packets, but only during a particular startup STATE wiindow.  Hrm... I thought I had left a comment here about why I was bracketing the work between `STATE_AGENT_SEND` and `STATE_PRECACHE`, but I seem to have lost it.  I will add it back.

akleshchev (approved):

PR #3608, opened 2025-02-24
viewer-private#398 Fix reference to a dead id

akleshchev:


marchcat (approved):

akleshchev:
Checked documentation for lambda and this shouldn't be an issue.

PR #3624, opened 2025-02-27
Resubmit PR#2849 New flags on ScriptTeleportRequest message.

Rider-Linden:
Accept new flags in ScriptTeleportRequest message. The new flags indicate if the viewer should open and optionally focus the world map when setting a marker beacon. If the flags are missing from the message, the viewer will default to opening the world map floater and giving it focus (this is the old behavior as triggered from [llMapDestination](https://wiki.secondlife.com/wiki/LlMapDestination).)

Issue https://github.com/secondlife/viewer-private/issues/297
Issue https://github.com/secondlife/server/issues/1496


scripts/messages/message_template.msg.sha1
```diff
@@ -1 +1 @@
-d7915d67467e59287857630bd89bf9529d065199
\ No newline at end of file
+b98fc0af5fa88601f5afa4f3c83f08188316e9a8
```
Rider-Linden, line 1: Note, sha may change depending on what other changes are included elsewhere in message_template.msg

akleshchev (approved):

akleshchev:
Shouldn't this be targeted at 2025.03?

Geenz:
@akleshchev I'm thinking we should bring this in yes.  Let's get it into develop, and merge back into 2025.03.

Issue #3644, opened 2025-03-03
Fix dropped packets logging affecting FPS


akleshchev:
Fix 'dropped packet' logging to work like lost ones do, current implemetation hits performance.

Viewer is spamming warnings like:
```
LLPacketRing::drainSocket : dropped 193 UDP packets
LLPacketRing::drainSocket : dropped 128 UDP packets
LLPacketRing::drainSocket : dropped 66 UDP packets
LLPacketRing::drainSocket : dropped 75 UDP packets
LLPacketRing::drainSocket : dropped 85 UDP packets
LLPacketRing::drainSocket : dropped 44 UDP packets
```
Since warnings happen once per frame, it severely affects FPS (need to make our logging faster and fix packets from dropping, but that work is scheduled for later)

Expected behavior:
Instead of spaming, periodically print out packet ring stats. Ex: look through logs for "Dropped packets total", should be each 60s

PR #3651, opened 2025-03-04
#3644 Fix new logging of packet drops hitting performance

akleshchev:
Not sure if that's how we want this reporting to work, but at least now it isn't hitting performance.


marchcat (approved):

maxim-productengine (approved):

indra/llmessage/llpacketring.cpp
```diff
@@ -304,7 +304,7 @@ S32 LLPacketRing::drainSocket(S32 socket)
     S32 num_dropped_packets = (num_loops - 1 + old_num_packets) - mNumBufferedPackets;
     if (num_dropped_packets > 0)
     {
-        LL_WARNS("Messaging") << "dropped " << num_dropped_packets << " UDP packets" << LL_ENDL;
```
akleshchev, line 307: Main cause of the fix, don't spam it each frame.

Issue #3314, opened 2024-12-19
The Quality & Speed numbers in the Improve Graphics Speed floater are off by one.


Dan-Linden:
### Environment

Second Life Release 7.1.11.12363455226 (64bit)
Release Notes

You are at 118.7, 89.3, 22.0 in By Design located at simhost-041cad5509e032fc9.agni
SLURL: http://maps.secondlife.com/secondlife/By%20Design/119/89/22
(global coordinates 261,495.0, 246,873.0, 22.0)
Second Life Preflight 2024-10-19.11413863173
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.91 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5247)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6590
OpenGL Version: 4.6.0 NVIDIA 565.90

Window size: 1799x1346
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (40.1% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 11/30,292 (0.0%)
December 19 2024 11:16:52

### Description

The Quality & Speed numbers in the Improve Graphics Speed floater should be 0 - 6 to match the other places that this number exists - eg. About Second Life, the SecondLife.log

### Reproduction steps

1. Open World menu > Improve Graphics Speed > Graphics Settings.
2. Note which number the Quality & Speed is set to.
3. Open Help > About Second Life and look at the "Render quality" value

Observed: The  Quality & Speed number is different than Render quality.
![Image](https://github.com/user-attachments/assets/fbc4251a-e1ab-4f30-992d-4575288ec1d0)

[...]

Issue #3072, opened 2024-11-12
The Apply button is not activated correctly on the Region/Estate floater when working with terrain textures


valeriiaaproductengine:
<details><summary>Environment</summary>
<p>
Second Life Release 7.1.12.11386494836 (64bit)
Release Notes

You are at 187.9, 85.2, 21.5 in Product Engine located at simhost-039ac35d0127be4e0.agni
SLURL: http://maps.secondlife.com/secondlife/Product%20Engine/188/85/21
(global coordinates 336,828.0, 307,285.0, 21.5)
Second Life Server 2024-10-15.11356152186
Release Notes

CPU: 11th Gen Intel(R) Core(TM) i5-11260H @ 2.60GHz (2611.2 MHz)
Memory: 16111 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5073)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 3050 Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.5612
OpenGL Version: 4.6.0 NVIDIA 556.12

Window size: 1920x1001
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.25
Render quality: 2
Texture memory: 4096MB
Disk cache: Max size 1638.4 MB (0.7% used)

RestrainedLove API: (disabled)
J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Not Connected
Packets Lost: 83/26,334 (0.3%)
November 12 2024 01:54:42

</p>
</details> 

### Reproduction steps

1. Login to the Second Life Release 7.1.12.11386494836 Viewer on Agni.
2. Select World > Region/Estate > Terrain tab.
3. Сlick on any pick terrain texture from 4 levels.
4. In the Pick:terrain floater click the **cancel** button.
5. Observe the Apply button.

**Actual result:** The Apply button became active in the Region/Estate floater even if no changes were made to the Pick:terrain floater.

[...]

Issue #2711, opened 2024-09-26
About SL still showing ALM on some languages


AtlasLinden:
### Environment

DeltaFPS 7.1.10.10800445603

### Description

Advanced Lighting Model is still being shown in About Second Life for multiple languages. Value is provided as [GPU_SHADERS].

Deutsch: 
![image](https://github.com/user-attachments/assets/9dfc9541-ecc0-4e82-b3b6-47086329d809)

Spanish:
![image](https://github.com/user-attachments/assets/134e4733-93bb-4b4a-bc04-ef3851891fcb)

French:
![image](https://github.com/user-attachments/assets/1d326610-6d10-448e-b0c0-5c8db3bd1152)

Also observed on Italian and Polish. 
Not spotted on Danish or English.

### Reproduction steps

Observe Advanced Lighting Model in About SL for each language other than English or Danish

dmytrohproductengine:
Failed QA.

Verified on the Second Life Release 7.1.13.13816156139 (64bit) on Win10/OSX in the scope of [IQA#464](https://github.com/secondlife/iqa/issues/464).

This issue is still reproduced for the Japanese and Chinese languages (see "actual_JP.png" and "actual_CN.png").

**actual_JP.png**
![Image](https://github.com/user-attachments/assets/8cd80d66-0eac-4734-919f-7092a2ff8df9)

**actual_CN.png**
![Image](https://github.com/user-attachments/assets/27f68c85-0782-4927-a969-f37161041ff4)

DmitrySProductengine:
Passed QA.
Verified on the Second Life Release 7.1.13.13906285233 (64bit) on OSX/Win11 in the scope of [IQA #469](https://github.com/secondlife/iqa/issues/469).
Also, during the verification of this ticket, 2 bugs were found. Perhaps, they could have been described in one ticket (since they are reproduced on the same locales), but I filed two different tickets:
1. https://github.com/secondlife/viewer/issues/3770
2. https://github.com/secondlife/viewer/issues/3771

PR #3657, opened 2025-03-05
Cherry-picking develop fixes into 2025.03

maxim-productengine:
Cherry-picking a batch of my low-risk fixes from archive/develop branch for the following bugs:
#3256 Moving in mouselook mode may cause attached linksets to visually break apart. (Example shows firing a gun while moving/jumping in that mode).
#2626 When you're calling to a resident with active DND, call is not automatically rejected, and it will end only after a timeout.
#3051 Wrong chat history file is used if you receive a notice before opening group chat window for that group
#3029 Log spamming when dragging a folder in inventory, instead of showing the tooltip
#3072 UI is not updated correctly after interaction with texture picker
#3314 Make Quality & Speed labels consistent with "Render quality" value in Debug settings/About SL
#3052 Pointless log spam when group notice is received
#2711 Clear ALM from About land info

### DIFF
--- a/indra/newview/lldrawable.cpp
+++ b/indra/newview/lldrawable.cpp
@@ -33,6 +33,7 @@
 
 // viewer includes
 #include "llagent.h"
+#include "llagentcamera.h"
 #include "llcriticaldamp.h"
 #include "llface.h"
 #include "lllightconstants.h"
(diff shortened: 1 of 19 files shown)


akleshchev (approved):

marchcat (approved):

Geenz:
This generally seems low risk enough.  @AtlasLinden Can we get a spot check on this for the above issues please?

AtlasLinden:
Spot check looks good. Tested on build with above fixes on both Win11 and Mac

Issue #3654, opened 2025-03-04
Searchable disabled settings in Preferences unreadable due to highlighting


AtlasLinden:
### Environment

Second Life Release 7.1.12.13509103967 (ForeverFPS)

### Description

Disabled settings in Preferences are difficult to read if they are highlighted via search.

![Image](https://github.com/user-attachments/assets/9d716ebc-7959-4351-83ad-a8afe271f53d)

### Reproduction steps

There are many disabled settings in Preferences that are disabled on the login screen.
Here's a few to search for:
- General>Show favorite locations on login screen
- Sound & Media>Toggle speak on/off with toolbar button
- Chat>Only friends and groups can call or IM me

kylelinden:
Change background to something less extremely "red".

irinabproductengine:
Passed QA.
Verified on the Second Life Release 7.1.13.13906285233 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/469.

This ticket is passed.
The color of the background is changed and settings can be read.

<details><summary>Screenshot</summary>
<p>

<img width="1158" alt="Image" src="https://github.com/user-attachments/assets/2bb202c2-1cce-4154-9996-fb0f85e9fcd5" />

</p>
</details> 

During verification, it was found when searching that the highlighted text is very dim, merges with the background, and is unreadable in the Preferences.
The new issue https://github.com/secondlife/viewer/issues/3778 was created.

PR #3660, opened 2025-03-05
#3654 Searchable disabled settings not readable when highlighted

akleshchev:


marchcat (approved):

maxim-productengine:
We should consider adding color picker for that setting some day...

maxim-productengine (approved):

Issue #3653, opened 2025-03-04
Preferences search missing or incorrectly highlighting some settings


AtlasLinden:
### Environment

Second Life Release 7.1.12.13509103967 (ForeverFPS)

### Description

Some settings are not found when searching in Preferences.
E.g. searching "avatar" does not highlight the Avatar Display setting.

![Image](https://github.com/user-attachments/assets/e88273f3-049f-44da-aad2-904fbe05ae87)

Some settings are also appearing for different searches. 
E.g. searching "draw" highlights "Avatar maximum complexity" instead of "Draw distance"

![Image](https://github.com/user-attachments/assets/1b3893e9-a444-43c6-87c2-4efb71408de5)

### Reproduction steps

Search for various settings.
Here are some known settings that do not appear in search:
- Draw Distance
- Brightness (exposure)
- Away timeout

AtlasLinden:
See also: https://github.com/secondlife/jira-archive-internal/issues/68467

AtlasLinden:
Passed QA.
Verified on the Second Life Release 7.1.13.14542338120 (2025.04) on Win11/MacOS

Note that Preferences will still highlight settings where searched text is found in tooltips.
For instance searching "draw" will highlight "Avatar Maximum Complexity".
This is being filed as a separate low priority issue.

PR #3682, opened 2025-03-06
Fix preferences search not finding text defined in text boxes (#3653)

VectorMutt:
# Rationale
Being unable to find certain preferences via search. This fixes the first case in issue #3653 The second case is due to preferences search also considering the text in tooltips as part of its results.

# Related issues
#3653 

# Testing
Verify that Avatar Display is now highlighted on the graphics preferences tab when searching for Avatar

# Future work
Examine if tooltips should or should not be included in the preferences text search


Geenz (approved):
LGTM

PR #3719, opened 2025-03-11
fix: Correctly update shadows on RenderShadowResolutionScale change

williamweaver:
Shadows were not updating correctly after a shader change occurred in-session and then the RenderShadowResolutionScale setting was adjusted. This issue is present in Second Life Release 7.1.12.1355088671 (64-bit) and Second Life Test 7.1.12.250701803 (64-bit).

**Specifically, after any shader-related setting is changed in-session (such as toggling Advanced Graphics options like SSAO, HDR, Depth of Field, SSR, Antialiasing, or changing the Graphics Quality preset), subsequent adjustments to `RenderShadowResolutionScale` via Debug Settings result in broken shadow rendering.** The shadows become corrupted or disappear entirely and do not reflect the new resolution scale. Correct shadow rendering is only restored by toggling a shader or restarting the viewer. This behavior is inconsistent with other render settings that update immediately after modification and degrades the user experience when dynamically adjusting shadow quality.

This commit changes the signal listener for "RenderShadowResolutionScale" in **llviewercontrol.cpp** from `handleShadowsResized` to `handleSetShaderChanged`.

`handleSetShaderChanged` ensures a full shader update, which is necessary for this setting to take effect immediately—similar to other render settings like RenderDeferredSSAO.

This change ensures that shadows update correctly and immediately when the resolution scale is changed in Debug Settings, even after prior shader changes in the session, without requiring additional shader toggling or viewer restarts. This provides a smoother and more responsive experience for advanced users adjusting shadow quality in various rendering scenarios.

---

**Steps to Reproduce (Bug)**
_Verified in Second Life Release 7.1.12.1355088671 (64-bit) and Second Life Test 7.1.12.250701803 (64-bit):_

1. **Fresh Install Preparation:** Ensure a clean Second Life installation state. Ideally, uninstall and reinstall the viewer or clear/rename all folders in `AppData\Local\Second Life` and `AppData\Roaming\Second Life` before launching.
2. Launch the Second Life Viewer and log in.
3. Enable the Debug Menu: Open Preferences (Ctrl+P), go to the "Advanced" tab, and check "Show Advanced Settings".
4. Open Debug Settings: Click the "Advanced" menu in the menu bar and select "Debug Settings."
5. Locate the `RenderShadowResolutionScale` setting (which should be set to 1.0 on a clean install).
6. Initially, changing `RenderShadowResolutionScale` at this point may not exhibit the bug. Proceed to the next steps to reliably trigger it.
7. **Trigger the Bug:** Open Preferences (Ctrl+P) again and go to the "Graphics" tab.
8. Click the "Advanced Settings" button.
9. **Toggle *any* of the following Advanced Graphics options:**
   - Screen Space Ambient Occlusion
   - HDR and Emissive Rendering
   - Depth of Field
   - Screen Space Reflections
   - Antialiasing
   - *Alternatively*, change the "Graphics Quality" preset slider (e.g., Low to Ultra or any other change).
10. Return to the Debug Settings floater.
11. Change the value of `RenderShadowResolutionScale` to a different value (e.g., from 1.0 to 0.5 or 2.0).
12. **Observe the Bug:** Notice that shadow rendering does not update correctly—shadows become corrupted or disappear.
13. **Workaround (in buggy version):** To restore correct shadow rendering without the fix, either:
    - Toggle a different shader (e.g., change graphics presets in Preferences, or toggle SSAO, SSR, etc.), or
    - Restart the viewer.

**Steps to Verify (Fix):**

1. Build the viewer with this commit applied.
2. Launch the viewer and log in.
3. Repeat steps 1–9 from "Steps to Reproduce (Bug)" to ensure an Advanced Graphics setting is toggled before proceeding.
4. Open Debug Settings and locate `RenderShadowResolutionScale`.
5. Change the value of `RenderShadowResolutionScale` (e.g., from 1.0 to 0.5 or 2.0).
6. **Verify the Fix:** Confirm that shadow rendering updates immediately and correctly—even after toggling Advanced Graphics settings—with the shadows visibly changing resolution in real time. No shader toggling or viewer restart is required.

---

- No specific regression testing is required for this targeted fix. However, standard viewer functionality should be verified after building to ensure no unintended side effects have been introduced. Pay particular attention to shadow rendering in various environments and lighting conditions to confirm the fix has not negatively impacted other shadow-related features.

---

- No documentation changes are needed as this is a bug fix for an existing debug setting.


williamweaver:

williamweaver:
recheck

Geenz:
Thanks for the contribution!

Changing the shadow resolution _shouldn't_ be necessary, which indicates a bug elsewhere that's cropped up over time.  Resetting shaders has other implications on the pipeline - namely thrashing the shader cache as an obvious one, and states getting reset more broadly in other parts of the pipe that may be a bit sensitive.  It's basically taking a sledgehammer to a nail here.

I'll give this a spin locally to make sure we're good here regardless and consider this a temporary fix if so until the actual problem is determined.  We have some other stuff that would benefit from shadow resolution scale being another user exposed setting.

Also, per the pre-commit, we're a spaces house.

williamweaver:
@Geenz Thanks for the review and feedback!

Understood on the "spaces house" point, and thanks for pointing that out. I've corrected the whitespace in the latest commit (specifically trailing whitespace that `git diff --check` flagged).  Hopefully, it's all good now style-wise.

I also want to acknowledge and agree with your point about the previous approach of forcing a shader refresh being a "sledgehammer." I understand now that it was not an ideal solution and could have negative side effects. I've taken that feedback on board, and this guard approach is intended to be a more targeted and lightweight interim solution until the underlying issue causing the mRT invalidation can be properly investigated and resolved.

I've intentionally left the debug logging in the code for now (the `LL_WARNS` and `LL_INFOS` messages with the frame counter in `resizeShadowTexture()`). My thinking was that this logging could be helpful during the review process and for anyone reproducing or further testing the fix, as it provides explicit confirmation of when and how many frames are skipped due to the mRT being invalid.  However, please let me know if you'd prefer that debug logging to be removed before merging – I'm happy to remove it if that's the project's preference for final code.

I appreciate you considering this guard as a temporary fix to prevent the broken shadows in the meantime while the root cause is investigated.  I'm happy to help with further investigation into that underlying problem as well.

Let me know if there's anything else I can do or if you have further feedback!

Geenz:
@williamweaver Thinking about this more - let's punt on this and find a better fix.  From my perspective this is low priority as far as where my current time commitments are.  So let's do this.

So here's some criteria:
- Shader recompiles are not really ideal
- Runtime switching of shadows, for example in the AGP, is a nice to have feature
  - Looking at just general better balancing of SL's visuals here - it'd be great to have a `Low/Mid/High/Stupid` shadow setting
  - Note that resolution probably isn't the _biggest_ drag when it comes to shadows, but I know a lot of people who would love to just have this be an easy to access setting
- Viewer stalls are bad and should be minimized - even when messing with graphics settings

If you'd like to continue pursuing this - you're more than welcome to.  I'll write up some notes when I have some time to give you some potential leads on trying to track this down.

Geenz:
@williamweaver Checking in on this one.  Any updates?

williamweaver:
Hi @Geenz,

Thank you again for your patience on this PR (#3719) and for the opportunity to delve into this shadow rendering issue. Apologies for the delay in getting back to this; my focus was split with other work, and as I'll touch on below, I also spent some time considering the broader context around RenderShadowResolutionScale.

Following your feedback, I shifted from the interim guard fix and focused on identifying the root cause. Using detailed logging (primarily LL_WARNS tagged ShadowDebug to keep console noise down), I was able to trace the state inconsistency. The investigation confirmed that the issue stemmed from LLPipeline::createGLBuffers incorrectly zeroing out mRT->width and mRT->height after allocateScreenBuffer had already properly set them. This created the window where an immediate call to resizeShadowTexture (triggered by changing RenderShadowResolutionScale) would read invalid dimensions and fail.

The latest commit removes those two erroneous lines from createGLBuffers. Post-fix logs and visual testing confirm this resolves the root cause cleanly – resizeShadowTexture now reads the correct dimensions immediately after a graphics setting change and successfully resizes the shadow buffers without delay, errors, or the need for the guard condition. I've removed the diagnostic logging from the code submitted in this PR for cleanliness, but I'm happy to share the instrumented code or raw logs if you'd like to see the detailed trace.

I've documented the investigation process, log analysis, root cause identification, and verification in more detail on a GitHub Wiki page here:
[RenderShadowResolutionScale: Fixing Shadow Updates After Graphics Setting Changes (#3719)](https://www.google.com/url?sa=E&q=https://github.com/ApertureViewer/Aperture-Viewer/wiki/RenderShadowResolutionScale:-Fixing-Shadow-Updates-After-Graphics-Setting-Changes-(%233719))

I hope this root cause fix now meets the expectations and criteria you outlined earlier, specifically by avoiding shader recompiles and minimizing stalls.

Regarding your point about the potential usefulness of runtime shadow quality adjustments (like Low/Mid/High presets): I agree that making RenderShadowResolutionScale more accessible could be beneficial, especially for users taking snapshots or seeking better performance tuning.

Part of the reason I temporarily shifted focus from this PR was to fully develop a related concept in my Aperture Viewer codebase. I have coded and tested a redesigned shadow frustum splitting and scaling system aimed at significantly improving perceived shadow quality (especially closer to the camera where detail matters most, ensuring even a scale of 1.0 looks sharp) while intelligently balancing the computational cost. The core idea is to better align shadow resource allocation with visual impact and the resolving power of pixels at different distances, using fixed distance ranges for primary focus areas. The system is functional and actively being finalized in Aperture.

Before potentially exposing RenderShadowResolutionScale more broadly to users outside of debug settings, perhaps considering such an optimized frustum/scaling approach might be beneficial. It ensures the setting works not just correctly (as this PR now achieves), but also intelligently, providing the best visual trade-off for the performance cost. Visual impact is paramount for a setting like this.

Would you be open to reviewing this complementary frustum/scaling system (which would need porting from Aperture) at a later time as a potential enhancement following this fix? For now, hopefully, this PR provides the robust fix needed for the immediate issue in #3719.

Thanks again for your guidance!

Geenz:
@williamweaver This is much better.

Regarding the shadow frustum work you've done - absolutely! Feel free to submit a PR for that - I'm always happy to consider graphics changes from the community (provided they don't break existing content).  I'll go ahead and merge this now.  Thanks!

Geenz (approved):

PR #3726, opened 2025-03-12
#3364 Fix a case of excessive texture updates when bias fluctuates a bit

akleshchev:
Make bias' effect a bit more gradual instead when it strictly goes over 2.f.


Geenz:
@akleshchev This looks fine, but I was curious if you've tested this with the texture console up?  How hard are we hitting textures when things start to fluctuate?

akleshchev:
>This looks fine, but I was curious if you've tested this with the texture console up? How hard are we hitting textures when things start to fluctuate?

Goes from ~1.90 to ~2.05 and back roughly each 20 seconds. Video in slack.

Geenz (approved):

akleshchev:
Users complain that fluctuation does hit fps and I hope having to update less textures (softer curve) will help there as well.

Issue #3685, opened 2025-03-06
Add two new slapps to support wearing contents of folders - Add and Remove


kylelinden:
_What does this feature/enhancement do? Why is it needed?_

Add two new slapps -- one to add the contents of a folder to the avatar, and one to remove the contents of a folder.

## Test Plan

Older QA SLapp test page (untrusted)
https://sl-viewer-media-system.s3-us-west-2.amazonaws.com/html/slappr.html#

Need to add a new viewer-based menu or floater to test SLapps from a Trusted source.

akleshchev:
Wearing or removing by name is not practical.
Inventory is not warrantied to not have dupplicate names. And there might be two different outfits, one unmodified in library and one modified in invenory. Library copies folder to inventory so we can no longer remove it by name 'from library'.

maxim-productengine:
Agree, wearing/adding from Library is OK, but not from user's Innventory.

DarlCat:
It'd be really nice for individual use if we could address folders by UUID outside of the library; just two cents from a resident. Potential for abuse there is basically zero imo, and would be greatly appreciated as an extension of the functionality.

maxim-productengine:
Two slapps where added: `add_folder` and `remove_folder`.

`add_folder` can be used similar to `wear_folder`: by name for Library folders and by ID for any folders(Library and non-Library).
Cardboard Boxbot by name:
`secondlife://app/add_folder/?folder_name=Cardboard%20Boxbot`
Musician Guy by ID:
`secondlife://app/add_folder/?folder_id=59219db2-c260-87d3-213d-bb3bc298a3d8`

To use `remove_folder` actual folder ID(non-Library) should be provided
`secondlife://app/remove_folder/?folder_id=USER_FOLDER_ID`

Keep in mind, that both new commands are disabled for non-trusted sources(_UNTRUSTED_BLOCK_), similar to old `wear_folder` (which is used in Complete Avatars floater). 
Let me know if anything should be changed.

igorlproductengine:
Deferred QA.
Verified on the Second Life Release 7.1.13.13906285233 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/469.

This functionality has not been implemented in this build version.

katerynamproductengine:
Passed QA.

Verified on the Second Life Release 7.1.13.13999474331 (64bit) on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/472.

PR #3732, opened 2025-03-12
#3685 Add two new slapps to support wearing contents of folders - Add and Remove

maxim-productengine:


marchcat (approved):

indra/newview/llappearancemgr.cpp
```diff
@@ -4724,46 +4724,64 @@ void wear_multiple(const uuid_vec_t& ids, bool replace)
...
+    LLUUID folder_uuid;
+
+    if (query_map.has("folder_name"))
+    {
+        std::string outfit_folder_name = query_map["folder_name"];
+        folder_uuid = findDescendentCategoryIDByName(gInventory.getLibraryRootFolderID(), outfit_folder_name);
+        if (folder_uuid.isNull())
+            LL_WARNS() << "Couldn't find " << std::quoted(outfit_folder_name) << " in the Library" << LL_ENDL;
+    }
+    if (folder_uuid.isNull() && query_map.has("folder_id"))
```
akleshchev, line 4738: If uuid is provided shouldn't it take priority over name?

indra/newview/llappearancemgr.cpp
```diff
@@ -4724,46 +4724,64 @@ void wear_multiple(const uuid_vec_t& ids, bool replace)
 
 // SLapp for easy-wearing of a stock (library) avatar
 //
+bool wear_library_category(const LLSD& query_map, bool append)
```
akleshchev, line 4727: Whis is no longer a 'library' only fuction.

akleshchev (approved):

indra/newview/llappearancemgr.cpp
maxim-productengine: It was this way for 'wear_folder', so keeping that logic for now.

PR #3769, opened 2025-03-19
Maint B leftover commits

marchcat:
This can go into either 2025.03 or 2025.04.
@Geenz, which one do you prefer?


Geenz:
Seems reasonably low risk. Lets get this into 2025.03.

Geenz (approved):

PR #3777, opened 2025-03-20
#1964 fix for reflection probe doesn't update position when the linkset is moved by the script

maxim-productengine:


akleshchev (approved):

marchcat (approved):

indra/newview/llviewerobject.cpp
```diff
@@ -4820,6 +4820,18 @@ void LLViewerObject::setPositionParent(const LLVector3 &pos_parent, bool damped)
...
         setPositionRegion(pos_parent, damped);
+
+        // #1964 mark reflection probe in the linkset to update position after moving via script
+        for (LLViewerObject* child : mChildList)
+        {
+            if (child && child->isReflectionProbe())
+            {
+                if (LLDrawable* drawablep = child->mDrawable)
+                {
+                    gPipeline.markMoved(drawablep);
```
akleshchev, line 4831: Should viewer be the one doing this? Shouldn't server inform viewer about the move?

maxim-productengine: I'm not sure, maybe. Anyway, we may remake this quickfix for probes later.

Issue #3596, opened 2025-02-20
Parts on my avatar are missing on 7.1.12.13416292187


Dan-Linden:
### Environment

Second Life Test 7.1.12.13416292187 (64bit)
Release Notes

You are at 121.5, 116.1, 23.1 in Lag Me 5 located at simhost-0d4a841119c1e8c19.agni
SLURL: http://maps.secondlife.com/secondlife/Lag%20Me%205/121/116/23
(global coordinates 260,473.0, 249,972.0, 23.1)
Second Life Server 2025-02-07.13191265457
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.91 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5487)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6590
OpenGL Version: 4.6.0 NVIDIA 565.90

Window size: 1725x1084
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 256m
Bandwidth: 3000kbit/s
LOD factor: 2
Render quality: 6
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (8.7% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/1,675 (0.0%)
February 20 2025 15:28:46

### Description

I logged in on 7.1.12.13416292187 and pieces of my shoulders are not rendering.

### Reproduction steps

1. Logged in with 7.1.12.13416292187 viewer

Observed: Some of the mesh around my shoulders did not load

![Image](https://github.com/user-attachments/assets/95f02b17-3658-433f-ac19-0a6a9eb6c712)
[missing shoulder pieces SecondLife.log](https://github.com/user-attachments/files/18896403/missing.shoulder.pieces.SecondLife.log)

[...]

PR #3802, opened 2025-03-25
#3596 Don't supress exceptions in WorkQueue

akleshchev:
If there is a crash we should know about it, otherwise we won't learn abut problems in our code. Especially if it's a crash in a window thread.

I'm not sure if this is the best way to handle this. On windows we can send crash directly to bugsplat without crashing, but mac won't have the same benefit.

P.S. There is a test under develop menu to cause a crash in a worker thread.

### DIFF
--- a/indra/llcommon/workqueue.cpp
+++ b/indra/llcommon/workqueue.cpp
@@ -17,6 +17,7 @@
 // std headers
 // external library headers
 // other Linden headers
+#include "llapp.h"
 #include "llcoros.h"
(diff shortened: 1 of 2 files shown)


marchcat (approved):

Geenz (approved):

maxim-productengine (approved):

akleshchev:
Test failed to reach bugsplat so new change will send crash directly on windows and will rethrow on Mac.

PR #3807, opened 2025-03-25
SLua editor: multi-line strings and comments

marchcat:


indra/llui/llkeywords.cpp
```diff
@@ -578,7 +587,121 @@ void LLKeywords::findSegments(std::vector<LLTextSegmentPtr>* seg_list, const LLW
...
+            if (!mRegexTokenList.empty())
+            {
+                S32 seg_start = (S32)(cur - base);
+
+                std::string text_to_search(wtext.begin() + seg_start, wtext.end());
+
+                for (token_list_t::iterator iter = mRegexTokenList.begin();
+                    iter != mRegexTokenList.end(); ++iter)
+                {
+                    LLKeywordToken* regex_token = *iter;
```
akleshchev, line 601: Suggestion:
`for (const LLKeywordToken* regex_token : mRegexTokenList)`

indra/llui/llkeywords.cpp
```diff
@@ -866,6 +984,14 @@ void LLKeywords::dump()
         LLKeywordToken* delimiter_token = *iter;
         delimiter_token->dump();
     }
+
+    LL_INFOS() << "LLKeywords::sRegexTokenList" << LL_ENDL;
```
akleshchev, line 988: Add a tag, please.

indra/newview/app_settings/keywords_lua_default.xml
```diff
@@ -104,47 +104,42 @@
          <key>boolean</key>
          <map>
             <key>tooltip</key>
-            <string>Lua boolean: represents a true or false value.</string>
+            <string>Boolean: represents a true or false value.</string>
```
akleshchev, line 107: Should lua tooltips really be in app_settings? We need them to be localizable.

akleshchev (approved):

maxim-productengine (approved):

indra/llui/llkeywords.cpp
marchcat: Thanks! This is mostly the proof of concept code now, I'll optimize the whole thing later.

indra/newview/app_settings/keywords_lua_default.xml
marchcat: We have a default LSL keywords file in the same location, so this is correct.
For LSL, we also receive one from the server, but that isn't supported for SLua yet.

akleshchev: Default keyword for LSL updates from server, I'm not sure it's something that should be replicated for Lua.

marchcat: It's planned to be implemented later. For now we just use a placeholder class LLSyntaxLua.

Issue #3803, opened 2025-03-25
BugSplat Crash: SecondLifeViewer!LLTabContainer::setTab(1560)


sl-service-account:
**BugSplat** [Key Crash 71339](https://app.bugsplat.com/v2/keycrash?stackKeyId=71339&database=SecondLife_Viewer_2018)

**Sample Version:**
7.1.13.13906285233

**Sample Callstack:**
[...]

PR #3813, opened 2025-03-26
#3803 Crash at LLTabContainer::setTab

maxim-productengine:


akleshchev (approved):

indra/newview/llfloaterimcontainer.cpp
```diff
@@ -1959,6 +1959,10 @@ bool LLFloaterIMContainer::removeConversationListItem(const LLUUID& uuid, bool c
             }
         }
     }
+    else
+    {
+        LL_WARNS() << "Conversation widgets: " << mConversationsWidgets.size() << LL_ENDL;
```
akleshchev, line 1964: I think 'warn' is a bit too severe.

PR #3817, opened 2025-03-26
Small FPS related adjustments

akleshchev:
Tectures (Leaves?) in our meeting spot were pointlesly calculating vsize fpr dozens of faces per texture, couldn't resist cutting it a bit (probably can be optimized a lot more, ex: record last heaviest face). LLViewerTextureList::updateImagesFetchTextures dropped from 200mks to 130mks.

Also on_screen was working wrongly using last face only.

@Geenz if you are fine with this I will find a ticket to attach this to.


Geenz:
@akleshchev Lemme grab the build on this one, and I'll get back to you.

Geenz (approved):
LGTM.

PR #3852, opened 2025-04-01
VRAM adjustments

akleshchev:
1. Clean up unused and malfunctioning DX+WMI memory test. That detection was capped at 4Gb so on practice it was never used and fell to other methods. DX+WMI can take seconds, too pricey to keep (but P0 might be removing it too)
2. Cap DXGI detected VRAM to 75% RAM size, since it returns shared memory on integrated GPUs, not dedicated memory. Dedicated GPUs should have used GL extentions for VRAM.
3. Shrink draw distance when VRAM is very low (and fix sRenderDistance). This should help with VRAM usage, but it is not reporting this disatance to interest list (but may be should).


marchcat (approved):

maxim-productengine (approved):
LGTM

Issue #3855, opened 2025-04-02
Texture repeats per meter incorrect for normals and specular


Hecklezz:
### Environment

Second Life Release 7.1.12.13550888671 (64bit)
<details>
<summary>Rest of environment-</summary>
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5608)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.7283
OpenGL Version: 4.6.0 NVIDIA 572.83

Window size: 2220x1250
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (32.2% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.23.1 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version:  

April 01 2025 23:24:12
</details>

### Description

PR with fix available here-
https://github.com/secondlife/viewer/pull/3856

When adjusting Blinn-Phong normal and specular scale values by using the repeats per meter slider, depending on the scale of the object and which face is selected, the scale values which gets calculated is completely wrong.
Take for example a 2x1x1 cube and you adjust the top face, if you attempt to adjust the repeats for normal or specular, the values will continually scale upwards roughly doubling.
However if you take for example a 1x2x1 cube and adjust the top face, everything behaves as expected.
Depending on the size of the object and which face you are editing, the behaviour of setting the repeats may be correct or incorrect.

What should be expected to happen:
The normal and specular scales should increase and decrease correctly when being adjusted via repeats per meter regardless of object size and face selected.

Video showing the bug-
https://github.com/user-attachments/assets/eaaf980e-ac1f-4a54-8b1f-8b44482f3e54

### Reproduction steps

1. Create a 2x1x1 cube (or any size where X > Y)
[...]

PR #3856, opened 2025-04-02
[#3855] Fix normal and specular repeats per meter scaling

Hecklezz:
https://github.com/secondlife/viewer/issues/3855

Fixes the problem described in the above issue.
Previously, normal and specular were not using the correct s and t axis scale values depending on the face.

Video showing it fixes (bug video in issue above)-
https://github.com/user-attachments/assets/67ae268b-ceb9-4b08-8705-ae61de76f246

### DIFF
--- a/indra/newview/llpanelface.cpp
+++ b/indra/newview/llpanelface.cpp
@@ -3592,7 +3592,7 @@ void LLPanelFace::onCommitRepeatsPerMeter()
     bool identical_scale_t = false;
 
     LLSelectedTE::getObjectScaleS(obj_scale_s, identical_scale_s);
-    LLSelectedTE::getObjectScaleS(obj_scale_t, identical_scale_t);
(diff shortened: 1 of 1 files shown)


marchcat (approved):

marchcat:

Hecklezz:
> Thank you for the contribution!

No Problem!
I actually have a more to come in another PR regarding the texture panel, but this one was important first.

I have already implemented working repeats per meter for PBR.
I have also improved the way repeats per meter works when used on an entire object, or just more than 1 face in general. Currently, the normal and specular values don't respect the dimensions of the face they are currently on and set their values incorrect, whereas diffuse is correct as it does determine the values per face. With what I have implemented, it will correctly apply the normal and specular scales to all faces. This same improvement will also apply to PBRs with its repeats per meter.

PR #3883, opened 2025-04-07
Fix: Remove potentially redundant RenderAutoHideSurfaceAreaLimit sett…

williamweaver:
This pull request removes a seemingly duplicated `connectRefreshCachedSettingsSafe` call in `LLPipeline::init()` for the `RenderAutoHideSurfaceAreaLimit` setting.

**Context:**

During a routine code review of `LLPipeline::init()`, a duplicated registration for the `RenderAutoHideSurfaceAreaLimit` setting was identified. While the impact is likely minimal, double registration can lead to:

* Potential for the associated callback function to be executed twice, increasing CPU overhead.
* The second registration overriding the first, rendering the first call ineffective.

It's *possible* this duplication was introduced with commit 440c7b2 (Added CollectFontVertexBuffers feature); however, further investigation by other developers is encouraged.

**Testing:**

To ensure the intended functionality remains intact, I performed the following tests after removing the duplicate registration:

* Verified that setting the value to 0 (zero) causes all objects to appear, regardless of size.
* Confirmed that values slightly above zero result in only small objects appearing, with all others hidden.
* Demonstrated that increasing the value causes objects of increasing size to appear, while smaller objects remain visible (additive behavior).

A video screencapture demonstrating this testing process has been embedded within this pull request for easy review.

SecondLifeViewer_GzaEnDWrrS2

This change is submitted for careful review by other developers to ensure it has no unintended side effects. Please verify these observations and provide any feedback. The goal is to improve code quality without impacting user experience.

### DIFF
--- a/indra/newview/pipeline.cpp
+++ b/indra/newview/pipeline.cpp
@@ -599,7 +599,6 @@ void LLPipeline::init()
     connectRefreshCachedSettingsSafe("RenderMirrors");
     connectRefreshCachedSettingsSafe("RenderHeroProbeUpdateRate");
     connectRefreshCachedSettingsSafe("RenderHeroProbeConservativeUpdateMultiplier");
-    connectRefreshCachedSettingsSafe("RenderAutoHideSurfaceAreaLimit");
 
     LLPointer<LLControlVariable> cntrl_ptr = gSavedSettings.getControl("CollectFontVertexBuffers");
     if (cntrl_ptr.notNull())


Geenz:
This looks fine to me, by I'm adding @akleshchev to take a look just in case as the commit mentioned was his originally.

akleshchev (approved):

akleshchev:
Looks good to me.

>It's possible this duplication was introduced with commit https://github.com/secondlife/viewer/commit/440c7b20dab3aa09c04bf3e72b4997181e585cbb (Added CollectFontVertexBuffers feature); however, further investigation by other developers is encouraged.

I just changed an existing call. Both calls were already there and I didn't notice a dupplicate.

Geenz:
Okay sounds good! Thank you @akleshchev! I'll go ahead and merge.  Thank you again @williamweaver!

PR #3889, opened 2025-04-08
AA Slowing down the viewer on Mac FIX

mobserveur:
AA was causing the viewer to slow down when used with hdr & emissive option ON. This commit fixes the issue, by making sure mipmap generation is inactive on the Luminance framebuffer.


Geenz:
Thanks @mobserveur for the contrib!

So there's one glaring problem with this - and that's we do actually use a mipmap from the luminance map.

Specifically when we generate the exposure map, we bind the luminance map to the emissive uniform (line 7116).  The exposure generation shader uses it like this:
```GLSL
    float L = textureLod(emissiveRect, tc, 8).r;
                                        // ^ the mip in question
```

What would be a better approach is to instead just reduce the resolution of the luminance map since we only ever really use it here, and at a much higher mip.

mobserveur:
> Thanks @mobserveur for the contrib!
> 
> So there's one glaring problem with this - and that's we do actually use a mipmap from the luminance map.
> 
> Specifically when we generate the exposure map, we bind the luminance map to the emissive uniform (line 7116). The exposure generation shader (exposureF.glsl) uses it like this:
> 
> ```glsl
>     float L = textureLod(emissiveRect, tc, 8).r;
>                                         // ^ the mip in question
> ```
> 
> What would be a better approach is to instead just reduce the resolution of the luminance map since we only ever really use it here, and at a much higher mip. In this context, being a 256x256 texture, mipmap 8 is going to be 1x1.

I understand. I did actually notice frametime spikes related to exposure and luminance in my own tests, sometimes .. after the "fix" has been applied. It did fix the slow down but those spikes were appearing. So i understand the reason now. I wonder why this mipmap generation was causing the issue though. I will investigate on my side too, trying to understand what's going on. Also i posted an other feedback on the canny, this might very well be related somehow : https://feedback.secondlife.com/bug-reports/p/regular-and-high-frametime-spikes-on-mac

Geenz:
@mobserveur  Since it seems there needs to be more investigation around this - I'm gonna move this into a draft PR.

Geenz:
Hey @mobserveur! Given the length of time since you've last poked this, we're gonna go ahead and close this.  Feel free to open a new one once you're ready.

PR #3896, opened 2025-04-10
Refactor tonemap blending to preserve HDR detail during mix

williamweaver:
The blending operation for the `tonemap_mix` uniform in `postDeferredTonemap.glsl` previously used the linear input color, clamped to the [0, 1] LDR range, as the target for the linear portion of the `mix()` function when `tonemap_mix < 1.0`.

This approach meant that High Dynamic Range (HDR) information above `1.0` in the linear input was not fully represented in the blend target. As `tonemap_mix` was reduced, the blend proceeded towards this clamped representation, causing highlights to visually compress or lose intensity relative to the original linear scene data earlier in the blend process.

This commit modifies the `toneMap` and `toneMapNoExposure` functions to adjust this behavior:

*   The original linear input color is preserved before exposure/processing.
*   The appropriate exposure factor is calculated and applied separately.
*   The chosen tone mapping operator is applied to the exposed color, storing the result.
*   The `mix()` function now blends between the appropriately scaled, *unclamped* linear input color and the fully tone-mapped result.
*   The final clamp to the `[0, 1]` LDR range occurs *after* the blend operation.

This change ensures that HDR information from the linear input is considered throughout the blending process. This results in a visually smoother transition as `tonemap_mix` is adjusted, allowing bright highlights to retain more of their original intensity and detail during the blend. The `tonemap_mix` control now functions more closely to an intensity blend between the linear (scaled) and tonemapped representations, reducing the appearance of highlight compression during the transition.

The computational cost is negligible.

### DIFF
--- a/indra/newview/app_settings/shaders/class1/deferred/tonemapUtilF.glsl
+++ b/indra/newview/app_settings/shaders/class1/deferred/tonemapUtilF.glsl
@@ -117,27 +117,34 @@ uniform float exposure;
 uniform float tonemap_mix;
 uniform int tonemap_type;
 
+
 vec3 toneMap(vec3 color)
 {
 #ifndef NO_POST
+    vec3 linear_input_color = color;
+
     float exp_scale = texture(exposureMap, vec2(0.5,0.5)).r;
+    float final_exposure = exposure * exp_scale;
+    vec3 exposed_color = color * final_exposure;
 
-    color *= exposure * exp_scale;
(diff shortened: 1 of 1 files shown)


Geenz (approved):
LGTM!  Thank you!

Dan-Linden:
@williamweaver Do you have any test content or environment that clearly demonstrates this improvement?

PR #3901, opened 2025-04-12
Refactor LLWindowWin32::LLWindowWin32Thread::wakeAndDestroy()

marchcat:
Key improvements:
1. Better error handling for thread detachment
2. More careful sequencing of operations
3. Ensuring the window is hidden before proceeding with other shutdown steps
4. Proper checking of thread joinability before detaching

### DIFF
--- a/indra/llwindow/llwindowwin32.cpp
+++ b/indra/llwindow/llwindowwin32.cpp
@@ -4846,39 +4846,60 @@ bool LLWindowWin32::LLWindowWin32Thread::wakeAndDestroy()
 {
     if (mQueue->isClosed())
     {
-        LL_WARNS() << "Tried to close Queue. Win32 thread Queue already closed." <<LL_ENDL;
+        LL_WARNS() << "Tried to close Queue. Win32 thread Queue already closed." << LL_ENDL;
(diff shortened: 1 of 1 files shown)


marchcat:
Oops, I accidentally merged it instead of enabling post-review auto-merge. 
In any case, this change shouldn't be harmful, but it actually seems to improve shutdown behavior under debugging.

indra/llwindow/llwindowwin32.cpp
```diff
@@ -4846,41 +4846,62 @@ bool LLWindowWin32::LLWindowWin32Thread::wakeAndDestroy()
...
-    SetWindowLongPtr(old_handle, GWLP_USERDATA, NULL);
 
-    // Let thread finish on its own and don't block main thread.
-    for (auto& pair : mThreads)
+    // Clear the user data to prevent callbacks from finding us
+    if (old_handle)
     {
-        pair.second.detach();
+        SetWindowLongPtr(old_handle, GWLP_USERDATA, NULL);
     }
```
akleshchev, line 4868: I think we need this callback to handle the destruction.

Issue #3897, opened 2025-04-11
Starred items/folders can be hidden when opening folders with default Inventory floater size


AtlasLinden:
### Environment

Maint C Test build 7.1.12.14392491737
Or any viewer with Inventory Favorites

### Description

Favourite items include a star symbol to the very right of their names in Inventory.
If a folder with long-named (+permissions) items is opened, the star symbols can be "hidden" due to possibly needing a scroll bar to move right. 

https://gyazo.com/e4621db48830950c1b7a42182555226f

Likely expected behaviour.

### Reproduction steps

See gyazo clip above

kylelinden:
Feature is pointless without the stars being visible all the time.

marchcat:
According to the [comment](https://github.com/secondlife/viewer/commit/154d15d42d4ca200dec9e404464cf86eaa384a32#commitcomment-155483782):

> If the inventory window is not wide enough to show the full item name, the icon is 
> 1. overflowing the text
> 2. moving into the inventory panel if you scroll to the right using the horizontal scrollbar


Overflowing is intended by this patch, but still not sure about the horizontal scrolling.

Dan-Linden:
Passed QA.
Verified on Second Life Release 7.1.15.15352812358 on Win11/OSX in the scope of https://github.com/secondlife/iqa/issues/500

![Image](https://github.com/user-attachments/assets/18e5f15f-1e71-400e-b3b5-ecf1b41a5c38)

PR #3921, opened 2025-04-16
#3897 Fix favorite's stars being hidden in some cases #2

akleshchev:
1. Star should follow scroll
2. Account for a star in label's width


marchcat (approved):
Thank you!

maxim-productengine (approved):

PR #3926, opened 2025-04-17
Added favorites' star for individual items in appearance floater

akleshchev:


akleshchev:
Does not yet subscribe to dynamic color changes, but neither does the gallery, so should be enough for now.

maxim-productengine (approved):

marchcat (approved):

PR #3927, opened 2025-04-17
Add a bunch of old and new math improvements

Ansariel:
These include:

- Use standard library functions for llisnan and replace obvious NaN checks that don't work using /fp:fast floating point behavior under MSVC
- Fix lerp issues on newer GCC versions (from old `develop` branch)
- Make eligible functions constexpr (from old `develop` branch)
- Use constants for vector indices where applicable (from old `develop` branch)
- Reformat to match actual coding conventions (from old `develop` branch)

### DIFF
--- a/indra/llappearance/llpolymorph.cpp
+++ b/indra/llappearance/llpolymorph.cpp
@@ -550,12 +550,12 @@ void LLPolyMorphTarget::apply( ESex avatar_sex )
 
     mLastSex = avatar_sex;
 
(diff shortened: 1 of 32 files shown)


marchcat (approved):
This one is great. Thank you, @Ansariel!

I had planned to do this `2025.05`, but your PR already covers an even wider scope.

marchcat:
That build error is fixed in `2025.04`, but not in `develop` yet.
I checked this build on my setup and it's fine. Merging.

PR #3930, opened 2025-04-17
Restore currently entered text in chat entry textbox after going through history with Ctrl-PgUp/PgDown

Ansariel:
Another contribution that got lost.


Geenz (approved):
This looks good!  Thanks!

Issue #3954, opened 2025-04-20
OpenJPEG issue uploading small textures can't convert to j2c


Hecklezz:
### Environment

Second Life Test 7.1.13.251101243 (64bit)
<details>
<summary>Rest of environment-</summary>
Release Notes

You are at 14.2, 140.4, 27.3 in Cloud Sandbox 3 located at simhost-08e46df2ed0f4d348.aditi
SLURL: secondlife://Aditi/secondlife/Cloud%20Sandbox%203/14/140/27
(global coordinates 305,934.0, 235,404.0, 27.3)
Durian Scones 2025-03-14.13862207703
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5737)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6636
OpenGL Version: 4.6.0 NVIDIA 566.36

Window size: 2220x1250
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (27.5% used)

J2C Decoder Version: OpenJPEG runtime: 2.5.2
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/10,840 (0.0%)
April 20 2025 06:18:07
</details>

### Description

PR with potential fix: https://github.com/secondlife/viewer/pull/3955

While using a viewer without KDU support (therefore using OpenJPEG), uploading small textures (those smaller than 64 pixels in either axis after being biased scaled to power of two, therefore anything < 57 pixels) results in it failing to upload and giving the error "Couldn't convert the image to jpeg2000".

[...]

PR #3955, opened 2025-04-20
[#3954] Fix uploading small textures with OpenJPEG

Hecklezz:
[#3954](https://github.com/secondlife/viewer/issues/3954)

Fixes the above issue by ensuring that as long as the image's width and height are >= MIN_IMAGE_SIZE, set their tile counts to a minimum of 1 and not 0. Previously they were being set to 0 which caused encoding to fail, and textures that were <64 pixels in either width or height would not upload while using OpenJPEG. 

Also contains a minor refactor of an old log that was incorrectly logging a warning when encoding succeeded but said it was returning false.


akleshchev (approved):

marchcat:
Thank you!

PR #3970, opened 2025-04-24
Reduce cost of joint lookups by reducing string allocations via use of std::string_view and heterogeneous map lookups

Ansariel:


indra/llcharacter/llbvhloader.h
```diff
@@ -227,7 +227,7 @@ class LLBVHLoader
     friend class LLKeyframeMotion;
 public:
     // Constructor
-    LLBVHLoader(const char* buffer, ELoadStatus &loadStatus, S32 &errorLine, std::map<std::string, std::string>& joint_alias_map );
+    LLBVHLoader(const char* buffer, ELoadStatus &loadStatus, S32 &errorLine, std::map<std::string, std::string, std::less<>>& joint_alias_map );
```
akleshchev, line 230: '>>' probably going to need a space inside to not fail on mac.

akleshchev (approved):

indra/llcharacter/llbvhloader.h
Ansariel: That hasn't been the case for ages.

marchcat:
Thank you, @Ansariel!

marchcat (approved):

Issue #3997, opened 2025-04-29
Crash when near or in sight of newly uploaded Collada Mesh


aiaustin:
### Environment

Second Life Release 7.1.13.14343205944 (64bit)
Release Notes

You are at 128.0, 128.0, 32.5 in Morris located at simhost-0103149d6e3418fac.aditi
SLURL: secondlife://Aditi/secondlife/Morris/128/128/32
(global coordinates 255,360.0, 256,384.0, 32.5)
Bleeding Edge 2025-04-24.14649052994
Release Notes

CPU: Intel(R) Xeon(R) CPU E5-1620 v3 @ 3.50GHz (3491.92 MHz)
Memory: 32689 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5796)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.7283
OpenGL Version: 4.6.0 NVIDIA 572.83

Window size: 1249x864
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 10000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (7.8% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/6,745 (0.0%)
April 29 2025 03:58:51

### Description

A newly uploaded Collada mesh to the Beta grid was placed in world at this location by avatar "Be Austin" on 29-Apr-2025.  Nominal mesh upload cost was L$211. As it stared to rez in world the viewer crashed.

secondlife://Aditi/secondlife/Rumpus%20Room%202048/128/128/22

If I try to relog to that location it instantly crashes the viewer using Second Life Release 7.1.13.14343205944  or Firestorm Beta Firestorm 7.1.13 (77930) with the original avatar that placed the object (Be Austin) or an alternative (Ai Austin).

[...]

PR #3999, opened 2025-04-29
#3997 Crash in a gltf asset enabled region

akleshchev:


marchcat (approved):

Hecklezz:
Related to this same crash, on [line 1672](https://github.com/secondlife/viewer/blob/65d70a8d8f211b462481e93f919a100c8b3b2af5/indra/llrender/llvertexbuffer.cpp#L1672) in llvertexbuffer.cpp, LLGLSLShader::sCurBoundShaderPtr is a nullptr so this assert fails, and then [line 1674](https://github.com/secondlife/viewer/blob/65d70a8d8f211b462481e93f919a100c8b3b2af5/indra/llrender/llvertexbuffer.cpp#L1674) will throw a null reference exception.

aiaustin:
Question, should there not be something server side to protect against such a crash as the viewers Colada Mesh Uploader does allow the example mesh to be uploaded, save and rezzed without indicating any error at all.

Hecklezz:
> Related to this same crash, on [line 1672](https://github.com/secondlife/viewer/blob/65d70a8d8f211b462481e93f919a100c8b3b2af5/indra/llrender/llvertexbuffer.cpp#L1672) in llvertexbuffer.cpp, LLGLSLShader::sCurBoundShaderPtr is a nullptr so this assert fails, and then [line 1674](https://github.com/secondlife/viewer/blob/65d70a8d8f211b462481e93f919a100c8b3b2af5/indra/llrender/llvertexbuffer.cpp#L1674) will throw a null reference exception.

Fixing this then causes a nullptr exception also on [line 761](https://github.com/secondlife/viewer/blob/65d70a8d8f211b462481e93f919a100c8b3b2af5/indra/newview/gltfscenemanager.cpp#L761) in gltfscenemanager.cpp since it is accessing LLGLSLShader::sCurBoundShaderPtr again which is nullptr as described above.

akleshchev:
> should there not be something server side

1. Server side is not responsible for rendering
2. This crash happens because viewer uses an experimental feature on this region.

> LLGLSLShader::sCurBoundShaderPtr

Thank you, will check that.

Hecklezz:
The latest hotfix works and the viewer is no longer crashing due to this specific bug.

PR #4018, opened 2025-05-04
calcDataSizeJ2C Adjust curve for more than 6 layers

TommyTheTerrible:
KDU is uploading 2k files with 7 and 8 layers which is shifting the location of discard 1 and 2.

To accommodate, this commit adds a max_layer check based on max_dimension and the MAX_BLOCK_SIZE to allow the extra layers for 2k and larger sizes.

Also shifted the starting total size to the MIN_LAYER_SIZE instead of MAX_BLOCK_SIZE's area to allow smaller files to be decoded at discard 5 completely.

Finally able to walk around Fantasy Faire without any gray blobs!


marchcat (approved):
Thank you!

PR #4019, opened 2025-05-04
updateImageDecodePriority - Avoid Long Face Loop

TommyTheTerrible:
To avoid running a long loop on thousands of faces, some textures were being set to a BOOST level to avoid the updateImageDecodePriority function entirely but this was causing many of them to never be deleted over the course of a user's travels.

Instead of relying on BOOST, this commit changes the logic of the texture channel loop such that the face loop will only run if the number of faces is below the threshold.

To do this, we move the face_count incrementing outside of the face loop into the channel loop and increment it using the getNumFaces function instead.

We then check the face_count against the maximum number of faces we want to check and if it exceeds the number we set the number of faces for the face loop to check down to zero.

This avoids branch prediction misses and the long face loop issue.

Later, if the face_count is above the threshold, we assign the virtual size to the maximum.

I personally believe the max_faces_to_check should be lower than 1024, but I left that value in for continuity. I use 64 faces as my max on my compiled version of the viewer without any noticeable issues for memory use.

### DIFF
--- a/indra/newview/llviewertexturelist.cpp
+++ b/indra/newview/llviewertexturelist.cpp
@@ -926,7 +926,7 @@ void LLViewerTextureList::updateImageDecodePriority(LLViewerFetchedTexture* imag
         {
             face_count += imagep->getNumFaces(i);
             S32 faces_to_check = (face_count > max_faces_to_check) ? 0 : imagep->getNumFaces(i);
-            
+
             for (S32 fi = 0; fi < faces_to_check; ++fi)
             {
                 LLFace* face = (*(imagep->getFaceList(i)))[fi];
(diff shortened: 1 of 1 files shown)


TommyTheTerrible:
Sorry for the additional small commit. I did not realize how much compilers like knowing when the incrementing value will increment. Reduces the number of instructions on GCC, most likely Clang as well, and might allow for parallelization.

indra/newview/llviewertexturelist.cpp
```diff
@@ -923,13 +924,15 @@ void LLViewerTextureList::updateImageDecodePriority(LLViewerFetchedTexture* imag
         LL_PROFILE_ZONE_SCOPED_CATEGORY_TEXTURE;
         for (U32 i = 0; i < LLRender::NUM_TEXTURE_CHANNELS; ++i)
         {
-            for (S32 fi = 0; fi < imagep->getNumFaces(i); ++fi)
+            face_count += imagep->getNumFaces(i);
+            S32 faces_to_check = (face_count > max_faces_to_check) ? 0 : imagep->getNumFaces(i);
+
```
marchcat, line 929:
```suggestion
            S32 faces_to_check = (face_count > max_faces_to_check) ? 0 : imagep->getNumFaces(i);

```

marchcat (commented):
Thank you for the contribution.

Please fix the trailing whitespace so that this change passes the pre-commit check.

Issue #4022, opened 2025-05-05
Windows 11 24H2 fullscreen mode failing


Hecklezz:
### Environment

Second Life Release 7.1.13.14343205944 (64bit)
<details>
<summary>Rest of environment-</summary>
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5737)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6636
OpenGL Version: 4.6.0 NVIDIA 566.36

Window size: 2220x1250
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 64m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (9.3% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version:  

May 05 2025 01:35:08
</details>

### Description

PR with fix: https://github.com/secondlife/viewer/pull/4023

Since Windows 11 24H2, FullScreen mode no longer works and errors with the message "Unable to run fullscreen at [WIDTH] x [HEIGHT]. Running in window."

This is caused by a change by Microsoft where the call to "ChangeDisplaySettings" now fails if correct width and height values are not provided that match available FullScreen dimensions. 

Since Second Life Viewer does not have the FullScreen setting exposed in preferences (but is in debug settings), this issue is unlikely to effect those users, however most TPV's such as Firestorm do have the FullScreen setting available in preferences to use.

### Reproduction steps

1. Have Windows 11 24H2 installed
2. Ensure the viewer window is not already maximised and is at a random window size
3. Enable FullScreen (currently only available in debug settings, but TPV's have this exposed in preferences)
4. Restart the viewer and it should present with an error saying "Unable to run fullscreen at [WIDTH] x [HEIGHT]. Running in window."

[...]

PR #4023, opened 2025-05-05
[#4022] Fix FullScreen mode in Windows 11 24H2

Hecklezz:
https://github.com/secondlife/viewer/issues/4022

Fixes the issue mentioned above by ensuring WindowWidth and WindowHeight in Windows are set to the current displays width and height before initialising the viewer window.

### DIFF
--- a/indra/newview/llappviewerwin32.cpp
+++ b/indra/newview/llappviewerwin32.cpp
@@ -816,6 +816,29 @@ bool LLAppViewerWin32::reportCrashToBugsplat(void* pExcepInfo)
     return false;
 }
 
+bool LLAppViewerWin32::initWindow()
+{
+    // This is a workaround/hotfix for a change in Windows 11 24H2 (and possibly later)
+    // Where the window width and height need to correctly reflect an available FullScreen size
+    if (gSavedSettings.getBOOL("FullScreen"))
+    {
+        DEVMODE dev_mode;
+        ::ZeroMemory(&dev_mode, sizeof(DEVMODE));
+        dev_mode.dmSize = sizeof(DEVMODE);
+        if (EnumDisplaySettings(NULL, ENUM_CURRENT_SETTINGS, &dev_mode))
+        {
+            gSavedSettings.setU32("WindowWidth", dev_mode.dmPelsWidth);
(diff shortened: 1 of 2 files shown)


akleshchev:
I already made a fix for that before:
https://github.com/secondlife/viewer/pull/3337/files

I assume that the same issue.

Ansariel:
> I assume that the same issue, so commit just needs to be cherry picked into new develop.

Can you do that so we can see if that fixes the issue? Because as of now the fix is only in the archived `develop`  branch.

akleshchev:
Done.

Hecklezz:
With a viewer build from develop with those changes merged in, and on firestorm with that commit cherry-picked in, the error still occurs that I described in the issue.
Recording of error-
https://github.com/user-attachments/assets/2bb66be4-2b42-45a2-943f-619b21d80eb8


If at all possible, could you test my reproduction steps on Windows 11 24H2? (I installed it on a VM since I run Windows 10 still on my host)
I can record and upload a short video if it helps to reproduce so you can test both your and my changes.

akleshchev (approved):

akleshchev:
> If at all possible, could you test my reproduction steps on Windows 11 24H2?

Tested, can repro again despite my fix(

Hecklezz:
Here is a recording with my changes (without your changes from the above commit, and I fully deleted the Secondlife folder in appdata before testing again, only adjust the window so it isn't maximised and enabled advanced menu)-
https://github.com/user-attachments/assets/520b59e9-2490-494f-944a-0ea7338bde64

akleshchev:
Result of WindowWidth setting being overwritten is not nice. But given this is a debug setting not exposed to users should be fine.

But I suggest using some kind of
```
LLAppViewer::getWindowDimentions(S32 &width, S32 &height)
{
    width = gSavedSettings.getU32("WindowWidth");
    height = gSavedSettings.getU32("WindowHeight");
}
```
and overload it for LLAppViewerWin32

Hecklezz:
I can double check with a breakpoint, but I believe they get overridden anyway here-
https://github.com/secondlife/viewer/blob/6e6e105f01726208d60c060c4de653cd15470db3/indra/newview/llviewerwindow.cpp#L2550

I'll make some code changes and do that now and edit this post with the result

PR #4033, opened 2025-05-06
GLTF import texture upload

marchcat:


Geenz (approved):

marchcat:
@Geenz, thank you!  
Actually, scaling to 1/100th helps with some models, but it makes things worse for others (mostly the ones I downloaded from Sketchfab).  
I'll add another commit for dynamic scaling once I resolve the conflicts with your branch — rigging seems to have affected it somehow.

aiaustin:
I mentioned this to Geenz already, but I wonder if either of you are seeing GLTF mesh import messages that indicate "Error: Material of model is not a subset of reference model"?

Seen on all model uploads from Blender  4.4.3 GKTF/GLB built in exporter, even simple untextured cube using geenz/gltf-mesh-import as at approx. 6-May-2025. Maybe its fixed in recent work?

[Cube-GLB.zip](https://github.com/user-attachments/files/20210945/Cube-GLB.zip)

Geenz:
@aiaustin Please do not clutter up PRs with bug reports.  File an issue instead.

Issue #4037, opened 2025-05-07
World Map Find removing results undesirably


Hecklezz:
### Environment

Second Life Release 7.1.13.14343205944 (64bit)
<details>
<summary>Rest of environment-</summary>
Release Notes

You are at 135.2, 133.0, 27.2 in Cloud Sandbox 1 located at simhost-0204807ba3eb587d2.aditi
SLURL: secondlife://Aditi/secondlife/Cloud%20Sandbox%201/135/133/27
(global coordinates 305,799.0, 235,397.0, 27.2)
Elderberry Eclairs 2025-04-24.14648572313
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5737)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6636
OpenGL Version: 4.6.0 NVIDIA 566.36

Window size: 1634x928
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 64m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (10.1% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 0/1,919 (0.0%)
May 06 2025 22:00:02
</details>

### Description

PR with fix: https://github.com/secondlife/viewer/pull/4038

Using the Find region functionality on the world map, depending on the response time of the Sim and the time it takes for the user to let go of the button, can result in the resulting regions not displaying all desired results.

[...]

PR #4038, opened 2025-05-07
[#4037] Fix world map Find button potentially removing results undesirably

Hecklezz:
https://github.com/secondlife/viewer/issues/4037

Fixes the above issue by ensuring the World Map Find button only triggers the callback on mouse down and not mouse up, resulting in it only sending 1 request, disallowing it to update the search query before the user lets go of mouse down and it sending a 2nd request with a now-changed search query on mouse-up.

Video showing fix-
https://github.com/user-attachments/assets/59877f8d-6063-4106-95f6-a4cd3b05b51a


marchcat (approved):
Thank you!

PR #4040, opened 2025-05-07
Fix map search bug causing duplicate requests properly

Ansariel:
This fix merged in #4038 is wrong as well as the explanation to it. The duplicated request stems from the fact that the search editor `"location"` triggers the same commit callback as the "Search" button, which happens as soon as the search editor loses focus. All that change does is reducing the time between the execution of the callbacks from the search editor and the button (the callback from the button is now called immediately as the button is pressed and not when the mouse button is released), the search is still triggered twice. That only works purely by coincidence because the second search request is sent before the result of the first request is received, Effectively it is the same if you click the search button fast (as in press and release the mouse button quickly).

The actual cause is - as already pointed out - the slightly delayed execution of the second search request after the result of the first request has already been received: A result containing at least one entry will select the an entry in the result list, which then updates the search editor. The then executed search request will use the updated text in the search editor and will then obviously return only the one entry with the exact match.

This fix causes the actual issue by removing the unnecessary commit callback for the search editor and instead fixing the incorrect setting of the default button, which is applied to the floater. However, the panel actually evaluating the default button in its `LLPanel::handleKeyHere` method is `layout_panel_4`, which is the parent of all tracking-related controls. Removing the commit callback for the search editor and fixing the correct default button handling will now properly invoke the commit callback from the default search button without sending a duplicate request. This also fixes the unexpected behavior of actually triggering a search when the search editor loses focus.

Also changed it so the pointer to the widgets is directly passed into the `LLPanel::setDefaultBtn` method as passing the name will cause a `getChild` XUI lookup each time it is called - and it is called each time you enter anything in the search editor.

### DIFF
--- a/indra/newview/llfloaterworldmap.cpp
+++ b/indra/newview/llfloaterworldmap.cpp
@@ -406,9 +406,12 @@ bool LLFloaterWorldMap::postBuild()
     F32 slider_zoom = mMapView->getZoom();
     mZoomSlider->setValue(slider_zoom);
 
+    mTrackCtrlsPanel = getChild<LLPanel>("layout_panel_4");
+    mSearchButton = getChild<LLButton>("DoSearch");
+
     getChild<LLPanel>("expand_btn_panel")->setMouseDownCallback(boost::bind(&LLFloaterWorldMap::onExpandCollapseBtn, this));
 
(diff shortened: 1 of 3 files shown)


marchcat (approved):
Thank you, @Ansariel!

PR #4044, opened 2025-05-08
Clean up worldmap: Remove LLCtrlListInterface and reduce number of getChild calls

Ansariel:


marchcat (approved):
Thank you, @Ansariel!

PR #4048, opened 2025-05-08
LLTextureFetchWorker::doWork - Add Decode Queue Size Check

TommyTheTerrible:
Make sure there is room in the decode queue before submitting work from the texture fetch worker.

Used a greater than or equal comparison just in case something weird happens, against 1024 since it's the default capacity value for LLThreadSafeQueue and returning false to make sure it attempts the decode again until decode queue length reduces. Threw a message into debug so we can see when it is happening.

This might not seem necessary but I have hit this limit in the last several days while testing ideas for the texture fetch queue and it caused the viewer to seize.

Obligatory, OpenJPEG 2.5.3 and X3D cache goes BRRRRR!

### DIFF
--- a/indra/newview/lltexturefetch.cpp
+++ b/indra/newview/lltexturefetch.cpp
@@ -1758,7 +1758,7 @@ bool LLTextureFetchWorker::doWork(S32 param)
             LL_DEBUGS(LOG_TXT) << mID << " DECODE_IMAGE abort: mLoadedDiscard < 0" << LL_ENDL;
             return true;
         }
-        if ((S32)LLAppViewer::getImageDecodeThread()->getPending() + 1 >= 1024)
+        if ((S32)LLAppViewer::getImageDecodeThread()->getPending() + 1 >= 512)
         {
             // No room in decode queue, wait in state for opening. LLThreadSafeQueue default is 1024.
             LL_DEBUGS_ONCE(LOG_TXT) << mID << " DECODE_IMAGE wait: Decode queue full!" << LL_ENDL;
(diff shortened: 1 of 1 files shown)


TommyTheTerrible:
I forgot to increment the decode queue count by one to make sure a new job would fit. Sorry for the tiny commit.

TommyTheTerrible:
Adding one was not enough because of the number of threads in use contributed to the real number but there was no way to find out how many threads easily so I just halved the 1024 limit down to 512.

TommyTheTerrible:
This is not the right way to do this. The getPending function is not reading from an atomic, so the compiler does not know to check each time and each fetch threadpool check is reading old information.

Because of this, the only way to keep the decode pool from being overloaded is to keep the fetch pool from processing more textures than slots available in the decode threadpool and this cannot be done inside the fetch threadpool itself.

For those wondering, this is the error (and seizure) I am trying to resolve:

`2025-05-10T05:02:58Z WARNING #ThreadPool# llcommon/llthreadsafequeue.h(314) LLThreadSafeQueue<struct tagMSG,class std::queue<struct tagMSG,class std::deque<struct tagMSG,class std::allocator<struct tagMSG> > > >::push_ : ONCE: Threadsafe queue push_(lockacquired) queue full 1024 >= 1024`

I was trying to run the fetch threadpool as fast as possible, but I see now that it has to be paced depending on the decode threadpool size because the one is passing directly to the other and we cannot rely on what each thread in one knows about the other threadpool's state.

The specific situation that seems to cause this reliably is allowing the processing of hundreds of small fetch threads (discard 5 of 1024x1024 or smaller images) that can finish quickly from fast local storage (SSD).

I will test to see if a volatile atomic solves the issue but for now just going to limit the number of times the updateFetch function is run per frame.

Closing this PR but thought an explanation was worth sharing.

PR #4053, opened 2025-05-09
#4046 Fix hud vs inworld text color mismatch

akleshchev:
This reverts commit e2cf375179b807b847414206bf79617e4e9889ae. Commit e9889ae originaly was meant to fix color mimatch between text and face, but looks like it is no longer needed and instead adds mismatch between inworld and hud.


marchcat (approved):

maxim-productengine (approved):

Geenz:
Merging.

PR #4063, opened 2025-05-12
Modernize away possibly missing std::less comparator

Ansariel:


marchcat (approved):
Thank you, @Ansariel!

PR #4078, opened 2025-05-13
Fix for Mesh missing after TP. Assume MeshRezEnabled if SimFeatures not yet received

beqjanus:
This PR addresses one of the causes of the increased missing meshes following a TP. I do not believe it will fix the "missing mesh on login" which may be related but still occurs. It has completely eliminated the loss of body parts and other meshes going missing after a TP in all of my testing and regular usage.

The check in llvovolume for gMeshRepo.meshRezEnabled() has the side-effect of testing the region asset caps. After TP this is frequently failing and any mesh items in rebuildGeom at that moment will be excluded. 

Another option is to remove the check in llvovolume.cpp but as that was deliberately added as part of SL-5161 it is not clear whether that would be acceptable. 

Monty Linden confirmed that no current server version will decline that asset cap. 

Note that all this does is change the expected response to optimistic. If the region returned false, subsequent updates would be treated correctly.

### DIFF
--- a/indra/newview/llviewerregion.cpp
+++ b/indra/newview/llviewerregion.cpp
@@ -3721,6 +3721,10 @@ bool LLViewerRegion::bakesOnMeshEnabled() const
 
 bool LLViewerRegion::meshRezEnabled() const
 {
+    if(!mSimulatorFeaturesReceived)
+    {
+        return true;
+    }
     return (mSimulatorFeatures.has("MeshRezEnabled") &&
                 mSimulatorFeatures["MeshRezEnabled"].asBoolean());
 }


Ansariel:
As already discussed internally, `LLViewerRegion::meshRezEnabled()` should either be removed entirely if it is guaranteed mesh reziing will never be disabled again.

Otherwise the correct way to handle this situation would be to check for `LLViewerRegion::simulatorFeaturesReceived()`  in `LLVolumeGeometryManager::rebuildGeom(LLSpatialGroup* group)`. If the simulator features haven't been received yet, register for the callback when these arrive via `LLViewerRegion::setSimulatorFeaturesReceivedCallback` and store the affected LLSpatialGroups to have their geometry rebuild when the callback is triggered.

indra/newview/llviewerregion.cpp
```diff
@@ -3721,6 +3721,10 @@ bool LLViewerRegion::bakesOnMeshEnabled() const
 
 bool LLViewerRegion::meshRezEnabled() const
 {
+    if(!mSimulatorFeaturesReceived)
```
akleshchev, line 3724: Suggestion: add an in-code comment that explains why this hack is needed. 

P.S. By this point meshRezEnabled is mission critical. Perhaps just deprecate it and assume meshes are always supported?

akleshchev (approved):

akleshchev:
Thank you for your contribution. But it appears that meshRezEnabled is just another 'new feature off button' that lost it's purpose once feature was deployed fully. After discussion it was decided that it's better to just remove the check.

akleshchev:
https://github.com/secondlife/viewer/pull/4084

PR #4125, opened 2025-05-20
Fix CodeQL errors in 2025.05

marchcat:
The diff is larger than expected due to mixed line endings.


AtlasLinden (approved):
I believe this change was already made in a separate PR with some extra updates directed at Develop but if this is just to suppress errors on 25.05 then LGTM

Issue #4109, opened 2025-05-20
GLTF Mesh Import: Model Parts Wrongly Positioned, Scaled and Rotated


aiaustin:
### Environment

Second Life Project Gltf-mesh-import 7.1.14.15122149339 (64bit)
Release Notes

You are at 124.6, 196.7, 23.0 in Rumpus Room 2048 located at simhost-030b947afda1f10ad.aditi
SLURL: secondlife://Aditi/secondlife/Rumpus%20Room%202048/125/197/23
(global coordinates 36,476.6, 14,276.7, 23.0)
Second Life Server 2025-04-24.14648572313
Release Notes

CPU: Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz (3491.58 MHz)
Memory: 32713 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5854)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 980/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.7652
OpenGL Version: 4.6.0 NVIDIA 576.52

Window size: 1278x889
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 256m
Bandwidth: 2400kbit/s
LOD factor: 3
Render quality: 6
Texture memory: 4096MB
Disk cache: Max size 1638.4 MB (30.1% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Not Connected
Packets Lost: 0/34,756 (0.0%)
May 20 2025 03:17:18

### Description

Several glTF models after upload have parts which are wrongly positioned, scaled or rotated resulting in the model looking "exploded". This may be due to wrong handling of some transformations in the model. This can lead to extremes of some parts being very large indeed and even limiting the scale at which the model can be imported.

A test model for this might be the GLB Mesh Model of the NASA Mars Ingenuity Helicopter

https://github.com/nasa/NASA-3D-Resources/tree/master/3D%20Models/Ingenuity%20Mars%20Helicopter

[...]

Issue #4080, opened 2025-05-14
GLTF Import: Rigged meshes disappear when attached


Geenz:
### Environment

Second Life Test 7.1.14.251291909 (64bit)
Release Notes
Build Configuration RelWithDebInfo

You are at 92.2, 128.0, 22.9 in Avalon Aviary located at simhost-0ed1af4f6c09b0a7a.agni
SLURL: http://maps.secondlife.com/secondlife/Avalon%20Aviary/92/128/23
(global coordinates 252,764.0, 288,128.0, 22.9)
Second Life Preflight 2025-04-24.14649112505
Release Notes

CPU: AMD Ryzen 9 7950X3D 16-Core Processor           (4192.04 MHz)
Memory: 97939 MB
OS Version: Microsoft Windows 11 64-bit (Build 26100.4061)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 3090/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.7640
OpenGL Version: 4.6.0 NVIDIA 576.40

Window size: 2558x1360
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 256m
Bandwidth: 3000kbit/s
LOD factor: 2
Render quality: 6
Texture memory: 24576MB
Disk cache: Max size 1638.4 MB (25.2% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 0/288 (0.0%)
May 14 2025 08:53:49

### Description

Uploading a rigged mesh will result in the mesh not being visible in the upload preview.  However, it will appear in world.  Attempting to attach it will result in the mesh disappearing.

### Reproduction steps

1. Upload rigged mesh
2. Rez it in-world
3. Attempt to attach it

akleshchev:
> This may or may not be related, but uploading any simple GLTF/GLB mesh (unrigged) and attaching it to an avatar, the mesh does not appear.

[...]

PR #4151, opened 2025-05-22
#4109 #4080 Apply coordinate rotation to GLTF inverse bind matrices

marchcat:


marchcat:
Looks like "Joint position override" is still broken.

marchcat:
"Joint position override" is still broken, but now it looks like it was before my changes for #4109.

Issue #4155, opened 2025-05-24
Sending Group notices previous text problems


Hecklezz:
### Environment

Second Life Release 7.1.13.14343205944 (64bit)
<details>
<Summary>Rest of Environment Here-</Summary>
Release Notes

You are at 147.6, 131.0, 27.2 in Cloud Sandbox 1 located at simhost-0204807ba3eb587d2.aditi
SLURL: secondlife://Aditi/secondlife/Cloud%20Sandbox%201/148/131/27
(global coordinates 305,812.0, 235,395.0, 27.2)
Elderberry Eclairs 2025-04-24.14648572313
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5854)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6636
OpenGL Version: 4.6.0 NVIDIA 566.36

Window size: 1710x999
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (5.7% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 1/4,378 (0.0%)
May 24 2025 13:16:29
</details>

### Description

PR with fix-
https://github.com/secondlife/viewer/pull/4159

The culprit commit is this one, since reverting it, the issue goes away-
https://github.com/secondlife/viewer/commit/442ee9795ecde5a89df254b78465c8f4ed46845f
More specifically, the change on [this line](https://github.com/secondlife/viewer/blob/442ee9795ecde5a89df254b78465c8f4ed46845f/indra/llui/lllineeditor.cpp#L1989) is the culprit.

[...]

PR #4159, opened 2025-05-25
[#4155] Add font buffer resetting in LLLineEditor.clear()

Hecklezz:
https://github.com/secondlife/viewer/issues/4155

Fixes the above issue and possibly others undiscovered by adding font buffer resetting into the Line Editor clear function.
This appears to have been an oversight to not clear them here, as they get cleared whenever characters are being added to or removed from mText, but never when it is fully cleared.


marchcat (approved):

akleshchev:
Thank you!

PR #4167, opened 2025-05-26
#4109 Improve LLGLTFLoader::computeCombinedNodeTransform()

marchcat:


indra/newview/gltf/llgltfloader.cpp
```diff
@@ -231,30 +231,35 @@ bool LLGLTFLoader::parseMeshes()
...
-                node_transform = parent_transform * node_transform;
-                break;
-            }
+        if (it != potential_parent.mChildren.end())
+        {
+            // Found parent - recursively get its combined transform and apply it
+            glm::mat4 parent_transform;
+            computeCombinedNodeTransform(asset, static_cast<S32>(i), parent_transform);
+            combined_transform = parent_transform * combined_transform;
+            return; // Early exit - a node can only have one parent
```
akleshchev, line 260: Might be better to just expand 'node' with a parent field and build a tree once.

PR #4201, opened 2025-05-31
Use latest release for changelog

marchcat:


Geenz:
Looking at this - wouldn't this mean that it'll go off of whatever was tagged previously whether it's a release or pre-release build?

Geenz (approved):

PR #4211, opened 2025-06-03
Follow-up fix for bb9adf332950a6495a41aab8c6c4b2436f83f53e: Pressing enter while location spinners have focus now starts a teleport

Ansariel:


marchcat (approved):
Thank you!

Issue #4214, opened 2025-06-04
Support mesh splitting for meshes with more than 8 materials


Geenz:
Adds support for splitting glTF meshes when they have more than 8 materials.  This works around a platform limitation where objects may have up to 8 faces.

This adds parity for the COLLADA importer's ability to split meshes with more than 8 materials.

## Test Plan

1. Pick a glTF mesh that has more than 8 materials on a specific sub-mesh
2. Upload the mesh - ensure that the mesh was split into multiple meshes

aiaustin:
Models uploaded with first Project Viewer 7.1.14.15361077240 - Second Life Project were given base colour and only materials with a texture for base colour were white. This was very useful especially for models with a lot of materials.

Models (unrigged) uploaded with latest version Second Life Project Gltf Mesh Import 7.1.14.15453847512 are all white.

Image shows yesterday's uploaded item as it appears when uploaded before proper materials application, and the white model in front is today's upload of the same model. 

![Image](https://github.com/user-attachments/assets/ba0f7216-8fe7-4477-abe9-6a9078a62b22)

akleshchev:
> Could the changes being made for this issue have caused this regression?

Yes, I had to change how they distribute, but missed that they were initialized. Fix is trivial, preparing it for review.

AtlasLinden:
Tested on Second Life Project Gltf-mesh-import 7.1.14.15119648860 Windows 11
A model was made in Blender that uses 9 materials that are applied to the faces of 2 separate objects.

![Image](https://github.com/user-attachments/assets/6046aefb-f139-4c0b-bef2-f018aa7ee887)

Uploading the model in-world shows that the faces using the last material, "Material9", are treated as a separate mesh 
https://gyazo.com/2902e3a8ea5abc0c68787aa481e23653
Note that Material9 is white. In this clip, the beveled cube has a single white face which is treated as a separate linked object.
The same goes for the white textured faces on the icosphere.

This comment will be updated after testing on MacOS

PR #4215, opened 2025-06-04
#4214 Support mesh splitting for meshes with more than 8 materials

akleshchev:


marchcat (approved):
Thanks. I was about to add something like this as well.

maxim-productengine (approved):

marchcat (approved):

PR #4227, opened 2025-06-07
[#4221] 

Hecklezz:
https://github.com/secondlife/viewer/issues/4221

Fixes the above issue which was caused by mAbsorptionConfigs, mMieConfigs and mRayleighConfigs becoming null since interpolateSDMap wasn't interpolating and combining them properly as they are arrays and not maps. This causes the sky settings to fail the validation check when trying to save them, which resulted in the added sky not being saved into the day cycle. Now those 3 configs correctly get interpolated and should no longer end up null and failing the validation, and allow them to be saved correctly.

### DIFF
--- a/indra/llinventory/llsettingsbase.cpp
+++ b/indra/llinventory/llsettingsbase.cpp
@@ -361,14 +361,12 @@ LLSD LLSettingsBase::interpolateSDValue(const std::string& key_name, const LLSD 
                 new_array = q.getValue();
             }
             else
-            {   // TODO: We could expand this to inspect the type and do a deep lerp based on type.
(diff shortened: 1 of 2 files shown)


Hecklezz:
1 thing I noticed is that the skip contains these 3 configs in it. Are they meant to just be skipped entirely now (which just takes the value from other)? Because last year before this became broken, they weren't being skipped since it was using the skip list inside of LLSettingsBase and not LLSettingsSky. 

If they are meant to be skipped, I'll need to make another small modification to the code to account for this, because right now the skip list is only used inside interpolateSDMap.

marchcat (approved):

PR #4228, opened 2025-06-09
Fix GLTF import missing mesh parts

marchcat:
@akleshchev Please check the part in addModelToScene(), it looks like that should be done inside the loop this way.


akleshchev:
Feels like we should be building a proper parent-child tree instead of doing a backward search.

marchcat:
Yeah, that makes sense. Should be more robust and efficient.

Issue #4170, opened 2025-05-27
GLTF Mesh Import: Meshes Created by GLTF Transform tools do not show all parts


aiaustin:
### Environment

Second Life Project Gltf-mesh-import 7.1.14.15263094146

### Description

While exploring upload of meshes with many parts, I have been using some GLTF Mesh Transformation and Cleanup tools which can flatten and join mesh parts and transform the meshes in various ways. Two are https://gltf.report and https://glb.babylonpess.com both based on the same GitHub available code library. For background see https://blog.inf.ed.ac.uk/atate/2025/04/21/second-life-gltf-mesh-uploader-tests/#tools

While the original Blender 4.4.3 .glb mesh imports correctly (1465 pasts, with multi-part storage in inventory), the tool processed meshes only partially include the parts and have many parts missing.  Both show fine in the Khronos glTF Viewer. glb.babylonpess.com (39 mesh count) and gltf.report (19 mesh count)...

![Image](https://github.com/user-attachments/assets/58f26d01-6fff-4bea-a690-e4c07c3338d6)

![Image](https://github.com/user-attachments/assets/8377e895-b40f-4d53-88b7-0f6041de8906)

But on attempt to upload to Second Life Project Gltf-mesh-import 7.1.14.15263094146 the tool processed meshes show only some of the parts both in the uploader preview and if actually uploaded and rezzed in world.

I attach a zip file with three models, the Blender exported 1465 mesh mode, and the two exported from the tools.

[2025-Supercar-V27-Parts-Test-via-Tools.zip](https://github.com/user-attachments/files/20463251/2025-Supercar-V27-Parts-Test-via-Tools.zip)

### Reproduction steps

1. Examine the three models in Khronos Sample Viewer.  All should look the same, note the number of mesh parts in each and note none have validation errors.
2. Attempt to use the glTF mesh uploader on each and note the tool processed meshes only show a limited number of parts.

aiaustin:
Yes, works for me too now in latest build I just downloaded.  Second Life Project Gltf-mesh-import 7.1.14.15281774530. Great.  This issue should be close as resolved.  Thanks.

![Image](https://github.com/user-attachments/assets/5587c1b9-5651-42ed-a0d5-b973d748d35b)

marchcat:
This processed model is much cheaper to upload and rezzes faster as well.

aiaustin:
[...]

PR #4234, opened 2025-06-09
#4170 Use GLTF scene definition for node traversal

marchcat:
Process nodes through the scene hierarchy as defined in the GLTF file instead of attempting to reconstruct parent-child relationships. This ensures proper import of models created by GLTF transform tools.


akleshchev:
Should we even support more than one scene?

akleshchev:
They do, but should they really be uploaded all at once? instead of something like 'Select a scene to upload'

PR #4239, opened 2025-06-10
macOS 26 build changes

Geenz:
You may need to set LL_SKIP_REQUIRE_SYSROOT to ON if you're using Xcode 26 beta.

Takes care of the AGL deprecation (this was removed from the macOS 26 SDK - and I can't find any notes or deprecation notices for it).
Takes care of VLAs now emitting a warning on Xcode 26 creating an error in the viewer compile.

Still determining the best and fastest solution to the CEF missing Info.plist error when you attempt to run the viewer from Xcode - but produced builds will still work.  May just add a post-build fixup to create the missing versions folder for the framework for now until we can produce a new CEF build that has this fix built-in.  Will produce a separate PR for that.
 
Carbon is still (somehow) present despite AGL's removal.

### DIFF
--- a/indra/cmake/Linking.cmake
+++ b/indra/cmake/Linking.cmake
@@ -72,7 +72,6 @@ else()
   find_library(COCOA_LIBRARY Cocoa)
   find_library(IOKIT_LIBRARY IOKit)
 
-  find_library(AGL_LIBRARY AGL)
   find_library(APPKIT_LIBRARY AppKit)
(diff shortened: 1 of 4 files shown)


Geenz:
@brad-linden Yeah - I'll probably remove that in a follow up.  There's a few things we should honestly just purge here - lots of deprecated stuff that we really just don't care about anymore that honestly is just gonna be noise when we eventually need to rip some of this stuff out anyways for a Metal port of the viewer.

Issue #4148, opened 2025-05-22
GLTF Mesh Import: deformed rigged mesh


akleshchev:
### Environment

GLTF Mesh Import project branch

### Description

Some meshes appear to be deformed when skinned.

![Image](https://github.com/user-attachments/assets/96e4f162-5244-47db-8492-96eb21e964be)

### Reproduction steps

Mesh in question is private

P.S. It is also compex and has multiple, potentially separate issues (deformed overrides, rotated joints and whatever is causing deformations), it's better to find something Much simpler that reproes this issue.

AtlasLinden:
Multiple models tested and no deformations spotted. If deformations are spotted on a new model we can look into that on a new ticket. 

Passed QA.
Verified on Second Life Release 7.2.0.16356654553 (2025.05) on Win11/MacOS

A new ticket will be made for the Z offset being added in the Upload Model preview.

PR #4240, opened 2025-06-10
#4148 Skeleton Translation #2

akleshchev:
A cheap hack to detect when skeleton is in a wrong rotation and adjust it.

I'm unwiling to dump more time into it. It works on models I have access to and I can return to it once I understand what's going on and what I'm doing.

### DIFF
--- a/indra/newview/gltf/llgltfloader.cpp
+++ b/indra/newview/gltf/llgltfloader.cpp
@@ -76,6 +76,13 @@ static const glm::mat4 coord_system_rotation(
 );
 
 
+static const glm::mat4 coord_system_rotationxy(
+    0.f, 1.f, 0.f, 0.f,
+    -1.f, 0.f, 0.f, 0.f,
+    0.f, 0.f, 1.f, 0.f,
+    0.f, 0.f, 0.f, 1.f
+);
+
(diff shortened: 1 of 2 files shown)


marchcat (approved):

indra/newview/gltf/llgltfloader.cpp
```diff
@@ -1140,9 +1160,73 @@ glm::mat4 LLGLTFLoader::computeGltfToViewerSkeletonTransform(const LL::GLTF::Ski
...
+void LLGLTFLoader::checkForXYrotation(const LL::GLTF::Skin& gltf_skin)
+{
+    // HACK: figure out model's rotation from shoulders' matrix.
+    // This is wrong on many levels:
+    // Too limited (only models that have shoulders),
+    // Will not work well with things that emulate 3 hands in some manner
+    // Only supports xy 90 degree rotation
+    // Todo: figure out how to find skeleton's orientation Correctly
+    // when model is rotated at a triangle level
+    constexpr std::string right_shoulder_str = "mShoulderRight";
```
Ansariel, line 1192: GCC will probably be angry about this because GCC...

For now, it's better to use `constexpr char right_shoulder_str[] = "mShoulderRight";` instead.

Issue #4147, opened 2025-05-22
GLTF Mesh Import: Joint override support


akleshchev:
### Environment

GLTF Mesh Importer branch

### Description

Joint overrides result in deformed avatars

### Actual:

![Image](https://github.com/user-attachments/assets/1b68d67b-b654-4366-8450-230b6b0cd344) ![Image](https://github.com/user-attachments/assets/9f7e201b-00e9-4b66-9e89-fa03f72533b7)

### Expected:

![Image](https://github.com/user-attachments/assets/5af75788-fc56-4e64-947d-1238a54289cb) ![Image](https://github.com/user-attachments/assets/3cd02719-7e95-42b9-855e-3f44da3c46b8)

### Reproduction steps

I used 'bento boots' and 'angel with tail' models for thsoe screenshots, but issue reproes with almost everything that provides joint positions.

maxim-productengine:
Some info about that particular model: head position is mismatched only on model preview. After uploading it with 'Include joint positions' and attaching it to the avatar - everything looks correct.

maxim-productengine:
Misplaced head for angel model should be fixed.
Bento boots model seems to be improved as well. (maybe it can be improved further, but I'm not sure).
We still have issues with joint override for other models, so I'm not closing this ticker for now. It may be closed, if separate issues will be created.

AtlasLinden:
Failed QA.
The rigged angel model is not appearing correct on MacOS (Image incoming)

Uploading the rigged angel model appears correct on Win11 but selecting "Include join positions" is adding a Z offset to the model preview: https://gyazo.com/532201ca0b8eda6ac52f81186c63d3c1
This Z-offset is not included upon uploading and wearing in-world.

[angel_rigged.zip](https://github.com/user-attachments/files/20986217/angel_rigged.zip)

PR #4251, opened 2025-06-10
#4147 Joint override

maxim-productengine:
For the models mentioned in 4147: fixes the issue with 'detached head' for angel model and improves bento-shoes model.


akleshchev:
This is very confusing...

maxim-productengine:
> This is very confusing...
> 
> You are basically doing joint_mat = glm::mat4(1.f);

updated, so it's clearly visible

Issue #4242, opened 2025-06-10
GLTF Mesh Import: Support face splitting for faces with more than 65532 vertices


akleshchev:
Collada importer splits meshes that don't fit into viewer's limits into multiple faces.

At the moment gltf import drop prims with a warning when prim.getVertexCount() is over limit.

Task:

For parity sake make GLTF importer split extra triangles into separate faces. See 'verts.size() >= 65532' check in load_face_from_dom_triangles

aiaustin:
I think this enhancement if implemented would be very useful indeed.

The main issue I have been having with my glTF test meshes is finding (resolved now with a Blender python script that shows the highest vertex count mesh parts) and then sensibly breaking up mesh parts into < 64K vertex chunks.  This is especially problematic after using a glTF transformation tool to reduce mesh part count before upload, which is essential in many cases. At this post Blender export stage it not a problem if mesh parts are mangled up as its not the "maintained model. But the parts that get flattened, joined and welded really can be quite large and almost impossible to sensible split by the creator into 64K vertex chinks.

Also, is the comment about just ignoring prims over the limit actually correct?  I think the importer actually gives the error about the mesh having vertices over the "16-bit limit".

Just for curiosity, Geenz mentioned a vertex limit of "65,363" that triggered the 16-bit limit error. I have been wondering where that figure came from.  So is that the limit or is the limit 65,532 (i.e. three less than the 16-bit limit or actually 16,535?

AtlasLinden:
The same behaviour is observed on MacOS.

The following explains said behaviour for the models under 65535 vertices.
> Condition is 65535, butonce condition applies it aims at 65532.

In regards to the 65535 model, this becomes 131071 with 3 faces. The same is also seen when importing the same model as a collada file.
If the viewer aims at 65532 after _revamping_ the model, then would 2 of the faces in the uploaded model should both have that many vertices?
131071 = 65532*2 + 7 extra vertices on the third face. This can be observed after texturing and zooming in on the 65535 mesh.

![Image](https://github.com/user-attachments/assets/dd93399c-4bc3-4864-98ff-f60dfa0368b9)

Given the condition this passes QA
Verified on Second Life Project 7.1.14.15976006598 (GLTF Mesh Import) on Win11/MacOS

Note that separate issues will made for the 65535 model being raised to a mighty 131071 vertices and the disappearing models mentioned above.

akleshchev:
[...]

PR #4270, opened 2025-06-17
#4242 Support splitting of gltf faces that are over 16bit limit

akleshchev:
If a face has too many vertices, split it into multiple faces.

P.S. Only did synthetic testing, like limiting maximum to 1000 vertices and uplaoding a 3K vertices model.


aiaustin:
Looking forward to trying this. I have various test meshes where one part is around 100K vertices (i.e. 2 parts needed), but another that has one complex part welded together as the original parts share one material by glb.babylonpress.com that is 136K vertices which would have to be split into three parts to stay under the 64K limit. 

Just so I know what to expect in testing, will the method you are using split into as may faces as needed or will it be limited to one split into 2 parts?

maxim-productengine (approved):

akleshchev:
>will the method you are using split into as may faces as needed or will it be limited to one split into 2 parts?

Multiple.

PR #4276, opened 2025-06-19
Fix loading name of banned residents in group bulk ban panel

Ansariel:
Requesting more than one avatar name in the group bulk ban panel causes avatars not getting resolved properly, because each time a name request is issued the callback for the previous request is canceled. Instead of using a single signal connection for the callbacks, use a connection for each request.


marchcat (commented):
Thank you!

marchcat (approved):

Issue #4257, opened 2025-06-11
GLTF Mesh Import: Model Name Initial Value


aiaustin:
### Environment

Second Life Project Gltf Mesh Import 7.1.14.15583209293 (64bit)
Release Notes

You are at 138.0, 102.6, 23.0 in Rumpus Room 2048 located at simhost-030b947afda1f10ad.aditi
SLURL: secondlife://Aditi/secondlife/Rumpus%20Room%202048/138/103/23
(global coordinates 36,490.0, 14,182.6, 23.0)
Second Life Server 2025-04-24.14648572313
Release Notes

CPU: Intel(R) Xeon(R) CPU E5-1620 v3 @ 3.50GHz (3491.91 MHz)
Memory: 32689 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.5854)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.7640
OpenGL Version: 4.6.0 NVIDIA 576.40

Window size: 1300x882
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 512m
Bandwidth: 10000kbit/s
LOD factor: 2
Render quality: 6
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (42.4% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.5942f08

Packets Lost: 1/45,706 (0.0%)
June 11 2025 09:50:15

### Description

When using the glTF Model Importer the Model Name is initially filled in with a mesh name within the model (presumably something like the first mesh part found).

It would be better if the Filename was used as the initial value of the mesh model, which is also already the root name used for materials and textures associated with the model if those are uploaded.

The Model Name can be changed and the uploaded model in inventory does have that changed model name, which is expected.

### Reproduction steps

1. Upload any mullt-part glTF model. E.g  Modelname-Test.glb
[...]

PR #4282, opened 2025-06-23
#4257 Use filename as model initial name

marchcat:


indra/newview/gltf/llgltfloader.cpp
```diff
@@ -465,6 +465,9 @@ void LLGLTFLoader::computeCombinedNodeTransform(const LL::GLTF::Asset& asset, S3
 
 bool LLGLTFLoader::populateModelFromMesh(LLModel* pModel, const LL::GLTF::Mesh& mesh, const LL::GLTF::Node& nodeno, material_map& mats, S32 instance_count)
 {
+    // Set the requested label for the floater display and uploading
+    pModel->mRequestedLabel = gDirUtilp->getBaseFileName(mFilename, true);
```
akleshchev, line 469: I think you need to sanitize that. Otherwise upload might fail or there will be a mismatch between uploaded model and importer.

P.S. We are using models' name because when there is only one mesh in file generaly we want resulting object to be named after the model.

akleshchev (approved):

indra/newview/gltf/llgltfloader.cpp
marchcat: It works with no issues, just like a shortcut to renaming the object right after uploading. The internal structure of the object remains unchanged.

Issue #4184, opened 2025-05-28
BugSplat Crash: SecondLifeViewer!LLConversationItemSession::clearAndDeparentModels(372)


sl-service-account:
## BugSplat Crash Group 98721
[View Report Group](https://app.bugsplat.com/v2/group?stackKeyId=98721&database=SecondLife_Viewer_2018)
[View Report Details](https://app.bugsplat.com/v2/crash?id=1633416&database=SecondLife_Viewer_2018)
[View Report Attachments](https://app.bugsplat.com/v2/crash?id=1633416&database=SecondLife_Viewer_2018#attachments)

**Application:** Second Life Release
**Version:** 7.1.14.15192634334
**Error Code:** c0000005 Access violation

**Callstack:**
[...]

PR #4295, opened 2025-06-26
#4184 clearAndDeparentModels crash

akleshchev:
To resolve dual ownership switched children to LLPointer.

I expect it to cause issues and crashes early on, but no apparent way around it.


maxim-productengine (approved):

marchcat (approved):

akleshchev:
Added a bunch of macOS build fixes, it was complaining about overrides.

PR #4310, opened 2025-06-28
Revert "Fix LLCharacter base class constness."

Ansariel:
And instead re-apply all my improvements from 10a324a1034c177b95545ac7ffaa6aa6abed65ff somebody managed to lose somehow.

How did this even got mangled up? My original commit seems to be entirely missing in this branch even though it was applied to `develop` and that branch apparently got merged into `gtlf-mesh-import`???


Geenz (approved):
Looks good - thank you for catching this.  I was on a tight deadline yesterday, so this got missed in the merge.  Merging it in - thank you!

Issue #4315, opened 2025-07-01
Invalid GLTF file crashes viewer


AtlasLinden:
### Environment

Second Life Project 7.1.14.15976006598 (GLTF Mesh Import)

### Description

Selecting the the following invalid GLTF file for upload crashes the viewer.
[Test_001_Invalid.zip](https://github.com/user-attachments/files/21000481/Test_001_Invalid.zip)

When attempting to upload this file as a material, the following prompt is provided:
![Image](https://github.com/user-attachments/assets/b5905fd7-be51-4128-9525-aba03495cd1e)

This file does not crash Blender when attempting to import. Instead the following error message is given:
`Bad glTF: json error: Expecting ',' delimiter: line 45 column 25 (char 1037)`

Last lines of SecondLife.log before crashing:
```
WARNING # llui/llview.h(722) LLView::getChild : Making dummy class LLView named "preview_panel" in floater
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Simplify Method - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Retain% - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Detail Scale - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Combine Quality - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Method - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Decompose Quality - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Close Holes (Slow) - 
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Simplify Coplanar - 
WARNING # llui/llview.h(722) LLView::getChild : Making dummy class LLCheckBoxCtrl named "Simplify Coplanar" in Model Preview
INFO # newview/llfloatermodelpreview.cpp(1185) LLFloaterModelPreview::initDecompControls : Cosine% - 
INFO # llui/llfloater.cpp(658) LLFloater::openFloater : Opening floater Model Preview full path: /main_view/menu_stack/world_panel/Floater View/Model Preview
[...]

PR #4321, opened 2025-07-01
#4315 Crash in GLTF uploader

akleshchev:
Properly handle importer's crashes
Update GLTF Mesh Import's version to 7.2


indra/newview/VIEWER_VERSION.txt
```diff
@@ -1 +1 @@
-7.1.14
+7.2.0
```
marchcat, line 1: Finally :)

marchcat (approved):

Issue #4323, opened 2025-07-01
Rezzing meshes with 65533 or 65534 vertices appear transparent in-world


AtlasLinden:
### Environment

Second Life Project Gltf Mesh Import 7.1.14.15976006598

### Description

As mentioned [here](https://github.com/secondlife/viewer/issues/4242#:~:text=65533%20and%2065534%20vertex%20models%20are%20separately%20uploaded%20and%20each%20show%20the%20same%20amount%20of%20vertices%20in%20the%20importer%20with%20the%20log%20LOD%203%20Faces%3A%201%0AThe%20uploaded%20mesh%20appears%20transparent%20when%20rezzed%20in%2Dworld%20unlike%20the%20previous%202%20models.%20The%20mesh%20can%20be%20observed%20in%20the%20importer%20preview%20panel.), uploading a GLB model with 65533 or 65534 vertices appears transparent when rezzed in-world.
It's possible that this only occurs with these single models. 
https://gyazo.com/d365e6aa38802f409d5e7d7fbda57dd0

### Reproduction steps

Upload and rez the following models: [65kvertices.zip](https://github.com/user-attachments/files/21008334/65kvertices.zip)

AtlasLinden:
It seems to be only LOD3 that is appearing transparent.
Note that the same issue does not repeat with collada as the importer vertices limits are different

akleshchev:
But does it repro on collada with 65530 models?

AtlasLinden:
These 65530 models as dae are treated different by the importer. 
The 65533 dae model for example is bumped up to 133913 vertices when imported to the viewer.

After playing around with the amount of vertices on the model, exporting and importing into the viewer I have stumbled on a dae model that also triggers this issue. 
65021vertices.dae is bumped up to 65532 vertices and also experiences transparency on its highest LOD.

[65021vertices.zip](https://github.com/user-attachments/files/21025637/65021vertices.zip)

PR #4340, opened 2025-07-04
#4323 fix for transparent meshes

maxim-productengine:
Force the models with 65533+ vertices into the splitting.


marchcat (approved):

akleshchev (approved):
Are you sure it doesn't repro on a 65530 or 65500 model?

Issue #4319, opened 2025-07-01
Windows crash not appearing in bugsplat


AtlasLinden:
### Environment

Second Life Project Gltf Mesh Import 7.1.14.15976006598

### Description

As mentioned in https://github.com/secondlife/viewer/issues/4315, this crash does not appear in bugsplat.

### Reproduction steps

Reproduce crash in https://github.com/secondlife/viewer/issues/4315
Or use the new and upcoming "force crash" menu item being added to a future viewer near you! Develop->Force an Error

PR #4344, opened 2025-07-07
#4319 Thread crashes not being reported to bugsplat

akleshchev:
Like with coroutines, start processing SEH exceptions.

Works as expected locally (triggers reportCrashToBugsplat), but going to wait for PR's build to see it work with bugsplat.


indra/llcommon/llthread.cpp
```diff
@@ -106,6 +108,31 @@ namespace
...
+            // C++ exception, go on
+            return EXCEPTION_CONTINUE_SEARCH;
+        }
+        else
+        {
+            // handle it, convert to std::exception
+            return EXCEPTION_EXECUTE_HANDLER;
+        }
+
+        return EXCEPTION_CONTINUE_SEARCH;
```
marchcat, line 133: This `return` looks unreachable.

akleshchev: Yeah, probably should remove it.

Issue #4290, opened 2025-06-25
GLTF Mesh Import: Unresponsive viewer when attempting to upload certain model


AtlasLinden:
### Environment

Second Life Project Gltf Mesh Import 7.1.14.15830455952 Windows 11
Does not appear to reproduce on M1

### Description

Attempting to upload the [blender exported mesh](https://github.com/user-attachments/files/20463251/2025-Supercar-V27-Parts-Test-via-Tools.zip) from https://github.com/secondlife/viewer/issues/4170 causes the viewer to become unresponsive for approximately 5 minutes as it attempts to process the model. If other actions are performed during this time, the viewer remains frozen.

Sometimes the model can be completely processed but yet to be successful at calculating weights and fees in the Upload Model floater. This is recorded in the following log before the user was logged out with a "This region may be experiencing trouble" error.
[SecondLife.log](https://github.com/user-attachments/files/20910192/SecondLife.log)

### Reproduction steps

Attempt to upload the model provided above.

AtlasLinden:
Viewer no longer becomes unresponsive upon selecting the given model.
Calculating weights & fees does take some time (and will also hit a MAV_BLOCK_MISSING error if physics isn't selected) but it does eventually complete. The same goes for actually uploading the model. 
Viewer also remains responsive during these times. 

Passed QA.
Verified on Second Life Release 7.2.0.16356654553 (2025.05) on Win11/MacOS

Bonus! Clip of very slow rezzing of said model: https://gyazo.com/9d8888681159d58fb99444d1635f476e

PR #4345, opened 2025-07-07
#4290 Unresponsive viewer when uploading models

akleshchev:
Fix severe log spam, model with hundreds of meshes was taking minutes to load.


marchcat:
Thank you!

maxim-productengine (approved):

Issue #4314, opened 2025-07-01
GLTF Mesh Import: Unable to use the same GLB model for physics


AtlasLinden:
### Environment

Second Life Project 7.1.14.15976006598 (GLTF Mesh Import)

### Description

A GLB used for Physics cannot be applied if it matches LOD_HIGH in the mesh importer.
This is not the case when using a DAE file for physics.

### Reproduction steps

Use the cube files mentioned in #4157 as these all have an identical shape: https://github.com/user-attachments/files/20431144/Cubes.zip

1. Select to upload a cube GLB file
2. For the physics model, select From file in the drop down 
3. Click Browse and select the same GLB file
4. Observe no Physics shown in preview. Log will also mention `No physics model specified for Cube`
5. Repeat the above but initially select a cube DAE file (which will use a GLB physics file)

PR #4348, opened 2025-07-08
#4314 Follow-up: duplicate logic for mSubmodelID

marchcat:
This is needed to avoid conflicts if the model already has `mesha` and similarly named objects alongside `mesh`.


akleshchev (changes requested):
I think instance name should match the one the one assigned to the model at string 243.

Suggestion: move 'instance' logic to string 243 then do
`next->mLabel = model_name + (char)((int)'a' + next->mSubmodelID) + "_copy_" + std::to_string(duplicate_count);`
Without the lod_suffix[mLod]. Then 
```
std::string instance_name = model->mLabel
model->mLabel + lod_suffix[mLod];
mScene[transformation].push_back(LLModelInstance(model, instance_name, transformation, materials));
```

indra/newview/gltf/llgltfloader.cpp
```diff
@@ -290,12 +310,9 @@ void LLGLTFLoader::addModelToScene(
...
             }
         }
-        std::string base_name = model_name;
-        if (model->mSubmodelID > 0)
-        {
-            base_name += (char)((int)'a' + model->mSubmodelID);
-        }
-        mScene[transformation].push_back(LLModelInstance(model, base_name, transformation, materials));
+        // Compose instance name from model label and add lod suffix
+        std::string instance_name = model->mLabel + lod_suffix[mLod];
```
akleshchev, line 314: suffix should not get into mScene. It should only be present in model->mLabel for model matching to work.
std::string instance_name = model->mLabel;
model->mLabel += lod_suffix[mLod];
mScene[transformation].push_back(LLModelInstance(model, instance_name, transformation, materials));

akleshchev (approved):

akleshchev:
Actually this is going to bug out because original models already have a suffix...

indra/newview/gltf/llgltfloader.cpp
```diff
@@ -240,7 +241,26 @@ void LLGLTFLoader::addModelToScene(
             LLModel* next = new LLModel(volume_params, 0.f);
             next->ClearFacesAndMaterials();
             next->mSubmodelID = ++submodelID;
-            next->mLabel = model_name + (char)((int)'a' + next->mSubmodelID) + lod_suffix[mLod];
+
+            std::string instance_name = model_name;
+            if (next->mSubmodelID > 0)
```
akleshchev, line 246: In this case mSubmodelID is always above 0

PR #4352, opened 2025-07-09
#4314 Fix model suffixes #2

akleshchev:
Follow up on 'duplicate logic for mSubmodelID'


marchcat:
The file still has some redundant code that we'll have to clean up sooner or later. It should also help avoid situations like this in the future.

Issue #4365, opened 2025-07-13
Emoji's in the recent emoji bar have incorrect hitboxes at certain window scales


Hecklezz:
### Environment

Second Life Release 7.1.15.15596336374 (64bit)
<Details>
<Summary>Rest of environment-</Summary>
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600.01 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.6093)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.7652
OpenGL Version: 4.6.0 NVIDIA 576.52

Window size: 1512x940
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 512m
Bandwidth: 3000kbit/s
LOD factor: 4
Render quality: 5
Texture memory: 8192MB
Disk cache: Max size 1638.4 MB (99.2% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.14.0.202408091639
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.21
Voice Server Version:  

July 13 2025 07:06:43
</Details>

### Description

Depending on how the IM floater's width is scaled, the emojis in the recent emoji list have their hitboxes incorrect, and mousing over them or clicking on them can click the emojis next to the one you meant. This appears to be caused by the recent emojis trying to centre themselves instead of be left align, but the hitboxes are always left aligned.

Video showing the bug-

https://github.com/user-attachments/assets/a9f94a0d-7b15-4469-8a3a-2323c513e487

### Reproduction steps

1. Use emoji's in chat so they appear in the recently used emojis list
2. Shrink/expand the chat floater and observe the positions of the emojis move
3. While the emojis are offset to the right of where they should be, mouse over them and observe the hitboxes not lining up.

AtlasLinden:
Passed QA.
Verified on Second Life Release 7.2.0.16841984053 (2025.06) on Win11/MacOS

PR #4366, opened 2025-07-13
[#4365] Fix emoji hitboxes in the emoji history list

Hecklezz:
## Description

This PR makes a change to the emoji history at the bottom of the IM floater by ensuring they are always left aligned which fixes their hitboxes when mousing over and clicking them. This is done by ensuring their render rectangle has a left value of 0, and the right value gets offset by how much the left got offset by.

<Details>
<Summary>Video showing new behaviour-</Summary>

https://github.com/user-attachments/assets/cbfe6474-0664-4dd5-9fe4-2700335b99ca

</Details>

## Related Issues

Issue: https://github.com/secondlife/viewer/issues/4365

---

## Additional Notes

The reason the hitboxes don't match where they are rendered is probably worth investigating too, however in my opinion, having them left aligned is more visually ideal regardless. That way an arbitrary gap depending on the floater size doesn't exist before the emoji history, and instead as the window grows in width, it simply adds another emoji if it can fit instead of trying to constantly centre them all.


marchcat (approved):
This looks good. Thank you!

Issue #4357, opened 2025-07-09
Crash: LLMeshRepoThread::lodReceived


sl-service-account:
## BugSplat Crash Group 454
[View Report Group](https://app.bugsplat.com/v2/group?stackKeyId=454&database=SecondLife_Viewer_2018)
[View Report Details](https://app.bugsplat.com/v2/crash?id=1644767&database=SecondLife_Viewer_2018)
[View Report Attachments](https://app.bugsplat.com/v2/crash?id=1644767&database=SecondLife_Viewer_2018#attachments)

**Application:** Second Life Release
**Version:** 7.1.15.15596336374
**Error Code:** c0000005 Access violation

**Callstack:**
[...]

PR #4398, opened 2025-07-21
Fix a crash bug and bogus calculations in LLMeshRepoThread::lodReceived()

vldevel:
## Description

When trying to update the rigging info for a newly received mesh LOD, a wrong usage of LLVolume::getNumFaces() is done to get the number of volume faces, causing the code to iterate over the number of faces in the underlying LLProfile instead. LLVolume::getNumVolumeFaces() must be used here.

This fixes a crash bug seen with low LODs in some meshes (when the number of mesh faces is smaller than the number of faces in the LLProfile), and also properly updates the rigging info for all mesh faces, as it should, when the mesh got more faces than the LLProfile.

## Related Issues

This may relate to crashes seen in LLMeshRepoThread::lodReceived()...

Issue Link: #4357


marchcat:
Thank you!

akleshchev:
https://github.com/secondlife/viewer/issues/4357

PR #4404, opened 2025-07-21
Properly warn when scaling textures down instead of being entirely silent.

akleshchev:
1. Notification for material upload through floater (bulk upload not covered)
2. Uploader Log for model upload

Needs a way to be localizable, but that's a general problem for addStringToLog.


maxim-productengine (approved):

maxim-productengine:
> Needs a way to be localizable

We may place these log text to strings.xml

akleshchev:
Added localization support for this specific log message, not for general output.

PR #4438, opened 2025-07-25
Make inventory creation from cache faster

akleshchev:
Switch from slow notation parser to faster binary parser.
Still has a lot of space for improvements, but should be 2.5 times faster.
Targeted at 2025.06 due to that branch having related changes.


marchcat (approved):

bennettgoble:
Sounds like a nice performance change. Thanks. Will this improve startup speed or another user-facing operation?

maxim-productengine (approved):

akleshchev:
> Sounds like a nice performance change. Thanks. Will this improve startup speed or another user-facing operation?

Loging in speed. Inventory cache gets loaded after viewer gets agent id from login response.

Issue #4388, opened 2025-07-18
Automatically Resize Large Images Before Upload


bennettgoble:
As a Second Life resident, I want the viewer to automatically downscale images larger than 2048×2048,
So that I can easily upload high-resolution images (like 4K or 5K) without needing to preprocess them manually.
Description

Currently, the viewer displays an error when attempting to upload images larger than 2048×2048 pixels. However, Second Life only supports a maximum texture resolution of 1024×1024. This creates unnecessary friction for users who commonly work with high-resolution images, particularly on modern displays.

## Acceptance Criteria

- Images larger than 2048×2048 are automatically downscaled by the viewer before upload.
- The viewer prompts the user or displays a notification when an image is being downscaled.
  - Possibly a note on the image upload preview 
- Downscaled images must respect SL’s maximum image resolution of 2048×2048.
- The upload process remains consistent with current pricing and compression settings.
- Edge cases like non-square images and images with alpha channels are handled appropriately.
- Both bulk and single-image upload paths updated
<br/>
<hr/>

## Screenshots

<img width="288" height="496" alt="Image" src="https://github.com/user-attachments/assets/8456765d-1739-477b-b545-5844cc4cb938" />

_The current viewer image uploader is unable to handle images larger than 2048x2048._

## Reward

Please follow our [reward guidelines](https://github.com/secondlife/.github/blob/main/REWARDS.md) if you would like to be compensated for completing this task.

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4388</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @bennettgoble</code> (replace <code>20</code> with the amount, and <code>@bennettgoble</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

bennettgoble:
Talked briefly on the opensource channel in Discord with @FelixWolf about this. Granting a **two week exclusive period** to finalize the work in https://github.com/secondlife/viewer/pull/4444. Looking forward to getting these changes integrated. :)

[...]

PR #4444, opened 2025-07-26
2k image resize

FelixWolf:
## Description

This resizes the images (both singular and bulk) to be under 2048px by 2048px, and alert the resident via dialog if the image is being resized.
This is done in-memory, and preserves the original.

New notifications for translation:
* `ImageUploadResized` - AlertModal, Ignorable

Video: https://www.youtube.com/watch?v=oyN9Etoptz4

## Related Issues

Issue Link: #4388

## Test cases:
- [x] Images smaller than 2048x2048 do not trigger dialog.
- [x] Images taller than 2048px trigger the dialog.
- [x] Images wider than 2048px trigger the dialog.
- [x] Images maintain aspect ratio when resized.
- [x] Bulk upload resizes images.
- [x] Ensure bulk upload cost is correct.

# Additional information
The `allow_2k` check had to be moved to a different location, specifically to line `591` in `llviwermenufile.cpp`. It preserves the same functionality as before, by clamping to 1024 if the resident does not want 2k uploads.


marchcat (approved):
Thank you for the excellent patch!

Geenz (approved):

bennettgoble:
Nice and thorough PR, thanks @FelixWolf. Merging now. If we find any bugs within the scope of the original write-up we'll let you know. Works well when I was testing it! :)

Issue #4448, opened 2025-07-27
Remove First Login Screen


bennettgoble:
### Description

The current Second Life viewer presents users with two distinct login screen variants: one that appears during a clean installation (the "First Login" screen) and a slightly different one that appears on subsequent logins. This dual approach is unnecessary and can be confusing for users, particularly those who are new to the client or are reinstalling it. It also comprises additional code and complexity.

## Solution

- Remove First Login specific UI: Eliminate any UI elements or prompts that are exclusive to the "First Login" screen.
- Clean up/remove code specific to first login screen

## Screenshots

<img width="1469" height="918" alt="Image" src="https://github.com/user-attachments/assets/2226b242-5a10-4a48-9098-be8ff56d0e29" />

_First login screen_

<img width="1470" height="918" alt="Image" src="https://github.com/user-attachments/assets/20135aaf-8e9d-4743-86c2-304674859d9a" />

_Normal login screen_

### Reward

Please follow our [reward guidelines](https://github.com/secondlife/.github/blob/main/REWARDS.md) if you would like to be compensated for completing this task.

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4448</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @bennettgoble</code> (replace <code>20</code> with the amount, and <code>@bennettgoble</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

FelixWolf:
Working on it at https://github.com/FelixWolf/viewer/tree/first-login-gone

Question regarding the code though, at https://github.com/secondlife/viewer/blob/develop/indra/newview/llpanellogin.cpp#L589, it is mentioned that a check is done because it is necessary for panel_login_first.xml, but it isn't clear what that is needed for.
I believe it is a left over comment from when there was two fields for the username, one for the first name, and one for the last name, but now the login_panel.xml uses one field for both. In which case I will remove the comment as redundant, but I just want to verify before I do that.

bennettgoble:
[...]

PR #4451, opened 2025-07-28
Remove the first login screen

FelixWolf:
## Description
Removes the first login screen and it's resources.
This still has code for the setting `FirstLoginThisInstall`, as it is used to inform the login server of first installs.

Test case:
1. Clean install or set `FirstLoginThisInstall` to `False`.
2. Launch the viewer, make note of the login screen.
3. Log in and log out, ensure the viewer closes gracefully.
4. Launch the viewer, ensure that that the login screen is the same as before.

Test case fails if the login screen is different.

## Related Issues
Issue Link: #4448

## Additional Notes
viewer_manifest.py had to be edited to remove the search for *.jpg, because there are no more jpgs in the texture folder now.

Additionally, I previously brought up questions about the credential loader and the "agent" vs "account" type. It seems "agent" is what we use today, where as "account" is a older field type in the bin_conf.dat. In theory, "account" can just be removed, as it doesn't appear to be used anymore, but I left it in because it is irrelevant to this PR, and would like approval before removing that.


bennettgoble:
Thanks, @FelixWolf. I kicked off the builds and will have this reviewed tomorrow.

marchcat (approved):

Issue #4449, opened 2025-07-27
Default Media Auto-Play to "Ask"


bennettgoble:
### Description

Currently, when new and existing users log into Second Life, media (such as music or sounds from in-world objects) can begin playing automatically. This often leads to a confusing and sometimes startling experience, as users are left wondering "Where is it coming from?" This unexpected audio can detract from the initial immersion and create a negative first impression, especially for new users who are still learning the environment.

## Solution

Change Default Setting: Modify the default value for the "Auto-play media" setting in the "Sound & Media" section of the Preferences to "Ask."

## Screenshots

<img width="650" height="533" alt="Image" src="https://github.com/user-attachments/assets/042d759d-0fa2-4fed-a227-5983b7f03791" />
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4449</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @bennettgoble</code> (replace <code>20</code> with the amount, and <code>@bennettgoble</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

DarlCat:
I want to comment that I believe that the setting for HUD media autoplay which is a different setting that I added in a recent PR should remain enabled because it makes things like body huds more responsive and enjoyable to use, and users are typically far more aware of their own attachments than they are the world around them which is usually outside of their control.

bennettgoble:
@DarlCat Yep, I agree. This issue is primarily for parcel media: music streams primarily.

kylelinden:
I'm hesitant to flip the default to "Ask" as it could impede or confuse the new user experience.  I like the 'ask if not trusted' option instead. This will need specific design first.

bennettgoble:
Closing while reviewing.

PR #4453, opened 2025-07-28
Change parcel media autoplay to ask by default

FelixWolf:
## Description
Simple XML change to make autoplay default to ask instead of always

## Related Issues

Issue Link: #4449

## Additional Notes
I assume we don't want to change what a resident's setting currently is, so there is no logic to change what people currently have. Fresh installs / new residents only!


marchcat (approved):

bennettgoble:
@FelixWolf thanks for the small PR. I need to confer with our product team on whether we want to implement an allow-list system first. I'll close this for now while we figure it out.

Issue #4418, opened 2025-07-23
Discord Rich Presence


bennettgoble:
<img width="2040" height="600" alt="Image" src="https://github.com/user-attachments/assets/8e840d7e-c001-4c3f-9cbc-b0836f914dbb" />

_Rich Presence Examples from [Discord's Overview](https://discord.com/developers/docs/rich-presence/overview)_

## Problem

Many Second Life users engage in shared activities and communities that overlap with Discord, such as live events, social hangouts, and collaborative building. Discord Rich Presence provides a way to surface Second Life session activity directly in users' Discord profiles, improving discoverability, engagement, and community presence outside of the viewer.

Rich Presence is **already implemented in several third-party viewers** (e.g., Firestorm), and users have come to expect this feature. Without it, the official viewer appears outdated and disconnected from broader social gaming trends.

## Details

Add Discord Rich Presence in the official Second Life Viewer, allowing users to optionally share activity information through Discord while using the viewer.

The feature will be accessible via a new "Discord" tab in the **Privacy** section of the Preferences floater.

### UI Mock

The preferences UI will include:

- **Enable Discord Integration** (default: on)
- **Show Location** (default: off)

<img width="1019" height="756" alt="Image" src="https://github.com/user-attachments/assets/35a6667a-8cc8-4664-86de-85d3318c0a7b" />

### Technical Approach

- Integrate the Discord Rich Presence API into the viewer.
- Follow [Discord Rich Presence Best Practices](https://discord.com/developers/docs/rich-presence/best-practices).
- Support presence data such as:
  - Current region (if location sharing is enabled)
  - Time spent in session
- (Optional) Investigate use of [Discord's Social SDK](https://discord.com/developers/docs/rich-presence/overview) as opposed to deprecated SDKs --for future extensibility.

### Requirements

- [ ] Bottle SDK as an autobuild package
- [ ] Add Discord tab to Privacy section of Preferences floater.
  - [ ] Include checkbox to enable/disable Discord presence (default: enabled).
  - [ ] Include checkbox to enable/disable sharing location with presence (default: disabled).
- [ ] Implement Rich Presence support following Discord's API and best practices.
- [ ] Log presence updates at viewer startup, region change.
- [ ] Use Social SDK if possible to future-proof the integration.

### Goals and KPIs

- **Adoption rate**: % of users with Discord Presence active
- **Engagement**: Track correlation between Discord visibility and return sessions (if measurable)

### Discord Developer Settings

[...]

PR #4457, opened 2025-07-28
Rich Presence support using Discord Social SDK

erikkundiman:
## Description

- New tab in privacy panel called "Discord", with three buttons which are all unchecked by default.
- The first button is for enabling the feature, and when checked, it will redirect the user to the Discord app (or the browser if the app isn't open) for authorisation, and an access token will be retrieved and the viewer will connect to Discord. When the button gets unchecked, the access token will be revoked and the viewer will disconnect from Discord.
- The 2nd & 3rd buttons are only enabled when the 1st button is checked, and they are for hiding/showing the avatar names (display name and/or username) and location (region and coordinates) respectively.
- These buttons correspond to new settings with the keys `EnableDiscord`, `ShowDiscordActivityDetails` and `ShowDiscordActivityState`, respectively.
- The access token is saved the way passwords are saved (though without a username), so we can have some persistence without having to implement an OAuth2 backend server.
- The "Playing" field automatically shows the viewer's name.
- The "Details" field is used for showing the username (and display name if different).
- The "State" field is used for showing the region name and coordinates.
- The "Party" field is used for showing the number of people including the user within the chat radius and the number of people including the user within the near range.
- The time elapsed is not reset on region change.
- Time elapses right away after the viewer launches even before login.
- Location gets updated on region change, or by unchecking and re-checking the "Show location" button.

## Related Issues

Issue Link: [Discord Rich Presence #4418](https://github.com/secondlife/viewer/issues/4418)

---

## Checklist

I have tested:

- [x] checking and unchecking the preferences buttons, before and after login;
- [x] changing regions by teleporting and crossing;
- [x] the persistence of the preferences across logins.

---

## Additional Notes

- Download DiscordSocialSdk-1.4.9649.zip from [https://discord.com/developers/applications/1394782217405862001/social-sdk/downloads](https://discord.com/developers/applications/1394782217405862001/social-sdk/downloads)
- The Discord app needs to be set to be a public client in the OAuth2 tab.
Screenshot 2025-07-29 at 00 35 18

- Linden Lab still needs to add the Autobuild installable entry with a key named `discord`.
- Linden Lab still needs to add the discord_partner_sdk libraries to the viewer manifest.
- Add `-DUSE_DISCORD:BOOL=ON` to the CMake configuration line to enable the feature.

### DIFF
--- a/indra/newview/llpanelpeople.cpp
+++ b/indra/newview/llpanelpeople.cpp
@@ -844,7 +844,8 @@ void LLPanelPeople::updateNearbyList()
     LLWorld::getInstance()->getAvatars(&mNearbyList->getIDs(), &positions, gAgent.getPositionGlobal(), gSavedSettings.getF32("NearMeRange"));
     mNearbyList->setDirty();
 #ifdef LL_DISCORD
-    LLAppViewer::updateDiscordPartyMaxSize(mNearbyList->getIDs().size());
+    if (gSavedSettings.getBOOL("EnableDiscord"))
+        LLAppViewer::updateDiscordPartyMaxSize(mNearbyList->getIDs().size());
 #endif
 
     DISTANCE_COMPARATOR.updateAvatarsPositions(positions, mNearbyList->getIDs());
--- a/indra/newview/llspeakers.cpp
+++ b/indra/newview/llspeakers.cpp
@@ -1027,7 +1027,8 @@ void LLLocalSpeakerMgr::updateSpeakerList()
     std::vector<LLVector3d> positions;
     LLWorld::getInstance()->getAvatars(&avatar_ids, &positions, gAgent.getPositionGlobal(), CHAT_NORMAL_RADIUS);
 #ifdef LL_DISCORD
-    LLAppViewer::updateDiscordPartyCurrentSize(avatar_ids.size());
+    if (gSavedSettings.getBOOL("EnableDiscord"))
+        LLAppViewer::updateDiscordPartyCurrentSize(avatar_ids.size());
 #endif
     for(U32 i=0; i<avatar_ids.size(); i++)
     {
--- a/indra/newview/llviewermessage.cpp
+++ b/indra/newview/llviewermessage.cpp
@@ -3053,7 +3053,8 @@ void process_agent_movement_complete(LLMessageSystem* msg, void**)
     }
 
 #ifdef LL_DISCORD
-    LLAppViewer::updateDiscordActivity();
+    if (gSavedSettings.getBOOL("EnableDiscord"))
+        LLAppViewer::updateDiscordActivity();
 #endif
 
     if ( LLTracker::isTracking(NULL) )


indra/newview/llappviewer.cpp
```diff
@@ -5862,3 +5873,129 @@ void LLAppViewer::metricsSend(bool enable_reporting)
...
+    if (gSavedSettings.getBOOL("EnableDiscord"))
+    {
+        gDiscordClient->UpdateToken(discordpp::AuthorizationTokenType::Bearer, gSecAPIHandler->loadCredential("Discord")->getAuthenticator()["token"].asString(), [](discordpp::ClientResult result) {
+            if (result.Successful())
+                gDiscordClient->Connect();
+        });
+    }
+}
+
+void LLAppViewer::handleDiscordSocial(const LLSD& value)
```
akleshchev, line 5897: This appears to be called only on change of debug settings. No tracking if 'handled' or not, so probably going to bug out if some external factor gets involved (like user moving settings from a different PC, it will just start 'enabled').
Also this does not take into account possiblity of viewer losing creds. loadCredential("Discord")->getAuthenticator() likely is just going to crash.

erikkundiman: Thank you again.. So, for this, I should assign a variable to contain `gSecAPIHandler->loadCredential("Discord")` first, and check whether is null or not before I try to get any authenticator out of it, right? So, something like lines 590 and 676 in indra/newview/llpanellogin.cpp..

indra/newview/llappviewer.cpp
```diff
@@ -5862,3 +5873,129 @@ void LLAppViewer::metricsSend(bool enable_reporting)
...
+            pos_x -= pos_x % 2;
+            pos_y -= pos_y % 2;
+        }
+        auto location = llformat("%s (%d, %d, %d)", gAgent.getRegion()->getName().c_str(), pos_x, pos_y, pos_z);
+        activity.SetState(location);
+
+        auto world = LLWorld::getInstance();
+        uuid_vec_t chat_radius_uuids, near_me_uuids;
+        auto position = gAgent.getPositionGlobal();
+        world->getAvatars(&chat_radius_uuids, NULL, position, CHAT_NORMAL_RADIUS);
```
akleshchev, line 5989: getAvatars is likely to get pricey in crowds and being called on each frame will affect performance.

indra/newview/llappviewer.cpp
akleshchev: Yes. As an example user can upgrade their motherboard=>creds will get wiped.
Code should be a bit more robust to handle such unexpected factors on both init and cleanup.

indra/newview/llappviewer.cpp
```diff
@@ -1319,6 +1326,10 @@ bool LLAppViewer::frame()
 
 bool LLAppViewer::doFrame()
 {
+#ifdef LL_DISCORD
+    discordpp::RunCallbacks();
```
akleshchev, line 1330: Suggestion:
```
{
    LL_PROFILE_ZONE_NAMED("discord_callbacks");
    discordpp::RunCallbacks();
}
```
So that it will be visible in Tracy

indra/newview/llappviewer.cpp
akleshchev: Sorry, i didn't notice that it's in the idle startup loop, not the general one. STATE_PRECACHE is in the  idle_startup loop. Still important, but not generaly precache doesn't care about FPS much.

But the question becomes: shouldn't it update when extra agents rez/move closer/move away?

indra/newview/llappviewer.cpp
```diff
@@ -5862,3 +5873,129 @@ void LLAppViewer::metricsSend(bool enable_reporting)
...
+                gDiscordClient->Disconnect();
+            auto cred = new LLCredential("Discord");
+            gSecAPIHandler->deleteCredential(cred);
+        });
+    }
+}
+
+void LLAppViewer::updateDiscordActivity()
+{
+    discordpp::Activity activity;
```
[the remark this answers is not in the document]
erikkundiman: Oh, question. That macro is just put at the beginning of the function like this?
```
void LLAppViewer::updateDiscordActivity()
{
    LL_PROFILE_ZONE_SCOPED;
    discordpp::Activity activity;
```

indra/newview/llappviewer.cpp
erikkundiman: It's okay! 😃 
The call in STATE_PRECACHE, I kept that to get correct values of coordinates (not just 10,10,10) right away after login, so that it doesn't have to wait until the avatar changes region, but I'll try to find another spot for the call that is not in a loop.
About the question, the requirement listed on [the linked issue](https://github.com/secondlife/viewer/issues/4418) concerning this is "Log presence updates at viewer startup, region change."
That's why I didn't make it update every time the avatar moves, even if it is far but still within the region.

akleshchev: > About the question, the requirement listed on the linked issue concerning this is "Log presence updates at viewer startup, region change."

I think it meant actual logging, LL_INFOS(), but might be wrong, I suggest clarifying that.

indra/newview/llappviewer.cpp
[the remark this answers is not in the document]
akleshchev: Yes.

indra/newview/llappviewer.cpp
erikkundiman: Ah okay.. I asked this on the GH issue just now, and I'll let you know after I get the reply.
(Also I just pushed more commits for the suggestions so far, I hope that's the right thing to do)

akleshchev: > (and not every time the avatar moves within the region)

Updating at every twich would be pushing things too far,

indra/newview/llappviewer.cpp
```diff
@@ -5862,3 +5876,143 @@ void LLAppViewer::metricsSend(bool enable_reporting)
...
+    });
+    if (gSavedSettings.getBOOL("EnableDiscord"))
+    {
+        auto credential = gSecAPIHandler->loadCredential("Discord");
+        if (credential.notNull())
+        {
+            gDiscordClient->UpdateToken(discordpp::AuthorizationTokenType::Bearer, credential->getAuthenticator()["token"].asString(), [](discordpp::ClientResult result) {
+                if (result.Successful())
+                    gDiscordClient->Connect();
+            });
```
akleshchev, line 5899: Doesn't it need something like
```
if (credential.isNull())
{
 handleDiscordSocial();
}
```
to generate new creds if EnableDiscord is true?

erikkundiman: Oh.. hmm. How about just force setting EnableDiscord to false like in line 5929?
`gSavedSettings.setBOOL("EnableDiscord", false);`
So the user won't get any authentication request automatically without manually checking the enable button..

akleshchev: Should be fine, but at minimum that needs to be logged.

indra/newview/llappviewer.cpp
erikkundiman: So, part of the reasons why I let it only report the initial state for the 'party' field, and not update it dynamically periodically within the region, is that the field almost never changes either when I observe, on Discord, other TPVs' implementations. So I just chose what I thought would be the least hurtful for now, but open to improvements in case LL, other contributors, or myself in the future have an idea on how to make it more accurate but still efficient.

About `near_me_uuids`, I just thought of a case where a user arrives at a specific spot in a region to have a local chat with people within chat radius without moving much, and the other people in the region aren't moving much either, that they could be representative enough 😬 
The party field can have 2 numbers (eg. 3 of 10), but it can also have just 1 number, which is the former. If the latter is set to 0, then it won't show up at all. So, we could also just lose `near_me_uuids` and just report `chat_radius_uuids`.

indra/newview/llappviewer.cpp
```diff
@@ -5862,3 +5876,143 @@ void LLAppViewer::metricsSend(bool enable_reporting)
...
+            else
+            {
+                gSavedSettings.setBOOL("EnableDiscord", false);
+            }
+        });
+    }
+    else
+    {
+        auto credential = gSecAPIHandler->loadCredential("Discord");
+        if (credential.notNull())
```
akleshchev, line 5936: Suggestion: if creds are null, log failure to 'disconnect'.

erikkundiman: I moved the disconnect line so that it will try to disconnect from Discord when the integration gets disabled regardless whether there are credentials to delete or not and whether there's an access token to revoke or not.
But I still added logging for when credentials are null, it's just saying that they're nonexistent already.

indra/newview/llappviewer.cpp
erikkundiman: So, by setting the Party, we get, what I consider, the desired layout for the State (region & coords) field:
Screenshot 2025-07-30 at 14 00 34
with the location on the right of the time elapsed field.

A Party requires an ID, CurrentSize & MaxSize, and if we miss one then it will cause the Party field **and** both the Details (username) and State (location) fields to not show somehow, that we might as well not set any Party.
But not setting any Party somehow causes the State field to show up above the time elapsed field, instead of on the right of it:
Screenshot 2025-07-30 at 13 59 25

erikkundiman: I'm thinking.. I could try to move the `updateDiscordActivity` to places where `getAvatars` are already called, and make `updateDiscordActivity` pass those numbers, even if it means separately (eg. having different update functions with different parameters), so not updating Discord Activity all at once.

akleshchev: Ok, thanks. There is probably no point to change it yet. It probably will trip QA, but sounds minor enough that we probably should get it out and and see how it works and get feedback first. For all we know users will want it to work not based of proximity, but based of voice/group chat or from some experience.

P.S. The best way to implement it proximity wise is probably to subscribe to the same thing that populates nearby chat.

Issue #3667, opened 2025-03-06
Improve usability and experience of the Nearby People pane


AshWeststar:
## Issue

Currently, the **Nearby People** panel (Ctrl + Shift + A) in the Second Life viewer does not always display all individuals who appear on the minimap. This inconsistency can be frustrating, as users are unable to view a complete list of people in the same region without manually hovering over each minimap marker.  

For example, in the image below, users are visible on the minimap but do not appear in the Nearby People list:  

![Image](https://github.com/user-attachments/assets/dc5717aa-c5e1-449e-96f7-e37839eab4d8)  

While I understand that this behavior is intentional, since these individuals are outside the "zoomed-in" range, I believe the user experience could be improved.  

## Proposed Solution
I suggest that all individuals within the same region still be listed in the **Nearby People** panel, but with a visual distinction to indicate their distance. Specifically:  
- Users outside the zoom range could appear **grayed out** or **italicized** to signal that certain interactions are unavailable.  
- When right-clicking on these users, any **restricted options should be grayed out** rather than completely hidden.  

## Benefits
- Provides a **more intuitive** and **consistent** experience.  
- Eliminates the need to manually hover over each minimap marker to identify users.  
- Maintains clarity while still reflecting interaction limitations.  

I believe this small UI enhancement would significantly improve usability and experience within the Second Life viewer.

AshWeststar:
I just learned that the debug setting `NearMeRange` manipulates this. I've changed it to `4096` which relatively has the same effect. I think that this probably should be the default since it is likely the largest distance you will have from someone on the same sim.

DarlCat:
The nearby section being overall more useful and informative is a common expectation across most all TPVs, showing data like a numerical distance, account age, or even payment info used/on file status icons is common.

It's also advantageous to move the range limit for this from debug `NearMeRange` to its own setting (or no setting and use a constant) to prevent it from being inadvertently made more frustratingly useless after usage of the avatar picker and shortening the range there since they share the same value. This specific issue is a common support issue among TPVs that have not performed this same modification.

AshWeststar:
> It's also advantageous to move the range limit for this from debug `NearMeRange` to its own setting (or no setting and use a constant)

Making the range configurable via a slider under the cog in Nearby's pane would certainly be a cool approach.

However, I really think that the default is pretty bad and confusing when you can see users on the embedded map that isn't on the Nearby list below and would strongly recommend that this be enhanced to cover all users.

[...]

PR #4490, opened 2025-08-06
Nearby tab slider

secretfoxclaw:
## Description

Add a slider to the "NEARBY" tab of `panel_people.xml` allowing users to adjust `NearMeRange`. 

## Related Issues

"Improve usability and experience of the Nearby People pane": https://github.com/secondlife/viewer/issues/3667

## Screenshot (as implemented in a TPV)

Screenshot_20250806_042033

### DIFF
--- a/indra/newview/skins/default/xui/en/panel_people.xml
+++ b/indra/newview/skins/default/xui/en/panel_people.xml
@@ -181,14 +181,21 @@ Learn about [https://community.secondlife.com/knowledgebase/joining-and-particip
               control_name="NearMeRange"
               decimal_digits="0"
               increment="1"
-              follows="top|left"
+              follows="left|top"
               left="5"
               min_val="0"
               max_val="4096"
               label="Range:"
               name="near_me_range"
               tool_tip="Nearby people range"
-              width="250" />
+              width="230"/>
+            <text
+              type="string"
+              follows="left|top"
+              left_delta="225"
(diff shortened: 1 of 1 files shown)


secretfoxclaw:
> NearMeRange range isn't specific to 'people' panel, shouldn't it be in settings?

It may not be entirely specific, but the people panel is potentially one of the more front-facing instances of where it used. Perhaps a slider or  stepper could be placed in settings, but some users may prefer the convenience of being able to adjust the value directly from the nearby tab.

bennettgoble:
I think this makes sense --having a slider on the nearby tab provides immediate feedback compared to placing it in settings. This also starts building out functionality on the nearby floater, which is what most users want (TPV people floaters have a lot more functionality.)

secretfoxclaw:
> Suggestion: add 'm' at the end of the range slider.

Took some fiddling but I got it done!
Screenshot_20250806_185448

> Perhaps detach it from the global setting and make it 'people' specific?

Like how many TPVs have a separate setting besides `NearMeRange` for their radar / nearby tab?

secretfoxclaw:
@erikkundiman pointed out that adding the slider had caused the bottom of the avatar list to be cut off. I have now reduced the height of the avatar list.

Screenshot_2025-08-07_at_12 20 31
Screenshot_20250807_013250

PR #4501, opened 2025-08-08
Chore: Move message.xml to scripts/messages

bennettgoble:
Get rid of the top-level etc/ directory by moving its only contents, message.xml, to scripts/messages


vldevel:
Why not moving it instead where it gets packaged in the end, i.e. in indra/newview/app_settings/ ?

bennettgoble:
> Why not moving it instead where it gets packaged in the end, i.e. in indra/newview/app_settings/ ?

Yeah, that's even better. Thanks, @vldevel.

Issue #4334, opened 2025-07-03
BugSplat Crash: SecondLifeViewer!LLReflectionMap::getIsDynamic(249)


sl-service-account:
## BugSplat Crash Group 99278
[View Report Group](https://app.bugsplat.com/v2/group?stackKeyId=99278&database=SecondLife_Viewer_2018)
[View Report Details](https://app.bugsplat.com/v2/crash?id=1643243&database=SecondLife_Viewer_2018)
[View Report Attachments](https://app.bugsplat.com/v2/crash?id=1643243&database=SecondLife_Viewer_2018#attachments)

**Application:** Second Life Release
**Version:** 7.1.15.15596336374
**Error Code:** c0000005 Access violation

**Callstack:**
[...]

PR #4507, opened 2025-08-09
#4334 Crash at getIsDynamic #2

akleshchev:
A TPV dev asked to change this, so that ref wouldn't be called on some environments.


Ansariel:
Reason for this change: GCC and older Clang versions will cause an incomplete type compiler error because LLViewerObject is only forward-declared at the point the overloaded constructor of LLPointer is called (in llreflectionmap.h), that internally calls the `ref` method which then tries to access the object handled by LLPointer.

maxim-productengine (approved):

PR #4517, opened 2025-08-11
Fix clang compiler issues due to unused variables

Ansariel:
## Description

A bunch of fixes for clang compiler issues that showed when building Firestorm with the latest clang version and that are also present in the LL viewer, mysteriously haven't been triggered yet.

## Related Issues

No related issue. Do we have to create issues for every tiny fix now?

---

## Checklist

- [X] I have provided a clear title and detailed description for this pull request.
- [X] If useful, I have included media such as screenshots and video to show off my changes.
- [X] I have tested the changes locally and verified they work as intended.
- [X] All new and existing tests pass.
- [X] Code follows the project's style guidelines.
- [X] Documentation has been updated if needed.
- [X] Any dependent changes have been merged and published in downstream modules
- [X] I have reviewed the [contributing guidelines](https://github.com/secondlife/viewer/blob/develop/CONTRIBUTING.md).

---

### DIFF
--- a/indra/newview/gltf/asset.cpp
+++ b/indra/newview/gltf/asset.cpp
@@ -589,7 +589,9 @@ bool Asset::prep()
 
                 for (U32 variant = 0; variant < LLGLSLShader::NUM_GLTF_VARIANTS; ++variant)
                 {
+#ifdef SHOW_ASSERT
                     U32 attribute_mask = 0;
+#endif
                     // for each mesh
                     for (auto& mesh : mMeshes)
                     {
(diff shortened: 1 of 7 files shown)


indra/newview/gltf/asset.cpp
```diff
@@ -607,7 +609,9 @@ bool Asset::prep()
 
                                 // all primitives of a given variant and material should all have the same attribute mask
                                 llassert(attribute_mask == 0 || primitive.mAttributeMask == attribute_mask);
```
Ansariel, line 611: Could alternatively makes this `llassert_always`, but then the check would always be evaluated, even in release builds.

Issue #4519, opened 2025-08-11
SLURL handling code


fmartian:
I was looking at the constructor of a SLURL and one of the earliest things it is doing is to detect if the passed in string has a schema. It does so by simply scanning for a colon (:). This is of course not quite right as the schema of a URL is in fact separated by :// from the actual address part.

I do understand that SLURLs are special URLs and that the standard SLURLs in SecondLife do not contain any other colon as they do newer specify a port number. What is the chance of having a PR which such a (minor) change accepted in the official viewer sources?

PR #4543, opened 2025-08-12
improvement(SLURL): Check for "://" as schema separator in a SLURL

fmartian:
## Description

The passed in string may not contain a schema. Instead of just searching for a ':' to see if there is a schema, do an explicit search for the normal "://" schema separator. If the string is a hostname with port, but with no schema, the parsing goes wrong.

## Related Issues

- [*] relates to #4519


fmartian:

Issue #4541, opened 2025-08-12
Cap image upload dimensions


AtlasLinden:
Some issues are being encountered upon attempting to upload very large images into the viewer now that it is downscaling files larger than 2k: https://github.com/secondlife/viewer/issues/4388

Let's cap the size of images that can be accepted to 8x8k to avoid these issues.

## Test Plan

Verify images higher than 8x8k in dimension are not accepted from the Image Uploader.
Verify images less than or equal to 8k are still accepted.

akleshchev:
Our limit is 256MB, so instead  of limiting strictly by 8Kx8K, I'm limiting by area.

PR #4546, opened 2025-08-13
#4541 Cap image upload size to prevent issues

akleshchev:
Our limit is 256MB, so instead of limiting strictly by 8Kx8K, I'm limiting by area.

### DIFF
--- a/indra/newview/skins/default/xui/en/strings.xml
+++ b/indra/newview/skins/default/xui/en/strings.xml
@@ -3970,7 +3970,7 @@ Abuse Report</string>
   <string name="AvatarBirthDateFormatShort">[mthnum,datetime,slt]/[day,datetime,slt]</string>
 
   <string name="DefaultMimeType">none/none</string>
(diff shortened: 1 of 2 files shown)


marchcat (approved):

indra/newview/skins/default/xui/en/strings.xml
```diff
@@ -3970,7 +3970,7 @@ Abuse Report</string>
   <string name="AvatarBirthDateFormatShort">[mthnum,datetime,slt]/[day,datetime,slt]</string>
 
   <string name="DefaultMimeType">none/none</string>
-  <string name="texture_load_dimensions_error">Can't load images larger than [WIDTH]*[HEIGHT]</string>
+  <string name="texture_load_dimensions_error">Can't load images larger than [PIXELS] pixels</string>
```
marchcat, line 3973: This might still be confusing. Maybe we could simplify it by stating the limit using only the larger dimension of the image.

PR #4563, opened 2025-08-15
Add new LastModified option to HttpRequest handling

fmartian:
## Description

Add a LastModified: option to the HttpOptions class and handle it properly in HttpOpRequest::prepareRequest()


bennettgoble:
Thanks, @fmartian this seems reasonable. Do you plan on using LastModified for a specific purpose?

Issue #4566, opened 2025-08-17
Add Memory Map and 64 bit file support to LLAPRFile class


minerjr:
Currently the LLAPRFile class only supports 32 bit files and 2 GB file limits.

Propose to add additional 64 bit methods which do not affect existing code, but allow newer code to use 64 bit versions of all APR commands.

The APR library is 64 bit native and the LLAPRFile class down casts to 32 bit.

Propose to expose native 64 bit methods using new 64 bit named versions of the existing LLAPRFile methods.
As well, memory map files are an existing feature of the APR library, but are not currently exposed to the LLAPRFile code base.

Propose to add new methods for loading a memory map file, creating and zeroing out a memory map file as well as assign object pointers to a memory map region.

With a memory map, you could map a data structure directly to memory, with out have to first load the object to RAM, then create a new object, then copy from the read in memory from file to the new object.

This could be helpful for such things as the Texture Fast Cache, or other resources which are cached in a native format that does not require conversion of the data. Could be also used for normal Texture Cache for mapping to a Raw texture to be decoded which would remove the need for File IO to read the data in. If a cache is larger then 2GB for a unified file, 64 file support would allow the file to be used or memory mapped.

Also, the #defines of the LLAPR.h file are using deprecated APR defines. Propose to update to the correct naming from the library which will remove deprecated warnings.

## Test Plan

1. Regression

- Perform regression testing to validate that Textures, which use LLAPRFIle, load as they have done in the past.
- This will show that the code does not affect the existing calls the LLAPRFile class and its methods.

2. 64 Bit file support

- Can create a piece of test code where load a large single file. 
- Can create a simple file where you seek to past the 2GB file size location and write a known value to the location.
- Then read back the value by opening the file for reading 64 bit and see if the value returned is the same.

3 Memory Map file support

- Create a piece of test code which opens the file from the second test as a memory map. 
- Take a pointer of the data type saved to the file and assign it the location of the saved data. 
- Then print out the value stored in the pointer to validate it matches the value from the second test.
<br/>
<hr/>

[...]

PR #4567, opened 2025-08-17
Add support for Memory Map and 64 bit files in LLAPRFile class

minerjr:
## Description

Currently the LLAPRFile class only supports 32 bit files and 2 GB file limits. Also, memory mapped files are not supported.

This PR adds additional native 64 bit methods using new 64 bit named versions of the existing LLAPRFile methods. This will allow files over 2 GB from being accessed and loaded.

This PR also adds new methods for loading a memory map file, creating and zeroing out a memory map file as well as assign object pointers to a memory map region.

With a memory map, you could map a data structure directly to memory, with out have to first load the object to RAM, then create a new object, then copy from the read in memory from file to the new object. This also removes the need for File IO commands for accessing data stored in the file.

This could be helpful for such things as the Texture Fast Cache, or other resources which are cached in a native format that does not require conversion of the data. Could be also used for normal Texture Cache for mapping to a Raw texture to be decoded which would remove the need for File IO to read the data in. If a cache is larger then 2GB for a unified file, 64 file support would allow the file to be used or memory mapped.

Also, the #defines of the LLAPR.h file are using deprecated APR defines. Propose to update to the correct naming from the library which will remove deprecated warnings.

Change summary:

Added support for 64 Bit files to llapr.cpp
- open64
- seek64
- read64
- write64
- static size64
- static seek64
- static readEx64
- static writeEx64

Added support for Memory map files
- LLARPFile with mmap Read/Write flags
- openMemoryMap
- openMemoryMap64
- openMemoryMap - no size specified
- memoryMapAssign
- memoryMapAssign64

Updated #define's for LL_APR_R etc. to use newer ARP_FOPEN_READ as the old ARP_READ is deprecated

Updated llassert_always to use signed 64bit limit instead of signed 32bit

## Related Issues

Issue Link: Address issue: https://github.com/secondlife/viewer/issues/4566

## Additional Notes

This will be a basis for a possible enhanced Texture Fast Cache system.

### DIFF
--- a/indra/llcommon/llapr.cpp
+++ b/indra/llcommon/llapr.cpp
@@ -341,8 +341,8 @@ LLAPRFile::LLAPRFile(const std::string& filename, apr_int32_t flags, apr_int32_t
 // Create a 32 bit memory map file pointer and initalize it.
 LLAPRFile::LLAPRFile(const std::string& filename, apr_int32_t flags, apr_int32_t mmap_flags, S32 init_file_size, bool zero_out, LLVolatileAPRPool* pool)
     : mFile(NULL),
-    mMMapFile(NULL),
-    mCurrentFilePoolp(NULL)
+      mMMapFile(NULL),
+      mCurrentFilePoolp(NULL)
 {
     openMemoryMap(filename, flags, mmap_flags, init_file_size, zero_out, pool);
 }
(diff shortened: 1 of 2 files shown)


minerjr:

minerjr:
Hi thanks for the feedback. I will make the noted changes. 

@bennettgoble I wrote my thoughts on 2 possible solutions in this feature request form.

https://feedback.secondlife.com/feature-requests/p/improved-texture-fast-cache-system

Let me know what you think of the ideas as any feedback is welcome.

I have the second proposal working mostly implemented as a proof of concept. There are a few things to work out when it comes to replacing the need to load from normal cache when the fast cache can handle it at times.

bennettgoble:
@minerjr I hope you don't mind --I've requested a copilot code review on these changes. Copilot reviews can generate some useful feedback, but can also be incorrect at times. Feel free to review and resolve/address commentary from the bot. :)

minerjr:
@bennettgoble Sure sounds good, I will take a look at implementing in the morning. 

I quickly scanned the report and saw a few that I will put in right away and they make sense. Mostly typo's in the comments and a few extra formatting and error handling things.

vldevel:
APR file stuff needs APR pools: those are a PITA to manage and have been the cause for many crashes (especially on shutdown) and memory leaks in the viewer code over years.

Why not getting fully rid of LLAPRFile and instead implement everything through LLFile (i.e. delegating to the OS the file concurrency access management, something all modern OSes can deal with without problem) ?

FYI, this is what I did in the Cool VL Viewer, years ago, and I am so happy I did it (so much stabler and so much easier to maintain) !

What I did is to implement LLFile() constructors and methods able to deal with LLFile instances, in excess of the static methods. I also migrated LLAPRFile::readEx() and LLAPRFile::writeEx() as static methods of LLFile (as for the 2GB limit with them, removing it is just a matter of using S64's for offsets, instead of S32's).

Feel free to borrow the Cool VL Viewer code for all this (you also have permission to re-license it as LGPL/whatever you need) !  You will also find interesting touch(), [un]lock() and createFileSymlink() methods in my improved/expanded code, as well as work around for [a Wine bug](https://bugs.winehq.org/show_bug.cgi?id=56837) (existing in current Wine 9 release), and since, unlike most TPVs, LL's viewer does not yet have a Linux build, it is relevant to it.

PS: oh, and I also got fully rid of LLLFSThread, which was barely used anyway, and can advantageously be replaced with my new LLFile instances and the LL:WorkQueue's for threading file operations.

minerjr:
@bennettgoble @Geenz I pushed up the copilot changes. I did it as a single update instead of trying to use the suggestions directly as some needed some tweaks. Nice feature though. I should try to use it offline before submitting in the future. 

I pushed up the changes so that it's at least in the best version of my PR. If you want to close the PR down and use @vldevel solution that is fine as well.

PR #4574, opened 2025-08-18
Improved open performance for some web based UI floaters by preloading the web content during login 

callumlinden:
## Description
This work moves some of the web based UI floaters (Search and perhaps Marketplace) into a simplified floater (out of the generic but overly complex WebContent Floater) and then preloads the content during login.  This doesn't add any meaningful delay to login but means that when activated, they appear almost instantly versus now, where it can take several seconds to full render.

In addition, the Search floater has to support the SLAPP Search feature described [here](https://lindenlab.atlassian.net/wiki/spaces/QA/pages/2696380641/FUI+SLApps+Test+Plan#3.5-Search)

### Outstanding Issues
* The documentation for Search SLAPPs seems to (a) exist in many places and is different in each place and (b) not fully describe what a "category" is. The behavior with this change seems to match but we may need to uncover a better description of what's expected before this can be called complete.

## Testing
At a high level the criteria for approving this change is "Do Search and Marketplace load faster than they do in the current release Viewer".

#### Search
* After you log into the Viewer and open Search using the same keyboard shortcut or menu option, the Search floater should appear much more quickly.
* Confirm you are logged in as expected (there will be a 'log in' button if you are not)
* You will notice that the extraneous browser-eqque UI has been removed. No more URL entry bar that is disabled. No more back/forward buttons. No more status bar at bottom. This is all by design.
* Confirm that Search works as it did before.  I wasn't able to find much in the way of docs - this [FUI doc](https://lindenlab.atlassian.net/wiki/spaces/QA/pages/2699493708/FUI+Viewer+Search+Floater) was the best I could do. You will notice that some of that plan is now invalid E.G. there are no longer back/forward arrows in the Search floater.

#### Marketplace
* Notice the same type of changes with the Marketplace floater you see when pressing the "Buy" button in the top menu bar.  It should open much more quickly that before and present the same content.
* Confirm you are logged in as expected (there will be a 'log in' button if you are not)
---

## Checklist

- [x] The PR is linked to a relevant issue with sufficient context.
- [x] Documentation has been updated if needed.
- [x] Any dependent changes have been merged and published in downstream modules

---


akleshchev (approved):
> and then preloads the content during login

How much memory do those instances take? Pehaps make them preload only with mid and higher graphical settings if it's anything notiecable?

marchcat (approved):

indra/newview/llfloatersearch.cpp
```diff
@@ -182,30 +132,41 @@ void LLFloaterSearch::search(const SearchQuery &p)
...
+    search_browser->navigateTo(url, HTTP_CONTENT_TEXT_HTML);
+}
+
+bool LLFloaterSearch::postBuild()
+{
+    enableResizeCtrls(true, true, false);
+
+    // This call is actioned by the preload code in llViewerWindow
+    // that creates the search floater during the login process
+    // and navigates to the SearchURLPreload from settings.xml
```
maxim-productengine, line 168: It seems that 'SearchURLPreload' setting actually is not used at all?

callumlinden: Done.

PR #4577, opened 2025-08-19
#4284 Reorder window shutdown

akleshchev:
Not sure what exactly caused a freeze, not enough info in those logs, so I'm enabling extra logging and making sure thread survives till detach.


maxim-productengine (approved):

marchcat (approved):

akleshchev:
Thank you.

That makes very little sense. It didn't get to detaching thread, so it either got stuck on ShowWindow or on SetWindowLongPtr.
Can you comment out ShowWindow and see if it still freezes?

Issue #3223, opened 2024-12-06
LLCoprocedurePool uses a great deal of memory


monty-linden:
[ Moved from https://github.com/secondlife/viewer-private/issues/320 ]

Part of the viewer avatar experience improvement effort going on in 4Q24.  Diving into causes of slow/delayed avatar appearance as described here:  https://docs.google.com/document/d/1FMLp1B3JOnQYm5Jvz52YweggMF6hJQrwmAYQuHwWTis/edit?tab=t.0

Here I was familiarizing myself with the coro adoption that occured 10 years ago.  I stumbled across a multi-thread request buffer in the form of LLCoprocedurePool.  We have four of these which present during startup as follows:

```
2024-11-15T21:33:24Z INFO #CoProcMgr# llmessage/llcoproceduremanager.cpp(355) LLCoprocedurePool::LLCoprocedurePool : Created coprocedure pool named "Upload" with 1 items, queue max 1048576
```

The four classes of things handled by these pools are:  Upload, AIS, AssetStorage, ExpCache.  The request queue is based on boost::fiber::buffered_channel.  The typical implementation of this is as a ring buffer operating on a single allocation of maximum size, here 1048576 elements in all four cases.  This amounts to something like 64-256MB of memory that is typically unused.  When it is used, it slowly crawls around the ring buffer dirtying pages and requiring residency.

There are other reasons (to show up in other GHIs) not to use buffered_channel and LLCoprocedureManager as a work queue and as an interface between multi-threads and multi-coros.  But while we are using them, tuning this down to a correct sizing for each class would recover a good chunk of memory.

vldevel:
The new LLCoprocedurePool code, introduced with the "performance viewer" changes is ill-coded and a catastrophe.

FYI, in the Cool VL Viewer I kept (and just refreshed/modernized a bit) the old code for LLCoprocedureQueue, which is using a std::deque storage queue (protected with boost::fibers::mutex) with a default capacity capping at 1024 elements only; it requires additional checks in the consumer code, when calling LLCoprocedurePool::enqueueCoprocedure(), to check that the queuing did happen and was not rejected due to a full queue (in which case the queuing must be retried at next frame by the consumer code), but it avoids the madness seen in the new code.

akleshchev:
> here 1048576 elements in all four cases.

Upload and experience cache now get 2048.

> There are other reasons (to show up in other GHIs) not to use buffered_channel and LLCoprocedureManager

[...]

PR #4579, opened 2025-08-19
#3223 Trim coroutine queues a little

akleshchev:
Queue appears to use a fair bit of space. It looks like the plan is to switch to other mechanisms, but for now queue gets an easy and cheap trim.


marchcat (approved):

vldevel:
The new LLCoprocedurePool code, introduced with the "performance viewer" changes is ill-coded and a catastrophe.

FYI, in the Cool VL Viewer I kept (and just refreshed/modernized a bit) the old code for LLCoprocedureQueue, which is using a std::deque storage queue (protected with boost::fibers::mutex) with a default capacity capping at 1024 elements only; it requires additional checks in the consumer code, when calling LLCoprocedurePool::enqueueCoprocedure(), to check that the queuing did happen and was not rejected due to a full queue (in which case the queuing must be retried at next frame by the consumer code), but it avoids the madness seen in the new code.

PR #4599, opened 2025-08-22
Websockets Server implementation for script editor integration with VSCode or other external editors.

Rider-Linden:
## Description

Implements a websocket server that may be launched in the viewer to allow external programs two way communication with systems in the viewer.  

Allows external code editors to communicate with the viewer and receive compile status and runtime information.  It also allows for transfering the current language definitions from the viewer to the editor. 
(See: doc/external-editor-json-rpc.md for full protocol.)

## Related Issues
Needs a full test plan, should test in coordination with https://github.com/secondlife/sl-vscode-edit/

### DIFF
--- a/indra/newview/CMakeLists.txt
+++ b/indra/newview/CMakeLists.txt
@@ -1653,7 +1653,8 @@ set_source_files_properties(${viewer_SHADER_FILES}
                             PROPERTIES HEADER_FILE_ONLY TRUE)
 list(APPEND viewer_SOURCE_FILES ${viewer_SHADER_FILES})
 
-
+# note that the scripts which publish these files are sensitive to the file 
+# extensions. Be cautious about adding new file types here.
 set(viewer_APPSETTINGS_FILES
     app_settings/anim.ini
     app_settings/autoreplace.xml
--- a/indra/newview/llpreviewscript.cpp
+++ b/indra/newview/llpreviewscript.cpp
@@ -1113,11 +1113,11 @@ void LLScriptEdCore::doSave( bool close_after_save )
 
 void LLScriptEdCore::openInExternalEditor()
 {
-    if (mContainer->mLiveFile)
-    {
-        // If already open in an external editor, just return
-        return;
-    }
+    //if (mContainer->mLiveFile)
+    //{
+    //    // If already open in an external editor, just return
+    //    return;
+    //}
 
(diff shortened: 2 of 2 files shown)


autobuild.xml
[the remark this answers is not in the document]
bennettgoble: Let's avoid the churn if possible --manual edit or figure out the sorting (presuming dictionary sorting differences with python llsd.)

bennettgoble:
@Rider-Linden could we get a summary of the JSON-RPC API this exposes? Perhaps it'd be useful to provide a separate *.md documenting the interface for developers to reference.

marchcat (approved):

bennettgoble:
> I'll make a copy of the one included in the vscode extension. I usually dislike having two copies of the same document, but I'm not sure there is a way around it.

The server defines the spec: so I would think it would be appropriate to place in here, then link to the document from the extension repo.

indra/newview/llscripteditorws.cpp
```diff
@@ -0,0 +1,830 @@
...
+        //    }
+        //    else
+        //    {
+        //        error_message = "Unknown script runtime error";
+        //    }
+        //}
+        //else
+        //{
+        //    error_message = "Unknown script runtime error";
+        //}
```
Rider-Linden, line 601: Ok... This bit has me flumoxed.

HaroldCindy: Yeah I've been meaning to merge the last two into a single message, it just means that I need to add a new function to the server that allows sending a chat message to everyone in the sim except a single person. Then we can send most users the general error message, then a more specific one to the owner.

No idea about the first though. Things in-world might be listening for that message currently? Do Mono objects currently get 2 messages on error?

Rider-Linden: Yes, Mono gets two messages.  So, that will simplify things.  
From the slack conversation:
- We'll consolidate this so that it is always 2 messages. 
- We can not go to one message since that may break script crash monitors.

indra/llcorehttp/lljsonrpcws.cpp
Geenz, line 1: I'd want to see a little bit more separation here of concerns into individual files (for example, LLJSONRPCServer and LLJSONRPCConnection), but that's not a blocker.  Generally looks fine.

PR #4635, opened 2025-09-05
viewer-cef build fixes

akleshchev:
Build fixes to make mac and windows pass cmake stage.

### DIFF
--- a/indra/CMakeLists.txt
+++ b/indra/CMakeLists.txt
@@ -11,8 +11,12 @@
 ##   We're not there yet, but once done, there is a kludge in Linking.cmake
 #    "if(${CMAKE_VERSION} VERSION_LESS "3.20.0")" that can also be removed
 cmake_minimum_required(VERSION 3.16.0...4.0 FATAL_ERROR)
-cmake_policy(SET CMP0156 NEW)
-cmake_policy(SET CMP0179 NEW)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.29.0")
+    cmake_policy(SET CMP0156 NEW)
+endif()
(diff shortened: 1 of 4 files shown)


indra/cmake/Boost.cmake
Ansariel, line 1: This is wrong. The lib names start with `libboost` for all 3 OS now.

indra/CMakeLists.txt
```diff
@@ -11,8 +11,12 @@
 ##   We're not there yet, but once done, there is a kludge in Linking.cmake
 #    "if(${CMAKE_VERSION} VERSION_LESS "3.20.0")" that can also be removed
 cmake_minimum_required(VERSION 3.16.0...4.0 FATAL_ERROR)
-cmake_policy(SET CMP0156 NEW)
-cmake_policy(SET CMP0179 NEW)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.29.0")
```
Ansariel, line 14: The policies should probably only be set for CMake versions less than 4.0 and the minimum required CMake version should only specify a minimum version. Specifying a maximum version here is going to produce empty lines in the build output at least if using CMake 4.1.0.

indra/cmake/LLKDU.cmake
Ansariel, line 1: Have you checked the content of the 3p packages? Pretty sure the lib names got unified.

indra/cmake/Boost.cmake
akleshchev: Old variant builds just fine on mac os (intel) but wasn't configuring on windows. I agree that logically mac should ahve been fine with libboost (files in the package are libboost), will try unifying.

callumlinden (approved):

PR #4643, opened 2025-09-05
Avoid using boost's wstring path to string conversion

akleshchev:
Boost's wstring path to string conversion is supposed to work based of system's locale, avoid it and convert to utf-8 explicitly.

Draft because I need to wrap this in windows/not windows checks.

P.S. Will make a ticket with a scenarion on a fresher head.


marchcat (approved):

akleshchev:
I thought it was a problem, but apparently it just doesn't hit any path that can be a problem. Or at least I couldn't find such a scenario, closing for now.

Issue #4670, opened 2025-09-12
Apple Silicon users cannot be heard by others in p2p calls


maestrolinden:
### Environment

Second Life Test 7.2.2.17662967096 (64bit)
Release Notes

You are at 98.5, 252.3, 23.0 in webRTC1 located at simhost-0462d668818ec3f3b.aditi
SLURL: secondlife://Aditi/secondlife/webRTC1/99/252/23
(global coordinates 259938.0, 245500.0, 23.0)
WebRTC-Voice 2025-08-21.17114312399
Release Notes

CPU: Apple M1 Pro (2400 MHz)
Memory: 16384 MB
OS Version: macOS 15.6.1 Darwin 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 arm64
Graphics Card Vendor: Apple
Graphics Card: Apple M1 Pro

OpenGL Version: 4.1 Metal - 89.4

Window size: 1173x815
Font Size Adjustment: 96pt
UI Scaling: 0.75
Draw distance: 296m
Bandwidth: 10000kbit/s
LOD factor: 1.125
Render quality: 0
Texture memory: 10922MB
Disk cache: Max size 2150.4 MB (73.8% used)
HiDPI display mode:

J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.21.0.202508272158
CEF: 139.0.28+g55ab8a8+chromium-139.0.7258.139
Chromium: 139.0.7258.139
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway

Packets Lost: 0/11736 (0.0%)
September 12 2025 09:46:35

### Description

If a user is on the viewer build with special apple silicon support, their voice cannot be heard by others in p2p calls, though others see their voice indicator waves in that case.   This issue does not affect spatial voice or multiagentchat calls.

### Reproduction steps

1. UserA (who is on apple silicon): visit a webRTC region
2. UserB (who is not on apple silicon; my UserB was on the 2025.06 viewer on Windows): visit UserA's location
3. UserA and UserB: speak in voice, and verify that both voice chat audio and visual indicators are working for both users
4. UserA: Initiate a p2p call to UserB
[...]

Issue #4669, opened 2025-09-12
Apple Silicon: changing the active input or output device for voice chat has no effect


maestrolinden:
### Environment

Second Life Test 7.2.2.17662967096 (64bit)
Release Notes

You are at 98.5, 252.3, 23.0 in webRTC1 located at simhost-0462d668818ec3f3b.aditi
SLURL: secondlife://Aditi/secondlife/webRTC1/99/252/23
(global coordinates 259938.0, 245500.0, 23.0)
WebRTC-Voice 2025-08-21.17114312399
Release Notes

CPU: Apple M1 Pro (2400 MHz)
Memory: 16384 MB
OS Version: macOS 15.6.1 Darwin 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 arm64
Graphics Card Vendor: Apple
Graphics Card: Apple M1 Pro

OpenGL Version: 4.1 Metal - 89.4

Window size: 1173x815
Font Size Adjustment: 96pt
UI Scaling: 0.75
Draw distance: 296m
Bandwidth: 10000kbit/s
LOD factor: 1.125
Render quality: 0
Texture memory: 10922MB
Disk cache: Max size 2150.4 MB (73.8% used)
HiDPI display mode:

J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.21.0.202508272158
CEF: 139.0.28+g55ab8a8+chromium-139.0.7258.139
Chromium: 139.0.7258.139
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway

Packets Lost: 0/11736 (0.0%)
September 12 2025 09:46:35

### Description

When a user on Second Life Test 7.2.2.17662967096 (64bit), who is also using Apple silicon changes their audio input and output device for voice chat to a non-Default value, the change has no effect.

### Reproduction steps

1. Login with the viewer and connect to voice chat in a webRTC region
2. Engage in voice chat with another user
3. Open Preferences -> Sound & Media -> Voice input and output devices
4. Change both the Input and Output dropdowns to a different device than before
5. Hit "OK" to apply
6. Engage in voice chat with another user; observe if the settings from (4) are respected

Expected results:
[...]

PR #4676, opened 2025-09-13
Fix indexing problem with mac devices

roxanneskelly:
This resulted in the wrong device being selected.

Also, fix a shutdown crash where recording was not being stopped, hence the recording thread was still running on shutdown and crashed because it lost access to resources.

Fix an issue with p2p calls where they're coming up muted even though the button indicates they are unmuted.

Issues:
#4669
#4670 
#4684

## Description

Test guidance
- repro the issues in the above tickets.

Also
- Bring up a viewer on windows, in a webrtc region.
- Click on the speak button
- shut down the viewer
- validate there is no crash.

### DIFF
--- a/indra/llwebrtc/llwebrtc.cpp
+++ b/indra/llwebrtc/llwebrtc.cpp
@@ -350,6 +350,16 @@ void LLWebRTCImpl::init()
 
 void LLWebRTCImpl::terminate()
 {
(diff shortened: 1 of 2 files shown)


roxanneskelly:
Yeah, it's kind of a mix between webrtc.lib's coding style and ours.  I'll make a note to clean that up.

PR #4678, opened 2025-09-13
LLLeap: handle partial lines in stderr

marchcat:
Attempt to fix that annoying test failure.

Let's see if this build succeeds.


marchcat:
No test failures so far.
Sign/package step is still red due to the NSIS issue, but that's out of the scope.

akleshchev (approved):

Issue #4695, opened 2025-09-17
voice causing an apple crash report after logging out


Dan-Linden:
### Environment

..

### Description

Voice causing an apple crash report after logging out

### Reproduction steps

1. Log into a WebRTC region.
2. Engage the voice button.
3. Teleport to a Vivox region
4. Log out. 
5. The logout seems normal. Wait for it...

Observed: After a few seconds or a minute the Apple Crash reporter will pop up.

[apple crash report.txt](https://github.com/user-attachments/files/22391950/apple.crash.report.txt)

[SecondLife.log](https://github.com/user-attachments/files/22391953/SecondLife.log)

Expected: No Apple Crash report

akleshchev:
> Teleport to a Vivox region.

Apparently there was no cleanup happening, fix in progress.

PR #4718, opened 2025-09-19
#4695 Fix missing voice cleanup

akleshchev:
Fixed cleanup in LLVoiceClient::terminate(), only one module was cleaned instead of both.
gWebRTCImpl was never deleted
Added mDeviceModule security


roxanneskelly (approved):
Good Stuff

PR #4725, opened 2025-09-20
Fix #4195: Preserve transforms when switching PBR materials

### LINKED ISSUE #4195: Make Scale/Offset/Rotation PERSIST when switching PBR Materials
In the texture/material properties of an item, the scale, offsets and rotations are set in the edit window. However, when any PBR material is applied, it resets to the default values of 1.0x1.0, 0u/0v, 0 degrees, cancelling any previous settings. This makes it very tedious to switch between different PBR materials.

This is a request to make the scale/offset/rotation of an item *PERSIST* after they are set in the edit window when a PBR material is switched, instead of defaulting to 1x1, etc. with the new material, as is the case when any blinn-phong texture maps are changed.

https://secondlife.canny.io/admin/board/bug-reports/p/make-scale-offset-rotation-persist-when-switching-pbr-materials

Aqil-Ahmad:
## Description

Fixes texture transforms being reset when switching from Blinn-Phong to PBR materials and between PBR. Previously, custom scale, offset, and rotation settings would be lost, making it tedious to switch between PBR materials.

## Related Issues

Issue Link: #4195 

---

## Checklist

- [x] The PR is linked to a relevant issue with sufficient context.
- [x] Documentation has been updated if needed.
- [x] Any dependent changes have been merged and published in downstream modules

---

## Additional Notes

https://github.com/user-attachments/assets/8c602efb-e7f8-454e-b89b-2a1c7ffc82df

### DIFF
--- a/indra/newview/llgltfmateriallist.cpp
+++ b/indra/newview/llgltfmateriallist.cpp
@@ -380,17 +380,6 @@ void LLGLTFMaterialList::queueApply(const LLViewerObject* obj, S32 side, const L
         LLGLTFMaterial* material = new LLGLTFMaterial(*material_override);
         sApplyQueue.push_back({ obj->getID(), side, asset_id, material });
     }
-
-    if (sUpdates.size() >= MAX_TASK_UPDATES)
-    {
-        LLCoros::instance().launch("modifyMaterialCoro",
-            std::bind(&LLGLTFMaterialList::modifyMaterialCoro,
-                gAgent.getRegionCapability("ModifyMaterialParams"),
-                sUpdates,
-                std::shared_ptr<CallbackHolder>(nullptr)));
-
-        sUpdates = LLSD::emptyArray();
-    }
 }
 
 void LLGLTFMaterialList::queueUpdate(const LLSD& data)
--- a/indra/llprimitive/llgltfmaterial.cpp
+++ b/indra/llprimitive/llgltfmaterial.cpp
@@ -923,3 +923,34 @@ void LLGLTFMaterial::updateTextureTracking()
     // setTEGLTFMaterialOverride is responsible for tracking
     // for material overrides editor will set it
 }
+
+void LLGLTFMaterial::convertTextureTransformToPBR(
+    F32 tex_scale_s,
+    F32 tex_scale_t,
+    F32 tex_offset_s,
+    F32 tex_offset_t,
+    F32 tex_rotation,
+    LLVector2& pbr_scale,
+    LLVector2& pbr_offset,
+    F32& pbr_rotation)
+{
+    pbr_scale.set(tex_scale_s, tex_scale_t);
+    pbr_rotation = -(tex_rotation) / 2.f;
+    const F32 adjusted_offset_s = tex_offset_s;
+    const F32 adjusted_offset_t = -tex_offset_t;
+    F32 center_adjust_s = 0.5f * (1.0f - tex_scale_s);
+    F32 center_adjust_t = 0.5f * (1.0f - tex_scale_t);
+
+    if (pbr_rotation != 0.0f)
+    {
+        const F32 c = cosf(pbr_rotation);
+        const F32 s = sinf(pbr_rotation);
+        const F32 tmp_s = center_adjust_s * c - center_adjust_t * s;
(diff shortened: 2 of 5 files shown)


indra/newview/llpanelface.cpp
```diff
@@ -5188,6 +5208,17 @@ void LLPanelFace::onTextureSelectionChanged(LLInventoryItem* itemp)
...
 void LLPanelFace::onPbrSelectionChanged(LLInventoryItem* itemp)
 {
+    // Commit transform values before switching PBR materials
+    if (mPBRScaleU && mPBRScaleV && mPBROffsetU && mPBROffsetV && mPBRRotate)
+    {
+        onCommitGLTFTextureScaleU();
+        onCommitGLTFTextureScaleV();
+        onCommitGLTFTextureOffsetU();
+        onCommitGLTFTextureOffsetV();
+        onCommitGLTFRotation();
```
akleshchev, line 5218: This causes 5 individual updates, 5 different overrides will be racing each other to the server.

Aqil-Ahmad:
> If that's indeed the cause, the solution is to compare current material id and if it didn't change then not send anything at 'OK', nothing changed after all.

You were right: the OK issue came from calling queueApply with the same material ID, which cleared overrides.
I tried to preserve values in the PBR→PBR case but couldn’t get it to work; I suspect a clear-then-override ordering, even though the local logic looks correct.

akleshchev:
> I tried to preserve values in the PBR→PBR case but couldn’t get it to work;

I think it did work in your closed PR with a 'drag&drop material over previous material' case.

indra/newview/llgltfmateriallist.cpp
```diff
@@ -371,24 +371,19 @@ void LLGLTFMaterialList::queueApply(const LLViewerObject* obj, S32 side, const L
 {
     if (asset_id.isNull() || material_override == nullptr)
     {
-        queueApply(obj, side, asset_id);
-    }
-    else
-    {
-        LLGLTFMaterial* material = new LLGLTFMaterial(*material_override);
-        sApplyQueue.push_back({ obj->getID(), side, asset_id, material });
+        return;
```
akleshchev, line 374: Please, consider adding llassert or LL_WARN here. Something along the lines of 'tried to use wrong queueApply variant to reset material'.

indra/newview/llgltfmateriallist.cpp
```diff
@@ -371,24 +371,19 @@ void LLGLTFMaterialList::queueApply(const LLViewerObject* obj, S32 side, const L
...
     {
-        LLCoros::instance().launch("modifyMaterialCoro",
-            std::bind(&LLGLTFMaterialList::modifyMaterialCoro,
-                gAgent.getRegionCapability("ModifyMaterialParams"),
-                sUpdates,
-                std::shared_ptr<CallbackHolder>(nullptr)));
-
-        sUpdates = LLSD::emptyArray();
+        queueModify(obj, side, material_override);
+        return;
```
akleshchev, line 381: Our local id being accurate is an assumption not a fact. Generaly we assume that viewer might be not up to date and sensing an id is a good thing, but since that's a workaound for a server issue, please mark it as such.

It also adds a pitfall for future maintainers, because people will try to set render id first then will send it to server and it will fail.

Exmple:
`// Server side has a bug that makes server ignore override if asset id did not change, untill that is fixed, avoid re-applying the same id; just modify override`
LL_WARNS() << "No apparent change, Ignoring asset id and sending overrides only" << LL_ENDL;

P.S. I will create a server ticket to fix it.

indra/newview/llgltfmateriallist.cpp
```diff
@@ -380,17 +380,6 @@ void LLGLTFMaterialList::queueApply(const LLViewerObject* obj, S32 side, const L
         LLGLTFMaterial* material = new LLGLTFMaterial(*material_override);
         sApplyQueue.push_back({ obj->getID(), side, asset_id, material });
     }
-
-    if (sUpdates.size() >= MAX_TASK_UPDATES)
```
akleshchev, line 384: Why are you removing this part?

akleshchev (changes requested):

Issue #4715, opened 2025-09-19
Zoom In behavior to Object via Chat After Region Change


calebrosia:
Steps to Reproduce:
User A and User B are in the same region
User A rezzes a prim and enters script

default
{
    touch_start(integer total_number)
    {
        llSay(0, "Touch: llSay");
        llWhisper(0, "Touch: llWhisper");
        llShout(0, "Touch: llShout");
    }
}

User A touches prim to emit and Object chat event
User B is able to Zoom In via Object chat event (right-click, Zoom In)
User B teleports away to a region that is not bordering the original region
User B right-clicks on the Object chat event 

Actual Result
Object chat is still allowing User B to "Zoom In"
During this stale window, clicking "Zoom In" attempts to zoom the camera, but since you are not connected, the camera only moves within the region in the general direction of the object
Impact: This can confuse users, as the UI suggests the feature is available when it is not.

Expected Result
Consider immediately invalidating or disabling the "Zoom In" option when the region connection is dropped or the agent moves away, rather than waiting for a timeout.

Reference: https://github.com/secondlife/viewer/issues/4474

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4715</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @calebrosia</code> (replace <code>20</code> with the amount, and <code>@calebrosia</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

Dan-Linden:
Passed QA.
Verified on Second Life Release 7.2.3.19118091373 on Win10/OSX in the scope of https://github.com/secondlife/iqa/issues/566

Teleporting about 1.5 regions away will invalidate the zoom in link. If you make the object visible again, then the zoom in link becomes active.

PR #4726, opened 2025-09-22
#4715 disable ZoomIn an object item when the region is not connected to the current region

maxim-productengine:
Disable 'Zoom In' menu item if the object's region is not reachable from the current agent's region.

### DIFF
--- a/indra/newview/llviewerobject.cpp
+++ b/indra/newview/llviewerobject.cpp
@@ -7717,6 +7717,51 @@ void LLViewerObject::clearTEWaterExclusion(const U8 te)
     }
 }
 
+bool LLViewerObject::isReachable()
+{
+    LLViewerRegion* agent_region = gAgent.getRegion();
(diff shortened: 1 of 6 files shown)


indra/newview/llviewerobject.cpp
```diff
@@ -7717,6 +7717,51 @@ void LLViewerObject::clearTEWaterExclusion(const U8 te)
...
+bool LLViewerObject::isReachable()
+{
+    LLViewerRegion* agent_region = gAgent.getRegion();
+    LLViewerRegion* object_region = getRegion();
+
+    if (!agent_region || !object_region)
+    {
+        return false;
+    }
+    if (agent_region == object_region)
```
akleshchev, line 7729: Isn't it too harsh? Object might be visible and even right next to you, just across the border.

akleshchev (approved):

Issue #4805, opened 2025-10-09
LLFile::rename() and LLFile::remove() are not Posix compliant under Windows


fmartian:
### Environment

Current develop branch HEAD

### Description

LLFile::rename() fails under Windows when the new_name already exists while the Posix standard for rename() says that it should simply create the new file anyhow, such that at no point there does not exist any new_name for other processes.
This last one may be a bit hard to guarantee under Windows without a lot of testing and debugging and low level file tinkering, but making it not fail when the new_name already exists is fairly trivial by calling directly MoveFileExW() with the according flags rather than the cludgy Windows C runtime replacement _wrename(), which does actually nothing else than calling that function too, but without the according flag. This results in quite some places of the viewer source code in conditional code on LL_WINDOWS that removes the new_name before calling LLFile::rename(). That seems very unnecessary.

Similarly while Posix remove() is documented to work on both files and directories, LLFile::remove() on Windows will only work on files since the used _wremove() does simply call DeleteFileW(), which can't delete empty directories. While this seems to be not something that has been worked around in code to much, it still is an inconsistency.

There are two possible fixes to LLFile::remove():
1: adopt Posix behaviour and find out if the path is for a directory and then call _wrmdir() and otherwise call _wremove() 
2: or make LLFile::remove() deviate from the Posix way and change the call in the non-Windows portion to unlink() to only work on files.

I personally tend to favor the first option, not at least because of the comment in the beginning of llfile.cpp about this trying to follow the Posix standard.

### Reproduction steps

look at llfile.cpp source
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4805</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @fmartian</code> (replace <code>20</code> with the amount, and <code>@fmartian</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

fmartian:
[...]

PR #4831, opened 2025-10-14
Improve LLFile to be consistent between Windows and Linux/Mac

fmartian:
LLFile::remove() will now remove both files and empty directories under Windows, just as its Posix counter part on other platforms

LLFile::rename() will rename the file even if there is already another file at the new location whose permissions allow it to be overwritten.

This allows to remove multiple remove() calls throughout the source code just before rename(), which were placed to circumvent the Windows limitation of rename()

It also prepends "\\\\?\\\" for long filenames to the filenames before passing them to the Windows functions, which removes the limitation of Windows file functions to filenames with not more than MAX_PATH characters. It also tries to sanitize filenames by replacing slashes with backslashes, since the above prefix skips certain path sanitation steps.

And it adds a function getattr() which on Windows can query the file attributes without some of the overhead of _wstat(). This function is then used in isdir() and isfile().

## Related Issues

- [X] **Feature/Enhancement:** #4805

### DIFF
--- a/indra/llcommon/llfile.cpp
+++ b/indra/llcommon/llfile.cpp
@@ -422,7 +422,7 @@ int LLFile::remove(const std::string& filename, int suppress_error)
     }
     else
     {
-        // get_filattr() failed and already set errno, preserve it for correct error reporting
+        // get_fileattr() failed and already set errno, preserve it for correct error reporting
     }
 #else
     int rc = ::remove(filename.c_str());
@@ -503,7 +503,7 @@ unsigned short LLFile::getattr(const std::string& filename, bool dontFollowSymLi
 #if LL_WINDOWS
     // _wstat64() is a bit heavyweight on Windows, use a more lightweight API
     // to just get the attributes
-    int            rc            = -1;
+    int rc = -1;
     std::wstring utf16filename = utf8path_to_wstring(filename);
     unsigned short st_mode = get_fileattr(utf16filename, dontFollowSymLink);
     if (st_mode)


indra/llcommon/llfile.cpp
```diff
@@ -275,28 +418,38 @@ int LLFile::remove(const std::string& filename, int supress_error)
 int LLFile::rename(const std::string& filename, const std::string& newname, int supress_error)
 {
 #if LL_WINDOWS
-    std::wstring utf16filename = ll_convert<std::wstring>(filename);
-    std::wstring utf16newname = ll_convert<std::wstring>(newname);
-    int rc = _wrename(utf16filename.c_str(),utf16newname.c_str());
+    // Posix rename() will gladly overwrite a file at newname if it exists, the Windows
+    // rename(), respectively _wrename(), will bark on that. Instead call directly the Windows
+    // API MoveFileEx() and use its flags to specify that overwrite is allowed.
```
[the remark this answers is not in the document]
fmartian: If you check in the source code you will find that there are 16 uses of LLFile:rename(). In at least 12 of them there is an (almost) direct LLFile::remove() (or in one case a LLAPRFile::remove()) call before, often in a conditional #if LL_WINDOWS precompiler clause (because of the Posix incompatibility in _wrename(). Most of them I removed in this change too, but I left some in because there exists a codepath that could result in remove() being called without rename() being called, so removing that remove() would change existing behavior (although likely not causing bad behavior as far as I could find).

The notable exceptions are:

llxfer_file.cpp (363): LLFile::remove() called a few lines before but might be called without LLFile::rename() being called because of the if clause in between.

llappviewer.cpp (2387): there is a remove() for the old_log_file about 50 lines above, and much logic in between that should not cause the dataflow to miss the rename() but not removing the file might cause some changes in the code in between. Likely not a problem but I did not try to assess every possible condition just yet, so left the remove in.

llappviewer.cpp (2402): Missed the remove() before this one and it could be removed too.

llappviewer.cpp (4258): only done once during cache migration, the new cache location should not exist at this point, why otherwise try to migrate it.

llconversationlog.cpp (466 + 470): creates a unique backup name to move old_file into, then moves current_file to old_file

lllogchat.cpp (771 + 778): Again creates a unique backup name to move old_file into, then moves current_file to old_file

lllogchat.cpp (288): only called if new_name is not a file (potential problem if there would be a directory with that name) and old file exists.

llsechandler_basic.cpp (1482): there is a remove() in the line just before but it is part of a compound condition and I was not sure if removing remove() would be fully safe here, so left it in.

There could be of course race conditions if running with multiple viewers at the same time, but they exist in the old situation just as much or even more, since the window of opportunity is larger.

Also if this change was causing trouble it would have certainly caused trouble in Linux and Mac for a long time already.

indra/llcommon/llfile.cpp
```diff
@@ -49,6 +48,86 @@ static std::string empty;
 // variants of strerror() to report errors.
 
 #if LL_WINDOWS
+namespace
+{
+    struct errentry
+    {
+        unsigned long oserr; // OS return value
```
akleshchev: I think comment should be a bit more blatant, like 'OS Windows system error code' and repeat that before the errtable, like 'Fill error translation table with Windows system error codes'. I know it's easy to figure out, but it took me a moment to realize what those values are.

fmartian: Fair point. It's obvious if you work regularly with Windows APIs but I can see your point. I was tempted to use Windows datatypes (notably DWORD in this case) but refrained from it seeing that this is carefully avoided in many places even if it is purely LL_WINDOWS only compilation.

akleshchev: Yeah, it's a crossplatform app, avoiding system specific 'solutions' helps to make it readable for everyone.

indra/llcommon/llfile.cpp
```diff
@@ -275,28 +418,38 @@ int LLFile::remove(const std::string& filename, int supress_error)
...
-    LLFILE* in = LLFile::fopen(from, "rb");     /* Flawfinder: ignore */
+    LLFILE* in = fopen(from, "rb");     /* Flawfinder: ignore */
     if (in)
     {
-        LLFILE* out = LLFile::fopen(to, "wb");      /* Flawfinder: ignore */
+        LLFILE* out = fopen(to, "wb");      /* Flawfinder: ignore */
         if (out)
         {
-            char buf[16384];        /* Flawfinder: ignore */
+            char buf[LLFILE_COPY_BUFFER_SIZE]; /* Flawfinder: ignore */
```
akleshchev: Just for the info: flawfinder hadn't been in use for over a decade, at least not viewer side. And there are other places where we allocate char/byte buffers with no 'ignore' and no issues so it's safe to just drop.

maxim-productengine (approved):

indra/llcommon/llfile.h
```diff
@@ -41,8 +41,9 @@ typedef FILE    LLFILE;
 #include <sys/stat.h>
 
 #if LL_WINDOWS
-// windows version of stat function and stat data structure are called _stat
-typedef struct _stat    llstat;
+// The Windows version of stat function and stat data structure are called _stat64
+// We use _stat64 here to support 64-bit st_size and time_t values
+typedef struct _stat64  llstat;
```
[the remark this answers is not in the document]
fmartian: This is not quite right. _stati64 is an alias to _stat64 when compiling for 64-bit and to _stat32i64 only when compiling for 32-bit. _stat is an alias to _stat64i32 on 64-bit and _stat32 on 32-bit Windows. Since we only compile for 64-bit we do not have to look at the 32-bit situation at all. So what Copilot is claiming would only be right for 32-bit compilation and also definitely not what we want here.

We do want both 64-bit st_size and 64-bit time_t values here if possible, since on Linux and Mac, those values in stat are nowadays all resolving to a long type and since we compile for 64-bit only and these systems are LP64, this means those values are 64-bit integers, so it's beneficial to be 64-bit for both since it matches the non-Windows situation and also supports 64-bit file sizes under Windows.

As far as potential problems with the current code, there are none as far as I can tell. In most cases LLFile:stat() is only used to detect if a file or directory exists. There are a few exceptions:

lllivefile.cpp (125): stores the st_mtime in the class private variable of type time_t but time_t is also 64-bit on 64-bit Windows. (64-bit compilation under Windows does not allow to have a 32-bit time_t).

llmodelloader.cpp (231): compares st_mtime of two stat structures.

llmodelloader.cpp (261): typecasts st_size to a S32 for further use

llappviewer.cpp (2356): compares st_ctime of two stat structures.

llfloaterpreference.cpp (2044): compares st_size to be > 0

akleshchev:
Please pull an update from develop, that should have a build fix for Mac.

Edit: Thank you, everything looks to be in order, once it finishes building, will merge.

PR #4877, opened 2025-10-22
Ensure final EOF in LLProcess test

marchcat:
This should unblock the ReadPipe "eof" integration test.


maxim-productengine (approved):

akleshchev (approved):

akleshchev:
Suggestion: add better logging. Like make and inherit from LLProcess a class that counts number of events, output that 'count'.

PR #4892, opened 2025-10-23
Implement querying the system certificate store on each platform

### LINKED ISSUE #4505: The viewer has 3 expired certs
### Environment

Second Life Release 7.1.15.15596336374 (64bit)
Windows and Mac

### Description

The viewer has 3 expired certs.

### Reproduction steps

1. Start the Release viewer 7.1.15.15596336374
2. Search the SecondLife.log for Certificate.

Observed: There are 3 Certificate Errors

```
2025-08-08T20:15:45Z WARNING #SECAPI# newview/llsecapi.cpp(168) LLCertException::LLCertException : Certificate Error: CertExpired
2025-08-08T20:15:45Z WARNING #SECAPI# newview/llsecapi.cpp(168) LLCertException::LLCertException : Certificate Error: CertExpired
2025-08-08T20:15:45Z WARNING #SECAPI# newview/llsecapi.cpp(168) LLCertException::LLCertException : Certificate Error: CertExpired
2025-08-08T20:15:45Z INFO #SECAPI# newview/llsechandler_basic.cpp(629) LLBasicCertificateStore::load_from_file : loaded 145 good certificates (rejected 3) from C:\\Program Files\\SecondLifeViewer\\ca-bundle.crt
```

Expected: We should keep these up to date.
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4505</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Dan-Linden</code> (replace <code>20</code> with the amount, and <code>@Dan-Linden</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

fmartian:
## Description

- Move the enumerations of the certificates into the LLBasicCertificateVector and implement a derived class LLSystemCertificateVector
- For each platform try to query the system certificate store and fall back to the application provided certification bundle if that fails
- Be more restrictive in issuing warning messages about rejected certificates since the system certificate store on Windows and Mac at least, generally also contains several expired certificates

This could eventually allow to not having to ship the application provided CA certificate bundle anymore and rely on the certificates from the Operating System, which are normally regularly updated.

## Related Issues

- [X] Also fixes (or at least relieves) #4505

## Checklist

- [X] I have provided a clear title and detailed description for this pull request.
- [X] The PR is linked to a relevant issue with sufficient context.
- [X] I have tested the changes locally and verified they work as intended.
- [X] Documentation has been updated if needed.
- [X] Any dependent changes have been merged and published in downstream modules
- [X] I have reviewed the [contributing guidelines](https://github.com/secondlife/viewer/blob/develop/CONTRIBUTING.md).


indra/newview/llsechandler_basic.cpp
```diff
@@ -561,90 +577,243 @@ LLPointer<LLCertificate> LLBasicCertificateVector::erase(iterator _iter)
...
+    }
+}
 
-LLBasicCertificateStore::~LLBasicCertificateStore()
+// Enumerate the certificates in the system CA store
+bool LLSystemCertificateVector::load_from_system(const std::string& storename, bool suppress_expire_warning)
 {
+    bool success = false;
+#if LL_WINDOWS
+    HCERTSTORE cert_store = CertOpenSystemStoreA(0, storename.empty() ? "ROOT" : storename.c_str());
```
Ansariel, line 702: Is there a reason to call the ANSI version explicitly instead of using `CertOpenSystemStore`?

fmartian: Actually yes. It avoids having to make a std::string to std:wstring conversion. Basically std::string is always an 8-bit string while the undecorated Win32 API may use wchar or char, depending on some "obscure" UNICODE preprocessor define. So if someone changes that define you end up with compiler errors. The undecorated Win32 API should only be used if you also pass TCHAR/LPTCHAR variables to them. Otherwise you create a potential type conversion error.

Since this string is anyhow only for special cases (needs more work to make this a somewhat useful parameter across all three platforms) and the various security stores on each platform have varying ideas about how they organize it (on my Mac only the kSecTrustSettingsDomainSystem really returns any significant number of certificates, it turned out to be a somewhat academic parameter. However I left it in as an option for now.

The idea was that by passing in different keywords, one could select a specific sort of keystore to query. On Windows this would by for instance ROOT, MY and CA, on Mac one of the three enums, on Linux ... oh well it's a mess! :-)
If we really want to use that eventually, an enum similar to what the Mac uses may be more prudent but more likely it just can go away eventually.

akleshchev: I suggest adding this as a comment in-code.

fmartian: MacOS isn't quite my expertise. But I see that in the test build it fails because of these:

Undefined symbols for architecture arm64:
  "_SecCertificateCopyData", referenced from:
      LLSystemCertificateVector::load_from_system(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, bool) in llsechandler_basic.o
  "_SecTrustSettingsCopyCertificates", referenced from:
      LLSystemCertificateVector::load_from_system(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, bool) in llsechandler_basic.o

Does this mean that these SecCertificate.. APIs have been depreciated, not implemented, forgotten by Apple or what for the ARM platform?

akleshchev: According to AI it might be because security framework isn't linked to the test in question.

akleshchev: It's probably not a good idea to load certs when running unit tests.

Check how viewer uses control variables, LLMachineID and LLCredential in indra\newview\tests\llsechandler_basic_test.cpp, you probably want to do something similar - move load_from_system into some namespace/class outside of llsechandler_basic then provide an empty on in the test.

P.S. Member's name like load_from_system does not match LLs coding standard.

fmartian: > It's probably not a good idea to load certs when running unit tests.
> 
> Check how viewer uses control variables, LLMachineID and LLCredential in indra\newview\tests\llsechandler_basic_test.cpp, you probably want to do something similar - move load_from_system into some namespace/class outside of llsechandler_basic then provide an empty on in the test.

Unless you intend to do some actual certificate verification it should not really matter aside from the (minor) runtime delay to actually query the certificate store. If the test machine does not have any valid certificates it would simply fallback to the previous solution, trying to load the application provided certificates and if that can't be found, it would load none.

But will see how it could be disabled for test builds.

> P.S. Member's name like load_from_system does not match LLs coding standard.

I was just trying to do it similar to the existing function load_from_file() in the already present code. Should that be changed?

fmartian: > According to AI it might be because security framework isn't linked to the test in question.

Does the test machine run on arm64 CPU? Why would it only complain during the arm64 and not the x86_64 linking? Building is supposed to be universal, isn't it?

akleshchev: It complains for both
```
Undefined symbols for architecture x86_64:
  "_SecCertificateCopyData", referenced from:
```
and
```
Undefined symbols for architecture arm64:
  "_SecCertificateCopyData", referenced from:
```

> I was just trying to do it similar to the existing function load_from_file() in the already present code.

It's preferable, but since it's already there feel free to keep it that way. I just noticed a 'weird' name, so pointed it out.

fmartian: Missed the x86_64 failure. Those logs are unwieldly long. :-(

What about an option to disable the system store with an optional parameter to the LLSecAPIBasicHandler() constructor? If this parameter is set to true, it instantiates the LLBasicCertificateVector with the existing certificate file rather than the derived LLSystemCertificateVector.
Although that won't solve the linking issue. Have to think about this a bit more.

RyeMutt: You'll need to add a `find_library(SECURITY_LIBRARY Security)` in Linking.cmake and add it to the list of linked libraries.

akleshchev: > You'll need to add a find_library(SECURITY_LIBRARY Security)

Thanks! I knew that it needed a framework or library, but I just don't see where viewer is adding it or how to do it.

> What about an option to disable the system store with an optional parameter to the LLSecAPIBasicHandler() constructor?

It's a linking error, not a runtime one.

fmartian: > > You'll need to add a find_library(SECURITY_LIBRARY Security)
> 
> Thanks! I knew that it needed a framework or library, but I just don't see where viewer is adding it or how to do it.

Already found it. It is in indra/cmake/Linking.cmake where it needs to be added. Also needs a similar addition for crypt32 in the Windows case. Wonder why it compiled and linked for me on both Mac and Windows without that.

> > What about an option to disable the system store with an optional parameter to the LLSecAPIBasicHandler() constructor?
> 
> It's a linking error, not a runtime one.

Yes I know. But my concern is more about the fact that a half working system store (not sure how the test systems are exactly setup and if they are a fully installed OS image) might actually make the tests fail for some reason as they are working with specific pre configured certificate resources.

akleshchev: > But my concern is more about the fact that a half working system store

It only uses old LLBasicCertificateVector's constructor so that part should be fine. And if not the right way might be to update test<5> to account for extra certs.

akleshchev:
@marchcat Please take a look at this, you are far more familiar with certs.

marchcat:
Thanks for the contribution!

This mostly LGTM. However, I’d address at least these two Copilot comments:
- handle the possible nullptr returned by getenv();
- return full paths from getFilesInDir().

marchcat:
And we still need to fix that macOS build failure by adding something like
```cmake
if(APPLE)
    find_library(SECURITY_FRAMEWORK Security)
    target_link_libraries(<your_target> PRIVATE ${SECURITY_FRAMEWORK})
endif()
```
to CMakeLists.txt (for every test target).

fmartian:
> And we still need to fix that macOS build failure by adding something like
> 
> ```cmake
> if(APPLE)
>     find_library(SECURITY_FRAMEWORK Security)
>     target_link_libraries(<your_target> PRIVATE ${SECURITY_FRAMEWORK})
> endif()
> ```
> 
> to CMakeLists.txt (for every test target).

I want to review this but am still working on the other LLFile, LLFileSystem part some. I was working on getting minimal changes added to LLFile, and then update LLFileSystem and any callers thereof to do the right thing. Once that is fixed the full LLFile change can be committed too, without causing anymore trouble.

This update was triggered because I did a synch of the relevant branch with latest changes from develop just to be up to date. Still will need to make the CMake changes as you correctly note, and will look at the Copilot comments. Some is cosmetic but nevertheless has some merits.

One additional thing I want to look at is the various certificate locations such as "ROOT", "CA" and "MY" under Windows and how they relate to each other. It may be useful to combine some of them in a similar way as on the Mac through the SecTrustSettingsDomain enum. But I need to look at the actual data returned for these to make a useful choice. Windows documentation is pretty vague about what is what.

akleshchev:
> I want to review this but am still working on the other LLFile, LLFileSystem part some.

MacOS build failure is still there. Should this be converted to draft? I think LLFile changes went to linux branch?

fmartian:
> > I want to review this but am still working on the other LLFile, LLFileSystem part some.
> 
> MacOS build failure is still there. Should this be converted to draft? I think LLFile changes went to linux branch?

The LLFile work was taken over by someone and indeed moved to the Linux branch, I moved to other non-SL work and completely forgot about this PR. I also heard rumors that custom certificate handling was going to be removed anyways. If this is still a desired functionality I can put in some work next week to make it fully compilable.

akleshchev:
> I also heard rumors that custom certificate handling was going to be removed anyways.

Sorry, I'm out of the loop on that one. And I'm not sure whom to ask about this either. I assume that we do want to use OS's cert store in some form when possible and practical.

@Geenz Any ideas on this one?

indra/newview/llsechandler_basic.cpp
```diff
@@ -561,90 +577,243 @@ LLPointer<LLCertificate> LLBasicCertificateVector::erase(iterator _iter)
...
+        {
+            CFIndex idx, count = CFArrayGetCount(cfCerts);
+            for (idx = 0; idx < count; idx++)
+            {
+                SecCertificateRef cfCert = (SecCertificateRef)CFArrayGetValueAtIndex(cfCerts, idx);
+                CFDataRef der = SecCertificateCopyData(cfCert);
+                if (der)
+                {
+                    const unsigned char* cert = reinterpret_cast<const unsigned char*>(CFDataGetBytePtr(der));
+                    success |= verify_and_add(cert, CFDataGetLength(der), suppress_expire_warning);
```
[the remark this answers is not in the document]
fmartian: Updated with an explicit test of the trust result for each certificate. Only adds a certificate to the store if it is either explicitly user trusted or not rejected by the user,

akleshchev:
Any suggestions for a test plan?

indra/llfilesystem/lldir.cpp
```diff
@@ -106,11 +105,23 @@ std::vector<std::string> LLDir::getFilesInDir(const std::string &dirname)
             {
                 v.push_back(dir_itr->path().filename().string());
             }
+            else if (includeExternalSymlinks && std::filesystem::is_symlink(dir_itr->status()))
```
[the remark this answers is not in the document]
fmartian: That is an oversight and for sure a valid observation.

indra/newview/llsechandler_basic.h
```diff
@@ -145,31 +143,61 @@ class LLBasicCertificateVector : virtual public LLCertificateVector
...
     virtual LLPointer<LLCertificate> erase(iterator _iter);
 
 protected:
-    std::vector<LLPointer<LLCertificate> >mCerts;
+    // verify the certificate and if it is valid and not expired add it to this certificate vector
+    int verify_and_add(const unsigned char* cert_bytes, long length, bool suppress_expire_warning = false);
+    int verify_and_add(X509* cert, bool suppress_expire_warning = false);
+    // load certificates from a file or directory
+    int load_from(const std::filesystem::path& path, bool suppress_expire_warning = false);
+    int load_from(const std::string& path, bool suppress_expire_warning = false)
```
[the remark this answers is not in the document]
fmartian: Good catch

fmartian:
> Any suggestions for a test plan?

Good question. In the new implementation it will be simply used when the test creates a trust store unless the system does not have any valid system certificates, in which case it falls back to the old application bundled certificate store file. It's probably safe to assume that any test runner inside the github infrastructure runs on a system with a valid system certificate store as they operate in the cloud and would be very vulnerable otherwise. So basically if the test reveals that there were certificates loaded it worked!

Trying to refer to specific certificates during the test to check if they are present sounds like an extremely brittle way of doing things. Anything from expired certificates to potential localization differences could throw such a test off the rails and it would end up being a maintenance nightmare.

We could add a test that tries to only load system certificates without the fallback, to see if we get any valid certificates that way. But anything beyond that seems to create more trouble in the long run.

indra/newview/llsechandler_basic.cpp
```diff
@@ -561,103 +577,384 @@ LLPointer<LLCertificate> LLBasicCertificateVector::erase(iterator _iter)
...
+int LLSystemCertificateVector::load_from_system(bool suppress_expire_warning)
 {
     int loaded = 0;
-    int rejected = 0;
+#if LL_WINDOWS
+    const static LPCSTR domains[] = {
+        "MY",
+        "CA",
+        "ROOT",
+    };
```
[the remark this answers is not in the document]
fmartian: Interesting observation. Not invalid but at the same time also a bit restrictive. I also think that if we do that we would have to also consider removing the kSecTrustSettingsDomainUser in the Mac system store enumeration. While the implementation of the two stores is not exactly the same it could lead to a similar issue there.

fmartian:
> Any suggestions for a test plan?

Ok, looking through the whole test again I see that it does not really try to instantiate a certificate store at all but only is concerned about creating a static store from source code inlined certificates and then do all kind of querying and manipulations on them to test that the certificate handling itself is correct.

What are the chances that the different test runner systems on the github CI infrastructure have a valid system certificate store installed? Should we try to add a test that tries to instantiate a LLSystemCertificateVector::LLSystemCertificateVector() object and then see if it got any valid certificates? We would probably need an additional optional parameter to this constructor to prevent it to try to fallback to the application bundled CA bundle. Is that CA bundle even present in the test environment?

indra/newview/llsechandler_basic.cpp
```diff
@@ -561,103 +577,402 @@ LLPointer<LLCertificate> LLBasicCertificateVector::erase(iterator _iter)
...
+// already exists in the store (based on CERT_SUBJECT_KEY_IDENTIFIER), it will be ignored
+int LLBasicCertificateVector::verify_and_add(X509* cert_x509, bool suppress_expire_warning)
 {
-    int loaded = 0;
-    int rejected = 0;
+    try
+    {
+        LLPointer<LLBasicCertificate> new_cert(new LLBasicCertificate(cert_x509));
+        LLSD                          validation_params;
+        _validateCert(VALIDATION_POLICY_TIME, new_cert, validation_params, 0, suppress_expire_warning);
```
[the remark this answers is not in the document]
fmartian: Shouldn't this be rather inside the _validateCert()?

PR #4893, opened 2025-10-24
Async Inventory Loading Implementation

### LINKED ISSUE #4972: Async Inventory: Improve Login Performance via Asynchronous Loading
_What does this feature/enhancement do? Why is it needed?_

This sub-issue coordinates validation and monitoring for the asynchronous inventory loading proposed in PR #4893. By moving inventory skeleton hydration and essentials fetch to an asynchronous workflow, login times are expected to improve, especially for residents with large inventories.

Key areas of focus:
- Confirm login speed increases by asynchronously loading inventory data
- Ensure user-facing inventory is accurate and complete post-login
- Test large inventory scenarios (100k+ items), edge cases, and cache integrity
- Follow regression scenarios to confirm no issues in folder hierarchy or item order

Reference: [Async Inventory Loading Implementation PR #4893](https://github.com/secondlife/viewer/pull/4893)

Parent Issue: [Quality of Life (#4969)](https://github.com/secondlife/viewer/issues/4969)
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4972</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Geenz</code> (replace <code>20</code> with the amount, and <code>@Geenz</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

marchcat:
### Summary:

Introduces asynchronous inventory skeleton loading via AIS to accelerate login (up to 12s faster for 100k+ item inventories) by hydrating from cache, fetching essentials on-demand, and throttling UI updates. Adds debug settings for concurrency/timeouts, fixes cache persistence/dehydration, and optimizes UI with lazy folder widget creation (O(n) parent-child mapping via caching).

### Testing: 

QA-only `ForceAsyncInventorySkeleton` toggle (enabled by default); monitor login times/cache integrity. No breaking changes.

Additional [test](https://github.com/secondlife/viewer/pull/4893#discussion_r2461587646):

1. Log in on machine A. Log out
2. Wear an outfit with multiple items of the same type on machine B. Log out.
3. Log in on machine A. Change outfit order. Log out.
4. Log in on machine B. Confirm order of items.


indra/llui/llpanel.cpp
```diff
@@ -489,58 +489,70 @@ bool LLPanel::initPanelXML(LLXMLNodePtr node, LLView *parent, LLXMLNodePtr outpu
         LL_RECORD_BLOCK_TIME(FTM_PANEL_SETUP);
 
         LLXMLNodePtr referenced_xml;
-        std::string xml_filename = mXMLFilename;
+        const std::string& xml_filename = mXMLFilename;
```
Ansariel: Holding a reference to `mXMLFilename` seems pointless as it's only used in the following if-check now.

indra/newview/llstartup.cpp
```diff
@@ -206,11 +215,704 @@
...
+#include <set>
+
 
 #if LL_WINDOWS
 #include "lldxhardware.h"
 #endif
 
+namespace
+{
+class LLAsyncInventorySkeletonLoader
```
Ansariel: Move this class into its own file maybe?

indra/newview/llstartup.cpp
```diff
@@ -206,11 +215,704 @@
...
+    LLUUID cof_id = LLAppearanceMgr::instance().getCOF();
+    if (cof_id.notNull()
+        && mFetchedCategories.count(cof_id) == 0
+        && mQueuedCategories.count(cof_id) == 0
+        && mActiveFetches.count(cof_id) == 0)
+    {
+        mSawCurrentOutfitFolder = true;
+        LLViewerInventoryCategory* cof = gInventory.getCategory(cof_id);
+        const S32 cached_version = gInventory.getCachedCategoryVersion(cof_id);
+        if (isCategoryUpToDate(cof, cached_version))
```
akleshchev: Please add to test plan:
- Log in on machine A. Log out
- Wear an outfit with multiple items of the same type on machine B. Log out.
- Log in on machine A. Change outfit order. Log out.
- Log in on machine B. Confirm order of items.

indra/newview/llinventorymodel.cpp
```diff
@@ -3188,61 +3582,133 @@ void LLInventoryModel::buildParentChildMap()
...
-            }
-            else
-            {
-                mIsAgentInvUsable = true;
+                LLPointer<LLInventoryValidationInfo> validation_info = validate();
+                if (validation_info->mFatalErrorCount > 0)
+                {
+                    // Fatal inventory error. Will not be able to engage in many inventory operations.
+                    // This should be followed by an error dialog leading to logout.
+                    LL_WARNS("AsyncInventory") << "Fatal errors were found in validate(): unable to initialize inventory! "
```
rohvani: There's a few places where I replaced `LL_WARNS("Inventory")` with `LL_WARNS("AsyncInventory")`, even for regular inventory operations -- worth checking them over.

indra/newview/lllogininstance.cpp
```diff
@@ -152,22 +155,71 @@ LLSD LLLoginInstance::getResponse()
...
 //#define LL_MINIMIAL_REQUESTED_OPTIONS
     LLSD requested_options;
     // *Note: this is where gUserAuth used to be created.
+    mForceAsyncSkeleton = gSavedSettings.getBOOL("ForceAsyncInventorySkeleton");
+    if (mForceAsyncSkeleton)
+    {
+        mAsyncSkeletonSuppressed = false;
+    }
+    mSupportsAsyncSkeleton = !mAsyncSkeletonSuppressed || mForceAsyncSkeleton;
+
```
rohvani, line 188: This is currently gated behind the `ForceAsyncInventorySkeleton` debug setting -- anyone testing should make sure this is set to 1 (or we should default it).

indra/newview/llasyncinventoryskeletonloader.cpp
```diff
@@ -0,0 +1,672 @@
...
+    {
+        FetchRequest request = mFetchQueue.front();
+        mFetchQueue.pop_front();
+
+        AISAPI::completion_t cb = [this, request](const LLUUID& response_id)
+        {
+            handleFetchComplete(request.mCategoryId, response_id);
+        };
+
+        LL_DEBUGS("AppInit") << "Async skeleton loader requesting AIS children for "
```
rohvani, line 237: Should this be logged under `AsyncInventory` instead of `AppInit`?

indra/newview/llasyncinventoryskeletonloader.cpp
```diff
@@ -0,0 +1,672 @@
...
+        mActiveFetches.emplace(request.mCategoryId, request);
+    }
+}
+
+void LLAsyncInventorySkeletonLoader::handleFetchComplete(const LLUUID& request_id, const LLUUID& response_id)
+{
+    auto active_it = mActiveFetches.find(request_id);
+    if (active_it == mActiveFetches.end())
+    {
+        LL_WARNS("AppInit") << "Async skeleton loader received unexpected completion for " << request_id << LL_ENDL;
```
rohvani, line 258: Should this be logged under `AsyncInventory` instead of `AppInit`?

There's a few other places here in the async loader where we aren't logging under `AsyncInventory`.

Not a blocker, just something I noticed while doing another review.

marchcat: Thank you, I'll double check these as well.

akleshchev (approved):

indra/llui/llview.cpp
```diff
@@ -253,6 +253,28 @@ const std::string& LLView::getName() const
...
+    if (name == mName)
+    {
+        return;
+    }
+
+    LLView* parent = mParentView;
+
+    if (parent && !mName.empty())
+    {
+        parent->mChildNameCache.erase(mName);
```
akleshchev, line 267: I don't think names are unique. A bunch of controls will just be named "name" and this has a chance to erase a wrong pointer.

indra/llcommon/llsingleton.h
```diff
@@ -524,13 +524,16 @@ class LLSingleton : public LLSingletonBase
...
-                // record the dependency, if any: check if we got here from another
+                // record the dependency: check if we got here from another
                 // LLSingleton's constructor or initSingleton() method
                 capture_dependency(lk->mInstance);
                 return lk->mInstance;
 
+            case INITIALIZED:
+                // normal subsequent calls - skip capture_dependency() for performance
+                // dependencies are only tracked during initialization
+                return lk->mInstance;
```
akleshchev, line 535: I think this goes against the purpose of singletons' dependency capture mechanics.

indra/newview/llasyncinventoryskeletonloader.cpp
```diff
@@ -0,0 +1,672 @@
...
+        gIdleCallbacks.addFunction(&LLAsyncInventorySkeletonLoader::idleCallback, this);
+        mIdleRegistered = true;
+    }
+}
+
+void LLAsyncInventorySkeletonLoader::removeIdleCallback()
+{
+    if (mIdleRegistered)
+    {
+        gIdleCallbacks.deleteFunction(&LLAsyncInventorySkeletonLoader::idleCallback, this);
```
akleshchev, line 142: I think gIdleCallbacks needs to be cleared in destructor in case user tries to quit early

indra/newview/llasyncinventoryskeletonloader.h
```diff
@@ -0,0 +1,129 @@
...
+#include <map>
+#include <set>
+#include <string>
+
+#include <boost/signals2/connection.hpp>
+
+class LLViewerInventoryCategory;
+class LLViewerRegion;
+
+class LLAsyncInventorySkeletonLoader
```
akleshchev, line 45: This class appears to be doing the same thing as LLInventoryModelBackgroundFetch, but slower and takes cache into account?

akleshchev (changes requested):
Viewer sometimes freezes for multiple seconds while spamming following messages:
LLInventoryModel::findCategoryUUIDForTypeInRoot : Tried to find folder, type 23 but category does not exist

akleshchev:
I suspect that when user searches for something, background fetch intiates and loads everything from net even if those items were supposed to be loaded from cache.

akleshchev:
I feel like this bundle of optimizations needs to be split into multiple separate ones and reviewer separately.
Ex: LLView name cache thing looks promissing if incomplete, but it should not be bundled with inventory changes.

indra/newview/llinventorypanel.cpp
```diff
@@ -1044,17 +1044,25 @@ void LLInventoryPanel::initializeViews(F64 max_time)
...
     }
     else
     {
         // Default case: always add "My Inventory" root first, "Library" root second
-        // If we run out of time, this still should create root folders
-        buildNewViews(gInventory.getRootFolderID());        // My Inventory
-        buildNewViews(gInventory.getLibraryRootFolderID()); // Library
+        // Build only root widgets - children will load on-demand when expanded
+        LLInventoryObject const* my_inv = mInventory->getObject(gInventory.getRootFolderID());
+        buildNewViews(gInventory.getRootFolderID(), my_inv, nullptr, BUILD_NO_CHILDREN);
```
akleshchev, line 1062: This is likely the cause behind 'Tried to find folder' spam, system folders weren't created.

Suggestion: use BUILD_ONE_FOLDER and load calling cards folder manually.

indra/newview/llasyncinventoryskeletonloader.h
marchcat: LLAsyncInventorySkeletonLoader serves a different lifecycle phase than LLInventoryModelBackgroundFetch. Background fetch runs post-login to gradually fill in the full inventory tree. The async skeleton loader runs *during* login to hydrate from cache, compare versions, and fetch only changed essential folders via AIS before the viewer considers inventory "ready." They could potentially be unified, but their concerns are distinct enough that merging them would complicate both.

marchcat:
Every file in this PR is directly related to async inventory loading. The three files that may look "unrelated" (llview.cpp/h, llpanel.cpp) are explained below.

#### llview.cpp/h: O(1) child view lookup cache

The existing findChildView() does a linear scan over mChildList. With synchronous loading, all inventory widgets are built in one batch at login, so this cost is paid once. With async loading, widgets are created lazily when folders are expanded (BUILD_ONE_FOLDER mode). This means findChildView() is called repeatedly during normal browsing, not just at startup — the linear scan became a bottleneck. The fix adds an unordered_map mChildNameCache to LLView for O(1) lookups, maintained in addChild(), removeChild() and setName().

#### llpanel.cpp: reduce repeated singleton/accessor overhead in initPanelXML()

initPanelXML() runs for every panel constructed from XML, including every inventory folder widget. With lazy creation, it fires much more often during the session. The changes cache three values that were re-fetched on every call: LLUICtrlFactory::getInstance(), node->getName() and child_registry_t::instance(). Also eliminates a redundant local string copy in favor of using mXMLFilename directly.

#### Why in this PR

All three files address performance issues caused by the lazy-creation approach. Without the cache, expanding a folder with hundreds of children would do O(n^2) string comparisons. The llpanel changes reduce per-widget overhead that compounds when widgets are created on-demand. They they are not unrelated work.

akleshchev:
> llview.cpp/h: O(1) child view lookup cache

My problems with this solution:
- It affects everything, not just inventory, names aren't unique and normally we do only one request in postBuild. With this change we keep the cache effectively forever after a first and only getChild request (we do have places that do repeated getChild, but the solution is to stop doing that). At a minimum this should be moved into LLFolderViewItem or whatever class it's used in, not be widespread.
- Some inventories are already massive, with a huge memory footprint, and we have multiple instances in the background in any moment. If this cashes all those names, the footprint will be massive and viewer already is heavy on ram.

marchcat:
You're right, thanks!

After tracing all 31 findChildView() call sites, none of them operate on LLFolderViewFolder or LLFolderView instances. The inventory code paths use direct mItemMap lookups by UUID, not name-based findChildView. So I removed the mChildNameCache from LLView entirely (d4a557a, in #5746).

The llpanel.cpp changes (caching getInstance/getName/instance locally in initPanelXML) are kept – those are local variable optimizations with zero blast radius beyond the function scope.

akleshchev:
> llpanel.cpp: reduce repeated singleton/accessor overhead in initPanelXML()

I agree that it is a good change, but not specific to async inventory.
In case of inventory it would be more effective to load inventory item's xml into memory and keep it there or to avoid parsing entirely and have some prototype params in memory, ready to be copied. We need it a lot and often so it would be justified and much faster

marchcat:
The prototype/template idea (parse inventory item XML once, clone instead of re-parse) is interesting and would be a big win. That's a separate effort though, let's do it as a follow-up.

akleshchev:
> The prototype/template idea (parse inventory item XML once, clone instead of re-parse) is interesting and would be a big win.

I'm planing to experiment on that today as a separate work.

akleshchev:
@marchcat inventory items do not appear to load from xml. Ex: createFolderViewFolder generates an item from code. initPanelXML does not get triggered and is not relevant to inventory generation, only to the initial inventory panel which is not relevant to async loading.

marchcat:
I looked deeper into it and found that inventory folder/item widgets inherit from LLView, not LLPanel, so they skip initPanelXML entirely. They use LLUICtrlFactory::create<T>(params) which fills from mParamDefaultsMap. XML parsing happens once when LLInventoryPanel loads; each widget copies from pre-parsed Params. So it's already doing the caching actually, no reparsing per widget.

marchcat:
Right. We can revert it since it's not a related change. But since we have it and it's a valid optimization with zero blast radius (just local variable caching), might as well keep it.

akleshchev:
> llpanel.cpp: reduce repeated singleton/accessor overhead in initPanelXML()

Please cut this out and move to a separate PR. This is good, but does not appear to be relevant. PR is already too large, those extra changes makes it hard to review for both human and AI.
Where did llfloatersettingsdebug.cpp‎ and llcontrol changes come from and why are they here?

At this point it's better to reform this PR entirely, rebased and with no unrelated changes. Sorry, but this looks like a mess.

indra/newview/llstartup.cpp
```diff
@@ -192,7 +198,9 @@
 #include "llagentlanguage.h"
 #include "llwearable.h"
 #include "llinventorybridge.h"
+#include "llfoldertype.h"
 #include "llappearancemgr.h"
```
akleshchev, line 202: Why are llfoldertype and boost/bind needed here?

indra/newview/llviewertexturelist.cpp
```diff
@@ -1014,6 +1014,38 @@ void LLViewerTextureList::updateImageDecodePriority(LLViewerFetchedTexture* imag
...
+            }
+
+            static LLCachedControl<LLVector4> channel_priority(gSavedSettings, "TextureChannelPriority",
+                LLVector4(10.0f, 20.0f, 40.0f, 20.0f));
+            F32 factor = llmax(channel_priority().mV[priority_channel], 0.1f);
+            if (factor != 1.0f)
+            {
+                imagep->mMaxVirtualSize /= factor;
+            }
+        }
```
akleshchev, line 1048: Appears to not be releavant to inventory in any way.

marchcat: Yep, that was due to wrong merge target.

indra/newview/llstartup.cpp
marchcat: Both are directly used in llstartup.cpp. They were previously included transitively through other headers, so including them reduces implicit dependencies.

akleshchev (approved):

marchcat:
Moved that optimization part into a dedicated PR (#5835).

While the other changes may be squashed into a single commit, it's probably worth keeping the commit log as it is here.

akleshchev:
>  it's probably worth keeping the commit log as it is here.

No it is not. It will cause merge conflicts or even erase those separate changes, since a revert from this pull request will be applied later. At least a rebase is needed.

marchcat:
Well, it won't cause any conflicts if we just merge this PR as it is. The "Merge" and "Squash" buttons are green.
Alternatively, we can use this "clean" PR: #5839.

marchcat:
@Geenz which one should we merge? #5839 is clean, but this one is "canonical" with all history and authorship attribution.

akleshchev:
I'm reviewing the clean one. It's a lot easier on the eyes.

Dzonatas:
I read this over and over, but I think the expected outcome is a valid request done with an OOP paradigm or a generic synopsis. I'm arguing about the parameters and alignment at this point. I mean, is it done here? With the delay implemented, that solves a lot. The profiler returns ~40% (where the idle thread needs to be). My C# version of the inventory pulls everything very quickly. It is faster and more uniform than C++, so I'll skip those details.

My suggestion is clamping it and reiterate oprofile (Linux) for the best return. I have a reason to test some of this under winehq... "oh my! Look at the size of that cache!" It's officially 32 GB now.

PR #4910, opened 2025-10-29
#3649 Incorrect camera aspec ratio

### LINKED ISSUE #3649: Incorrect aspect ratio used in camera setup in two places
### Environment

7.X viewers

### Description

In SL, the aspect ratio is always width/height
This is consistently applied throughout the code base with the exception of tow recently added situations
llterrainpaintmap.cpp (`bool LLTerrainPaintMap::bakeHeightNoiseIntoPBRPaintMapRGB(const LLViewerRegion& region, LLViewerTexture& tex)`)
and
llgltfmaterialpreviewmgr.cpp (`bool LLTerrainPaintMap::bakeHeightNoiseIntoPBRPaintMapRGB(const LLViewerRegion& region, LLViewerTexture& tex)`) 

where height/width is used.

In both these cases it is almost certainly 1:1 ratio anyway due to the fact they are working directly on textures but it would make sense to clean these up just in case there are future cases where that assumption is no longer true

### Reproduction steps

open the module in question
search for camera->setAspect
note the inverted ratio.

akleshchev:
Beq pointed out that everything else uses width/height ratio and I see no reason why this should be different.

### DIFF
--- a/indra/newview/llgltfmaterialpreviewmgr.cpp
+++ b/indra/newview/llgltfmaterialpreviewmgr.cpp
@@ -462,7 +462,7 @@ bool LLGLTFPreviewTexture::render()
     // Set up camera and viewport
     const LLVector3 origin(0.0, 0.0, 0.0);
     camera.lookAt(origin, object_position);
-    camera.setAspect((F32)(mFullHeight / mFullWidth));
+    camera.setAspect((F32)(mFullWidth / mFullHeight));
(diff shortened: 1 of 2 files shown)


Geenz (approved):
Seems fine.

marchcat (approved):

cosmic-linden (commented):
Width over height is the behavior for `getWorldViewAspectRatio`, so that's probably correct.

PR #4927, opened 2025-11-04
Linux support

Geenz:
Linux support for the viewer.  Still very much a WIP.

Please target develop-linux if you'd like to contribute to linux support.  We're trying to keep develop more or less clean until we've shipped 2025.08.


indra/newview/llfilepicker.cpp
```diff
@@ -167,7 +167,451 @@ void LLFilePicker::reset()
...
+        case LLFilePicker::FFLOAD_XML:
+            filter_vec.push_back({ "XML files (*.xml)", "xml" });
+            break;
+        case LLFilePicker::FFLOAD_SLOBJECT:
+            filter_vec.push_back({ "Objects (*.slobject)", "slobject" });
+            break;
+        case LLFilePicker::FFLOAD_RAW:
+            filter_vec.push_back({ "RAW files (*.raw)", "raw" });
+            break;
+        case LLFilePicker::FFLOAD_MODEL:
```
niciphilly, line 212: I noticed that I can only upload DAE files for models on Linux.
So I think this isn't quite correct. The Windows and Mac implementation also includes `*.gtlf` and `*.glb` files for `LLFilePicker::FFLOAD_MODEL`. I guess it's missing here, too.

indra/newview/llfilepicker.cpp
```diff
@@ -167,7 +167,451 @@ void LLFilePicker::reset()
...
+        case LLFilePicker::FFLOAD_IMAGE:
+            filter_vec.push_back({ "Images (*.tga; *.bmp; *.jpg; *.jpeg; *.png)", "tga;bmp;jpg;jpeg;png" });
+            break;
+        case LLFilePicker::FFLOAD_ANIM:
+            filter_vec.push_back({ "Animations (*.bvh; *.anim)", "bvh;anim" });
+            break;
+        case LLFilePicker::FFLOAD_GLTF:
+            filter_vec.push_back({ "glTF (*.gltf; *.glb)", "gltf;glb" });
+            break;
+        case LLFilePicker::FFLOAD_COLLADA:
```
niciphilly, line 200: Couldn't `LLFilePicker::FFLOAD_COLLADA` be dropped completely? I couldn't find any usage besides other `case` blocks anywhere.

indra/newview/linux_tools/refresh_desktop_app_entry.sh
```diff
@@ -13,18 +13,25 @@ function install_desktop_entry()
     local desktop_entry="\
 [Desktop Entry]\n\
 Name=Second Life\n\
+GenericName=Second Life Viewer\n\
 Comment=Client for the On-line Virtual World, Second Life\n\
-Exec=${installation_prefix}/secondlife\n\
+Path=${installation_prefix}\n\
+Exec=secondlife\n\
```
niciphilly: This doesn't work on Plasma. It complains `Could not find the program 'secondlife'`. And according to https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html `Exec` should be an absolute path or the executable needs to be in `$PATH`. `Path` only sets the working directory for the application but isn't used to lookup the executable.

Quinn-Elara: This is incorrect. Because the desktop entry is created from a bash script, the variable (represented by the $ sign) is substituted for the actual (absolute) path.

niciphilly: Yes, it is substituted. But not in the `Exec` line anymore. Right now it results in a desktop file like:
```
Path=/home/nici/installation/dir
Exec=secondlife
```
And this would require that I add `/home/nici/installation/dir` to $PATH. The version before the change was correct:
```
Exec=/home/nici/installation/dir/secondlife
```

Quinn-Elara: In which case, this is likely a difference in how Plasma and Gnome works (and likely different in COSMIC, et al.)
Seemingly, GNOME will interpret the PATH variable as the launch directory (and thus, the working directory), and if what you're saying is correct, then KDE is doing this another way. Easily fixed, will likely PR some point next week.

niciphilly: Yes, it might be something like that. The spec says absolute path or look up with `$PATH`. I guess Gnome does it a bit more relaxed and Plasma sticks to the letters.

RyeMutt:
Additional fixes in #4954

PR #4943, opened 2025-11-07
[Draft] Nearby chat voice moderation

maxim-productengine:
New menu options for Nearby chat moderation are added.
Also new notification will be displayed after being muted/unmuted in Nearby chat or clicking 'voice' button while being muted.


indra/newview/llnearbyvoicemoderation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    {
+        const std::string cap_url = getCapUrlFromRegion(avatar->getRegion());
+        if (cap_url.length())
+        {
+            const std::string operand = mute ? "mute" : "unmute";
+
+            LLSD body;
+            body["operand"] = operand;
+            body["agent_id"] = agent_id;
+            body["moderator_id"] = gAgent.getID();
```
AndrewMeadows: There is no need to send the `moderator_id`: remove it.  The server ignores that field and instead grabs the `agent_id` directly from the connection context.

indra/newview/llnearbyvoicemoderation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    LLViewerRegion* region = gAgent.getRegion();
+
+    const std::string cap_url = getCapUrlFromRegion(region);
+    if (cap_url.length())
+    {
+        const std::string operand = mute ? "mute_all" : "unmute_all";
+
+        LLSD body;
+        body["operand"] = operand;
+        body["moderator_id"] = gAgent.getID();
```
AndrewMeadows: Remove the `moderator_id` field.

indra/newview/llfloaterimcontainer.cpp
```diff
@@ -2267,6 +2337,31 @@ LLSpeaker * LLFloaterIMContainer::getSpeakerOfSelectedParticipant(LLSpeakerMgr *
     return speaker_managerp->findSpeaker(participant_itemp->getUUID());
 }
 
+bool LLFloaterIMContainer::isNearbyChatSpeakerSelected()
+{
+    LLFolderViewItem *selectedItem = mConversationsRoot->getCurSelectedItem();
+    if (NULL == selectedItem)
```
AndrewMeadows: Has the viewer switched yet to using `nullptr` (c++-ism) instead of `NULL` (c-ism)?

AndrewMeadows (changes requested):

indra/newview/llfloaterimcontainer.cpp
maxim-productengine: In this case it was probably just copy-paste from another place. It's more common just to do `if (!selectedItem) `check.
Speaking of `NULL `vs `nullptr` - there we no switch, but it makes sense to use `nullptr` at least in the new code.

indra/newview/llnearbyvoicemoderation.cpp
maxim-productengine: Will do.

maxim-productengine:
Separate project build will be used, so closing this PR.

Issue #4932, opened 2025-11-04
@ing someone that has a TM symbol in their display name will print their name twice


Dan-Linden:
### Environment

Second Life Release 7.2.2.18475198968 (64bit)
Release Notes

You are at 117.3, 88.9, 21.8 in By Design located at simhost-0caa2e4080ef1beab.agni
SLURL: http://maps.secondlife.com/secondlife/By%20Design/117/89/22
(global coordinates 261,493.0, 246,873.0, 21.8)
Second Life Preflight 2025-09-26.18046863877
Release Notes

CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz (3491.92 MHz)
Memory: 32610 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.6456)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce GTX 1080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.6590
OpenGL Version: 4.6.0 NVIDIA 565.90

Window size: 1831x1484
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 8192MB
Disk cache: Max size 11468.8 MB (10.9% used)

J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.24.0.202510081738
  CEF: 139.0.40+g465474a+chromium-139.0.7258.139
  Chromium: 139.0.7258.139
LibVLC Version: 3.0.21
Voice Server Version: Not Connected
Packets Lost: 0/5,059 (0.0%)
November 04 2025 13:04:30

### Description

@ing someone that has a TM symbol in their display name will print their name twice.

### Reproduction steps

1. Set your display name to Name™
2. Relog so you see your new display name.
3. In chat, type @ and choose your name

Issue 1 observed: your cursor is between your display name and the real name. 
Expected: The cursor should be at the end of the line.

4. Move your cursor to the end of the line, type hi, and hit enter.

Issue 2 observed: The chat output is: @Name™ (testerd1.tester)™@Name™ (testerd1.tester) hi
Expected: @Name™ (testerd1.tester) hi

[...]

PR #4953, opened 2025-11-10
#4932 fix @mention being displayed twice if the name has a TM symbol

maxim-productengine:
Emoji (TM) shouldn't split the segment with the mention.


akleshchev (approved):
Approving, but I'm not sure if that's sufficient. I feel like a 'mention' should either have own segment type or creation of a mention should set some kind of a 'mPermitsEmoji = false'

akleshchev:
Breaks emoji tooltips

PR #4959, opened 2025-11-10
Remove cpuid feature bits verification in llprocessor.cpp

Geenz:
Removes cupid feature bits verification - it doesn't look like we are actually using this.  Also just generally breaks debug viewer builds on macOS.


brad-linden (approved):
yep.  no longer useful under rosetta

Issue #3731, opened 2025-03-12
Create SLua scripts by default


bennettgoble:
### Description

When a script is created with the Lua Editor Project Viewer it should be a SLua script instead of a LSL script.

### Acceptance Criteria

- [ ] Selecting "Create New > New Script" in inventory creates a SLua script
- [ ] Clicking "New Script" from object inventory should create a SLua script
- [ ] SLua script has the same behavior as the default LSL "Hello, Avatar!" script

### Screenshots

<img width="178" alt="Image" src="https://github.com/user-attachments/assets/bf0b73a9-b8a7-4c21-9198-b3284da825e0" />

_New Script option in object inventory_

<img width="365" alt="Image" src="https://github.com/user-attachments/assets/6aee7706-b18f-4ae7-ba1f-1bd6c1cf0dfe" />

_New script option in user inventory_

FelixWolf:
Most residents on Second Life are familiar with LSL than they are with Lua. Doing this will be both confusing and cumbersome, especially since Lua hasn't even came out of alpha yet, and the decision to make it the new default before a good majority of scripter residents expect LSL as the default.
I propose that the default be moved to `LSL / Luau` (as opposed to mono), with the option for people wanting to use SLua as their default as a option in the Preferences.

bennettgoble:
@FelixWolf thanks for the feedback, the intention here is to make working with SLua as easy as possible for the duration of the Alpha and Beta processes. Likely the default for the mainline viewer will continue to be LSL.

FelixWolf:
@bennettgoble In that case, this is a completely reasonable request! Sorry for the mis-understanding, apparently others got confused about it in #scripting-lua as well, so I figured I'd put it here instead of let it get lost in discussion. I've informed everyone there of the intentions.
Thanks!

marchcat:
This should be done on the server side.  
The viewer-side solution is a temporary workaround that should be reverted once the server-side fix is implemented.

monty-linden:
See #5815 for some recent thoughts on this.

Dan-Linden:
Passed QA
Verified on Second Life Release 26.4.0.33417795895 on Win11/OSX in the scope of https://github.com/secondlife/iqa/issues/604 

Presently, when on a Slua region and the New Script button is clicked, a Slua script is created. The screenshot shows what's visible if the down arrow widget is clicked.

[...]

PR #5027, opened 2025-11-21
#3731 Remove temporary workaround for default script

marchcat:
Remove the hack (with longer description to pass the check).


marchcat:
It turned out we still need the hack. Closing for now.

PR #5034, opened 2025-11-21
Update websocketpp package to allow compiling on Linux with C++20

HaroldCindy:
The build was failing on Linux due to some code that is now illegal in C++20 included in websocketpp, upgrade the package with a patch from the websocketpp PRs to fix it.


HaroldCindy:
Looks like there's still a couple places that need patching, upstream creator of the patch probably wasn't using the server stuff. https://github.com/secondlife/viewer/actions/runs/19583332816/job/56086393691?pr=5034#step:11:650

marchcat (approved):
Thank you very much!

PR #5037, opened 2025-11-23
Prevent the creation of temporary objects that will be discarded.

### LINKED ISSUE #5049: Prevent the creation of temporary objects that will be discarded.
A contribution for code with undefined behavior. Potentially a crash fix.

https://github.com/secondlife/viewer/pull/5037

7.2.3.19375695301

P.S. Might be linux specific, in linux branch.

AiraYumi:
By returning the actual object instead of the temporary object's LLSD, we prevent reading undefined areas.

### DIFF
--- a/indra/newview/llpathfindingnavmesh.cpp
+++ b/indra/newview/llpathfindingnavmesh.cpp
@@ -139,7 +139,8 @@ void LLPathfindingNavMesh::handleNavMeshResult(const LLSD &pContent, U32 pNavMes
         ENavMeshRequestStatus status;
         if ( pContent.has(NAVMESH_DATA_FIELD) )
         {
-            const LLSD::Binary &value = pContent.get(NAVMESH_DATA_FIELD).asBinary();
+            LLSD nav_data = pContent.get(NAVMESH_DATA_FIELD);
+            const LLSD::Binary& value = nav_data.asBinary();
             auto binSize = value.size();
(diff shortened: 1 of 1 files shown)


indra/newview/llpathfindingnavmesh.cpp
```diff
@@ -139,7 +139,7 @@ void LLPathfindingNavMesh::handleNavMeshResult(const LLSD &pContent, U32 pNavMes
         ENavMeshRequestStatus status;
         if ( pContent.has(NAVMESH_DATA_FIELD) )
         {
-            const LLSD::Binary &value = pContent.get(NAVMESH_DATA_FIELD).asBinary();
+            LLSD::Binary value = pContent.get(NAVMESH_DATA_FIELD).asBinary();
```
RyeMutt: I'd suggest this method to suppress the related compiler warning. It avoids creating an additional copy of the binary data vector.
```suggestion
            LLSD nav_data = pContent.get(NAVMESH_DATA_FIELD);
            const LLSD::Binary& value = nav_data.asBinary();
```

AiraYumi: Thank you. I refixed it.

akleshchev:
Thank you. Is there anything our QA can test?

I see that it's navmesh related and that .asBinary() returned an object that can expire before newStr, but I'm not sure what scenario this can be tested in?

akleshchev:
Any suggestions for a test plan?

PR #5047, opened 2025-11-25
p#492 Set MacOS LSMinimumSystemVersion

akleshchev:
So that MacOS will properly warn about application not running on lower version


marchcat (approved):

akleshchev:
Never mind, looks like it's now getting reported correctly even without this.

Issue #4939, opened 2025-11-05
Update bugsplat to version 1.2.6


akleshchev:
Viewer curently uses bugsplat for MacOS which seems to be at least 2 years old. It might have missing features.

As an example there is a better attribute support that allows bugplat to store attributes from a previous session (so we will have accurate data about state of previous session) instead of adding what we can post factum.
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #4939</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @akleshchev</code> (replace <code>20</code> with the amount, and <code>@akleshchev</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

Dan-Linden:
Is this ready for QA in 26.1 or should it move to 26.2?
Is there anything QA can test?

akleshchev:
Ready. Please check that all synthetic crashes (develop menu, errors) on windows reach bugsplat (sans the loop one, that's for the watchdog). On Mac there might be one that didn't work before, it's fine if it still doesn't work (I think it had a ticket, if not a new ticket might be needed, should be fixable now).

Dan-Linden:
Mac 2026.1.0.20116586066
✅ Force Breakpoint
✅ Force LLError And Crash
✅ Force LLError, Message And Crash
✅ Force Bad Memory Access
✅ Force Bad Memory Access in Coroutine
✅ Force Infinite Loop - Watchdog caught it.
✅ Force Driver Crash -> FLEngine!glDeleteTextures_Exec
⛔ Force Software Exception -> nothing reported in Bugsplat
⛔ Force OS Exception - this exists on Mac, but not on Windows. It does nothing, the viewer continues to run.
⛔ Force a Crash in a Coroutine -> nothing reported in Bugsplat
✅ Force a Crash in a Coprocedure
⛔ Force a Crash in a Work Queue -> nothing reported in Bugsplat
✅ Force a LLError Crash in a Thread
[...]

PR #5054, opened 2025-11-26
#4939 Try out bugsplat's hang detection

akleshchev:
Statistics reports a lot of feezes, this is a test attempt to detect them (will try it out once build completes). If this does not work, will do a watchdog.


marchcat (approved):

akleshchev:
Not detecting any hangs, closing.

Issue #5120, opened 2025-12-09
WebRTC voice does not connect on the first session after boot-up.


Dan-Linden:
### Environment

Second Life Release 7.2.3.19375695301 (64bit)
Release Notes

You are at 118.2, 89.9, 21.9 in By Design located at simhost-0afd3b4358f8196b2.agni
SLURL: http://maps.secondlife.com/secondlife/By%20Design/118/90/22
(global coordinates 261,494.0, 246,874.0, 21.9)
Second Life Preflight 2025-10-29.18924347600
Release Notes

CPU: 12th Gen Intel(R) Core(TM) i7-12700F (2112 MHz)
Memory: 16221 MB
OS Version: Microsoft Windows 11 64-bit (Build 26200.7171)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 4070 Ti/PCIe/SSE2

Windows Graphics Driver Version: 32.0.15.9144
OpenGL Version: 4.6.0 NVIDIA 591.44

Window size: 1194x928
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.375
Render quality: 3
Texture memory: 12282MB
Disk cache: Max size 2150.4 MB (5.3% used)

J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.24.0.202510081738
  CEF: 139.0.40+g465474a+chromium-139.0.7258.139
  Chromium: 139.0.7258.139
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 0/3,179 (0.0%)
December 09 2025 08:05:00

### Description

WebRTC voice does not connect on the first session after boot-up.
I suspect there is something special about my hardware but we'll see.

### Reproduction steps

1. Boot up the computer.
2. Start SL and log into a WebRTC region.

Observed: No voice dot above my head. The Speak button is grayed out.
Disabling/re-enabling Voice Chat in Preferences > Sound & Media does not fix it. I must relog to fix it.

Expected: Voice should work on every login.

Failed voice session: [SecondLife.log](https://github.com/user-attachments/files/24059471/SecondLife.log)
[...]

Issue #5118, opened 2025-12-09
[WebRTC] Voice disconnects because of failure to successfully handle reconnect requests.


roxanneskelly:
### Environment

25.08 release viewer talking to Aditi region WebRTC1

### Description

Voice dropped, and the following logs were grabbed.
```
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (basic_ice_controller.cc:576): Conn[59d120f0:0:Net[Intel(R):10.0.0.x/32:Wifi:id=2]:wi/EXg10:1:0:srflx:udp:67.160.9.x:60000->ubAn1F8/:1:2015363327:host:udp:44.249.117.x:10855|C-wI|S|0|0|7240414910524424702|23]\n
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (p2p_transport_channel.cc:2230): Channel[0|1|_W]: Changed writable_ to 0\n
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (dtls_transport.cc:678): DtlsTransport[0|1|_W]: ice_transport writable state changed to 0 dtls_state: kConnected\n
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (dtls_transport.cc:1011): DtlsTransport[0|1|_W]: set_writable to: 0\n
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (jsep_transport_controller.cc:1283):  Transport 0 writability changed to 0.\n
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (jsep_transport_controller.cc:1344): 0 Transport 1 state changed. Check if state is complete.\n
2025-12-09T03:58:12Z DEBUG #Voice# newview/llvoicewebrtc.cpp(321) LLWebRTCVoiceClient::LogMessage : (peer_connection.cc:1865): Changing IceConnectionState 3 => 5\n
[...]

Issue #4743, opened 2025-09-25
Viewer does not automatically reconnect to voice if existing voice server crashes


maestrolinden:
### Environment

Second Life Release 7.2.2.17774206511 (64bit)
Release Notes

You are at 128.0, 128.0, 23.0 in webRTC1 located at simhost-0462d668818ec3f3b.aditi
SLURL: secondlife://Aditi/secondlife/webRTC1/128/128/23
(global coordinates 259968.0, 245376.0, 23.0)
WebRTC-Voice 2025-08-21.17114312399
Release Notes

CPU: Apple M1 Pro (2400 MHz)
Memory: 16384 MB
OS Version: macOS 15.6.1 Darwin 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 arm64
Graphics Card Vendor: Apple
Graphics Card: Apple M1 Pro

OpenGL Version: 4.1 Metal - 89.4

Window size: 1360x877
Font Size Adjustment: 96pt
UI Scaling: 0.75
Draw distance: 96m
Bandwidth: 10000kbit/s
LOD factor: 1.125
Render quality: 0
Texture memory: 10922MB
Disk cache: Max size 2150.4 MB (3.7% used)
HiDPI display mode: 

J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.21.0.202508272158
  CEF: 139.0.28+g55ab8a8+chromium-139.0.7258.139
  Chromium: 139.0.7258.139
LibVLC Version: 3.0.21
Voice Server Version: Secondlife WebRTC Gateway 

Packets Lost: 0/285 (0.0%)
September 24 2025 17:02:17

### Description

If the webRTC voice server a viewer is connected to crashes, the viewer remains disconnected from voice until the user manually disables and reenables the Voice Chat setting.  This user action should not be required - the viewer should automatically reconnect to voice after the simulator finds a new voice server instance.

### Reproduction steps

1. Visit a webRTC voice region
2. Confirm that nearby voice chat is working
3. Follow instructions in https://github.com/secondlife/server/issues/2140  to force a crash on the voice server the viewer's simulator is using for nearby voice chat
4. Note if nearby voice chat ever recovers (without user action)

Expected results:
[...]

PR #5126, opened 2025-12-09
Fix failure to reconnect after disconnect and occasional dropout issue

roxanneskelly:
We were occasionally seeing dropouts which may have been caused by ICE renegotiate requests.  The code is there to reconnect in that case, but there were a few bugs, some of which were likely due to the webrtc upgrade.

Also, we were seeing failures to reconnect after voice server restart.

There were some issues with the PTT button that came up after the above issue was fixed.

https://github.com/secondlife/viewer/issues/5120
https://github.com/secondlife/viewer/issues/5118
https://github.com/secondlife/viewer/issues/4743

Test Guidance:
* be in a voice region during a voice server deploy
* validate that you reconnect to voice automatically after the deploy.
* Do so with the speak button pressed.
* Validate that you can still be heard after the deploy.

### DIFF
--- a/indra/llwebrtc/llwebrtc.cpp
+++ b/indra/llwebrtc/llwebrtc.cpp
@@ -1004,7 +1004,7 @@ bool LLWebRTCPeerConnectionImpl::initializeConnection(const LLWebRTCPeerConnecti
             }
 
             webrtc::PeerConnectionInterface::RTCOfferAnswerOptions offerOptions;
-            this->AddRef();
+            this->AddRef(); // CreateOffer will deref this when it's done.  Without this, the callbacks never get called.
             mPeerConnection->CreateOffer(this, offerOptions);
             mPendingJobs--;
         });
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

indra/llwebrtc/llwebrtc.cpp
```diff
@@ -1004,6 +1004,7 @@ bool LLWebRTCPeerConnectionImpl::initializeConnection(const LLWebRTCPeerConnecti
             }
 
             webrtc::PeerConnectionInterface::RTCOfferAnswerOptions offerOptions;
+            this->AddRef();
```
akleshchev: I don't see the removal anywhere, is this right? This is confusing so might be better to add reason for this as a comment.

roxanneskelly: adding a comment.

akleshchev (approved):

PR #5151, opened 2025-12-11
Add more functionality to LLFile and cleanup LLAPRFile

akleshchev:
Draft, applying 'Add more functionality to LLFile and cleanup LLAPRFile' on top of a develop-linux branch to see how it goes.


akleshchev:
@RyeMutt are you fine with adding this to linux branch?

Seems like the only issue is 'out of memory' for mac builds which does not look to be specific to this change.

RyeMutt:
I'm currently working on a changeset to resolve the out of memory/diskspace issue on the macos runner by splitting the test and viewer builds into seperate jobs. I'd like to get that in first and have a few days to put this branch through a good round of regression testing before it gets merged into linux.

akleshchev:
> I'd like to get that in first and have a few days to put this branch through a good round of regression testing before it gets merged into linux.

File operations are mostly covered by unit tests, like llcontrol test - creates a settings file, writes, reads. If you think something isn't properly covered yet should be checked, perhaps you can suggest a unit test to implement?

indra/llcommon/tests/llfile_test.cpp
RyeMutt, line 1: I'd like to see better test coverage of path and file names with unicode characters. We've had particular trouble over the years with those issues.

indra/newview/llviewerdisplay.cpp
```diff
@@ -1139,15 +1139,12 @@ std::string getProfileStatsFilename()
...
-    std::error_code ec;
     do
     {
         // base + missing 2-digit seconds, append ".json"
         // post-increment sec in case we have to try again
         name = stringize(base, std::setw(2), std::setfill('0'), sec++, ".json");
-    } while (std::filesystem::exists(fsyspath(name), ec));
-    // Ignoring ec means we might potentially return a name that does already
-    // exist -- but if we can't check its existence, what more can we do?
+    } while (LLFile::exists(fsyspath(name)));
```
RyeMutt: fsyspath should be omitted here as that's a compatibility wrapper for direct use with std::filesystem that creates a std::filesystem::path.

indra/llcommon/llfile.cpp
```diff
@@ -256,319 +235,951 @@ static int warnif(const std::string& desc, const std::string& filename, int rc,
...
-        char sep;
-#if LL_WINDOWS
-        sep = '\\';
+        temppath = std::filesystem::temp_directory_path().string();
+    }
 
-        std::vector<wchar_t> utf16path(MAX_PATH + 1);
-        GetTempPathW(static_cast<DWORD>(utf16path.size()), &utf16path[0]);
-        utf8path = ll_convert_wide_to_string(&utf16path[0]);
+    char sep;
```
RyeMutt: The addition of the ending separator could be moved into the if block above

indra/llcommon/llfile.cpp
```diff
@@ -245,9 +183,50 @@ std::string strerr(int errn)
...
+                LL_CONT << '\n' << line;
+            }
+        }
+        LLFile::remove(tf);
+    }
+    LL_CONT << LL_ENDL;
+}
+#endif // LL_WINDOWS hack to identify processes holding file open
+
+static int warnif(const std::string& desc, const std::string& filename, int rc, int suppress_warning = 0)
```
RyeMutt: `const std::string& desc` can be `std::string_view desc` to reduce string temporaries

indra/llcommon/llfile.cpp
```diff
@@ -256,319 +235,951 @@ static int warnif(const std::string& desc, const std::string& filename, int rc,
...
         }
-#endif  // LL_WINDOWS hack to identify processes holding file open
+#endif
     }
     return rc;
 }
 
-// static
-int LLFile::mkdir(const std::string& dirname, int perms)
+static int warnif(const std::string& desc, const std::string& filename, const std::error_code& ec, int suppress_warning = 0)
```
RyeMutt: `const std::string& desc` can be `std::string_view desc` to reduce string temporaries

akleshchev:
Thanks, adding those changes.

indra/llcommon/tests/llfile_test.cpp
akleshchev: I feel like I was the one to fix most of those LL side (I think there is still one within our updater). Adding a test case for unicode.

RyeMutt: There's unfortunately been a lot of duplicated effort over the years between all the opensource viewers and LL 😅

Now when I run the test suite with the added unicode tests on windows I'm seeing failures like:
```
2>  PYTHON='C:/Python314/python.exe' \
2>  R:/newview/viewer/build-vc180-64/sharedlibs/Release/INTEGRATION_TEST_llfile.exe
2>  Unit test group_started name=llfile_test
2>  [llfile_test, 5] fail: 'LLFile::write() did not write correctly to Unicode file'
2>  Unit test group_completed name=llfile_test
2>  	Total Tests:	5
2>  	Passed Tests:	4
2>  *********************************
2>  Failed Tests:	1
2>  Please report or fix the problem.
```
and then if i re-run it so it has to do a directory cleanup I get

```
2>  PYTHON='C:/Python314/python.exe' \
2>  R:/newview/viewer/build-vc180-64/sharedlibs/Release/INTEGRATION_TEST_llfile.exe
2>  Unit test group_started name=llfile_test
2>  [llfile_test, 5] fail: 'Unicode test directory should not exist'
2>  Unit test group_completed name=llfile_test
2>  	Total Tests:	5
2>  	Passed Tests:	4
2>  *********************************
2>  Failed Tests:	1
2>  Please report or fix the problem.
2>  *********************************
```

akleshchev: I'm aware, and there is a new line in the file name. I lost power so had to push it half backed, just now fixed it.

RyeMutt:
After doing some testing and scanning through the log file I'm seeing some concerning failures from animation loading from these changes after running the viewer on a clean cache(Deleted the cache folder)

```
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(592) LLKeyframeMotion::onInitialize : Failed to allocate buffer: -12408fe9e-df1d-1d7d-f4ff-1384fa7b350f
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(600) LLKeyframeMotion::onInitialize : Can't open animation file 2408fe9e-df1d-1d7d-f4ff-1384fa7b350f
2025-12-17T14:32:52Z INFO # llcharacter/llmotioncontroller.cpp(368) LLMotionController::createMotion : Motion 2408fe9e-df1d-1d7d-f4ff-1384fa7b350f init failed.
2025-12-17T14:32:52Z WARNING #Motion# newview/llvoavatar.cpp(6131) LLVOAvatar::processSingleAnimationStateChange : Failed to start motion!
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(592) LLKeyframeMotion::onInitialize : Failed to allocate buffer: -1249010f1-fe5f-2533-9006-e6e1639ed0c5
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(600) LLKeyframeMotion::onInitialize : Can't open animation file 249010f1-fe5f-2533-9006-e6e1639ed0c5
2025-12-17T14:32:52Z INFO # llcharacter/llmotioncontroller.cpp(368) LLMotionController::createMotion : Motion 249010f1-fe5f-2533-9006-e6e1639ed0c5 init failed.
2025-12-17T14:32:52Z WARNING #Motion# newview/llvoavatar.cpp(6131) LLVOAvatar::processSingleAnimationStateChange : Failed to start motion!
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(592) LLKeyframeMotion::onInitialize : Failed to allocate buffer: -142b46214-4b44-79ae-deb8-0df61424ff4b
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(600) LLKeyframeMotion::onInitialize : Can't open animation file 42b46214-4b44-79ae-deb8-0df61424ff4b
2025-12-17T14:32:52Z INFO # llcharacter/llmotioncontroller.cpp(368) LLMotionController::createMotion : Motion 42b46214-4b44-79ae-deb8-0df61424ff4b init failed.
2025-12-17T14:32:52Z WARNING #Motion# newview/llvoavatar.cpp(6131) LLVOAvatar::processSingleAnimationStateChange : Failed to start motion!
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(592) LLKeyframeMotion::onInitialize : Failed to allocate buffer: -1538bea3f-787f-27fd-c141-01cc91fc6c50
2025-12-17T14:32:52Z WARNING # llcharacter/llkeyframemotion.cpp(600) LLKeyframeMotion::onInitialize : Can't open animation file 538bea3f-787f-27fd-c141-01cc91fc6c50
2025-12-17T14:32:52Z INFO # llcharacter/llmotioncontroller.cpp(368) LLMotionController::createMotion : Motion 538bea3f-787f-27fd-c141-01cc91fc6c50 init failed.
```

akleshchev:
> Failed to allocate buffer: -12490 10f1-fe5f-2533-9006-e6e1639ed0c5

So a negative value somehow got into getSize, checking

RyeMutt:
It looks like this is potentially caused by `LLFileSystem::getFileSize` now wrapping `LLFile::size` which can return `-1` where as the previous version a failure would return 0.

akleshchev:
> hich can return -1 where as the previous version a failure would return 0.

Yes, it's probably better to return 0 (it returns ec either way) instead of looking for everything that expects non zero value.

Lots of places do !vf.getSize()

fmartian:
> > hich can return -1 where as the previous version a failure would return 0.
> 
> Yes, it's probably better to return 0 (it returns ec either way) instead of looking for everything that expects non zero value.
> 
> Lots of places do !vf.getSize()

The idea was to fix the callers of LLFile::size() and I think I got them almost all but seems one or two escaped (or might be caused by Linux specific changes, the patch was developed against the development branch). For LLFileSystem::getSize() I think it is easiest for now to limit the returned value to 0 although the checks should be probably rather changed to "vf.getSize() > 0" for a more generic fix since that function is defined to return a signed value.

I do have more thorough changes that will avoid some of these problems (and completely do away with LLAPRFile) but it got somewhat hard to manage with the new Linux compatibility requirement too.

akleshchev:
Switched size functions to return 0. Double-checked logs and I see nothing suspicious.

> The idea was to fix the callers of LLFile::size()

I don't think it adds much value. If someone needs to check the error state, they can check ec variant.

> but it got somewhat hard to manage with the new Linux compatibility requirement too.

Sorry for adding more work, at the moment it might be wise to wait a bit with that.

fmartian:
> Switched size functions to return 0.
> 
> > The idea was to fix the callers of LLFile::size()
> 
> I don't think it adds much value. If someone needs to check the error state, they can check ec variant.

The problem is not in the places that call LLFile::size(). It's called only a handful times throughout so far and except in LLFileSystem the caller is actually properly checking for the value to be positive to determine if the file is valid.
The problem is with LLFileSystem::getSize() which is called all over the place without checking the return value. Most times the code just happens to not crash by chance while it does repeatedly do an allocation for 0 bytes/elements. That's not really safe.

#5172 is an attempt to fix this although testing has been limited.

A quick and dirty fix would be to limit the return value of LLFileSystem::getSize(). However that does only mask the real problem. There is more opportunity to clean/optimize things in several of those places, but I tried to make it not to big.

Hecklezz:
Seems you accidentality changed the permissions of 20 of the 33 files to be executable (chmod 755) from their original read and user owner write (chmod 644). There is really no need for that to have happened and seemed like it was done on accident.

akleshchev:
> Seems you accidentality changed the permissions of 20 of the 33 files to be executable

Thanks for noticing. No idea how that happened, probably was same way in the original commit. Will fix some time later, probably after the holidays.

Hecklezz:
> > Seems you accidentality changed the permissions of 20 of the 33 files to be executable
> 
> Thanks for noticing. No idea how that happened, probably was same way in the original commit. Will fix some time later, probably after the holidays.

After your recent commit, there are still 9 more that were changed to be executable that shouldn't be. 
I also noticed there are a number of files that shouldn't be executable that have snuck in with other commits over the years, but those are outside the scope of this PR.
When this PR gets merged I'll run a script to fix them all and put up a separate PR :)

akleshchev:
> After your recent commit, there are still 9 more that were changed to be executable that shouldn't be.

Fixed. For some reason those weren't visible locally.

RyeMutt (commented):
LGTM!

akleshchev:
@RyeMutt Thank you!

Issue #5217, opened 2026-01-06
BugSplat Crash: SecondLifeViewer!LLControlVariable::LLControlVariable(163)


sl-service-account:
## BugSplat Crash Group 101171
[View Report Group](https://app.bugsplat.com/v2/group?stackKeyId=101171&database=SecondLife_Viewer_2018)
[View Report Details](https://app.bugsplat.com/v2/crash?id=1695040&database=SecondLife_Viewer_2018)
[View Report Attachments](https://app.bugsplat.com/v2/crash?id=1695040&database=SecondLife_Viewer_2018#attachments)

**Application:** Second Life Release
**Version:** 7.2.3.19375695301
**Error Code:** c0000005 Access violation

**Some Customer Responses:**
> Must supply a comment for control Include<br/>Must supply a comment for control Include<br/>Must supply a comment for control Include<br/>Must supply a comment for control Include<br/>Must supply a comment for control Include<br/>

**Callstack:**
[...]

PR #5222, opened 2026-01-06
#5217 Small cleanup and const declarations

akleshchev:
Small cleanup and const declarations.

### DIFF
--- a/indra/newview/llavataractions.cpp
+++ b/indra/newview/llavataractions.cpp
@@ -29,8 +29,6 @@
 
 #include "llavataractions.h"
 
-#include "boost/lambda/lambda.hpp"  // for lambda::constant
-
 #include "llavatarnamecache.h"  // IDEVO
 #include "llsd.h"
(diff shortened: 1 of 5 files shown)


indra/newview/pipeline.cpp
```diff
@@ -2729,7 +2729,7 @@ void LLPipeline::clearRebuildGroups()
 
     mGroupQ1Locked = true;
     // Iterate through all drawables on the priority build queue,
-    for (LLSpatialGroup::sg_vector_t::iterator iter = mGroupQ1.begin();
+    for (LLSpatialGroup::sg_vector_t::const_iterator iter = mGroupQ1.begin();
```
Ansariel: This looks weird TBH. You make the iterator const, that then gets dereferenced to a pointer to `LLSpatialGroup` that is non-const (which should be `LLPointer<LLSpatialGroup>` in the first place), because otherwise you could not call `clearState`!? And why not change to range-based iterator anyway?

PR #5235, opened 2026-01-08
Simplify `--[[` Lua comment syntax highlighting, speed up SLua highlighting

HaroldCindy:
image

Here's a callgrind graph of the work currently done when inserting a new character to a very long script. As you can see, the regex match type is pretty expensive and dominates the frame time.

Since the regex match kind was only being used for Lua comments, I removed the regex match type and added a match type specifically for Lua's long bracket syntax, which helps a lot with the editor lag.

### DIFF
--- a/indra/llui/llkeywords.cpp
+++ b/indra/llui/llkeywords.cpp
@@ -28,7 +28,6 @@
 
 #include <iostream>
 #include <fstream>
-#include <regex>
 
 #include "llkeywords.h"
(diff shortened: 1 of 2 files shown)


marchcat (approved):
Using regex here was more of a temporary solution, since I wasn't sure if it would be needed for anything else. Since we ended up using it only to find brackets, removing it was planned anyway.

Thank you very much for this fix!

Issue #5083, opened 2025-12-01
Fallback to OS open handler in external editor


akleshchev:
If no external editor is set, spawn OS editor

## Test Plan

1. Open a Notecard from Inventory
2. Click the "Edit..." button on the bottom of the Notecard.
3. Confirm that external editor opens up even if not set (ex: used in xui preview, notecards, scripts, only one needs to be tested)
4. Confirm that setting a different editor uses the new editor (ExternalEditor debug setting)
Windows example: C:\Users\tester\Downloads\npp.8.9.portable.minimalist.x64\notepad++.exe "%s"
Mac Example: /usr/bin/open -a TextEdit %s

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #5083</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @akleshchev</code> (replace <code>20</code> with the amount, and <code>@akleshchev</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

Dan-Linden:
Reopened. 
Verified on Second Life Release 26.1.0.20374992269 on Win11/OSX in the scope of https://github.com/secondlife/iqa/issues/581

Mac:
Editing a NoteCard works on Mac. The file is opened in TextEdit.
Editing a Script does nothing.
How is the xui preview opened and tested?

Windows:
Editing a NoteCard or Script on Windows does not open an external editor and this dialog appears instead.
<img width="372" height="101" alt="Image" src="https://github.com/user-attachments/assets/5ea587df-c68e-43a2-8215-3a2ac1bad4e0" />
How is the xui preview opened and tested?

akleshchev:
> How is the xui preview opened and tested?

<img width="339" height="327" alt="Image" src="https://github.com/user-attachments/assets/03e553f8-3639-49d4-b9ef-deb43a68d511" />

It's on login screen. But if it fails for notecards, there is no point testing XUI.

Dan-Linden:
[...]

PR #5243, opened 2026-01-09
#5083 Fix windows not finding default external editor

akleshchev:
Looks like process couldn't handle %SystemRoot%, so getting it from environment first.


akleshchev:
Rye got a better and fuller fix, closing.

PR #5293, opened 2026-01-18
[#3013] Fix hollowed prims not shading correctly and appearing not smooth

Hecklezz:
Issue: https://github.com/secondlife/viewer/issues/3013

This is a long standing bug since the release of viewer version 7.1.9. 
The bug was introduced in this commit when a FALSE fake bool was mistakenly changed to a true bool-
https://github.com/secondlife/viewer/commit/c3e6f7b1643076df35a6960f735024078ddbb353

This PR simply just corrects this function call making sure it is passed false instead of true.

<details>
<Summary>Before and After Screenshots-</Summary>
Before:

Firestorm_2026-01-19_00-42-50

After:

Firestorm_2026-01-19_01-44-05

</details>


marchcat (approved):
Thank you!

Issue #5299, opened 2026-01-19
Improve keywords files for SLua and LSL


marchcat:
Issue to track XML file changes/contributions.

PRs: #5288
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #5299</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @marchcat</code> (replace <code>20</code> with the amount, and <code>@marchcat</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

marchcat:
Integrated #5288, keeping the issue open for further related PRs.

tapple:
All PR's in this project. I'll edit this comment if I add more
- https://github.com/secondlife/lsl-definitions/pull/38
- https://github.com/secondlife/viewer/pull/5288
  - https://github.com/secondlife/viewer/pull/5301
  - https://github.com/secondlife/viewer/pull/5325
- https://github.com/secondlife/sl-vscode-plugin/pull/57
- https://github.com/secondlife/python-llsd/pull/37

marchcat:
Thank you, @tapple!  
Shouldn't we just merge all these PRs into the SLua editor viewer, since they are related?

tapple:
All the PR's which are made against the viewer are made against the project/lua_editor branch

3 of them aren't against the viewer, and are made against the main branches in those repositories

jellyfish-linden:
Passed QA.
Verified on Second Life Release 26.4.0.34502396632 on Mac in the scope of https://github.com/secondlife/iqa/issues/604

PR #5327, opened 2026-01-24
Add luau-lsp files to the viewer

tapple:
- Part of #5299 

These are the two files that any installation of [the Luau Language Server (luau-lsp)](https://github.com/JohnnyMorganz/luau-lsp) needs to make sense of slua code. Currently, vscode plugin builds these two files based on these two files it fetches from the viewer:
-  [types_lua_default.xml](https://github.com/secondlife/viewer/blob/project/lua_editor/indra/newview/app_settings/types_lua_default.xml)
- [keywords_lsl_default.xml](https://github.com/secondlife/viewer/blob/project/lua_editor/indra/newview/app_settings/keywords_lsl_default.xml) (except it uses the version from the simulator capability)

Instead, I propose the viewer and/or simulator keep these two files on-hand so that [the vscode plugin](https://github.com/secondlife/sl-vscode-plugin) and other editors (including the viewer's builtin editor) can handle slua without complex file transformation.

These two files were generated by [the latest unreleased version](https://github.com/secondlife/sl-vscode-plugin/commit/251f9b840647a49c3a8255e20ba64d1ba419f42b) of [the vscode plugin](https://github.com/secondlife/sl-vscode-plugin). Further auto-generated updates to these files will be in a seperate PR #5328 ([View the diff](https://github.com/tapple/viewer/pull/1/files))

---

Part of a set:
- https://github.com/secondlife/lsl-definitions/pull/42
- https://github.com/secondlife/viewer/pull/5327
- https://github.com/secondlife/viewer/pull/5328

### DIFF
--- a/indra/newview/app_settings/slua_default.d.luau
+++ b/indra/newview/app_settings/slua_default.d.luau
@@ -1,0 +1,1962 @@
+
+----------------------------------
+---------- LSL LUAU DEFS ---------
+----------------------------------
+
+declare class quaternion
+  x: number
+  y: number
+  z: number
+  s: number
+  function __add(self, other: quaternion): quaternion
+  function __sub(self, other: quaternion): quaternion
+  function __mul(self, other: quaternion): quaternion
+  function __div(self, other: quaternion): quaternion
+  function __unm(self): quaternion
+  function __eq(self, other: quaternion): boolean
+  function __tostring(self): string
+end
+
+declare class uuid
+  istruthy: boolean
+  bytes: string?
+  function __tostring(self): string
+end
+
+declare class vector
+  x: number
+  y: number
+  z: number
+  function __add(self, other: vector): vector
+  function __sub(self, other: vector): vector
+  function __unm(self): vector
+  function __mul(self, other: number | vector | quaternion): vector
+  function __div(self, other: number | vector | quaternion): vector
+  function __mod(self, other: vector): vector
+  function __tostring(self): string
+end
+
+
+type rotation = quaternion
+type numeric = boolean | number
+type list = {string | number | vector | uuid | quaternion | boolean}
+type LLDetectedEventName = "collision" | "collision_end" | "collision_start" | "final_damage" | "on_damage" | "sensor" | "touch" | "touch_end" | "touch_start"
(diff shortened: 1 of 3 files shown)


marchcat:
Thank you, @tapple!

marchcat:
@Rider-Linden, please take a look.

Rider-Linden:
The `d.luau` files are uses only by Lua LSPs that are forked from JohnnyMorganz' luau-lsp plugin.  Although one of the most popular there are other plugins and editors that do not use the `.d.luau` format we would like to support (Neovim, Zed, Helix).  

Delivering the JSON and having the plugin generate the appropriate configs file strikes me the correct approach.

I don't object to providing these two files as a reference but it should be made clear that they are not canonical.

tapple:
These ones I found use JohnnyMorganz luau-lsp:
- neovim: https://github.com/lopi-py/luau-lsp.nvim
- zed: https://zed.dev/docs/languages/luau
-  helix: https://github.com/JohnnyMorganz/luau-lsp/discussions/463

I can't quickly find any editors supporting luau that don't use JohnnyMorganz luau-lsp

HaroldCindy:
Hmm, realistically, `luau-analyze` itself isn't sufficient for our purposes, and we need `luau-lsp` to be able to describe our API extensions correctly. `luau-lsp` is _the_ choice of LSP for Luau, so I wouldn't feel too badly about depending on it for type checking. It also provides `luau-lsp analyze` for that purpose.

HaroldCindy:
Spoke with @Rider-Linden about this. Seems like really what we want is to send down the current contents of the definitions YAML (maybe reserialized as JSON for the JSON-RPC stuff) to external consumers, rather than bundle a generated artifact from that source of truth that the viewer itself will not use. The keywords XML will probably have to remain in-repo because the in-viewer editor's syntax highlighting impl is a bit of a weird case, but otherwise we're looking at providing a way for the JSON-RPC clients to get the full definitions blob the server is currently using. They can generate whatever schema they want (`.d.luau`, Selene definition, etc) based on that blob.

tapple:
Oh. Selene uses a third format I was unaware of: https://kampfkarren.github.io/selene/usage/std.html

If that's the official consensus though, I guess I'll work on generating types_lua.xml and updating the vscode plugin to work with the changes I made to its schema

tapple:
This can be closed as harold is working on a better solution: https://github.com/secondlife/lsl-definitions/pull/46/files

PR #5341, opened 2026-01-28
#2975 PBR support for planar aligment

akleshchev:
Since checkbox is shared between pbr and normal textures, I'm making it affect both for now.

This might be first of multiple commits as I see an 'align' button that might need similar handling.


akleshchev:
Doesn't work yet. Top and bottom faces are a mess.

PR #5383, opened 2026-02-06
Rider/default script

Rider-Linden:
## Description

The viewer may now indicate to the server which default script to use when creating a new script in either task or agent inventory.
Optionally the viewer may also provide the ID of a script in the agent's inventory, this script will be used instead of the default script.

### Future Work
Default script choice is determined by what the server supports.  If it supports SLua it will use the lua default, otherwise it will select the LSL default.

We should provide some sort of user option to select which default to use.  There is also no interface at this time for selecting an alternate default template from inventory, this would be useful to provide as well.

## Related Issues
Simulator: https://github.com/secondlife/server/pull/2344
Message Template: https://github.com/secondlife/master-message-template/pull/4

---

## Additional Notes
This also cherry picks the `updateLoginButtons` fix.


indra/newview/llpanelcontents.cpp
```diff
@@ -254,99 +241,21 @@ void LLPanelContents::onClickNewScript(void *userdata)
         std::string desc;
         LLViewerAssetType::generateDescriptionFor(LLAssetType::AT_LSL_TEXT, desc);
 
-        // --------------------------------------------------------------------------------------------------
-        // Begin hack
```
marchcat, line 258: Yay!

marchcat (approved):

indra/newview/llviewerinventory.cpp
```diff
@@ -1816,7 +1784,7 @@ void create_new_item(const std::string& name,
                           desc,
                           asset_type,
                           inv_type,
-                          NO_INV_SUBTYPE,
+                          subtype,
```
HaroldCindy, line 1787: Is there anywhere else that uses `wearabletype` this way outside of wearables themselves?

indra/newview/llviewerinventory.h
```diff
@@ -42,6 +42,10 @@ class LLViewerInventoryCategory;
 class LLInventoryCallback;
 class LLAvatarName;
 
+constexpr U8 NO_INV_SUBTYPE{ 0 };
+constexpr U8 LANG_ID_LSL = 0;
+constexpr U8 LANG_ID_LUA = 1;
```
HaroldCindy: If scripts are created as one type, does that constrain the types of compile options you can use for the script once it's created? Is that something the viewer needs to be aware of?

indra/newview/llviewerinventory.cpp
Rider-Linden: Yes, subtype is also used for settings (Skies, Water, Day Cycles). 

As a bonus, the viewer does check this subtype when drawing wearables and settings items and could display a different icon based on script type.

Thinking about it, I'm make a change to the metadata branch that I have which will switch the subtype on a successful compile when it updates the inventory item.

indra/newview/llviewerinventory.h
Rider-Linden: No.  It does not.

Rider-Linden: I fixed subtype on the server side so that it is changed to match the last successful compile.  I still need the metadata to identify which VM but this should provide a good default.

AndrewMeadows (approved):

marchcat (approved):

PR #5385, opened 2026-02-06
Visual Polish Alpha

Geenz:
relnotes:

# Visual Polish Alpha

Welcome to the visual polish alpha!  In this viewer, we have various visual polish features we're working on that we're collecting community feedback on.  Stay tuned for a complete list!

## New Features

* Completely rewritten screen space reflections that are much higher quality and more performant
* Automatic reflection probes on water
* Improved sky reflections on reflection probes
* Reflection quality drop down with Low, Medium, and High settings
* Improved shadow sampling that's far less blurry
* Improved alpha blended shadows
* Tonemapping can now be controlled via EEP
* HDR can now be controlled via EEP
* Added a new "Use Environment" setting for HDR
* Added SMAA T2x temporal anti-aliasing
* Added motion blur support
* Improved texture streaming and discards
* Add support for physically based brightness of the sun and moon
* Support for PBR Specular glTF extension


akleshchev (approved):

indra/newview/pipeline.cpp
```diff
@@ -9350,34 +9353,50 @@ void LLPipeline::bindReflectionProbes(LLGLSLShader& shader)
...
 
+        static LLCachedControl<LLVector3> traceIterations(gSavedSettings, "RenderScreenSpaceReflectionIterations");
+        static LLCachedControl<LLVector3> traceSteps(gSavedSettings, "RenderScreenSpaceReflectionRayStep");
+        static LLCachedControl<LLVector3> traceDistanceBias(gSavedSettings, "RenderScreenSpaceReflectionDistanceBias");
+        static LLCachedControl<LLVector3> traceRejectBias(gSavedSettings, "RenderScreenSpaceReflectionDepthRejectBias");
+        static LLCachedControl<LLVector3> traceStepMultiplier(gSavedSettings, "RenderScreenSpaceReflectionAdaptiveStepMultiplier");
+        static LLCachedControl<LLVector3> traceSplitStart(gSavedSettings, "RenderScreenSpaceReflectionSplitStart");
+        static LLCachedControl<LLVector3> traceSplitEnd(gSavedSettings, "RenderScreenSpaceReflectionSplitEnd");
+        static LLCachedControl<F32> traceMaxDepth(gSavedSettings, "RenderScreenSpaceReflectionMaxDepth");
+        static LLCachedControl<F32> traceMaxRoughness(gSavedSettings, "RenderScreenSpaceReflectionMaxRoughness");
```
akleshchev: Is there a point for all these settings to be configurable in gSavedSettings?

Geenz: Tweakability for the time being.

Dan-Linden:
| Visual Polish Alpha 26.1.0.22463915568 | Result | Comment |
|---|---|---|
| secondlife/iqa#616 Acceptance Tests | Passed | Passed |
| **Overall Result** | **Passed** | **OK to ship** |

**NOTES**

Anti-virus false positive report:
| File | Security vendor analysis |
|---|---|
| SLVersionChecker.exe | SecureAge - Malicious |
| slplugin.exe | MaxSecure - Trojan.Malware.300983.susgen<br>  SecureAge - Malicious |

Crexon1:
Just adding in my of my observations and feedback from testing this out so far.

Frame times are a bit wild with SSR enabled. Im sure theres some improvement to be had but just adding my particular system on this. For this screen shot it was at a sim I normally go to and with ssr off frame times tighten up to normal levels.

image

If i go to a 100% empty sandbox frame times are better, but not great considering a completely empty scene. Performance wise even at the empty sim fps goes from average of 80fps with ssr off to 35-40 with ssr on (changing no other setting)

For the new sun controls is there a reason why we going to 2 million lux? Only need really between 0-150,000 (_maybe_ extreme of 200,000) Having it such high from an end user stand point makes the slider bar basically useless since what users will be using is only about 5% of the slide bar.

Crexon1:
Also the latest build does not like OBS. With OBS open the viewer will crash loading to login page. If I launch the viewer first and let it get to at least the login page then launch OBS it seems to be ok.  OBS running latest version 32.0.4.
[SecondLife.log](https://github.com/user-attachments/files/25649673/SecondLife.log)

**Update** it does seem to also crash sometimes if I launch OBS after logging in. 
[SecondLife.log](https://github.com/user-attachments/files/25650206/SecondLife.log)

Crexon1:
When "Use Environment HDR settings" is check tone mapper and mix should be greyed out as those no longer do anything. Same functionality as turning off HDR Emissive.
image

Also Exposure appears to not work as well. I dont see exposure in the upgraded environment settings. If exposure is being added to environment then this should revert back to a "brightness" scale.

Crexon1:
It appears mirrors are backwards in this version.

Mirror in 7.2.3. 
image


Same mirror in Second Life Project Visual Polish 26.1.0.22463898576 
image

if I go the opposite side of the mirror it appears to be correct on the other side. 
image

vldevel:
Non-planar mirrors are broken after commit b18e63f57c906657bd69a2ad0668c01c794b33e5
They only reflect sky and avatars eyeballs...

To repro, simply run latest SLVP build and stand before any current mirrors in SL (they almost all use probes with Z-size > 0.02m: IMHO this size should be made configurable too).

Here is a video I made, which also demonstrates issues with the CameraAngle not being properly taken into account by mirrors (*) with the water probe looking especially bad for all CameraAngle values above 1.0:
http://sldev.free.fr/misc/mirror-issues.mkv

Note that for this video, I used an experimental Linux build of my viewer (with commits up to b18e63f57c906657bd69a2ad0668c01c794b33e5 implemented), because I also added to it a setting to force all mirrors as planar (easing the demonstration), but I get the exact same results with the latest SLVP build under Windows.

-------

(*) This was already the case before the changes in SLVP: the mirrors mips selection is wrong, as demonstrates the fact that increasing the CameraAngle and then zooming back on mirror to get the same FOV, you get a much sharper mirror reflection: with proper mips selection it should be possible to get sharp mirrors without having to use ludicrous 2048x2048 mirror probes: 1024x1024 should be more than enough, saving about 250MB of VRAM per mirror probe (500MB in total, now that we have two probes in SLVP) in the process !

Geenz:
@vldevel @Crexon1 Should be fixed in the next build.  Along with multiple mirror probes _should_ work now*.

_* Assuming you really want to crank up the appropriate debug setting and have reflections set to ultra_

vldevel:
> @vldevel @Crexon1 Should be fixed in the next build.

Not quite: commit a55949c649ffa1fee87abe3c9b17cfc4d94d9132 does fix the non-rendering non-planar probes, but it is still broken in the sense that it renders a magnified reflection.

Here is what I'm seeing in my experimental build with the new commit; note that I implemented a configurable probe Z size limit to flag (non-water) probes as planar or not (*): when inferior or equal to this limit, the mirror probe is considered planar.
Non-planar mirror:
non-planar-mirror
Planar mirror:
planar-mirror

-----
(*) < 0.02 was too small: almost all mirrors in SL could be made planar (since they are indeed flat mirrors) but since the mirror probe object must enclose the surface of the mirror object, almost everyone is using a probe larger than 0.0199 (even for a flat mirror usually being 0.01 thick), like 0.05 (such as the mirror I use for my testing and seen in the above snapshots). Making this limit configurable also allows to disable (non-water) planar mirrors entirely (by setting the limit to 0 or less), or to turn every single mirror into a planar one (by setting the limit to 64 or more).

vldevel:
We also need some VRAM usage optimization for planar hero probes: since they only use face 0 in a cube map, the 5 other cube map faces are allocated for no purpose whatsoever, and given how ludicrously large the hero probes need to be (2Kx2K) to avoid blurry mirror reflections (due to bad mips selection, see my comments above), that's about 200MB of VRAM going to waste...
The cube map should therefore be replaced with planar render targets for planar probes.

Geenz:
@vldevel Part of the plan (when I have a moment) is to add another texture array just for mirrors.  Though, this is going to depend on specific platform capabilities (read: Apple's limping along OpenGL implementation) if we can get the extra sampler in the shaders.  There have been problems before about this - especially in PBR's development cycle where we would bust the sampler budget in the shaders.

If that doesn't work, then I'm likely to just repurpose the other faces of the cube map to better utilize what we're setting up here.  It's a cost you always pay for, but at least it's a predictable cost here.

In theory you'd think 16 is more than enough for us, but well.  Turns out we have quite a few uniforms and samplers in the main sun + atmospherics deferred pass.

vldevel:
> read: Apple's limping along OpenGL implementation)

That's an euphemism, considering Apple stopped all OpenGL maintenance over 7 years ago... :-D

I'd advocate for excluding macOS from planar mirrors rendering, if it means Windows and Linux can enjoy "low" VRAM budget planar probes.

Geenz:
No.

PR #5389, opened 2026-02-09
UI controls adjustments and selection color update

maxim-productengine:
Mostly adjustments for UI controls but some changes are worth mentioning:

- Set default alpha to 1 for `LLTextBase` selection background. After that `LLLineEditor` and `LLTextEditor` will have the same behavior and it's finally possible to assign the same selection background color and actually to get it.
- Hide 'General' button in Preferences>Privacy as we currenly doesn't show Discord tab at all
- Use old font for notification date/time - this is the only way to avoid overlaps with sender name (the width of date time is quite significant considering AM/PM addition),

### DIFF
--- a/indra/llui/lltextbase.cpp
+++ b/indra/llui/lltextbase.cpp
@@ -51,6 +51,9 @@ const F32   CURSOR_FLASH_DELAY = 1.0f;  // in seconds
 const S32   CURSOR_THICKNESS = 2;
 const F32   TRIPLE_CLICK_INTERVAL = 0.3f;   // delay between double and triple click.
 
+constexpr F32 FOCUSED_SELECTION_BG_ALPHA = 1;
+constexpr F32 UNFOCUSED_SELECTION_BG_ALPHA = 0.7f;
+
 LLTextBase::line_info::line_info(S32 index_start, S32 index_end, LLRect rect, S32 line_num)
 :   mDocIndexStart(index_start),
     mDocIndexEnd(index_end),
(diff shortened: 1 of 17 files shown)


indra/llui/lltextbase.cpp
```diff
@@ -529,7 +529,7 @@ void LLTextBase::drawSelectionBackground()
         // Draw the selection box (we're using a box instead of reversing the colors on the selected text).
         gGL.getTexUnit(0)->unbind(LLTexUnit::TT_TEXTURE);
         const LLColor4& color = mSelectedBGColor;
-        F32 alpha = hasFocus() ? 0.7f : 0.3f;
```
akleshchev, line 532: Magic numbers, perhaps add constexpr?

indra/newview/skins/default/xui/en/panel_notification_list_item.xml
```diff
@@ -63,7 +63,8 @@
                     use_ellipses="true" word_wrap="false" mouse_opaque="false" name="sender_or_fee_box" visible="false">
                     Sender: "Resident R e s i d e n t R e s i d e n t"
                   </text>
-                  <text allow_scroll="false" font="SansSerifSmall" top="0" right="-5" width="95" height="13" follows="right" halign="right" layout="topleft" left_pad="5"
+                  <!-- Old DejaVu font is used to avoid overlapping with sender name -->
+                  <text allow_scroll="false" font="DejaVu" font.size="LSmall" top="0" right="-5" width="100" height="13" follows="right" halign="right" layout="topleft" left_pad="5"
```
akleshchev, line 67: Why not give it more width instead?

akleshchev (approved):

indra/newview/skins/default/xui/en/panel_notification_list_item.xml
maxim-productengine: There is not enough space for both long (and even medium) usernames and notification time if the new font is used.

akleshchev: And what if whole panel becomes bigger?

Issue #4508, opened 2025-08-10
Proposal: Add unvendored third-party dependencies to `external/*` directory


bennettgoble:
### Description

As part of the effort to remove our custom package manager autobuild (see #4390), we need to organize third-party dependencies that are not available through popular open source package managers.

## Problem

Currently, some of our third-party dependencies, such as **tut**, are managed through autobuild and are not readily available via standard package managers like Conan or vcpkg. These dependencies need a clear organizational structure as we transition away from autobuild.

## Solution

Create a new `external/*` directory structure to house third-party dependencies that cannot be obtained through popular open source package managers (Conan, vcpkg, etc.).

## Dependencies

- [ ] tut
- [ ] libndofdev
- [ ] colladadom
- [ ] ...?

bennettgoble:
Most of these packages are incredibly unsupported and unmaintained, so keeping them up to date shouldn't be much of a problem. 🙃

Issue #4390, opened 2025-07-18
📦 RFP: Replace Autobuild with Modern Dependency Management System


bennettgoble:
We are requesting proposals for removing [autobuild](https://github.com/secondlife/autobuild) from the Second Life Viewer build process and replacing it with a modern, maintainable dependency management system such as vcpkg, Conan, or direct embedded dependencies.

Autobuild was developed in-house to solve a real problem at the time, but it has become a maintenance burden and is now limiting our ability to quickly iterate, support new platforms, update dependencies, and simplify build tooling.

## ✅ Acceptance Criteria

- A detailed written proposal is submitted (as a comment linking to a Gist, Google Doc, or similar)
- Proof of concept featuring a building indra/llcommon
- The proposal includes a recommended replacement for autobuild (e.g., vcpkg, Conan, embedded libraries)
- The proposed solution:
  - Simplifies local development environment setup
  - Offers native or improved integration with CMake
  - Reduces the effort needed to update or patch third-party libraries
  - Improves or maintains code signing compatibility (especially for Windows)
  - Supports all current Viewer platforms (Windows and macOS)
  - Easily supports platforms supported by forks: (Linux, arm64 builds)
- The proposal includes:
  - CI/CD migration strategy
  - High-level timeline and migration plan
  - Identification of high-risk dependencies or steps
  - Tradeoffs between proposed options

## 🧪 Evaluation

- Build verification across Windows and macOS
- CI pass with the new dependency management system
- Comparison of build time, dependency footprint, and setup complexity (before vs. after)
- Developer experience feedback during migration

## 📎 Resources

- [Autobuild repository](https://github.com/secondlife/autobuild)
- [vcpkg](https://github.com/microsoft/vcpkg)
- [Conan](https://github.com/conan-io/conan)

---

**Please post your proposal as a comment on this issue by linking to a document (Gist, Google Doc, etc.).**

sploid:
Hi @bennettgoble, 
I suggest creating a separate repository with the compiled dependencies so we don't have to download them every time we build the application. One pipeline will build the dependencies and update the link to them in the source code repository, while another will build the application.
I've answered you on Upwork.
Thanks,
Ivan

kylelinden:
25.09 or 26.01

sfunnytoad:
> This project is now available for bids/application at https://www.upwork.com/jobs/~021973518684571374385 (Requires an Upwork account to view...)

[...]

PR #5178, opened 2025-12-19
One Click Install Alpha

Geenz:
as of = 2026-01-26 for PV deploy
build = https://github.com/secondlife/viewer/releases/tag/Second_Life_Project%2385298ebb-OneClickInstall
cohort = One Click Install
deployed = https://github.com/secondlife/viewer/releases/tag/Second_Life_Project%2385298ebb-OneClickInstall
desired = 0
relnotes:

# One Click Install Alpha Viewer

Hey everyone!

This release we have the One Click Install Alpha Viewer.  This is an early alpha release to streamline the experience of setting up the viewer on Windows computers and getting in world.  This also will help ease the update process for viewers moving forward.

## What's different

* Viewers on Windows now use a new "one click" installer that installs the viewer to a local user directory.
* Viewers on macOS still use the disk image method for drag and drop install.
* All viewers will now use a new updater that makes updates more transparent.
  * Viewers will no longer prompt the user to install an update.  It will now install in the background.
* Installing to custom locations is supported by the `--installto` command line parameter
* Portable installations can be found on our GitHub: https://github.com/secondlife/viewer/releases/tag/Second_Life_Project%2385298ebb-OneClickInstall

At present, you will have two viewers installed: the one click viewer and the current viewer.  This is expected, and will be improved in a future alpha.

PR #5429, opened 2026-02-13
Replace autobuild with vcpkg and CMake build system modernization

RyeMutt:
## Description

This set of changes is to modernize the developer experience with vcpkg based dependency management and a cleaned up cmake build system.

## Related PR
https://github.com/secondlife/3p-colladadom/pull/17
https://github.com/secondlife/3p-open-libndofdev/pull/11

## Related Issues

Issue Link: #4390 #4508

---

## Checklist

- [ ] I have provided a clear title and detailed description for this pull request.
- [ ] I have tested the changes locally and verified they work as intended.
- [ ] I have reviewed the [contributing guidelines](https://github.com/secondlife/viewer/blob/develop/CONTRIBUTING.md).

---


Ansariel:
Is there a reason why it is not possible to do a SINGLE thing at a time per PR and NOT mix it with a bunch of other changes so we don't end up in unnecessary giant PRs?

Geenz:
Hey @Ansariel - I know how disruptive and annoying big changes like these are, however from the LL end there was the decision to group this into a bigger package for the work we commissioned Rye to do here.

That being said, it's my hope that we won't do this again.  Large chunks like this is disruptive to everyone, even our engineers.  So in the future I'll endeavor to ensure that our statements of work are sufficiently broken up and measurable from the start rather than what this started out as.

ChorazinAllen:
Hi @Geenz 

I believe that this and other changes already in the pipe are going to prove to be the breaking point for all but the best resourced TPVs. Let's look at what's coming down:-

- Removing autobuild will force TPVs to rework their build systems and, probably, do significant work to get their own Linux builds working again unless they pause their Linux releases until the LL one sees light of day
- There's a growing assumption that the only way the viewer is ever built is using GHA. That simply isn't so.
- The change in Windows packaging will again force TPVs to adapt it to their own requirements and package content
- The 2D UI will force TPVs to rework their own UI elements to match the 2D style to avoid a mix of 2D and 3D UI appearing

and that's on top of the routine work caused by LL breaking TPV changes that they're unaware of when something gets refactored.

The biggest elephant in the room though is RLVa. If that ever starts appearing in the LL viewer it's going to cause major issues for every viewer that already has RLV or RLVa included. These TPVs currently have a functional RLV or RLVa and wouldn't cut across to including a LL version until its functionality meets or exceeds what they already have. Assuming the LL development is done in phases, as was previously discussed, it's going to be a merge nightmare to keep LL RLVa elements out of TPVs until it's ready for a wholesale switch over.

This is a lot of disruption for minimal functional change or feature benefit in the viewer and it's putting TPVs with small or single person teams under huge strain for very little gain. I'm pretty sure many other TPVs are feeling the same.

--Chorazin

Hecklezz:
>   * Removing autobuild will force TPVs to rework their build systems and, probably, do significant work to get their own Linux builds working again unless they pause their Linux releases until the LL one sees light of day

The vcpkg changes won't be out until after the Linux changes, which will come with the SLua viewer, so at least that worry of yours can be crossed-off. It will still be a lot of work of course to merge everything, but official Linux support in my opinion is extremely important and worth it.

>   * There's a growing assumption that the only way the viewer is ever built is using GHA. That simply isn't so.

I don't believe this to be the case at all. I have already tested building with these vcpkg changes on my local machine and it is as simple as 1 single command without any other setup, or build environment variables, or any nonsense to create the project (I tested both MSVC and with Ninja). I don't see anywhere that there is an assumption that the only way people do it is with GHA.

>   * The 2D UI will force TPVs to rework their own UI elements to match the 2D style to avoid a mix of 2D and 3D UI appearing

For the TPVs that mostly follow LL's skin I can definitely see this being a pain and not something fun to do. At least in Firestorm's case, we will likely be throwing out all of the flat UI changes because it is a gigantic merge conflict that will break way too much of our own UI changes that are in place (plans may change of course).

___
  
Autobuild has been a maintenance burden for LL (they have said as much), and with Signal leaving I imagine even more of a burden now. It is more limiting and harder to work with compared to modern, widely used systems such as vcpkg. So getting it done now and getting through the struggles to move to an industry standard system that LL doesn't need to maintain, is better for everyone.

DarlCat:
> The change in Windows packaging will again force TPVs to adapt it to their own requirements and package content

I took a look again at #5178 and still don't see how the existing Windows packaging code is removed from the viewer with the new one-click project. TPVs already customize their installers, and I don't think many if any will choose to adopt this new approach as-is. Like the flat UI changes already addressed above me the installer will probably also just be dropped on merge by TPVs because there is no obligation to follow the lab's direction on it or really any of the things that were listed.

Third-party viewers exist that are one-person operations, and can't even merge modern Linden code due to license conflicts or in some cases nearing two decades of source code drift; they've objectively been doing fine.

Lastly, being nitpicky here, but with a PR prefixed "WIP do not run actions" how does one arrive at this conclusion when there aren't even finished GHA scripts to build it yet?

> There's a growing assumption that the only way the viewer is ever built is using GHA. That simply isn't so.

---

All this to say; I have actually played with these the changes in this PR and I find it MASSIVELY simplifies the development process. Between the whole thing being literally one command to get started, and having cmake integrations working in code editors without any fiddling to take advantage of them, this PR turns what has always been an esoteric and tribal knowledge laden nightmare into a modern batteries-included developer experience.

ChorazinAllen:
Thanks @Hecklezz  and @DarlCat  for the comments.

I do support the intent of this project, respect the amount of work that's gone into it and will be glad to move away from autobuild too. What troubles me though is the size of this PR and the number of merge conflicts it's likely to throw up when it first meets a TPV codebase.

Knowing that this will surface after Linux does indeed make that side of things much easier.

Windows packaging: I'm tempted to just exclude that too since TPVs (except Firestorm) are generally not what a new user will start out with, however it's usually the case that when old and new methods exist in parallel sooner or later there's a position where only the new is maintained and the old falls behind and eventually stops working until it's finally removed from the code base as a mercy killing. Avoiding that in a TPV keeping NSIS means an overhead of keeping an eye on changes to the new to see whether they need to be ported back to the old and dodging the bullet when old is eventually removed from upstream.

2D UI: I'm also tempted to exclude that too, however again it's not that simple. If a TPV chooses to remain 3D there's then another new maintenance overhead to identify assets created for future new features in the 2D UI style and make 3D versions. These would typically slip through in a merge (new item, so there's no merge conflict) and have to be checked for specifically whilst sorting out the merge conflicts.

--Chorazin

Hecklezz:
For any TPV devs who haven't ventured into merging this yet, as long as you have already merged in linux-develop, it honestly isn't that bad and the conflicts were fairly straightforward. I got this merged and all working in our internal firestorm development build for both Windows and Linux (awaiting Rye to finish Mac stuff before testing that), and everything is working and packaged almost identically how it was with autobuild. There was a few things here and there after the fact you notice and need to fix or were slightly wrong despite the viewer working, or doing a few little changes to fix debug builds, but they were all fairly easy to spot and fix.

Working with vcpkg is honestly pretty nice, and although I've always used my own python helper script to avoid the annoyances of autobuild and the initial setup of it, for those who are cloning and building the viewer from scratch or don't have a helper script, they will have a far easier time and it will feel more intuitive.

The only real downside of the vcpkg setup as it is right now, is if the packages aren't built and cached yet, it can take ~30+ minutes for it to do so on initial configuration before you even get to building the viewer. Ideally you only ever have to do this once (or at least rarely), so it _shouldn't_ be an issue, but it will be one of the main differences people will notice.

indra/cmake/BuildPackagesInfo.cmake
```diff
@@ -10,11 +14,11 @@ include(FindAutobuild)
...
-          ${CMAKE_SOURCE_DIR}/../autobuild.xml
-  COMMAND ${PYTHON_EXECUTABLE}
-          ${CMAKE_SOURCE_DIR}/cmake/run_build_test.py -DAUTOBUILD_ADDRSIZE=${ADDRESS_SIZE} -DAUTOBUILD=${AUTOBUILD_EXECUTABLE}
-          ${PYTHON_EXECUTABLE}
-          ${CMAKE_SOURCE_DIR}/../scripts/packages-formatter.py "${VIEWER_CHANNEL}" "${VIEWER_SHORT_VERSION}.${VIEWER_VERSION_REVISION}" "${AUTOBUILD_INSTALL_DIR}" > packages-info.txt
+  MAIN_DEPENDENCY ${INDRA_SOURCE_DIR}/../autobuild.xml
+  DEPENDS ${SCRIPTS_DIR}/packages-formatter.py
+          ${INDRA_SOURCE_DIR}/../autobuild.xml
+  COMMAND ${Python3_EXECUTABLE}
+          ${INDRA_SOURCE_DIR}/cmake/run_build_test.py -DAUTOBUILD_ADDRSIZE=${ADDRESS_SIZE} -DAUTOBUILD=${AUTOBUILD_EXECUTABLE}
```
brad-linden: we should also completely kill `run_build_test.py` in this cmake modernization as well if we can.  I think it can be completely replaced by `cmake -E env ...`

indra/cmake/LLTestCommand.cmake
```diff
@@ -4,14 +4,14 @@ MACRO(LL_TEST_COMMAND OUTVAR LD_LIBRARY_PATH)
   # cannot return a value. And yet, variables you set inside a FUNCTION are
   # local. Try a MACRO instead.
   SET(value
-    ${PYTHON_EXECUTABLE}
-    "${CMAKE_SOURCE_DIR}/cmake/run_build_test.py")
+    ${Python3_EXECUTABLE}
+    "${INDRA_SOURCE_DIR}/cmake/run_build_test.py")
```
brad-linden: repeating above thoughts:  consider replacing `run_build_test.py` with `cmake -E env ...`

indra/llrender/llvertexbuffer.cpp
```diff
@@ -325,8 +325,8 @@ class LLVBOPool
 {
     public:
     virtual ~LLVBOPool() = default;
-    virtual void allocate(GLenum type, U32 size, GLuint& name, U8*& data) = 0;
-    virtual void free(GLenum type, U32 size, GLuint name, U8* data) = 0;
+    virtual void allocateVBO(GLenum type, U32 size, GLuint& name, U8*& data) = 0;
```
brad-linden: seems inappropriate to be renaming functions in a PR for these purposes.  I agree with the change in this case, but makes things harder to review

Geenz (approved):
Approving with a note:

Due to the extensive nature of this change, and due to the Lua Editor viewer being based upon our linux branch, we will **not** be merging this into the Linux branch.  We will instead be pushing it to a vcpkg specific branch.

I know this is rather frustrating in terms of process flow, but we're trying to minimize risk on Lua given we're getting pretty down to the wire on that front.

The merge ordering I have in my head so far is:

vcpkg-viewer -> develop-vcpkg (branched from develop-linux) -> develop (after Lua makes it into develop)

Additionally given how disruptive this is likely to be downstream - I'd like to give this some time to settle with TPVs.  I feel that's important to the adoption of any significant build system change we make - and especially for one as big as this.

Issue #5084, opened 2025-12-01
Resurrect Watchdog


akleshchev:
Watchdog was disabled in 2011. After that there appears to be no intances of it being used.

1. Test and make sure it properly reports to bugsplat
2. Either enable it (for test/beta viewers?) or make use of simulator feature WatchdogDisabled
3. Apply a more dynamic timing approach, like 60s for login and shutdown, 5-10 for a running session.

Reason:

Statistics report a high percentage of freezes, but not the cause or 

Test plan
* Debug menu > Force an Error > Force Infinite Loop 
* Verify a report appears in BugSplat.

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #5084</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @akleshchev</code> (replace <code>20</code> with the amount, and <code>@akleshchev</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

Dan-Linden:
Is Debug > Force an Error > Force Infinite Loop the way to test this?

akleshchev:
Yes, but anything that freezes viewer should do. Expect a report in bugsplat.

Dan-Linden:
I see 2 simultaneous BugSplats from Windows: https://app.bugsplat.com/v2/crash?database=SecondLife_Viewer_2018&id=1696409 , https://app.bugsplat.com/v2/crash?database=SecondLife_Viewer_2018&id=1696410
I see 1 BugSplat from Mac: https://app.bugsplat.com/v2/crash?database=SecondLife_Viewer_2018&id=1696411

Dan-Linden:
Passed QA.
Verified on Second Life Release 26.1.0.20374992269 on Win11/OSX in the scope of https://github.com/secondlife/iqa/issues/581

Dan-Linden:
I can file a separate issue about the 2 simultaneous BugSplats if that's something worth changing. Let me know.

akleshchev:
[...]

PR #5502, opened 2026-03-06
#5084 Improve watchdog's behavior

akleshchev:
Draft

- On windows watchdog will report the freeze to bugsplat than just stop observing (unless thread wakes up), viewer will keep 'running'. If reporting isn't avaliable, it will fallback to crashing (for ease of testing).
- On mac watchdog will not gently go into the night, but instead will show a notification to user and will attempt to logout.
- On both platforms watchdog will create a marker. So that if viewer gets closed forcefully, next launch will know if problem was caught by watchdog and will be reported to statistics

To be resolved in separate PRs:
- Watchdog triggering on modal OS level notifications
- Find if Mac can report without callstack (dows not support runtime reporting with callstack)
- Decide how to handle mac vs windows difference. We probably want to report at 20s without crashing, then crash at 5m or something similar.

### DIFF
--- a/indra/llcommon/llapp.h
+++ b/indra/llcommon/llapp.h
@@ -285,6 +285,7 @@ public:
 
 #ifdef LL_WINDOWS
     virtual bool reportCrashToBugsplat(void* pExcepInfo /*EXCEPTION_POINTERS*/) { return false; }
+    virtual bool reportCustomToBugsplat(const std::string& desription) { return false; }
 #endif
 
 public:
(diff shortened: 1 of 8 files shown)


marchcat (approved):

akleshchev:
Getting to something I'm happy with. Tested and watchdog created a marker, then properly set 'froze' state for statistics report.

akleshchev:
Rebased to 2026.02

PR #5527, opened 2026-03-11
Restore optional updates, only restart the viewer if it were optional.

Geenz:
This restores optional updates, and will only restart the viewer if it's an optional update upon completion.

### DIFF
--- a/indra/newview/llvelopack.cpp
+++ b/indra/newview/llvelopack.cpp
@@ -35,6 +35,7 @@
 #include <fstream>
 #include <unordered_map>
 #include "llnotificationsutil.h"
+#include "llviewercontrol.h"
 #include "llcoros.h"
 
 #include "Velopack.h"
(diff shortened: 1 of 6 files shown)


indra/newview/llvvmquery.cpp
```diff
@@ -82,8 +82,8 @@ namespace
         }
 
         // Gather parameters for VVM query
-        std::string channel = LLVersionInfo::instance().getChannel();
-        // std::string channel = "QA Target for Velopack";
+        //std::string channel = LLVersionInfo::instance().getChannel();
```
akleshchev: Forgot to disable?

akleshchev (approved):

indra/newview/llvelopack.cpp
```diff
@@ -760,6 +762,44 @@ void velopack_check_for_updates()
...
+
+void velopack_check_for_updates(bool required, const std::string& version, const std::string& relnotes_url)
+{
+    if (required)
+    {
+        velopack_download_update();
+    }
+    else
+    {
+        LL_INFOS("Velopack") << "Optional update available (version " << version << "), prompting user" << LL_ENDL;
```
akleshchev: You probably want to check UpdaterServiceSetting here. There are 3 modes. 0 - install silently. 1 - ask, 2 - don't install.

akleshchev (approved):

akleshchev (approved):

PR #5584, opened 2026-03-26
Bring in an updated Dullahan autobuild package containing an updated version of CEF 146

callumlinden:
This change brings in a new version of the Dullahan autobuild package that was built was against CEF version 146.

This updated PR pulls in CEF 146.0.12 (vs versions in earlier inrevisions of this PR 146.0.7 and 146.0.10). Other than the requisite CEF -> Dullahan -> Viewer autobuild steps, no other code was modified. 

This change is for Windows and macOS only. Linux version to follow when we are able to create a compatible v146 version of CEF and remains on version 139.

No changes were required in Dullahan or Viewer C++ code - merely updated in Dullahan autobuild.xml to point to the new C++ binaries and in the Viewer autobuild.xml to point to the updated Dullahan.

For QA/testing, it's been 8 months since the version of CEF used in the canonical Viewer (v139) was released so there will be a very large number of changes in CEF itself.

I did a simple smoke test and confirmed that web pages appear to render as before, links work, keyboard works etc. but if there are issues to resolve, it will be in the more esoteric areas of integration - SLURL support, multiple windows, cookie management etc. I don't know if this warrants a full browser integration regression test - perhaps we wait for feedback in the alpha phase before making that call.


Geenz:
_Should_ be fine - though we might want to ask QA to run whatever regression tests they have available for this.

callumlinden:
Pulled in the latest CEF (146.0.10) via the usual CEF -> Dullahan -> Viewer `autobuild` route. No other code changes - just new Dullahan package definitions in `autobuild.xml`.

Confirmed that it appears to work as expected (simple smoke test with > 1 instances) and that the version is correct (`chrome//version` in a media floater).

callumlinden:
Superseded by 147 and 148

PR #5597, opened 2026-04-01
Roxie/cef accelerated

roxanneskelly:
Enable accelerated painting from CEF/Dullahan.  Dullahan will now hand the CEF media plugin a handle to the painted texture in the GPU, which can be shared with the viewer process.  This avoids the previous case where we needed to copy the texture into system RAM and then put it back to VRAM after the bits were transferred to the viewer.

### DIFF
--- a/indra/media_plugins/cef/media_plugin_cef.cpp
+++ b/indra/media_plugins/cef/media_plugin_cef.cpp
@@ -41,10 +41,13 @@
 // _getpid()/getpid()
 #if LL_WINDOWS
 #include <process.h>
+#include <d3d11_1.h>
+#include <dxgi.h>
 #else
 #include <unistd.h>
 #endif
 
+#include <unordered_map>
 #include "dullahan.h"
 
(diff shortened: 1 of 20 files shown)


indra/media_plugins/cef/media_plugin_cef.cpp
```diff
@@ -933,6 +1040,21 @@ void MediaPluginCEF::receiveMessage(const char* message_string)
             {
                 mEnableMediaPluginDebugging = message_in.getValueBoolean("enable");
             }
+#if LL_WINDOWS0
```
callumlinden, line 1043: Typo ?

callumlinden (approved):
Approved - with a single LL_WINDOWS0 typo to be fixed

indra/media_plugins/cef/media_plugin_cef.cpp
roxanneskelly: still a wip.  my way of temporarily commenting this section out

PR #5651, opened 2026-04-15
#4893 Async inventory review fixes

marchcat:
Added destructor to clean up callbacks and connection if loading is interrupted.

Fixed mChildNameCache to avoid removing wrong pointers when siblings share names.

Restored singleton initialization flow to preserve dependency tracking.

Ensured system folders are created at startup while keeping lazy loading for subfolders.

### DIFF
--- a/indra/newview/llinventorypanel.cpp
+++ b/indra/newview/llinventorypanel.cpp
@@ -1045,22 +1045,62 @@ void LLInventoryPanel::initializeViews(F64 max_time)
 void LLInventoryPanel::initRootContent()
 {
     // Optimization: Build root folder and its immediate children (system folders)
-    // but not grandchildren. Children will be built lazily when folders are expanded.
-    // This ensures system folder widgets exist while avoiding the full tree build.
+    // but not grandchildren. Essential system folders (Calling Cards, My Outfits,
+    // Marketplace, etc.) also get their immediate children built so that nested
+    // system content (e.g. Friends subfolder, outfit folders) is available at init.
+    // Everything deeper is built lazily when folders are expanded.
     LLUUID root_id = getRootFolderID();
     if (root_id.notNull())
     {
         LLInventoryObject const* objectp = mInventory->getObject(root_id);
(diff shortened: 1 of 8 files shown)


indra/newview/llinventorypanel.cpp
```diff
@@ -1044,25 +1044,23 @@ void LLInventoryPanel::initializeViews(F64 max_time)
...
-        buildNewViews(root_id, objectp, nullptr, BUILD_NO_CHILDREN);
+        buildNewViews(root_id, objectp, nullptr, BUILD_ONE_FOLDER);
     }
     else
     {
         // Default case: always add "My Inventory" root first, "Library" root second
-        // Build only root widgets - children will load on-demand when expanded
         LLInventoryObject const* my_inv = mInventory->getObject(gInventory.getRootFolderID());
-        buildNewViews(gInventory.getRootFolderID(), my_inv, nullptr, BUILD_NO_CHILDREN);
+        buildNewViews(gInventory.getRootFolderID(), my_inv, nullptr, BUILD_ONE_FOLDER);
```
akleshchev, line 1060: This won't build nested system folders like friend's calling cards. Also a good idea to build 'my outfits' fully and se if it has folder id for the outfit.

akleshchev: Also, please, make sure marketplace loads.

akleshchev (changes requested):
Spam, example:
2026-04-16T19:58:46Z WARNING #Inventory# newview/llinventorymodel.cpp(5039) LLInventoryModel::validate : invalid desc count for ae5834b2-7d47-6557-0cff-f7e05c763801 [/My Inventory/Objects/tmp7] cached 341 expected 1+5=6

That's a massive discrepancy between cached and expected, and it's reported each run, should it even complain if inventory isn't loaded?

P.S. Please make sure to test this with Inventory2 Tester and check marketplace

akleshchev (approved):

PR #5674, opened 2026-04-19
Respect custom logfile during startup

### LINKED ISSUE #1681: Sort out / maybe fix viewer "logfile" option
### Environment

Release viewer

### Description

The viewer has a bunch of code for the "logfile" setting and directly logs to a custom file name.    That, unfortunately, doesn't seem to work as expected.      Look into it, sort out what's working and fix what is not, so that we can:

- set up command shortcuts to start the viewer and log to different files
- invoke the handleLogFileChanged() function when the setting is changed and switch files as expected

### Reproduction steps

Use the log settings and look for files as you expect

mayconbekkers:
﻿This fixes the viewer logfile startup path so the effective log file is selected before initial log rotation and duration tracking run.

The viewer now:
- keeps the default `SecondLife.log` behavior when no custom log file is set
- respects `--logfile` / `-logfile` on Windows during startup
- keeps `.old` rotation and last-run duration checks aligned with the effective active log file
- stays consistent with the existing runtime `UserLogFile` switch behavior

Validated locally with:
- default startup
- startup with `--logfile`
- startup with `-logfile`
- runtime `UserLogFile` switching
- second-instance smoke check

### DIFF
--- a/indra/newview/llappviewer.cpp
+++ b/indra/newview/llappviewer.cpp
@@ -2380,7 +2380,7 @@ void LLAppViewer::initLoggingAndGetLastDuration()
         // Rename current log file to ".old"
         LLFile::rename(log_file, old_log_file);
 
-        // Set the log file to SecondLife.log
+        // Set the log file.
         LLError::logToFile(log_file);
         LL_INFOS() << "Started logging to " << log_file << LL_ENDL;
         if (!duration_log_msg.empty())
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

mayconbekkers:
Updated to handle dotfile log names correctly and to recognize --logfile= / -logfile= during the early Windows startup scan. The existing Win32 startup parsing path is unchanged otherwise.

indra/newview/llappviewer.cpp
```diff
@@ -2515,6 +2519,76 @@ namespace
                      LLStringUtil::null,
                      OSMB_OK);
     }
+
+    std::string getStartupLogFileName()
+    {
+        if (LLControlVariable* user_log_file = gSavedSettings.getControl("UserLogFile"))
```
akleshchev, line 2525: Is gSavedSettings loaded before we set up logging or after?

akleshchev:
Please update from develop, it seems like this is missing 'velopack' fix so it fails to build.

mayconbekkers:
After. `initLoggingAndGetLastDuration()` is called from `processMarkerFiles()` in the `LLAppViewer` constructor, before `initConfiguration()` loads settings or applies command-line options, so `UserLogFile` is not available there yet.

I also rebased onto current `develop` to pick up the Velopack fix.

akleshchev (approved):

marchcat (approved):

maxim-productengine (approved):

akleshchev:
Thank you, merging.

PR #5675, opened 2026-04-20
Replace the Snapshot to Disk shortcut with Ctrl+Shift+D

mayconbekkers:
﻿The old `Snapshot to Disk` shortcut relied on an OEM key, which does not behave consistently on international keyboard layouts.

This switches it to `Ctrl+Shift+D` and leaves the existing save-to-disk behavior as-is.

Tested on Windows with `pt-BR` and `US-International`.

### DIFF
--- a/indra/newview/skins/default/xui/en/menu_viewer.xml
+++ b/indra/newview/skins/default/xui/en/menu_viewer.xml
@@ -2467,7 +2467,7 @@ function="World.EnvPreset"
             <menu_item_call
              label="Snapshot to Disk"
              name="Snapshot to Disk"
-             shortcut="control|`"
+             shortcut="control|shift|D"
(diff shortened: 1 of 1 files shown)


akleshchev:
I think Ctrl+Shift+D is already in use for camera controls. And even if it isn't, consider everything WASD to be incident prone thus avoided. Example: User is shift-rinning (there is such an option in controls), then tries to ctrl-something (gesture?), you get your ctrl-shif-d by accident.

P.S. A better option for this one might be to edit german xui file only or to finally detach shortcuts from menus and have them be configurable in Controls.

akleshchev:
Sorry. Closing after confirmation from Lindens. Any of the WASD keys are incident prone thus should not be used for shortcuts.

PR #5731, opened 2026-04-28
FIX - Prejump regression on first jump

### LINKED ISSUE #4218: Pre-jump is sometimes skipped
### Environment

Second Life Release 7.1.14.15192634334 (2024.04)

### Description

When holding down the jump button in a no-fly area, the pre-jump animation is sometimes skipped. 
https://gyazo.com/92ab3f5681bde217f07b36c6bdd78a98

This is impacting the longstanding Pony Play community, as detailed in https://feedback.secondlife.com/bug-reports/p/disable-waiting-for-pre-jump-and-landing-setting-not-working

This issue also does not appear to occur on pre-PBR viewers including FS 6.6.17 and the LL GHA viewer 6.6.16.6566955269. 

This pre-jump animation skip appears to line up with causing https://github.com/secondlife/server/issues/2044. That is, when the pre-jump doesn't occur whilst jumping on an object, a super jump is observed. This can be confirmed instantly while enabling Firestorm's "Disable waiting for pre-jump and landing animation" setting.

### Reproduction steps

1. Go to a no-fly parcel/region or set one accordingly.
2. Hold down the jump key

trish-sl:
I inadvertently introduced a side-effect with #5409 that this PR fixes.

With #5409 included, if you are standing still and hold jump, the first prejump can be unnecessarily delayed, causing a noticeable "hitch" with some AOs.

This adjustment makes it so we only check jumps that follow a landing, so the first jump in a chain is unaffected and no longer suffers that hitch. (In other words, we do not need to suppress a AGENT_CONTROL_FINISH_ANIM that comes from the viewer determining the prejump anim is done)

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -492,7 +492,7 @@
     <key>RecentJumpThresholdSecs</key>
     <map>
       <key>Comment</key>
-      <string>Seconds after a jump input during which finish-anim is suppressed to avoid interrupting rapid successive jumps.</string>
+      <string>Seconds after jump input during which landing finish-anim is suppressed to avoid interrupting rapid successive jumps.</string>
       <key>Persist</key>
       <integer>1</integer>
       <key>Type</key>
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

akleshchev:
Thank you.

Unfortunately original change is already in 26.2 and we are unlikely to get anything else into 26.2. Will need to consult with lindens if this needs a new ticket or if something can be worked out with 26.2.

PR #5769, opened 2026-05-07
p#590 keep a spare browser plugin for login screen

maxim-productengine:
Don't merge for now, requires some more testing.

### DIFF
--- a/indra/newview/llviewermedia.cpp
+++ b/indra/newview/llviewermedia.cpp
@@ -48,6 +48,7 @@
 #include "llmutelist.h"
 #include "llnotifications.h"
 #include "llnotificationsutil.h"
+#include "llstartup.h"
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

Dan-Linden:
I tested on a personal potato here (Intel i5-4210U/win10). It's using Microsoft AV, with Real-time protection on. The 26.2 (26.2.0.25021396775 Beta) viewer is marginally slower (0.5 to 1s) than 26.1.1, just like my faster machines. The test viewer (26.2.0.25508111477) with this pr is about 3 seconds slower, so that change does affect potatoes.

On my potato, the time from the viewer appearing to the login screen content is beginning to load is ~14s on the 26.2 Beta. On the test viewer (26.2.0.25508111477) the time was ~17 seconds.

PR #5793, opened 2026-05-12
feat: Bounty solution for #5789 - fix: Every once in a while it seems everyone else gets muted

vulam1808:
## Bounty Solution

Fixes the issue where users get randomly muted in voice chat.

**Changes:**
- Fixed voice muting logic that caused users to be muted after ~1 second
- Updated mute state management to properly handle OS-level mute events

*Auto-generated by Bounty Hunter Pipeline*


akleshchev:
No actual fix present?

akleshchev:
Whatever this is, there is no fix inside, closing.

PR #5823, opened 2026-05-15
docs: add specification for Set UI Size to Default help menu item

### LINKED ISSUE #5820: Add 'Set UI Size to Default' to help menu on login screen
### Description

Issue:

User can set scale to be too large with no obvious way out of it. We have a shortcut inworld to fix it. But not only login button will be out of view, option is not visible due to being under 'advanced' menu inworld, while user is likely stuck on the login.

Task:

Dupplicate the option from 'advanced' inworld menu to 'help' menu on login screen.

Test plan:

1. Set UI scale to some large value
2. Click on menu's Help > Set UI Size to Default
3. Verify Scale resets to 1.0

zp6:
## Summary
Adds implementation spec for the 'Set UI Size to Default' help menu item on the login screen.

Fixes #5820

Wallet: zp6


akleshchev:
Sorry, but this doesn't even touch the help menu, closing.

PR #5829, opened 2026-05-18
Texture streaming rework

Geenz:
Reworks the texture streaming system for more consistent signal composition and better VRAM accounting.

 - Discard is now driven by a normalized 0..1 signal: distance x on-screen-size x per-channel exponent, with staleness / background / VRAM-pressure floors layered on top. Close textures stay less aggressively downrezzed than far ones at every pressure level
- Post-decode textures can downrez down to 1x1, independent of the j2c's encoded mip count.
- Terrain LOD now participates in streaming (was pinned to discard 0). Distance is sourced from the closest visible patch each frame.
- Per-texture last-bind time drives a progressive staleness floor that downrezzes long-idle textures.
- Backgrounded windows ramp a global discard floor over time, capped at (dim_max - TextureBackgroundDiscardOffset) so some baseline quality survives.
- Close-camera bubble: textures within TextureCloseBubbleMeters of the camera get dist_factor = 0; the ramp to 1 spans (bubble, draw_distance).
- New RenderTextureQuality preset (Low/Medium/High/Ultra) drives texture cap + per-channel exponents.
- Per-channel exponents are now four individual F32 debug settings (TextureChannelNormal/BaseColor/Specular/Emissive) instead of a Vector4

### DIFF
--- a/indra/llrender/llimagegl.cpp
+++ b/indra/llrender/llimagegl.cpp
@@ -715,7 +715,10 @@ void LLImageGL::dump()
 //----------------------------------------------------------------------------
 void LLImageGL::forceUpdateBindStats(void) const
 {
-    mLastBindTime = sLastFrameTime;
+    // Intentionally a no-op: mLastBindTime is written only by real bind
+    // paths so the staleness signal reflects actual GPU use. Callers that
+    // still invoke this (avatar "keep alive" sites, deleted-texture
+    // fallback) no longer falsely refresh staleness.
 }
 
 bool LLImageGL::updateBindStats() const
@@ -1650,7 +1653,6 @@ bool LLImageGL::createGLTexture(S32 discard_level, const LLImageRaw* imageraw, S
     {
         destroyGLTexture();
         mCurrentDiscardLevel = discard_level;
-        mLastBindTime = sLastFrameTime;
         mGLTextureCreated = false;
         return true ;
     }
@@ -1766,9 +1768,7 @@ bool LLImageGL::createGLTexture(S32 discard_level, const U8* data_in, bool data_
 
 
     mTextureMemory = (S64Bytes)getMipBytes(mCurrentDiscardLevel);
-
-    // mark this as bound at this point, so we don't throw it out immediately
-    mLastBindTime = sLastFrameTime;
+    mGLCreateTime = sLastFrameTime;
 
     checkActiveThread();
     return true;
--- a/indra/llrender/llimagegl.cpp
+++ b/indra/llrender/llimagegl.cpp
@@ -1895,7 +1895,7 @@ bool LLImageGL::readBackRaw(S32 discard_level, LLImageRaw* imageraw, bool compre
     LLGLint is_compressed = 0;
     if (compressed_ok)
     {
-        glGetTexLevelParameteriv(mTarget, is_compressed, GL_TEXTURE_COMPRESSED, (GLint*)&is_compressed);
+        glGetTexLevelParameteriv(mTarget, gl_discard, GL_TEXTURE_COMPRESSED, (GLint*)&is_compressed);
     }
 
     //-----------------------------------------------------------------------------------------------
--- a/indra/llrender/llimagegl.cpp
+++ b/indra/llrender/llimagegl.cpp
@@ -66,8 +66,8 @@ static LLMutex sTexMemMutex;
 static std::unordered_map<U32, U64> sTextureAllocs;
 static U64 sTextureBytes = 0;
 
-// track a texture alloc on the currently bound texture.
-// asserts that no currently tracked alloc exists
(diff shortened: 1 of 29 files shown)


indra/llrender/llimagegl.cpp
```diff
@@ -80,15 +80,46 @@ void LLImageGLMemory::alloc_tex_image(U32 width, U32 height, U32 intformat, U32
...
     sTexMemMutex.lock();
 
-    // it is a precondition that no existing allocation exists for this texture
-    llassert(sTextureAllocs.find(texName) == sTextureAllocs.end());
-
-    sTextureAllocs[texName] = size;
+    auto iter = sTextureAllocs.find(texName);
+    if (iter != sTextureAllocs.end())
+    {
+        iter->second += size;
```
akleshchev, line 86: I take it something substructs the old value? Where does it happen?

indra/newview/app_settings/settings.xml
```diff
@@ -11805,33 +11816,206 @@
...
       <key>Value</key>
-      <array>
-        <real>5</real>
-        <real>7.5</real>
-        <real>20</real>
-        <real>7.5</real>
-      </array>
+      <real>1.0</real>
+    </map>
+    <key>TextureChannelBaseColor</key>
```
akleshchev, line 11830: That's a lot of setting that get used in the same place for the same purpose. Please consider storing these 4 channel settings as a single vec4.

indra/newview/llviewertexture.cpp
```diff
@@ -2108,17 +2170,24 @@ bool LLViewerFetchedTexture::updateFetch()
         }
     }
 
-    desired_discard = llmin(desired_discard, getMaxDiscardLevel());
+    // Clamp the fetch request to what the codestream encodes; deeper
+    // discards are served from the GL mip pyramid via scaleDown.
+    desired_discard = llmin(desired_discard, (S32)mCodecMaxDiscardLevel);
```
akleshchev, line 2289: This might need to be conditional.
Ex: Some texture (profile icons) get intentionally requested at max resolution and then scaled down (icons) to a reasonable size. Some textures aren't JPEG2000 and don't discard well.

indra/newview/app_settings/settings.xml
Geenz: I did - however the new vec4 stuff will crash old viewers that don't have support for it from what I've seen.  For now I'm avoiding using that for this release - will consider bringing it back in 26.4 or something.

indra/newview/llviewertexture.cpp
Geenz: Good point - and FWIW the discards now better relate to how they work on the GPU with mip levels and such.  I'll go ahead and add some guard rails to make sure profile images and similar (like picks) don't get accidentally downrez'd.

indra/newview/app_settings/settings.xml
akleshchev: Sounds like viewer needs protections again unrecognized values, so that adding somethng new won't break older viewers (fallback to LLSD?).

TommyTheTerrible:
Why is this reverting the J2C size calculations back to the previous "pre-aug 2024" method? Was there an issue with the updated method? Why is this action not described in the scope of the PR?

Geenz:
@TommyTheTerrible So some of that might've been messing with some of the KDU code stream stuff which had a bit of a knock on effect in other parts of the texture pipeline.  It's also not quite accurate to say that it reverts it - we just use a different code path for KDU vs. OpenJPEG now.

TommyTheTerrible:
> @TommyTheTerrible So some of that might've been messing with some of the KDU code stream stuff which had a bit of a knock on effect in other parts of the texture pipeline. It's also not quite accurate to say that it reverts it - we just use a different code path for KDU vs. OpenJPEG now.

I'm quoting your commit which states "Bring back the pre-aug 2024 byte estimation for j2c streams".

Why would the amount of needed file loaded be different between the two decoders? What "knock on" effects are you alluding to? And if these effects are a problem then why is KDU still being supported when OpenJPEG works?

I think this is going to cause KDU to under-decode textures again and then viewers using KDU will miraculously use less VRAM all of a sudden without telling users that all those 2k textures they are paying to upload are never being decoded fully in 2k on KDU viewers.

Geenz:
RE: 2k textures - those still appear to work to me.  And I do have quite the library to test with.  And the knock on appears to be overallocation on the KDU end of things - I'm still profiling and validating various parts of this part of the pipeline.

TommyTheTerrible:
> RE: 2k textures - those still appear to work to me. And I do have quite the library to test with. And the knock on appears to be overallocation on the KDU end of things - I'm still profiling and validating various parts of this part of the pipeline.

Alright, well it would be nice to have an explanation but I guess I don't deserve one.

I wish you luck with your tinkering.

PR #5839, opened 2026-05-19
Async inventory - cleaned up

marchcat:
"Clean" version of #4893 - the same diff, but squashed into three commits.

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -3296,11 +3296,11 @@
       <key>Comment</key>
       <string>Force viewer to skip legacy login inventory skeleton and rely on async AIS fetching.</string>
       <key>Persist</key>
-      <integer>1</integer>
+      <integer>0</integer>
       <key>Type</key>
       <string>Boolean</string>
       <key>Value</key>
-      <integer>1</integer>
+      <integer>0</integer>
     </map>
     <key>AsyncInventoryMaxConcurrentFetches</key>
     <map>
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -3291,6 +3291,72 @@
       <key>Value</key>
       <integer>255</integer>
     </map>
+    <key>ForceAsyncInventorySkeleton</key>
+    <map>
+      <key>Comment</key>
+      <string>Force viewer to skip legacy login inventory skeleton and rely on async AIS fetching.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>Boolean</string>
+      <key>Value</key>
+      <integer>1</integer>
+    </map>
+    <key>AsyncInventoryMaxConcurrentFetches</key>
+    <map>
+      <key>Comment</key>
+      <string>Maximum number of concurrent AIS fetches used during async inventory skeleton loading.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>U32</string>
+      <key>Value</key>
+      <integer>2</integer>
+    </map>
+    <key>AsyncInventoryNotifyMinInterval</key>
+    <map>
+      <key>Comment</key>
+      <string>Minimum seconds between inventory observer notifications while async inventory loading is active.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>F32</string>
+      <key>Value</key>
+      <real>0.05</real>
+    </map>
+    <key>AsyncInventoryCapsTimeout</key>
+    <map>
+      <key>Comment</key>
+      <string>Maximum seconds to wait for inventory capabilities during async skeleton loading.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>F32</string>
+      <key>Value</key>
+      <real>45.0</real>
+    </map>
+    <key>AsyncInventoryFetchTimeout</key>
+    <map>
+      <key>Comment</key>
+      <string>Maximum seconds for the entire async inventory skeleton fetch process.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>F32</string>
+      <key>Value</key>
+      <real>180.0</real>
+    </map>
+    <key>AsyncInventoryEssentialTimeout</key>
+    <map>
+      <key>Comment</key>
+      <string>Maximum seconds to wait for essential inventory folders during async loading.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>F32</string>
+      <key>Value</key>
+      <real>90.0</real>
+    </map>
     <key>ForceLoginURL</key>
     <map>
       <key>Comment</key>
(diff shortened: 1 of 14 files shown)


indra/newview/llasyncinventoryskeletonloader.cpp
```diff
@@ -0,0 +1,712 @@
...
+    request.mCachedVersion = cached_version;
+    request.mRequestDepth = depth;
+
+    if (essential)
+    {
+        mFetchQueue.push_front(request);
+    }
+    else
+    {
+        mFetchQueue.push_back(request);
```
akleshchev, line 553: I can understand essential folders, but why are we always fetching non-essential folders with minimal depth?

indra/newview/llasyncinventoryskeletonloader.cpp
```diff
@@ -0,0 +1,712 @@
...
+    {
+        LL_WARNS("AppInit") << "Async inventory skeleton loader failed after essential-ready: " << mFailureReason << LL_ENDL;
+    }
+    else
+    {
+        LL_WARNS("AppInit") << "Async inventory skeleton loader failed: " << mFailureReason << LL_ENDL;
+    }
+}
+
+bool LLAsyncInventorySkeletonLoader::hasFetchedCurrentOutfit() const
```
akleshchev, line 609: Might be insufficient. "Current outfit" consists of links, we probably want to fetch those items first, otherwise some of the logic will fail due to 'broken' links.

akleshchev (approved):

indra/newview/app_settings/settings.xml
```diff
@@ -3291,6 +3291,72 @@
...
+    <key>ForceAsyncInventorySkeleton</key>
+    <map>
+      <key>Comment</key>
+      <string>Force viewer to skip legacy login inventory skeleton and rely on async AIS fetching.</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>Boolean</string>
+      <key>Value</key>
+      <integer>1</integer>
```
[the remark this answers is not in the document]
akleshchev: Yeah, this probably was there for debugging.

indra/newview/llasyncinventoryskeletonloader.cpp
marchcat: After the essential folders load, the rest are fetched gradually to build the full tree. Minimal initial depth keeps loading fast, while deeper folders load only when expanded.

indra/newview/llasyncinventoryskeletonloader.cpp
marchcat: It’s fine, but the flow is indirect: `hasFetchedCurrentOutfit()` only checks the COF folder, while linked folders are fetched separately through the essential queue before readiness is marked.

indra/newview/llasyncinventoryskeletonloader.cpp
```diff
@@ -0,0 +1,712 @@
...
+        LL_WARNS("AppInit") << "Async inventory skeleton loader failed to fetch "
+                             << request_id << " (library="
+                             << (request.mIsLibrary ? "true" : "false") << ")" << LL_ENDL;
+        if (request.mEssential)
+        {
+            markFailed("AIS skeleton fetch returned no data for essential category " + request_id.asString());
+        }
+        processQueue();
+        return;
+    }
```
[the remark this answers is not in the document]
marchcat: Non-essential folders failing is non-critical; they'll be fetched on-demand later by normal LLInventoryModelFetch mechanisms when the user expands them. Retries would add complexity for minimal gain.

indra/newview/llasyncinventoryskeletonloader.cpp
akleshchev: > while deeper folders load only when expanded

That's not how it works. For filtering to work properly bulk fetch always gets started and will 'bulk' request inventory using high depth, doing more partial requests slows everything down and breaks things apart for the bulk fetch.

indra/newview/llasyncinventoryskeletonloader.cpp
akleshchev: I think we have a cof specific request that fetches both cof and linked items in one go to avoid 'cloud' avatars where viewer has to slowly collect outfit by parts from different folders.

indra/newview/llstartup.cpp
```diff
@@ -2022,12 +2179,18 @@ bool idle_startup()
         // gInventory.mIsAgentInvUsable is set to true in the gInventory.buildParentChildMap.
         gInventory.buildParentChildMap();
 
+        const bool inventory_usable = gInventory.isInventoryUsable();
+        if (inventory_usable && gInventory.isAsyncInventoryLoading())
+        {
+            gInventory.setAsyncInventoryLoading(false);
```
akleshchev, line 2185: Log this please or log changes in setAsyncInventoryLoading at info level.

akleshchev:
> idle_startup : Using legacy inventory skeleton payload

What region can I test this at?

Also this, not sure if this is a problem, might be an aditi thing:

```
2026-05-20T16:09:30Z WARNING #Inventory# newview/llinventorymodel.cpp(940) LLInventoryModel::findCategoryUUIDForTypeInRoot : Tried to find folder, type 23 but category does not exist
2026-05-20T16:09:30Z WARNING #Inventory# newview/llinventorymodel.cpp(940) LLInventoryModel::findCategoryUUIDForTypeInRoot : Tried to find folder, type 14 but category does not exist
2026-05-20T16:09:30Z WARNING #Inventory# newview/llinventorymodel.cpp(940) LLInventoryModel::findCategoryUUIDForTypeInRoot : Tried to find folder, type 3 but category does not exist
2026-05-20T16:09:30Z WARNING #Inventory# newview/llinventorymodel.cpp(940) LLInventoryModel::findCategoryUUIDForTypeInRoot : Tried to find folder, type 14 but category does not exist
```

akleshchev:
I tried force enabling async inventory (supposed to work?) and got broken link spam, like this:
`2026-05-20T16:49:47Z INFO #Inventory# newview/llinventorymodel.cpp(2682) LLInventoryModel::addItem : Adding broken link [ name: test itemID: 141d0017-9c5a-3792-aa38-072cc8cfd634 assetID: 3589a0d0-b17c-940d-5e1f-5e836f681ded )  parent: 7a49997e-cd49-4e4f-9be3-28b37cdd7c68`

marchcat:
> What region can I test this at?

Any region should work (the async path uses existing AIS API), just set ForceAsyncInventorySkeleton=1 in debug settings.

> `LLInventoryModel::findCategoryUUIDForTypeInRoot : Tried to find folder, type 23 but category does not exist`

Those are FT_LANDMARK (3), FT_TRASH (14), and FT_FAVORITE (23). In force-async mode, createCommonSystemCategories() runs before AIS has finished fetching, so findCategoryUUIDForType can't find them yet. This should be harmless, they get discovered once the root fetch is completed.

> I tried force enabling async inventory (supposed to work?) and got broken link spam

This is expected in force-async mode: items are loaded from cache but their link targets may not exist in the model yet (loaded in a different order or not yet fetched from AIS). Once the full skeleton loads, these links should resolve.
The spam is cosmetic, addItem() logs "broken link" but it doesn't discard them.

Dzonatas:
@marchcat I think there is a magic hole -- it climbs in then it tries to climb out (with surprise). One mouse could break the code by doing runs on the wheel for awhile. If leftover volts are collected, then it is overpowered now.

But, we shouldn't do magic holes. Yes, I know others that wanted them. (if they still do then dunno) They were strict for awhile, but now there is no need to extrude them. It may be just the story and joy catching up to the lore.

PR #5862, opened 2026-05-22
Clarify restricted script permission message

### LINKED ISSUE #3430: Incorrect error message on viewing limited permissions script

aidevelopers2:
Summary:
- follow up on #5822 by replacing the explicit no-copy/no-modify wording with the issue's more general restricted-permissions wording
- remove the inside-an-object wording so the same message is accurate for inventory scripts too

Fixes #3430

Testing:
- xmllint --noout indra/newview/skins/default/xui/en/floater_live_lsleditor.xml
- xmllint --noout indra/newview/skins/default/xui/en/panel_script_ed.xml

### DIFF
--- a/indra/newview/skins/default/xui/en/floater_live_lsleditor.xml
+++ b/indra/newview/skins/default/xui/en/floater_live_lsleditor.xml
@@ -15,7 +15,7 @@
   width="508">
   <floater.string
     name="not_allowed">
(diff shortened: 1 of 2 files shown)


aidevelopers2:
recheck

aidevelopers2:
Updated the branch onto the latest `develop` commit after #3430 landed upstream. The PR now only keeps the English wording change from the current base text, and the XML files still validate with `xmllint --noout`.

PR #5865, opened 2026-05-22
#3430 Strip the 'inside an object' from message as it is not object specific

### LINKED ISSUE #3430: Incorrect error message on viewing limited permissions script

akleshchev:
Small improvement for wording.


akleshchev:
@copilot please, update other language files.

maxim-productengine (approved):

PR #5904, opened 2026-06-06
Fix inaccurate "no copy" wording on restricted-permission scripts

### LINKED ISSUE #3430: Incorrect error message on viewing limited permissions script

jakethehoffer:
Fixes #3430.

The error shown when viewing a restricted-permission script said it was set to "no copy" or "no modify" even when the script is +copy, which is inaccurate. This updates the wording to "set with restricted permissions" (matching the phrasing Firestorm uses) and clarifies that copy and modify permissions are required, in both floater_live_lsleditor.xml (not_allowed) and panel_script_ed.xml (can_not_view).


jakethehoffer:

PR #5918, opened 2026-06-11
Add drag-n-drop reordering for clothing layers

### LINKED ISSUE #5929: Add drag-n-drop reordering for clothing layers
Contribution by [gwigz](https://github.com/gwigz)
https://github.com/secondlife/viewer/pull/5918

Ability to reorder clothes using drag&drop
https://github.com/user-attachments/assets/2581dc20-3368-4886-ac18-96ef003b66ec

gwigz:
This also includes changes for scroll containers to preserve and restore scroll position during layout and resize. Without it, the scroll would jump and place the first selected item at the bottom of the scroll area, instead of holding position while the selection remained visible.

https://github.com/user-attachments/assets/2581dc20-3368-4886-ac18-96ef003b66ec

### DIFF
--- a/indra/llui/llflatlistview.cpp
+++ b/indra/llui/llflatlistview.cpp
@@ -968,6 +968,26 @@ S32 LLFlatListView::getInsertIndexAt(S32 x, S32 y) const
     return index;
 }
 
+LLFlatListView::item_pair_t* LLFlatListView::getReorderPairAt(S32 x, S32 y) const
+{
+    S32 panel_x, panel_y;
+    localPointToOtherView(x, y, &panel_x, &panel_y, mItemsPanel);
+
+    for (item_pair_t* pair : mItemPairs)
+    {
+        if (!pair->first->getVisible()) continue;
+
+        // Claim the padding above each row so gap presses still resolve to a row.
+        LLRect rc = pair->first->getRect();
+        rc.mTop += mItemPad;
+        if (rc.pointInRect(panel_x, panel_y))
+        {
+            return pair;
+        }
+    }
+    return NULL;
+}
+
 S32 LLFlatListView::constrainInsertIndex(S32 dest_index) const
 {
     if (!mReorderValidateCallback) return dest_index;
@@ -1052,6 +1072,24 @@ bool LLFlatListView::handleHover(S32 x, S32 y, MASK mask)
         }
     }
     return LLScrollContainer::handleHover(x, y, mask);
+}
+
+bool LLFlatListView::handleMouseDown(S32 x, S32 y, MASK mask)
+{
+    bool handled = LLScrollContainer::handleMouseDown(x, y, mask);
+
+    // A press in the padding between rows misses every item, so the per-item
+    // mouse-down never arms a reorder. Map it to the nearest row and arm there;
+    // armReorderDrag's capture guard ignores presses a child control already took.
+    if (mAllowReorder && !mReorderDragPair && !(mask & (MASK_CONTROL | MASK_SHIFT)))
+    {
+        if (item_pair_t* pair = getReorderPairAt(x, y))
+        {
+            onItemMouseClick(pair, mask);
+        }
+    }
+
+    return handled || (mReorderDragPair != NULL);
 }
 
 bool LLFlatListView::handleMouseUp(S32 x, S32 y, MASK mask)
--- a/indra/llui/llflatlistview.cpp
+++ b/indra/llui/llflatlistview.cpp
@@ -910,7 +910,15 @@ void LLFlatListView::finishReorderDrag()
 
         if (mReorderCallback)
         {
-            mReorderCallback(moved_value, mReorderInsertIndex);
+            S32 visible_index = 0;
+            for (item_pair_t* pair : mItemPairs)
+            {
+                if (!pair->first->getVisible()) continue;
+                if (pair == mReorderDragPair) break;
+                ++visible_index;
+            }
+
+            mReorderCallback(moved_value, visible_index);
         }
     }
 
@@ -933,11 +941,19 @@ void LLFlatListView::clearReorderDragState()
     {
         getWindow()->setCursor(UI_CURSOR_ARROW);
     }
+
+    bool was_armed = (mReorderDragPair != NULL);
+
     mReorderDragPair = NULL;
     mReorderDragGroup.clear();
     mDeferredSelectPair = NULL;
     mIsReordering = false;
     mReorderInsertIndex = -1;
+
+    if (was_armed && mReorderEndedCallback)
+    {
+        mReorderEndedCallback();
+    }
 }
 
 void LLFlatListView::onMouseCaptureLost()
@@ -1108,13 +1124,6 @@ bool LLFlatListView::handleMouseUp(S32 x, S32 y, MASK mask)
         mDeferredSelectPair = NULL;
 
         finishReorderDrag();
-
-        // The grab ended (moved or not); let owners re-run work they deferred
-        // while a drag was active.
-        if (mReorderEndedCallback)
-        {
-            mReorderEndedCallback();
-        }
 
         if (was_reordering)
         {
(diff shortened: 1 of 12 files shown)


gwigz:

indra/llui/llscrollcontainer.cpp
```diff
@@ -197,11 +200,8 @@ void LLScrollContainer::reshape(S32 width, S32 height,
         bool show_h_scrollbar = false;
         calcVisibleSize( &visible_width, &visible_height, &show_h_scrollbar, &show_v_scrollbar );
 
-        mScrollbar[VERTICAL]->setDocSize( scrolled_rect.getHeight() );
-        mScrollbar[VERTICAL]->setPageSize( visible_height );
-
-        mScrollbar[HORIZONTAL]->setDocSize( scrolled_rect.getWidth() );
-        mScrollbar[HORIZONTAL]->setPageSize( visible_width );
+        preserveScrollbarMetrics(VERTICAL, show_v_scrollbar, scrolled_rect.getHeight(), visible_height);
+        preserveScrollbarMetrics(HORIZONTAL, show_h_scrollbar, scrolled_rect.getWidth(), visible_width);
```
akleshchev: Floaters don't seem to have issues with resizes on reshape. Does it really need to call setDocPos? setDocPos issues callbacks as is.

indra/llui/llscrollcontainer.cpp
```diff
@@ -586,6 +586,24 @@ bool LLScrollContainer::addChild(LLView* view, S32 tab_group)
...
+void LLScrollContainer::preserveScrollbarMetrics(EOrientation axis, bool show, S32 doc_size, S32 page_size)
+{
+    // Snapshot the position before the resize, but only while the scrollbar is
+    // visible: during a transient full-content pass it hides and its position is
+    // forced to 0, which would otherwise clobber the remembered position.
+    if (show && mScrollbar[axis]->getVisible())
+    {
+        mStoredDocPos[axis] = mScrollbar[axis]->getDocPos();
+    }
+    mScrollbar[axis]->setDocSize(doc_size);
```
akleshchev: This function seems needlesly complicated. setDocSize already calls setDocPos(mDocPos), so does setPageSize(), neither seem to modify mDocPos unless maximum position changed. In theory this can result in 3 callbacks from setDocSize?

indra/llui/llscrollcontainer.cpp
```diff
@@ -605,6 +623,9 @@ void LLScrollContainer::updateScroll()
     calcVisibleSize( &visible_width, &visible_height, &show_h_scrollbar, &show_v_scrollbar );
 
     S32 border_width = getBorderWidth();
+    preserveScrollbarMetrics(VERTICAL, show_v_scrollbar, doc_height, visible_height);
+    preserveScrollbarMetrics(HORIZONTAL, show_h_scrollbar, doc_width, visible_width);
```
akleshchev: I feel like this might cause issues with some views, where user refreshes the list (like a different search request), and instead of the starting point will end up in the middle of nowhere because refreshing cleaned the list but did not reset the position?

indra/newview/skins/default/xui/en/strings.xml
```diff
@@ -225,6 +225,7 @@ If you feel this is an error, please contact support@secondlife.com</string>
     <string name="Toycamera">Toy Camera</string>
 	
     <!-- Tooltip -->
+	<string name="ReorderClothingTooltip">Drag to reorder layers, or use the arrows</string> <!-- Hint shown on clothing rows that can be dragged to reorder -->
```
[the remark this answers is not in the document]
akleshchev: Not immediately relevant, copilot seems to be good with translating simple strings, so I can ask it to translate later.

indra/llui/llscrollcontainer.cpp
gwigz: worth making it opt-in behavior? it's likely only COF that needs this to retain a nice UX when re-ordering

gwigz: when COF reorder triggers a full [`refresh()`](https://github.com/secondlife/viewer/blob/cdf102b8d0a9a9513473705602b6ce6757136669/indra/newview/llcofwearables.cpp#L440) (clear + repopulate), the [clear](https://github.com/secondlife/viewer/blob/cdf102b8d0a9a9513473705602b6ce6757136669/indra/llui/llflatlistview.cpp#L416) zeroes the items-panel height, [reshapes](https://github.com/secondlife/viewer/blob/cdf102b8d0a9a9513473705602b6ce6757136669/indra/llui/llflatlistview.cpp#L438), hides the scroll, and zeroes itself here:

https://github.com/secondlife/viewer/blob/cdf102b8d0a9a9513473705602b6ce6757136669/indra/llui/llscrollcontainer.cpp#L660-L661

`refresh()` saved & restored scroll, but it was one-shot, and a later re-layout re-ran `updateScroll` and wiped it.

the `mStoredDocPos` is an attempt to work over that at the container level, which is possibly too broad? tried at a higher level but you'd see the scroll jump to the top for a frame, some very fiddly logic around it.

akleshchev: > worth making it opt-in behavior?

Probably. It does make sense for position to persist that way, but not all lists going to need or want it.

akleshchev: > the mStoredDocPos is an attempt to work over that at the container level, which is possibly too broad?

While it is too broad in the sense that nothing else uses it, having optional position restoration on container level does make sense. But it would be fine to have it at cof level.

gwigz: will take a shot at simplifying it

gwigz:
spotted two other issues which I've attempted fixes for, which I'm happy with

first one is that a second attempt to re-order could be swallowed by refresh, to fix that I've added some logic to hold refresh until `handleMouseUp` is fired

the other is that the gaps between items would not match anything on mouse-down, if you attempted to drag from here it wouldn't grab anything... so to work around that I've got a new `getReorderPairAt` for my `LLFlatListView` changes, which pushes that selection into `onItemMouseClick` on `handleMouseDown`

image

indra/llui/llscrollcontainer.cpp
gwigz: hopefully that'll do it, kept it scroll container side -- may be nice for possibly conversations list

indra/llui/llflatlistview.cpp
```diff
@@ -715,8 +782,347 @@ void LLFlatListView::onItemRightMouseClick(item_pair_t* item_pair, MASK mask)
     if ( !(mask & MASK_CONTROL) && mMultipleSelection && isSelected(item_pair) )
         return;
 
-    // else got same behavior as at onItemMouseClick
+    // else got same behavior as at onItemMouseClick, but a right click must never start a drag
+    mProcessingRightClick = true;
     onItemMouseClick(item_pair, mask);
+    mProcessingRightClick = false;
```
akleshchev: This looks weird.

Why not do something like `onItemMouseClick(item_pair, mask, CLICK_RIGTH /*EMouseClickType*/);` or onItemLeftMouseClick that calls 'armdrag' and onItemMouseClick?

indra/llui/llflatlistview.cpp
```diff
@@ -715,8 +782,347 @@ void LLFlatListView::onItemRightMouseClick(item_pair_t* item_pair, MASK mask)
...
+    mReorderDragPair = NULL;
+    mReorderDragGroup.clear();
+    mDeferredSelectPair = NULL;
+    mIsReordering = false;
+    mReorderInsertIndex = -1;
+}
+
+void LLFlatListView::onMouseCaptureLost()
+{
+    clearReorderDragState();
```
akleshchev, line 961: What about floater or viewer loosing focus? Shouldn't that be cleared too?

akleshchev (approved):

akleshchev:

Ansariel:
Please see my remarks in the merge commit 9aaf3e7747dd92ecdbce9878d3ac19a0027845c4

PR #5919, opened 2026-06-11
Gesture auto-complete, similar to emojis/mentions

### LINKED ISSUE #5928: Gesture auto-completion
Feature request with a contribution by [gwigz](https://github.com/gwigz)
https://github.com/secondlife/viewer/pull/5919

New gesture picker for nearby chat slash gesture autocomplete. Shown when the user types / nearby chat, matching active gestures instead of relying only on inline text completion.

SCR-20260611-jvuu

Preview only shown in nearby chat, for active `/` prefixed triggers, like `/wave`.

SCR-20260611-jvww

gwigz:
New gesture picker for nearby chat slash gesture autocomplete. Shown when the user types `/` nearby chat, matching active gestures instead of relying only on inline text completion.

- `LLGestureAutocompleteHelper` following the existing `LLChatMentionHelper`/`LLEmojiHelper` autocomplete pattern
- `LLFloaterGestureAutocompletePicker` and XUI for displaying matching gesture triggers and names.
- Wires nearby chat to build gesture matches from active `LLGestureMgr` entries
- Supports keyboard selection through the same `LLTextEditor` helper key-forwarding path used by other autocomplete helpers
- Respects the existing `ChatAutocompleteGestures` chat preference

SCR-20260611-jvuu

Preview only shown in nearby chat, for active `/` prefixed triggers, like `/wave`.

SCR-20260611-jvww

Also ignores anything that doesn't match gestures, as you'd expect.

### DIFF
--- a/indra/llui/llgestureautocompletehelper.h
+++ b/indra/llui/llgestureautocompletehelper.h
@@ -1,0 +1,83 @@
+/**
+ * @file llgestureautocompletehelper.h
+ *
+ * $LicenseInfo:firstyear=2026&license=viewerlgpl$
+ * Second Life Viewer Source Code
+ * Copyright (C) 2026, Linden Research, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation;
+ * version 2.1 of the License only.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Linden Research, Inc., 945 Battery Street, San Francisco, CA  94111  USA
+ * $/LicenseInfo$
+ */
+
+#pragma once
+
+#include "llhandle.h"
+#include "llsingleton.h"
+
+#include <boost/signals2.hpp>
+#include <functional>
+#include <string>
+#include <vector>
+
+class LLFloater;
+class LLUICtrl;
+
+class LLGestureAutocompleteHelper : public LLSingleton<LLGestureAutocompleteHelper>
+{
+    LLSINGLETON(LLGestureAutocompleteHelper) {}
+    ~LLGestureAutocompleteHelper() override {}
+
+public:
+    struct Row
+    {
+        std::string value;
+        std::string trigger;
+        std::string name;
+    };
+
+    bool isActive(const LLUICtrl* ctrl) const;
+    void showHelper(
+        LLUICtrl* host_ctrl,
+        const std::vector<Row>& rows,
+        size_t total,
+        const std::string& empty_text,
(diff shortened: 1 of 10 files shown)


indra/llui/llgestureautocompletehelper.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include <boost/signals2.hpp>
+#include <functional>
+#include <string>
+#include <vector>
+
+class LLFloater;
+class LLUICtrl;
+
+class LLGestureAutocompleteHelper : public LLSingleton<LLGestureAutocompleteHelper>
```
akleshchev, line 39: LLFloaterGestureAutocompletePicker is effectively a singleton, since only a single instance can exist.

This is a not a request to change this, just a note, that code might have been simpler if LLFloaterGestureAutocompletePicker was doing all this instead of managing both host and picker manualy.

indra/newview/llfloaterimnearbychat.cpp
```diff
@@ -501,6 +569,30 @@ void LLFloaterIMNearbyChat::onChatBoxKeystroke()
 
     KEY key = gKeyboard->currentKey();
 
+    if (gSavedSettings.getBOOL("ChatAutocompleteGestures"))
```
akleshchev: While not particularly frequent, might still be better to use LLCachedControl

akleshchev:
@maxim-productengine I think you are familiar with with autocompletion in chat, please check this over.

This likely needs approval from product before merge, I will create a ticket for triage.

maxim-productengine (changes requested):
I was not sure that `mGestureList->setDoubleClickCallback` is the correct approach, because the first click may cause focus loss for the chat editor and that triggers `hideHelper`, so the picker is closed before the double-click is actually handled. I built this branch and it is indeed the issue: double-clicking the gesture won't put it into the chat.

Probably you need to update the `mHostCtrlFocusLostConn = host_ctrl->setFocusLostCallback` lambda: when focus goes into the picker's own floater, restore it to the host control instead of hiding. This way the picker stays open and the double-click commit fires correctly.
On the other hand, you may want to get rid of the double-click callback and handle single click similar to the name-mention helper. Both approaches are fine with me.

And Andrey is right, we'll need product approval before merge.

gwigz:
thanks, that should put it in a better position now

the lambda is a bit more than what mentions has, as gestures is using a scroll list which seems to pinch focus when it's clicked... could do it differently but that'll possibly require edits to `LLScrollListCtrl`?

maxim-productengine:
Looks good

indra/newview/llfloaterimnearbychat.cpp
```diff
@@ -501,8 +570,34 @@ void LLFloaterIMNearbyChat::onChatBoxKeystroke()
 
     KEY key = gKeyboard->currentKey();
 
+    static LLCachedControl<bool> autocomplete_gestures(gSavedSettings, "ChatAutocompleteGestures");
```
akleshchev: please, add a default value, just in case, like:
static LLCachedControl<bool> autocomplete_gestures(gSavedSettings, "ChatAutocompleteGestures",true);

P.S. Got approval from UI-UX,

maxim-productengine (approved):

indra/newview/llfloaterimnearbychat.cpp
gwigz:
```suggestion
    static LLCachedControl<bool> autocomplete_gestures(gSavedSettings, "ChatAutocompleteGestures", true);
```

akleshchev:
Thank you for your contribution!

PR #5927, opened 2026-06-15
Depth-splice rigged attachment alpha into world alpha

Viscerous:
The viewer currently renders alpha in two separate passes: rigged alpha first, grouped by avatar pointer, followed by unrigged alpha back-to-front. Since most avatar alpha is rigged and most world alpha is not, the two cannot be ordered against each other. This causes familiar failures such as foreground hair rejecting trees behind it, while pointer-based avatar ordering can also differ between viewers.

This PR interleaves both streams in one back-to-front walk. Avatars and standalone animesh are keyed from their animated extents, while rigged attachments retain attachment-slot order. Non-HUD unrigged attachments inherit the wearer’s depth, so an avatar drains as one block: rigged first, followed by its unrigged groups back-to-front. This preserves the ordering content has historically been authored against within an avatar while giving the avatar a meaningful position among world alpha. Avatar depth lives in a dedicated `mAvatarDepth` key; the existing bounds-derived `mDepth` remains unchanged for world alpha and ordering within an attachment. Exact-depth ties use explicit ordering, with pointer identity only as the final tie-breaker. 

Interleaving is enabled by `RenderInterleavedAlpha` for the normal post-water world pass. Pre-water, HUD, shadow, reflection, cube and other special renders retain the legacy ordering. Disabling the setting restores the legacy two-pass rendering path.

QA note: Feature-off is not byte-for-byte identical to upstream ordering. Attachment stamps now fan out from a bridge to every alpha group in its linkset, so multi-group rigged attachments can sort differently and more consistently, even with interleaving disabled. Stamps on unrigged groups are inert in the legacy world sort.

https://github.com/user-attachments/assets/775bf84a-cafe-4bae-8ea7-f2ddd952b863


Ansariel:
@Geenz any opinion on this?

Geenz:
@Ansariel @Viscerous I will be taking a closer look at this tomorrow.  Hopefully prior to the open source meeting.  I have a few concerns with the technique, but I just need to squint at it in realtime before I make a judgement here.

Viscerous:
@Geenz I had some more time to test this. The main regression I found was standalone animesh: it could be composited in front of world alpha regardless of depth, while its own alpha ordering could also be disturbed. I factored the depth calculation into `LLVOAvatar::calcRiggedAlphaDepth()` and now stamp it onto standalone animesh spatial bridges from `LLControlAvatar::idleUpdate()`. Worn attachments and worn animesh remain covered by the wearer’s attachment loop.

I also covered some of the other scenarios: the shared alpha lists stay in legacy order and are re-sorted only immediately before the eligible post-water world pass. Pre-water, HUD, shadow, reflection, cube and feature-off rendering retain the legacy two-pass path. Exact-depth ties now follow the same ensemble order in both sorted streams and the merge.

Within an avatar or animesh ensemble it deliberately retains the established behaviour: rigged first, then unrigged groups back-to-front, while the ensemble as a whole interleaves with world alpha. Unstamped world groups keep their normal bounds depth.

Remaining weak point: alpha groups within alpha groups (e.g. avatar hair inside building with windows) don't have the granularity to resolve deterministically, these will be view-dependent; will likely not get cleaner without going finer-grain like OIT.

Shadowolf7:
I've patched this into my own Alchemy derived build, and it works a charm, with no perceivable performance impact. Bravo.

akleshchev:
Don't know why build status got stuck. Suggestion: merge from develop to retrigger it?

@Geenz Should this be aimed at the graphics care package?

PR #5951, opened 2026-06-24
#3988 Don't render malformed sculpts

### LINKED ISSUE #3988: Set RenderAutoHideSurfaceAreaLimit to a lower value, and base it on GPU capabilities.
I recently became aware of the RenderAutoHideSurfaceAreaLimit debug setting and see it could be quite useful for blocking objects that have excessive surface area from bogging down the renderer.
In my brief testing I found the setting's default value of 10 million is of little use. 1000 seems the right order of magnitude to block the griefer objects that abuse overdraw while allowing everything else to render as normal. My testing was done with some griefer objects on private region http://maps.secondlife.com/secondlife/Danger!%20Danger!/216/32/39

This debug setting could be a really useful parameter to use in the Improve Graphics Speed floater. It does need to be based on GPU capability as graphics cards may someday be able to render the 10 million (of whatever units) this setting uses.

## Test Plan

* Verify "sim-surrounds" and other large objects render correctly when this value is set to Default.
* Toggle this feature between off (value = 0) and Default and verify there is no visual different.
* Verify objects that have excessive surface area do not bog down the renderer to single digit FPS. (eg. http://maps.secondlife.com/secondlife/Danger!%20Danger!/216/32/39)

akleshchev:
Not sure if this is a good idea, but malformed sculpts hit performance hard and this change seems to work with the example I have.

### DIFF
--- a/indra/llmath/llvolume.cpp
+++ b/indra/llmath/llvolume.cpp
@@ -1971,6 +1971,8 @@ LLVolume::LLVolume(const LLVolumeParams &params, const F32 detail, const bool ge
     mNumHullIndices = 0;
 
     // set defaults
+    mSculptValidationCache.fill(LLSculptValidationState::Unvalidated);
+
     if (mParams.getPathParams().getCurveType() == LL_PCODE_PATH_FLEXIBLE)
     {
         mPathp = new LLDynamicPath();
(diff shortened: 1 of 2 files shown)


maxim-productengine (approved):

Dzonatas:
I wrote a sculpt importer, and I never synthesized points. I thought it was broken; I thought it required calculus to process, adapt, and bake. Now, I am calling this a feature instead of something malformed. I don't see the bug?

akleshchev:
The bug is a clearly 'malformed' sculpt that slows viewer from 200fps to 10fps by making viewer calculate a lot of 'occlusions' https://github.com/secondlife/viewer/issues/3988

P.S. I think we had a mesh with a similar problem, but wasn't able to figure out how that was handled.

Dzonatas:
The cinematic sculpts were a surprise. I digress.

PR #5988, opened 2026-07-10
Replace apr with boost in LLProcess

### LINKED ISSUE #5994: Replace apr in llprocess with boost::process
APR does not support unicode, has lacking exception coverage, lacking documentation, was made for win32 and whenever issues occur we are strugling to figure out what's going on. Additionally one of the tests periodically fails and it's hard to diagnose, as it only reproes on GH agents.

Note! This is mostly AI made, I added extra unit tests and reviewed it, but things might have slipped through.

## Test Plan

1. Make sure cef works without issues (search, destinations, login page, media on a prim)
2. Same for vls based media on a prim
3. Leap is LLProcess based, so if there abnormalities there (like during regression testing), this might be to blame.

akleshchev:
Replace apr's process logic with boost::process::v2


akleshchev:
Last failure: at least 1 of 1 LLLeap instances timed out (180 seconds) without terminating

marchcat (approved):

maxim-productengine (approved):

PR #5997, opened 2026-07-14
Add a multi-threaded compositor

Geenz:
This adds support for a multi-threaded compositor in the viewer to do things like separating the UI into its own thread, and eventually moving the renderer onto its own thread.


Geenz:
A few things to note:
- Lots of on viewer thread asserts.  Can probably back off of those - was kinda necessary initially given how much of the viewer assumed OS main (mostly in LLThread and friends, which had a knock on effect elsewhere).  This helped to track a ton of races.
- There's a good bit of race chasing in here due to that - so keep that in mind
- The actual bits you should review are gonna be in LLCompositor and some of our platform specific bits (it'd be nice if someone could provide feedback on the macOS stuff)
- The LLAppViewer compositable establishes a mailbox to get frames lined up for the compositor.  When VSync is off this is gonna be upwards of 3 frames.  When it's on, this is gonna be two.  We always have a back and front buffer, all the third frame enables is us to "not care" what's going on with the compositor.  Otherwise, compositor gets bounded on how fast the viewer can finish a frame
  - I anticipate this will eventually make it into a XUI (or whatever we end up with) refactor.

akleshchev (approved):

indra/llrender/llcompositor.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+void LLCompositor::attachToWindow(LLWindow* window, U32 width, U32 height)
+{
+    mWindow = window;
+    mSwapChain.attachToWindow(window, width, height);
+}
+
+void LLCompositor::createRefreshOverlay()
+{
+    // Runs on our own thread - the squares each spin up a producer
```
akleshchev, line 58: > our own thread

What does 'our own thread' means? It's own thread? main thread?

indra/llrender/llcompositable.h
```diff
@@ -0,0 +1,103 @@
...
+#include <string>
+
+class LLRenderTarget;
+
+// One layer in the compositor's stack. A compositable owns its front buffer
+// (an LLRenderTarget); the compositor takes a shared lease on it and samples
+// it during present.
+//
+// Neither side blocks the other: frames hand over via tryAcquireNewFront,
+// and the per-RT fence pair orders the GPU reads and writes.
```
akleshchev, line 43: Please add an in-code comment about what this actually is, the name isn't intuitive enough. Arguably name is too generic.

indra/llrender/llgpuresource.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+    else
+    {
+        llassert(false); // a lease holder outlived this resource
+    }
+}
+
+// - LLSharedLease ----------------------------------------------------------
+
+LLSharedLease::LLSharedLease(LLGPUResource* res)
```
akleshchev, line 51: Isn't this basically `std::shared_lock<std::shared_mutex>`?

PR #6015, opened 2026-07-17
Add transparent MOAP background and click pass-through support

### LINKED ISSUE #6027: Click pass-through through transparent media
Allows for media-on-a-prim to have its background fully transparent if the website sets it via CSS.
This allow click pass-through to in world objects if clicking a fully transparent area on the moap.

Contribution: https://github.com/secondlife/viewer/pull/6015

shadoskill:
## Description

This allows for media-on-a-prim to have its background fully transparent if the website sets it via CSS.
This also allow click pass-through to in world objects if clicking a fully transparent area on the moap.

---

## Checklist

---

## Additional Notes

Video of the feature and link to basic test page used in the video.

https://fullarray-dev.emptylist.co/moap_test.html

https://github.com/user-attachments/assets/4f9e9cc7-7ba5-4c30-85df-8e774caa438c

### DIFF
--- a/indra/llprimitive/llmediaentry.cpp
+++ b/indra/llprimitive/llmediaentry.cpp
@@ -53,6 +53,9 @@
 #define MEDIA_PERMS_INTERACT_KEY_STR     "perms_interact"
 #define MEDIA_PERMS_CONTROL_KEY_STR      "perms_control"
 
+// "display" fields
+#define MEDIA_TRANSPARENT_BACKGROUND_KEY_STR "transparent_background"
+
 // "general" fields
 const char* LLMediaEntry::ALT_IMAGE_ENABLE_KEY  = MEDIA_ALT_IMAGE_ENABLE_KEY_STR;
 const char* LLMediaEntry::CONTROLS_KEY          = MEDIA_CONTROLS_KEY_STR;
@@ -74,6 +77,9 @@ const char* LLMediaEntry::WHITELIST_KEY         = MEDIA_WHITELIST_KEY_STR;
 const char* LLMediaEntry::PERMS_INTERACT_KEY    = MEDIA_PERMS_INTERACT_KEY_STR;
 const char* LLMediaEntry::PERMS_CONTROL_KEY     = MEDIA_PERMS_CONTROL_KEY_STR;
 
+// "display" fields
+const char* LLMediaEntry::TRANSPARENT_BACKGROUND_KEY = MEDIA_TRANSPARENT_BACKGROUND_KEY_STR;
+
 #define DEFAULT_URL_PREFIX  "https://"
 
 // Constructor(s)
@@ -93,6 +99,7 @@ LLMediaEntry::LLMediaEntry() :
     // mWhiteList
     mPermsInteract(PERM_ALL),
     mPermsControl(PERM_ALL),
+    mTransparentBackground(false),
     mMediaIDp(NULL)
 {
 }
@@ -166,6 +173,9 @@ void LLMediaEntry::asLLSD(LLSD& sd) const
     // "permissions" fields
     sd[PERMS_INTERACT_KEY] = mPermsInteract;
     sd[PERMS_CONTROL_KEY] = mPermsControl;
+
+    // "display" fields
+    sd[TRANSPARENT_BACKGROUND_KEY] = mTransparentBackground;
 }
 
 // static
@@ -263,6 +273,16 @@ bool LLMediaEntry::fromLLSDInternal(const LLSD& sd, bool overwrite)
         status |= setPermsControl( 0xff & (LLSD::Integer)sd[PERMS_CONTROL_KEY] );
     }
 
+    // "display" fields
+    // TODO: When the simulator adds "transparent_background" to ObjectMedia
+    // Change this block to match all other fields:
+    //   if ( overwrite || sd.has(TRANSPARENT_BACKGROUND_KEY) )
+    // Until then leave it session based.
+    if ( sd.has(TRANSPARENT_BACKGROUND_KEY) )
+    {
+        status |= setTransparentBackground( sd[TRANSPARENT_BACKGROUND_KEY] );
+    }
+
     return LSL_STATUS_OK == status;
 }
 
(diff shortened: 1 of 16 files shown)


shadoskill:

akleshchev:
This is likely to affect existing content, may be even render some content unreadable due to matching color of the background. I wonder if we can get some form of 'respect transparent background' checkbox to work on per-face basis. It should be trivial viewer side to store the value in LLMediaEntry, but server might will need to support it.

indra/media_plugins/cef/media_plugin_cef.cpp
```diff
@@ -652,7 +652,7 @@ void MediaPluginCEF::receiveMessage(const char* message_string)
                 // SL-15560: Product team overruled my change to set the default
                 // embedded background color to match the floater background
                 // and set it to white
-                settings.background_color = 0xffffffff; // white
+                settings.background_color = 0x00000000; // transparent
```
callumlinden: This line means the default background color will be #000000 (black) so pages that do not specify a background color will inherit a black background.

Is that what you intended?

callumlinden (commented):
As well as my comment on the background color, I would echo @akleshchev's point about existing content. We do not want to make pages unclickable because someone specified a transparent background color in older content.

indra/media_plugins/cef/media_plugin_cef.cpp
shadoskill: That was not the intention no, I must have forgot to change that doing some testing.

As for existing content, I did some testing and yeah any blank page that does not have a CSS set background does show as fully transparent.

I added a toggle in media general settings to toggle it per face, and it _should_ be ready for a future server message if I did everything correctly.
For now the checkbox is session based for testing so it wont get set back to false from the server not knowing about it.

callumlinden: Wonderful - thank you for your excellent contribution.  We'll ask our product folk to look at it since it has some potential content impacts and once we get their blessing, we'll merge it in. 

Thank you.

Geenz:
Popping in here - what we might want to do instead is have a meta tag in the HTML that "opts in" to this rather than a flag from the simulator.  Would require a little more plumbing I think to get that meta tag out, but I think it might be worth it to go this way instead of finding a server release for a change to an already pretty packed wire protocol.

indra/llprimitive/llmediaentry.cpp
```diff
@@ -166,6 +173,9 @@ void LLMediaEntry::asLLSD(LLSD& sd) const
     // "permissions" fields
     sd[PERMS_INTERACT_KEY] = mPermsInteract;
     sd[PERMS_CONTROL_KEY] = mPermsControl;
+
+    // "display" fields
+    sd[TRANSPARENT_BACKGROUND_KEY] = mTransparentBackground;
```
Geenz, line 181: Note for if we stick with having an entry flag for this: you will want to extend unit tests to account for this.  The existing tests are failing.

Dzonatas:
I tested MOAP, and it is broken. (That is the determination made.) It slows the game down, but it is the ability to click on something to restore all or parts of commits (with payload) that we want to look at. They get deleted when spawned. It is a bug that needs fixing. MOAP is highly insecure, as in $US Dollars. Sometimes everybody waits for a fix.

DarlCat:
I've found a way to avoid breaking content while not requiring Second Life-specific opt-in behavior from websites to enable transparency.

https://github.com/DarlCat/secondlife-viewer/tree/MOAP

The downside though is that it currently requires a patch to dullahan to facilitate injecting a very small piece of javascript into the webpage which means javascript must be enabled in CEF.

https://github.com/DarlCat/dullahan/tree/transparent-background

Since there is not much of any use for this transparency feature without dynamic webpages IMO; while not ideal in the slightest the tie-in isn't the worst tradeoff.

A cleaner approach without injection or a platform-specific meta tag detection would be patching CEF directly, something I'm hesitant to do. I see something that makes Second Life web content require nonstandard meta tags as equally undesirable to patching CEF. I did search around for a formal or informal standard to adopt for flagging webpages as supporting non-opaque rendering, but found none.

I know that there is currently an embedded content project underway in the dullahan repo, so it is possible there may duplicated work or different solutions to shared issues on both sides; hopefully we can share ideas with each other along the way if so.

Screenshot 2026-08-03 at 7 52 51 AM

akleshchev:
@Dzonatas that isn't relevant for current PR, if you have performance problems with moap, I suggest reporting them with specific scenarios and examples.

Dzonatas:
@akleshchev It has a security issue that is known.  I was just thinking about loudness and how to resolve it and its conflicts. There is expected behavior. Mine is about compiling/calculus. I could continue this conversation, but I'm not soliciting it.

Geenz:
@DarlCat I really don't think that a bit of JS is the way to go here.  A single SL specific meta tag I think is completely fine in this scenario.  Either way, we need something that is SL specific that we need to account for _somehwere_, and I'm not fond of using a media entry flag here.  We need an author-time flag that a content creator can mark pages with as "hey, do this".

I should also note that not every page needs this - so a content creator might have one set of pages that does this, and another set that does not.  So doing this page by page is not a bad idea I believe.  Really at the end of the day, to me the ideal setup is going to be something embedded by the author in the HTML.

DarlCat:
Respectfully, isn't the background-color of the `<html>` element being translucent as defined by CSS an intentional-enough author signal?

I've had a potion of my career in web development and I subjectively believe that it is. However if a specific meta tag is still the lab's desired way to go with this that's the direction this PR can go.

Geenz:
We would need to review existing content that makes use of MOAP to ensure that it's not suddenly going semi-transparent - which I don't think there's a 0% chance of.  This is far from standard functionality for a web browser - and you encounter all sorts of weird edge cases with this sort of thing.  Even if by accident (as in "eh, it doesn't do anything so no harm leaving it in"), we can't really risk breaking content with this.

Dzonatas:
(Debugging) MOAP is not fun while amid a crowded audience. That's not my argument or effort, but it's my experience. I played around with it until it was clear to me... [snip]... and the math review kept me busy.

Again, thank you for the consideration.

PR #6016, opened 2026-07-17
SLua PV <- 26.3 merge

marchcat:
Keeping the PV up to date.


marchcat:
That’s correct. The action uses the workflow file from the target branch, so it should pass after the merge.

maxim-productengine (approved):

Issue #6009, opened 2026-07-16
Race condition between agent ObjectUpdate and AvatarAppearance messages


AndrewMeadows:
### Environment

develop branch

### Description

Henri Beauchamp says a race condition was introduced by the UDP packet flood mitigation in PR #5565 .  When the dual queue sytstem (reliable=high_priority, unreliable=low_priority) gets backed up very early in login/arrival an agent's **AvatarAppearance** message ("semi-reliable", which means "reliable" in this context) may get handled before its first **ObjectUpdate** message (unreliable?) which can cause the **AvatarAppearance** message to be ignored (doesn't know about agent) and the avatar "never appears" (I think he means they remain cloudy).

### Reproduction steps

1. Use a potato
2. Login to region with lots of agents and objects, a busy/full neighbor region would also help with this
3. Look for agents that are cloudy, maybe even your own.

AndrewMeadows:
An slightly abridged version of my conversation with Henri went like this:

[12:32 PM] **Henri Beauchamp**: @Leviathan Linden: I have a bug to report about your new dual-queue UDP messages processing code in the viewer (which I did backport to my viewer a few weeks ago): it causes out-of-order messages processing and, in particular, the first **ObjectUpdate** message creating an avatar can be processed too late and after the **AvatarAppearance** message to rez that avatar: since the avatar object is not yet created (not in object list), the **AvatarAppearance** is ignored and the avatar does not rez. I had to change the code to use only one queue to fix this issue (see http://sldev.free.fr/forum/viewtopic.php?f=4&t=2722#p13704 ). Conclusion: the dual queue design is inappropriate since there are assumptions (both server and viewer sides) on the order in which some messages must be sent and processed for things to work properly.
[12:32 PM] **URL Reader HUD**: http://sldev.free.fr/forum/viewtopic.php?f=4&t=2722#p13704 = "Cool VL Viewer forum - View topic - Package out of order and login message"
[12:34 PM] **Leviathan Linden**: I will have to look into that.   Was it merged already?  I need to check...
[12:34 PM] **Henri Beauchamp**: @Leviathan Linden; It's in viewer-develop already. Not sure if it got merged to release branches
[12:35 PM] **Leviathan Linden**: Ok.  I might need to jump on that.
[12:35 PM] **Leviathan Linden**: @Henri Beauchamp what particular messages were being handled out of order?
[...]

PR #6017, opened 2026-07-17
Defer processing AvatarAppearance until after first ObjectUpdate

AndrewMeadows:
This PR fixes issue #6009 where the high- and low-priority UDP message queue would exacerbate a race condition where the **AvatarAppearance** message would be processed before the agent was created by its first **ObjectUpdate**.  It also changes where the packet sequence ID is processed to reduce the WARNING logs about out-of-order packets caused by asymmetric priority during packet floods.

### DIFF
--- a/indra/llmessage/llpacketbuffer.h
+++ b/indra/llmessage/llpacketbuffer.h
@@ -49,11 +49,18 @@ public:
     void init(S32 hSocket);
     void init(const char* buffer, S32 data_size, const LLHost& host);
 
+    // Whether LLCircuitData::checkPacketInID() has already been run for this
+    // packet (done at socket-read time, before this packet was sorted into
+    // the high/low priority inbound queue).
+    bool    getPacketIDChecked() const          { return mPacketIDChecked; }
+    void    setPacketIDChecked(bool checked)    { mPacketIDChecked = checked; }
+
 protected:
     char    mData[NET_BUFFER_SIZE]; // packet data       /* Flawfinder : ignore */
     S32     mSize;                  // size of buffer in bytes
     LLHost  mHost;                  // source/dest IP and port
     LLHost  mReceivingIF;           // source/dest IP and port
+    bool    mPacketIDChecked = false;
 };
 
 #endif
--- a/indra/llmessage/message.cpp
+++ b/indra/llmessage/message.cpp
@@ -897,7 +897,7 @@ bool LLMessageSystem::isHighPriorityMessage(const LLPacketBuffer& pkt) const
     //
     // AvatarAppearance is "Low 158" which means it is stored in four bytes: 0xff 0xff 0x00 0x9E
     // the last two bytes represent 158 in a BigEndian U16
-    return !(header[1] == 255 && header[2] == 0 && header[3] == 158);
+    return header[1] != 255 && header[2] != 0 && header[3] != 158;
 }
 
 void LLMessageSystem::dropPackets(U32 num_to_drop)
@@ -1070,7 +1070,7 @@ S32 LLMessageSystem::bufferInboundPacket()
             // Skip genuine duplicate resends, same as checkMessages()/logValidMsg()
             // would do further downstream.
             bool recv_resent = (data[0] & LL_RESENT_FLAG) != 0;
-            if (!(recv_resent && cdp->isDuplicateResend(recv_packet_id)))
+            if (!recv_resent || !cdp->isDuplicateResend(recv_packet_id))
             {
                 cdp->checkPacketInID(recv_packet_id, recv_resent);
                 pkt.setPacketIDChecked(true);
--- a/indra/llmessage/message.cpp
+++ b/indra/llmessage/message.cpp
@@ -519,10 +519,11 @@ bool LLMessageSystem::checkMessages(LockMessageChecker&, S64 frame_count )
         bool recv_resent = false;
         S32 num_acks = 0;
         S32 true_rcv_size = 0;
+        bool recv_packet_id_checked = false;
 
         U8* buffer = mTrueReceiveBuffer;
 
-        mTrueReceiveSize = receivePacketOrDrop((char *)mTrueReceiveBuffer);
+        mTrueReceiveSize = receivePacketOrDrop((char *)mTrueReceiveBuffer, recv_packet_id_checked);
         // If you want to dump all received packets into SecondLife.log, uncomment this
         //dumpPacketToLog();
 
@@ -686,7 +687,7 @@ bool LLMessageSystem::checkMessages(LockMessageChecker&, S64 frame_count )
 
             if ( valid_packet )
             {
-                logValidMsg(cdp, host, recv_reliable, recv_resent, num_acks>0 );
+                logValidMsg(cdp, host, recv_reliable, recv_resent, num_acks>0, recv_packet_id_checked );
                 valid_packet = mTemplateMessageReader->readMessage(buffer, host);
             }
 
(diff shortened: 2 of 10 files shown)


indra/llmessage/llpacketbuffer.h
```diff
@@ -49,11 +49,18 @@ class LLPacketBuffer
...
+    // the high/low priority inbound queue).
+    bool    getPacketIDChecked() const          { return mPacketIDChecked; }
+    void    setPacketIDChecked(bool checked)    { mPacketIDChecked = checked; }
+
 protected:
     char    mData[NET_BUFFER_SIZE]; // packet data       /* Flawfinder : ignore */
     S32     mSize;                  // size of buffer in bytes
     LLHost  mHost;                  // source/dest IP and port
     LLHost  mReceivingIF;           // source/dest IP and port
+    bool    mPacketIDChecked = false;
```
AndrewMeadows, line 63: We now check the packet sequence ID when we pull it off of the socket.  This infrastructure is used to track whether the packet has already been checked or not when it comes time to process it.

indra/llmessage/message.cpp
```diff
@@ -866,19 +867,37 @@ bool LLMessageSystem::computeDrop()
...
+
+    // BUG: The low-frequency AvatarAppearance message will be ignored if its agent is unknown
+    // and the agent is created upon receipt of its first high-frequency ObjectUpdate message.
+    // This race condition will be exacerbated by our default prioritization strategy.
+    //
+    // WORKAROUND: All middle- and low-frequency messages are high-priority except AvatarAppearance
+    //
+    // AvatarAppearance is "Low 158" which means it is stored in four bytes: 0xff 0xff 0x00 0x9E
+    // the last two bytes represent 158 in a BigEndian U16
+    return !(header[1] == 255 && header[2] == 0 && header[3] == 158);
```
AndrewMeadows: This is the main fix: we categorize **AvatarAppearance** messages as low-priority (same as **ObjectUpdate**) to reduce the race condition where **AvatarAppearance** is processed too soon.

indra/newview/llviewermessage.cpp
```diff
@@ -4184,7 +4184,11 @@ void process_avatar_appearance(LLMessageSystem *mesgsys, void **user_data)
     }
     else
     {
-        LL_WARNS("Messaging") << "avatar_appearance sent for unknown avatar " << uuid << LL_ENDL;
+        // WORKAROUND: this can arrive before the ObjectUpdate that creates the avatar (e.g. if
+        // the ObjectUpdate was lost and is being resent). Defer it until the avatar shows up,
+        // instead of dropping it.
+        LL_WARNS("Messaging") << "avatar_appearance sent for unknown avatar " << uuid << "; deferring" << LL_ENDL;
+        LLVOAvatar::addPendingAvatarAppearance(uuid, mesgsys);
```
AndrewMeadows: There are still ways by which **AvatarAppearance** can arrive before the first **ObjectUpdate**.  Here we queue the **AvatarAppearance** data, give it an expiry, and process it later (or expire it) when new avatars are created.

indra/llmessage/message.cpp
Ansariel: `return header[1] != 255 || header[2] != 0 || header[3] != 158;`

indra/llmessage/message.cpp
```diff
@@ -1034,10 +1057,24 @@ S32 LLMessageSystem::bufferInboundPacket()
...
+            {
+                cdp->collectRAck(recv_packet_id);
+            }
+
+            // Check packet sequencing here, in true socket-arrival order, before
+            // this packet is sorted into the high/low priority inbound queue.
+            // Skip genuine duplicate resends, same as checkMessages()/logValidMsg()
+            // would do further downstream.
+            bool recv_resent = (data[0] & LL_RESENT_FLAG) != 0;
+            if (!(recv_resent && cdp->isDuplicateResend(recv_packet_id)))
```
Ansariel: `if (!recv_resent || !cdp->isDuplicateResend(recv_packet_id))`

akleshchev (approved):

indra/newview/llvoavatar.cpp
```diff
@@ -9686,6 +9717,56 @@ void LLVOAvatar::parseAppearanceMessage(LLMessageSystem* mesgsys, LLAppearanceMe
...
+        {
+            LL_WARNS("Messaging") << "Expired AvatarAppearance for agent " << timer_iter->first << LL_ENDL;
+            sPendingAvatarAppearanceContents.erase(timer_iter->first);
+            timer_iter = sPendingAvatarAppearanceExpiries.erase(timer_iter);
+        }
+        else
+        {
+            ++timer_iter;
+        }
+    }
```
AndrewMeadows: I just now realize: we should check expiries first otherwise we could process an expired appearance for `avatarp`. Meanwhile, the expiry duration I picked (20 seconds) is mostly arbitrary.  I know we can see UDP packets get delayed ~10 seconds on the dual-queues.  The viewer will re-ask for AvatarAppearance data on a 120 second timer, so I figure the duration should be between those values.

AndrewMeadows: Meanwhile **Henri Beauchamp** suggests (via IM): instead of storing the whole **AvatarAppearance** data just store the **agent_id** and then when we reach this context just fire off a new **avatartexturesrequest**.

indra/newview/llviewermessage.cpp
```diff
@@ -4181,10 +4181,18 @@ void process_avatar_appearance(LLMessageSystem *mesgsys, void **user_data)
...
     {
         avatarp->processAvatarAppearance( mesgsys );
+        return;
+    }
+    // Remember that this avatar needs to see its appearance re-requested when
+    // they will finally be created. HB
+    LLVOAvatar::registerEarlyAppearance(uuid);
+    if (uuid == gAgentID)
+    {
+        LL_WARNS("Messaging") << "AvatarAppearance message received before avatar is created." << LL_ENDL;
```
akleshchev: Perhaps clarify that Own message was received before own avatar was created? Is it even possible?

indra/newview/llvoavatar.cpp
```diff
@@ -779,6 +781,14 @@ LLVOAvatar::LLVOAvatar(const LLUUID& id,
     mVisuallyMuteSetting = LLVOAvatar::VisualMuteSettings(LLRenderMuteList::getInstance()->getSavedVisualMuteSetting(getID()));
 
     sInstances.push_back(this);
+
+    uuid_list_t::iterator it = sEarlyAppearanceList.find(id);
+    if (it != sEarlyAppearanceList.end())
+    {
+        sEarlyAppearanceList.erase(it);
+        LL_INFOS("Avatar") << "Re-requesting AvatarAppearance for new avatar " << id << LL_ENDL;
+        LLAvatarPropertiesProcessor::getInstance()->sendAvatarTexturesRequest(getID());
```
akleshchev, line 796: What cleans sEarlyAppearanceList if matching object id was never received?

indra/newview/llviewermessage.cpp
AndrewMeadows: Yes, I believe this can happen even for the viewer's avatar: for each agent the simulator will send **ObjectUpdate** and **AvatarAppearance** messages to the viewer controlling that agent.

indra/newview/llvoavatar.cpp
AndrewMeadows: Yes, `sEarlyAppearanceList `is a slight resource leak but we expect it to be small enough for each session to not cause problems.  I will add a comment about it.

akleshchev: Perhaps reset it on teleports between regions? But personally I would log something like 'received appearance for id, but avatar haven't arrived in 15 minutes'

AndrewMeadows: I don't mind resetting it on TP, but logging the duration of staleness would require adding arrival time info which is too much overhead IMO.  An orphan on the list will be a uncommon occurrence: it would require a lost packet right before an agent vanishes server-side.  It would typically happen not more than once or twice each session, which isn't enough memory to worry about.

PR #6021, opened 2026-07-18
Fix undefined symbol link error in llcorehttp

nibbbl:
These static consts in `HttpStatus` were not actually defined --fails with a link error periodically on Xcode 26.6 and maybe some other toolchains.


akleshchev:
QA can't test this and this is effectively a build fix, so won't make a ticket, will just take as is for now.

Thank you. for your contribution!

PR #6034, opened 2026-07-23
Cap automatic texture memory on Apple Silicon

LisaScheers:
## Summary

- cap automatic texture and vertex-buffer budgeting on ARM64 macOS to one eighth of physical unified memory
- preserve the existing VRAM divisor and automatic-budget behavior on Windows, Linux, and Intel macOS
- preserve explicit manual VRAM budgets
- add focused unit coverage for dedicated and unified-memory calculations

## Why

A native profile on a 48 GB Apple Silicon Mac showed the viewer detecting 38,338 MB as VRAM. Because that memory is shared with the CPU, treating most of it as independently available graphics memory allows the automatic budget to create avoidable system-memory pressure.

The new cap produces a 6,144 MB automatic budget on the profiled machine before the existing 80% safety target is applied. Non-unified-memory platforms continue to use the existing calculation unchanged.

## Validation

- compiled the production budget helper and ran all seven regression cases with Nix Clang using `-std=c++17 -Wall -Wextra -Werror`
- covered dedicated VRAM, zero-divisor, minimum-budget, 48 GB unified-memory, low-memory, and missing-memory-data cases
- `git diff --check`
- `xmllint --noout indra/newview/app_settings/settings.xml`

A full viewer build was not run because the profiling host does not have the Linden dependency bundle or an accepted full-Xcode license.

### DIFF
--- a/indra/newview/CMakeLists.txt
+++ b/indra/newview/CMakeLists.txt
@@ -609,6 +609,7 @@ set(viewer_SOURCE_FILES
     lltexturefetch.cpp
     lltextureinfo.cpp
     lltextureinfodetails.cpp
+    lltexturememorybudget.cpp
     lltexturestats.cpp
     lltextureview.cpp
     llthumbnailctrl.cpp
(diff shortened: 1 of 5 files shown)


akleshchev:
Note, there were some vram-adjusted changes in 26.3 and more might come with https://github.com/secondlife/viewer/pull/6028.

akleshchev:
project/graphics-care-package remakes budget tracking

PR #6037, opened 2026-07-23
#5796 Add CEF remote debugging port debug setting

### LINKED ISSUE #5796: Add CEF remote debugging port debug setting for test automation
## Summary

Add a debug setting that enables Chrome DevTools Protocol (CDP) remote debugging on the embedded CEF browser. This allows test automation tools (e.g. Playwright) to connect to the CEF instance and interact with web page content — filling forms, clicking links, reading DOM state, scrolling, etc.

## Dullahan Already Supports This

The Dullahan library (`dullahan.h`) already has the required fields in `dullahan_settings`:

```cpp
// allow remote debugging
int enable_remote_debug = false;

// allow Chrome (or other CEF window) to debug via http://localhost::PORT_NUMBER
int remote_debugging_port = 1964;
```

**No Dullahan changes are needed.** The viewer simply isn't setting these fields today, so `enable_remote_debug` defaults to `false` and the port is never opened.

## Request

Add a viewer debug setting (e.g. `CEFRemoteDebuggingPort`) that, when set to a non-zero port number, enables CEF remote debugging by passing the port value through the existing plugin message pipeline to Dullahan's `enable_remote_debug` and `remote_debugging_port` settings. Default should be `0` (disabled).

This follows the same pattern used by existing CEF settings like `CefVerboseLog` — read via `gSavedSettings`, sent to SLPlugin via plugin message, applied before `mCEFLib->init(settings)`.

### Enabling

Via Debug Settings UI: **Develop → Debug Settings → set the value → 9222**

Or via command line: `./secondlife --set CEFRemoteDebuggingPort 9222`

**Note:** Multiple SLPlugin processes may be spawned (one per media instance). If they share the same port setting, only the first will successfully bind.  Could some function be added to LEAP to return a list of assigned SLPlugin processes and ports so we could find the correct port to target?

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #5796</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @kurtlinden</code> (replace <code>20</code> with the amount, and <code>@kurtlinden</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

maxim-productengine:
Added debug setting to enable and set CEF remote debugging port.
Added Leap api to get the list of media plugins with related info, including debugging port.

### DIFF
--- a/indra/llplugin/llpluginclassmedia.cpp
+++ b/indra/llplugin/llpluginclassmedia.cpp
@@ -41,6 +41,8 @@ extern bool gHiDPISupport;
 #endif
 
 static int LOW_PRIORITY_TEXTURE_SIZE_DEFAULT = 256;
+static const U32 MIN_DEBUG_PORT = 1024;
+static const U32 MAX_DEBUG_PORT = 65535;
 
 static int nextPowerOf2( int value )
 {
@@ -963,6 +965,34 @@ void LLPluginClassMedia::showPageSource()
     sendMessage(message);
 }
 
+
+static U32 assignCefDebuggingPort()
+{
+    U32 base_port = gSavedSettings.getU32("CEFRemoteDebuggingPort");
+    if (base_port == 0)
+    {
+        return 0;
+    }
+    base_port = llclamp<U32>(base_port, MIN_DEBUG_PORT, MAX_DEBUG_PORT);
+
+    static U32 last_base_port = 0;
+    static U32 offset = 0;
+    if (base_port != last_base_port)
+    {
+        last_base_port = base_port;
+        offset = 0;
+    }
+
+    U32 new_port = base_port + offset;
+    if (new_port > MAX_DEBUG_PORT)
+    {
+        return 0;
+    }
+
+    ++offset;
+    return new_port;
+}
+
 void LLPluginClassMedia::setUserDataPath(const std::string &user_data_path_cache,
                                          const std::string &username,
                                          const std::string &user_data_path_cef_log)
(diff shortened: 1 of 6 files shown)


marchcat (approved):

indra/llplugin/llpluginclassmedia.cpp
```diff
@@ -974,6 +974,10 @@ void LLPluginClassMedia::setUserDataPath(const std::string &user_data_path_cache
 
     bool cef_verbose_log = gSavedSettings.getBOOL("CefVerboseLog");
     message.setValueBoolean("cef_verbose_log", cef_verbose_log);
+
+    U32 cef_remote_debugging_port = gSavedSettings.getU32("CEFRemoteDebuggingPort");
```
akleshchev: Should the value be sanity checked?

indra/newview/app_settings/settings.xml
```diff
@@ -16170,6 +16170,17 @@
...
+    <map>
+      <key>Comment</key>
+      <string>Enable CEF remote debugging on the specified port ( set 0 to disable)</string>
+      <key>Persist</key>
+      <integer>1</integer>
+      <key>Type</key>
+      <string>U32</string>
+      <key>Value</key>
+      <integer>0</integer>
+    </map>
```
akleshchev, line 16183: I assume it's a tcp port? I suggest clarifying and reminding about sane limits. And if you if you are making it persist, definetely needs validation, just to not wonder why something isn't working later (because cef took voice's port)

indra/llplugin/llpluginclassmedia.cpp
akleshchev: Also do multiple instances work fine sharing the same port? Shouldn't it round robing over some range?

akleshchev (approved):

indra/llplugin/llpluginclassmedia.cpp
maxim-productengine: Valid point. It works when I manually change debug setting before opening the new web floater, but it may be inconvenient.
On the other hand, round robin sounds like overkill for QA purpose (I may be wrong here).
So, I think incrementing port for each new CEF plugin should be enough.

indra/newview/app_settings/settings.xml
maxim-productengine: Yep, it's better not to make it persistent.

PR #6043, opened 2026-07-23
Two issues.

Rider-Linden:
1. Was static constructing the LLCacheControl before gSystemOptions
2. Not correctly parsing metadata back from the task inventory api.

### Testing
1. It doesn't crash at startup.
2. Correct VM target is shown in edit box after close and reopen.

### DIFF
--- a/indra/newview/llscripteditorws.cpp
+++ b/indra/newview/llscripteditorws.cpp
@@ -156,11 +156,11 @@ namespace
         }
         return std::string(s);
     }
+
+    LLCachedControl<bool>* OPT_ENABLESCRIPTEDITORWS(nullptr);
+    LLCachedControl<bool>* OPT_TIGHTINTEGRATION(nullptr);
+
 }
-
-LLCachedControl<bool> LLScriptEditorWSServer::sEnableScriptEditorWS(gSavedSettings, "ExternalWebsocketSyncEnable", false);
-LLCachedControl<bool> LLScriptEditorWSServer::sTightIntegration(gSavedSettings, "ExternalEditorTightIntegration", false);
-
 
 class LLPublishedPrimListener : public LLVOInventoryListener
 {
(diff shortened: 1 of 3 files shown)


indra/newview/llscripteditorws.cpp
```diff
@@ -221,6 +221,24 @@ LLScriptEditorWSServer::ptr_t LLScriptEditorWSServer::getServer()
             wsmgr.findServerByName(LLScriptEditorWSServer::DEFAULT_SERVER_NAME));
 }
 
+bool LLScriptEditorWSServer::isEnabled()
+{
+    if (!OPT_ENABLESCRIPTEDITORWS)
+    {
+        OPT_ENABLESCRIPTEDITORWS = new LLCachedControl<bool>(gSavedSettings, "ExternalWebsocketSyncEnable", false);
```
akleshchev: isEnabled is static, but OPT_ENABLESCRIPTEDITORWS isn't?

Probably could have used
```
static LLCachedControl<bool> sync_enabled(gSavedSettings, "ExternalWebsocketSyncEnable", false);
return sync_enabled;
```

Rider-Linden: I guess I could have made it static within the function.
(My initial try was static at the module level and that causes badness.)

akleshchev (approved):

indra/newview/llscripteditorws.h
```diff
@@ -354,6 +354,6 @@ class LLScriptEditorWSServer : public LLJSONRPCServer
     static constexpr F32 CLEANUP_INTERVAL = 60.0f; // seconds
     static constexpr F32 CONNECTION_TIMEOUT = 300.0f; // 5 minutes
 
-    static LLCachedControl<bool> sEnableScriptEditorWS;
-    static LLCachedControl<bool> sTightIntegration;
+    LLCachedControl<bool> *mEnableScriptEditorWS;
+    LLCachedControl<bool> *mTightIntegration;
```
marchcat:
```suggestion
```

mac build fails due to these unused fields.

marchcat (approved):

indra/newview/llscripteditorws.h
Rider-Linden: Yup.  Just pushed the fix.

PR #6044, opened 2026-07-24
Defer script editor settings initialization

marchcat:
Defers script editor settings initialization until first use, preventing a pre-main crash that caused Velopack installation to partially fail and the viewer not to launch.


marchcat:
Fixed in #6043, closing.

PR #6049, opened 2026-07-25
Show five-digit precision for texture transforms

### LINKED ISSUE #1861: Expose 5 digits of precision for texture scale/offset values in edit window instead of rounding to 3 digits
The viewer only displays 3 digits after the decimal point for texture scale and offset values in the edit window. However, the server safely supports float precision down to at least 0.00001 (0.12345) for these values in my experience. (I use this a lot for sprite maps based on inverse powers of 2 - 0.5, 0.25, 0.125, 0.0625, 0.03125. I'm not sure if gets any more precise server-side, since llGetTextureOffset returns a vector that always rounds to 5 digits.)

The viewer is unnecessarily rounding off 2 digits on these values. Firestorm does not, so that data is definitely sent by the server and can fit in the existing UI. This limits safe texture scale/offset editing in the viewer to 0.001 granularity. The viewer will still set the values as-typed, but rounds them off immediately in the textbox when saved, so if you accidentally hit Enter on the field again, it will overwrite it with the rounded value. I haven't found a setting that can disable this behavior.

It would help to expose an additional 2 digits (0.12345) in these fields so that they correctly display the full precision available for texture scales and offsets.

https://secondlife.canny.io/admin/board/feature-requests/p/expose-5-digits-of-precision-for-texture-scale-offset-values-in-edit-window-inst

volksec:
## Summary

- show five decimal places for texture scale and offset controls
- apply the same precision to diffuse, normal, and specular transforms
- preserve precise values in the editor instead of redisplaying them rounded to three decimal places

## Root cause

The transform spinners inherited the global three-decimal default. Values with greater precision could be submitted, but the controls immediately displayed a rounded value, which could overwrite the original value on a subsequent commit.

## Validation

- `pre-commit run --from-ref upstream/develop --to-ref HEAD`
- parsed `panel_tools_texture.xml` and verified `decimal_digits="5"` on all 12 scale/offset controls
- `git diff --check upstream/develop...HEAD`

Fixes #1861


Dzonatas:
Using IEEE numbers are not accurate. They get used mostly by default -- being the root cause. The only solution, still, is using decimals, but that takes more engineering work. C# has fully implemented decimals if you want to try them. You'll get more digits with decimals. Not a showstopper but it will matter.

volksec:
Thanks for the context, @Dzonatas. This PR is intentionally scoped to #1861: it changes the display precision of the existing `LLSpinCtrl` fields from the inherited three-decimal default to five decimal places. It does not claim to provide exact decimal arithmetic or change the underlying `F32` storage used by the Viewer.

Replacing the numeric representation would be a substantially broader architectural change and would be better handled separately. For this issue, the goal is to avoid the editor immediately redisplaying entered transform values rounded to only three decimal places, while remaining consistent with the existing Viewer data model. If the maintainers prefer a broader approach, I’m happy to follow their guidance.

akleshchev (approved):

akleshchev (approved):

akleshchev:
The last digit doesn't fit in the field. There seems to be plenty of space, so I suggest making sliders wide enough to fit whole value.
image
image

volksec:
Thanks for pointing this out. I widened all 12 texture scale/offset spinners in commit `97e326e13a` by increasing the control width and reducing the unused label area, so the full five-decimal value now has enough room to display.

I also re-ran the XML and pre-commit checks successfully.

akleshchev:
Please, line this up (make sizes uniform), also that's too big, only one digit was invisible, there was no need to make this so big
image

Also makes me wonder if PBR's transform are supposed to change as well or not.

Dzonatas:
There are issues, yes, especially with what leftovers that are available each day. My default response is to move and resize the tab. If it is more than serious, block the mail port. Offer some bottled water while we wait. Do something for about a half hour.  Now type in a review. It will be appreciated.

volksec:
Addressed in 3a7c775fa0.\n\n- Restored the uniform 265 px total width so the transform spinners line up with the neighboring controls.\n- Kept the increase compact: the label width is now 195 px (instead of 175 px), adding only enough field space for the extra digit.\n- Applied the same five-decimal precision and compact sizing to the PBR scale and offset controls.\n\nThe XML and pre-commit checks pass locally.

akleshchev:
@Steeltoe-Linden
image

Steeltoe-Linden (approved):

akleshchev:
Thank you for your contribution! merging.

PR #6064, opened 2026-07-29
#5907 Fix textures for "Save" button in edit daycycle floater for flat UI appearance

### LINKED ISSUE #5907: Flyout button still uses old 3D-style textures
The flyout button control still uses the old 3D-style textures and did not yet get updated with the flat design update. The button is currently used in panel_snapshot_local.xml - see attached screenshot.

https://secondlife.canny.io/admin/feedback/bug-reports/p/flyout-button-still-uses-old-3d-style-textures

Ansariel:
Additional fix for the appearance of the "Save" button in `floater_edit_day_cycle.xml':

grafik

Button itself isn't an actual flyout button, but appears as one. As such:
- Don't overlap buttons to fix weird discolorations and in case of pressed state an odd texture mismatch
- Switch up arrow with down arrow to for consistant look & feel of a flyout button

### DIFF
--- a/indra/newview/skins/default/xui/en/floater_edit_ext_day_cycle.xml
+++ b/indra/newview/skins/default/xui/en/floater_edit_ext_day_cycle.xml
@@ -622,11 +622,13 @@
             <button
                     follows="top|left"
                     height="23"
+                    image_selected="SegmentedBtn_Left_Selected"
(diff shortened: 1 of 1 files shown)


akleshchev (approved):

akleshchev:
Thank you!

PR #6083, opened 2026-08-03
p682 Fix GetStats webrtc call being in a wrong thread

akleshchev:
Moved the blocking call into signaling thread. Fixed terminate to not wait indefinetely.

Sadly this doesn't fix the issue of GetStats blocking after an OS Sleep event, but avoids it. Since signaling thread is blocked, had to make terminate not wait indefinetely. But given that viewer is expected to only show 'lost connection, quit?' after a sleep event (we have a 'do not sleep' flag in develop) it is a good enough mitigation for now.

### DIFF
--- a/indra/llwebrtc/llwebrtc.cpp
+++ b/indra/llwebrtc/llwebrtc.cpp
@@ -26,6 +26,9 @@
 
 #include "llwebrtc_impl.h"
 #include <algorithm>
+#include <chrono>
+#include <future>
+#include <thread>
 #include <string.h>
 #include "api/audio/create_audio_device_module.h"
 #include "api/audio_codecs/audio_decoder_factory.h"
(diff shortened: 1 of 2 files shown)


roxanneskelly (changes requested):
I noticed that the GetStats call into libwebrtc is being made on a viewer thread (the main thread.)  Calls into libwebrtc should not be made on the main viewer thread, but should instead be made on the appropriate worker, signaling, or network webrtc thread.  This might be what's going wrong here.

akleshchev:
GetStats internally marshals work to the signaling thread. Moving GetStats to signaling task deadlocks signaling thread.

akleshchev:
Dan says this resolves his problem with the usb headset. Bluetooth headset instead of freezing viewer breaks voice only.

roxanneskelly (approved):
LGTM

PR #6091, opened 2026-08-05
#6068 Prioritize the user's own attachment textures

### LINKED ISSUE #6068: Viewer should not lower the resolution of my avatar's textures simply because I zoomed out
When I zoom out far from my own avatar at a club or other texture-heavy areas, say to simply look around, when I return the camera to my own avatar a lot of the textures on my avatar - including materials - are lowered drastically in resolution and it takes the viewer upwards of 20-30 seconds before they go back to full. A resident's own avatar should never have the textures lowered in resolution unless it's specifically set in graphics > advanced, and should always take priority in staying at max quality according to graphics settings.

1. Go to a club that is avatar heavy, so lots of textures are around - shopping events will probably work here too

2. Wait for the viewer to load everything it can do - draw distance was 128 with textures at 2k maximum for this

3. When your own avatar has loaded fully, alt+cam onto yourself and zoom out about half way to the clip plane - just looking around the area for a few seconds will work here too

4. Return the camera to your avatar

5. Lots of your avatar's textures, excluding BoM, are now super blurry, and will take ages to load back again

Image 1 shows how my avatar is supposed to look > image 2 is how far I zoomed out (just across the room at this club) > image 3 shows how blurry my avatar's textures become after zooming back in, which stay like this for 20-30 seconds.

https://secondlife.canny.io/admin/feedback/bug-reports/p/viewer-should-not-lower-the-resolution-of-my-avatars-textures-simply-because-i-z

marchcat:
Prioritize textures used by the agent’s own attachments so they recover faster after zooming out in crowded scenes.
The bounded boost preserves normal resolution caps and memory-pressure behavior.

### DIFF
--- a/indra/newview/app_settings/settings.xml
+++ b/indra/newview/app_settings/settings.xml
@@ -11887,6 +11887,17 @@
       <key>Value</key>
       <real>8.0</real>
     </map>
+    <key>TextureSelfAvatarBoost</key>
+    <map>
+      <key>Comment</key>
+      <string>Coverage multiplier for textures used by the agent's own attachments. 4.0 gives approximately one mip of extra priority without pinning the texture or bypassing the maximum texture resolution.</string>
+      <key>Persist</key>
+      <integer>0</integer>
+      <key>Type</key>
+      <string>F32</string>
+      <key>Value</key>
+      <real>4.0</real>
+    </map>
     <key>TextureDecodeDisabled</key>
     <map>
       <key>Comment</key>
--- a/indra/newview/llviewertexturelist.cpp
+++ b/indra/newview/llviewertexturelist.cpp
@@ -31,6 +31,7 @@
 #include "llviewertexturelist.h"
 
 #include "llagent.h"
+#include "llvoavatarself.h"
 #include "llgl.h" // fot gathering stats from GL
 #include "llimagegl.h"
 #include "llimagebmp.h"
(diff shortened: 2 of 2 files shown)


akleshchev:
I'm not sure this is the right approach. This will prevent textures from discarding in a limited set of cases (user can still zoom out completely out of view, which this does not cover). And avatar if it is not in the view or far enough should discard.

If I remember correctly textures have an 'important enough' delay.  If texture just came into view it will increase resolution gradually, based of what's in the focus and time spent visible. the right approach might be to ignore the time delay for own avatar, so that texture would load immediately instead of not discarding in some cases, may be grant texture higher priority group.

Geenz:
I should also note that we already have a ton of texture streaming changes coming in GCP - and this should really be lumped into that instead.  Lets not do this one right now.

maxim-productengine (approved):

marchcat:
Re https://github.com/secondlife/viewer/pull/6091#issuecomment-5205444131:

That is a reasonable alternative, but I don’t think this approach prevents normal discarding. We intentionally avoid BOOST_AVATAR_SELF or pinning, so the texture remains an ordinary LOD texture subject to resolution limits, off-screen handling and memory pressure. The 4x multiplier gives it some extra headroom and higher fetch priority, so the user’s own attachments recover faster and are less likely to blur than other content.
A separate priority group or bypassing progressive refinement could achieve the same goal and that approach may fit the GCP texture-streaming work better. Both are valid ways to give the user’s avatar preferential treatment without forcing maximum resolution everywhere.

PR #6101, opened 2026-08-06
p#682 Fix shutdown crash flushing peer connection stats

roxanneskelly:
LLWebRTCImpl::terminate() only posted the peer connection closes to the signaling thread, so they ran behind the BlockingCall that releases the factory -- after the factory and device module were gone, or not at all if the thread was destroyed with the task still queued.

That matters for stats: PeerConnection::Close() flushes any in-flight GetStats request and delivers the report inline, and our callback walks mSignalingObserverList to hand it to the viewer.  By then the viewer's LLVoiceWebRTCConnection objects have been destroyed -- on shutdown they deliberately skip unsetSignalingObserver, on the assumption that llwebrtc::terminate() already finished the job -- so the stats callback reaches into freed memory.

Split the close out of LLWebRTCPeerConnectionImpl::terminate() into closeOnSignalingThread() and run it from a BlockingCall, so connections are closed and destroyed on the signaling thread before the factory, the device module or the observers go away.  Also take ownership of the connection list up front so the detached thread in the timeout path isn't racing the main thread over it, clear the observer lists as part of the close, and skip the observer notification for a stats report delivered while shutting down.

### DIFF
--- a/indra/llwebrtc/llwebrtc.cpp
+++ b/indra/llwebrtc/llwebrtc.cpp
@@ -404,8 +404,14 @@ void LLWebRTCImpl::terminate()
     auto done_promise = std::make_shared<std::promise<void> >();
     std::future<void> done_future = done_promise->get_future();
 
+    // Hand ownership of the connections to the shutdown thread.  Nothing on
+    // this thread may touch them afterwards -- in the timeout case below the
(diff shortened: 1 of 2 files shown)


Geenz (approved):
LGTM

PR #6156, opened 2026-08-19
Rider/runtime msging

Rider-Linden:
Changes to the way runtime messages and compiler errors are routed.

### DIFF
--- a/indra/newview/llpublishedobjectmgr.cpp
+++ b/indra/newview/llpublishedobjectmgr.cpp
@@ -29,7 +29,9 @@
 
 #include "llscripteditorws.h"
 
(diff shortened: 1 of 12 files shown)


marchcat (approved):

indra/newview/llpublishedobjectmgr.cpp
```diff
@@ -39,6 +41,12 @@
 
 namespace
 {
+    static const boost::regex LUAU_LOCATION_PATTERN(
```
akleshchev, line 44: This appears to be dupplicated in two places. Is there some common space this can live in?

akleshchev (approved):

Issue #6157, opened 2026-08-19
BugSplat Report: LLParticlePartition::getGeometry


sl-service-account:
## BugSplat Report Group 94245
[View Report Group](https://app.bugsplat.com/v2/group?stackKeyId=94245&database=SecondLife_Viewer_2018)
[View Report Details](https://app.bugsplat.com/v2/crash?id=1766000&database=SecondLife_Viewer_2018)
[View Report Attachments](https://app.bugsplat.com/v2/crash?id=1766000&database=SecondLife_Viewer_2018#attachments)

**Application:** Second Life Release
**Version:** 26.3.0.31203661088 (26.3.0.31203661088)
**Error Code:** SIGSEGV SEGV_MAPERR at 0x88

**Callstack:**
[...]

PR #6160, opened 2026-08-19
#6157 Crash at LLParticlePartition::getGeometry

akleshchev:
Crash was reported for apr and not Boost.Process v2, but underlaying issue should be the same, deadlock on fork()/malloc.


akleshchev:
More work needed. May be I will just make those LLProcess allocations happen on main thread.

PR #6210, opened 2026-08-27
Move GLTF files into own project

akleshchev:
Move gltf related, non-rendering files into own project.


Geenz:
Seems reasonable @akleshchev - have you tried a viewer build and uploaded a GLTF mesh with this yet?  Or taken a look at some PBR materials in-world with this change?

akleshchev:
> have you tried a viewer build and uploaded a GLTF mesh with this yet?

Local build is fine. Haven't tried GH one. This simply moves files, if something was wrong it would have failed. 

But I'm constantly postponing moving llgltfmaterial, which is why this is till draft. Sadly primitive project depends onto llgltfmaterial. and it is not easy to decouple.

PR #6217, opened 2026-08-31
Fix combobox detachment on scroll in Game Control preferences

Alistar-Snook:
## Problem
Opening the Input dropdown on a row in Game Control preferences,then scrolling the panel with the side scrollbar, left the
dropdown frozen in place instead of following the row visually detached from the UI underneath it.

## Cause
The dropdown's position was only computed once, at open time (fitInRect() in initCombobox()). Nothing re-evaluated it if the panel scrolled afterward.

## Fix
Track the cell's on-screen position when the dropdown opens. On each draw() call, compare it to the current position; if it has moved (panel scrolled), close the selector via the existing clearSelectionState(), which commits/restores the value cleanly.

## Testing
Manually tested on Windows (RelWithDebInfoOS): dropdown now closes cleanly on scroll instead of staying detached; normal selection without scrolling still works as before.

### DIFF
--- a/indra/newview/llpanelpreferencegamecontrol.cpp
+++ b/indra/newview/llpanelpreferencegamecontrol.cpp
@@ -52,6 +52,9 @@ namespace
     static LLScrollListCtrl* sSelectedGrid { nullptr };
     static LLScrollListItem* sSelectedItem { nullptr };
     static LLScrollListCell* sSelectedCell { nullptr };
+    static S32 sSelectedRow { -1 };
+    static S32 sSelectedCol { -1 };
+    static LLRect sSelectedOpenRect;
 
     // The combobox popup currently overlaying sSelectedCell (one of the four
     // action-mapping/output selectors below).  Stashed so applyGameControlInput()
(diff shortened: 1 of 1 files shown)


Alistar-Snook:

Alistar-Snook:
recheck

indra/newview/llpanelpreferencegamecontrol.cpp
```diff
@@ -1843,6 +1873,8 @@ void LLPanelPreferenceGameControl::clearSelectionState()
     sSelectedItem = nullptr;
```
AndrewMeadows, line 1878: The selector table now vanishes upon scrolling the table, however the cell remains highlighted although it is not "selected" as far as this UI logic is concerned.  Please add this before clearing `sSelectedGrid` to remove the highlight on the selected cell:
```
if (sSelectedGrid)
    {
        sSelectedGrid->deselectAllItems(true);
    }
```

AndrewMeadows (changes requested):

indra/newview/llpanelpreferencegamecontrol.cpp
Alistar-Snook: Thank you!

AndrewMeadows (approved):

PR #6218, opened 2026-08-31
p#700 Fix issue with tittle bar on Windows

akleshchev:
Fixed an issue with tittle bar hiding menu on windows,

### DIFF
--- a/indra/newview/llviewerwindow.cpp
+++ b/indra/newview/llviewerwindow.cpp
@@ -6094,10 +6094,10 @@ void LLViewerWindow::checkSettings()
         mStatesDirty = false;
     }
 
-    // We want to update the resolution AFTER the states getting refreshed not before.
     if (mResDirty)
     {
-        LLCoordScreen window_size;
+        // Deferred resolution update after states have been refreshed.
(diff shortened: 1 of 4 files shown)


maxim-productengine (approved):

indra/newview/llviewerwindow.cpp
```diff
@@ -6094,9 +6094,11 @@ void LLViewerWindow::checkSettings()
         mStatesDirty = false;
     }
 
-    // We want to update the resolution AFTER the states getting refreshed not before.
     if (mResDirty)
     {
+        // Deferred resolution update after states have been refreshed.
+#ifdef LL_DARWIN
+        // On macOS, LLWindow::getSize(LLCoordScreen) reports the backing view size.
         LLCoordScreen window_size;
```
Hecklezz: Like I mentioned in the 26.4 PR, wouldn't just changing this line to `LLCoordWindow window_size;` work for all platforms and not require platform specific checks?

akleshchev: You are right, thank you. I jumped the gun again(

PR #6193, opened 2026-08-25
Release/26.4

Geenz:
as of = 2026-09-11 for RC deploy
build = https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%23d91cd7a0-26.4
cohort = 26.4
deployed = https://github.com/secondlife/viewer/releases/tag/Second_Life_Release%23d91cd7a0-26.4
desired = 1000
relnotes:

# **26.4 Lua Editor**

The **Lua Editor Viewer** takes another important step toward a more capable scripting workflow in Second Life. This release improves communication between the Viewer and the newly published **Second Life VS Code extension**, restores an experimental **Linux Viewer**, and includes the latest WebRTC Voice and viewer-maintenance updates.

## **Lua Editor and VS Code Support**

Note: The Lua Project itself is still in Beta and only available on a hand full of regions on the maingrid. Please refer to the Wiki link below this section for more details on how to try out Lua today. 

The Lua Editor integration now provides a smoother, more reliable workflow between Second Life and Visual Studio Code.

* Improved live exploration of objects and linksets from VS Code.  
* Added better tracking when the selected object or linkset changes.  
* Improved opening, reading, editing, and saving scripts and notecards stored in objects.  
* Improved reconnection and subscription handling between the Viewer and external editors.  
* Added structured compiler diagnostics so errors can be associated more accurately with the affected script, line, and column.  
* Improved forwarding of runtime errors, debug messages, owner messages, and stack information to VS Code.  
* Added clearer and more consistent error responses when an external-editor request cannot be completed.  
* Added Viewer menu commands for working with the external script editor.

These improvements accompany the [**Second Life VS Code extension**](https://marketplace.visualstudio.com/items?itemName=lindenlab.sl-vscode-plugin) and its tighter integration with the Viewer. The related Viewer work is documented in [PR \#6104](https://github.com/secondlife/viewer/pull/6104) and [PR \#6156](https://github.com/secondlife/viewer/pull/6156). With VSCode plugin setup instructions here: https://github.com/secondlife/sl-vscode-plugin/blob/develop/doc/installation.md

The Viewer continues to support both LSL and Lua scripts, including language-specific syntax information supplied by supported regions. 

For information about using Lua in Second Life, visit the [Lua wiki page](https://wiki.secondlife.com/wiki/Lua_Alpha).

## **Linux Viewer Returns**

An experimental **Linux build of the Second Life Viewer** is available again with this release.

[...]

PR #6230, opened 2026-09-01
#6225 Fix issues with scrollbar in chat

### LINKED ISSUE #6225: [26.4] Unexpected srollbar in chat
### Description

Scenario:
- Past any long string of numbers into chat's input field
- Resize chat

Image

Specific to 26.4 and develop branches

akleshchev:
Make maxDrawableChars behavior match render

### DIFF
--- a/indra/llrender/llfontgl.cpp
+++ b/indra/llrender/llfontgl.cpp
@@ -349,8 +349,9 @@ S32 LLFontGL::render(const LLWString &wstr, S32 begin_offset, F32 x, F32 y, cons
         F32 x_offset = 0.0f;
         if (mFontFreetype->getFontWeight() > 0 && fgi->mChar >= '0' && fgi->mChar <= '9' && mFontFreetype->getMaxDigitWidth() > 0.0f)
         {
-            // use mXAdvance directly here, since we don't want to get max width instead.
-            x_offset = (mFontFreetype->getMaxDigitWidth() - fgi->mXAdvance) * 0.5f;
+            // getXAdvance will return max digit width.
+            // use mXAdvanceRaw directly here, since we don't want to get max width instead.
+            x_offset = (mFontFreetype->getMaxDigitWidth() - fgi->mXAdvanceRaw) * 0.5f;
         }
 
         // Draw the text at the appropriate location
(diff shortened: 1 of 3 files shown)


maxim-productengine (approved):

Hecklezz:
> From https://github.com/secondlife/viewer/pull/6193#discussion_r3908776274 : What's your UI scale? I was working at 1.0 and it looks like yours is bigger than 1, something might be scale dependent.

UI and HUD Scale are 1.0 and medium chat size. 
I tested your PR right now and something definitely isn't right 😅
<details><summary>Screenshots with current state of the PR-</summary>

image

image

</details>

indra/llrender/llfontgl.cpp
```diff
@@ -613,10 +614,19 @@ S32 LLFontGL::maxDrawableChars(const llwchar* wchars, F32 max_pixels, S32 max_ch
...
+    F32 scaled_max_pixels = (F32)S32_MAX;
+    if (llfinite(max_pixels))
+    {
+        const F64 scaled = std::ceil((F64)max_pixels * (F64)sScaleX); // F64 to avoid overflow.
+        if (scaled > 0.f && scaled < (F64)S32_MAX)
+        {
+            scaled_max_pixels = (F32)scaled;
+        }
+    }
+
```
Hecklezz: I believe this would be better and fix the issue in my last comment-

```suggestion
    F32 scaled_max_pixels = 0.f;
    if (!llfinite(max_pixels))
    {
        scaled_max_pixels = (F32)S32_MAX;
    }
    else
    {
        const F64 scaled = std::ceil((F64)max_pixels * (F64)sScaleX);
        if (scaled > 0.0)
        {
            scaled_max_pixels = (scaled < (F64)S32_MAX) ? (F32)scaled : (F32)S32_MAX;
        }
    }

```

akleshchev: Fixed, but the right way is probably to make it be S32 instead of F32. Everything uses S32, why does this one function  wants F32 pixels...

Hecklezz: Yeah I saw the force push right as I made the suggestion.
However, even with my above change, OR with your latest change I am still getting the horizontal bar-
image

Hecklezz:
Testing the most latest changes still unfortunately can cause the horizontal bar to appear.

The group chat that I see it most frequently impact is the "Firestorm Support English" group.
If you open up debug settings and go to- ConversationsListPaneWidth and set this to 190
then go to and observe- floater_rect_im_container
Move the conversation floater so Left becomes 0 and then resize it so Right becomes a value between 714->720
This should be able to get the bar to appear if you have the Firestorm Support English group open.

<details><summary>Screenshot example of above steps-</summary>

image

</summary>

Hecklezz:
Upon further investigation and retrying 26.3, that is behaving the exact same as 26.4 does with this PR's changes. Prior to this PR's changes, the issue in 26.4 is much more noticeable and pretty much always present. So at least this PR brings it back in-line with 26.3 behaviour.

So the issue caused by the https://github.com/secondlife/viewer/commit/97b0c99e29a60e858f7328094de7e59e5ad02278 commit is fixed, but there still appears to be a bug elsewhere that is also present in 26.3 that can cause the conversation floater to have a horizontal scroll-bar at certain sizes depending on the chat contents.

akleshchev:
> Upon further investigation and retrying 26.3

I was about to ask that) I suspect that the issue with 26.3 is in llceil: if we have 101.1 pixels of space, why are we drawing till 102 instead of 101? We should have been using trunc.

Hecklezz:
I believe I found the culprit to the remaining scrollbar issue mentioned in my last comment, and have put up a PR with the fix here- https://github.com/secondlife/viewer/pull/6241

So combining both this PR you did, and the one I did, I can no longer reproduce the horizontal scrollbar issue.
However the change I did _may_ impact other places in the UI elsewhere, but with some small testing I haven't observed anything wrong.

What you said regarding llceil might still be worth looking into as well though.

akleshchev:
Thank you!

> What you said regarding llceil might still be worth looking into as well though.

Copilot agrees that font might be shited into those pixels, but otherwise disagrees with me, says that I souldn't be cutting pixels out. But it's counter suggestion to 'sync' everything to use ceil, not just fonts, sounds like a tall order.

PR #6234, opened 2026-09-02
FIX - Keep moving linkset children aligned with focused root

### LINKED ISSUE #6253: Child prims and avatars of a moving linkset can appear dislocated from the root prim if the camera is focused on the root.

trish-sl:
Related to [this canny](https://feedback.secondlife.com/bug-reports/p/child-prims-and-avatars-of-a-moving-linkset-can-appear-dislocated-from-the-root)

Alt-clicking on a root prim of a moving object results in a long standing bug where the root will move faster than its child prims. This causes visual dislocation of any other moving parts of the object such as an avatar sitting on a vehicle or the wheels turning. This also extends to the nametag and voice visual indicator of an avatar.

This bug is 100% reproducible by taking any vehicle and focusing on the root while its in motion. The old tickets (MAINT-1742 MAINT-2247, MAINT-2275) are referenced in the code and mention that some stuff was reverted in the past due to doors breaking. The nature of the changes were different, and I could not spot any undesirable side effect of my version of this fix (including with doors). If anyone has history on these old tickets or suggestions to rework the nature of the fix, I'd love to hear it.

With this fix in place, focusing on the root of a moving object keeps all of its parts and seated avatars together as you would expect.

### DIFF
--- a/indra/newview/llvoavatar.cpp
+++ b/indra/newview/llvoavatar.cpp
@@ -3015,8 +3015,16 @@ void LLVOAvatar::idleUpdateVoiceVisualizer(bool voice_enabled, const LLVector3 &
                 }
             }
         }
-        mVoiceVisualizer->setPositionAgent(position);
+        idleUpdateVoiceVisualizerPosition(position);
     }//if ( voiceEnabled )
+}
+
+void LLVOAvatar::idleUpdateVoiceVisualizerPosition(const LLVector3 &position)
+{
+    if (mVoiceVisualizer)
+    {
+        mVoiceVisualizer->setPositionAgent(position);
+    }
 }
 
 static void override_bbox(LLDrawable* drawable, LLVector4a* extents)
--- a/indra/newview/llagentcamera.cpp
+++ b/indra/newview/llagentcamera.cpp
@@ -1651,6 +1651,73 @@ LLVector3d LLAgentCamera::calcFocusPositionTargetGlobal()
                             gPipeline.updateMoveDampedAsync(drawablep);
                         }
                     }
+
+                    // Updating only a moving linkset root leaves its children on the normal
+                    // update pass, which makes them appear dislocated from the root every few
+                    // frames when the root was updated first.
+                    // This is particularly noticeable on moving children or avatars (and avatar attachments).
+                    if (mFocusObject->isRoot())
+                    {
+                        for (LLViewerObject* childp : mFocusObject->getChildren())
+                        {
+                            LLDrawable* child_drawablep = childp ? childp->mDrawable.get() : nullptr;
+                            if (!child_drawablep || child_drawablep->isDead() || !child_drawablep->isActive())
+                            {
+                                continue;
+                            }
+
+                            child_drawablep->clearState(LLDrawable::EARLY_MOVE);
+
+                            if (childp->isSelected() ||
+                                child_drawablep->isState(LLDrawable::MOVE_UNDAMPED) ||
+                                !childp->getAngularVelocity().isExactlyZero())
+                            {
+                                gPipeline.updateMoveNormalAsync(child_drawablep);
+                            }
+                            else
+                            {
+                                gPipeline.updateMoveDampedAsync(child_drawablep);
+                            }
+
+                            // Also apply transform to any seated avatars
+                            if (LLVOAvatar* avatarp = childp->asAvatar())
+                            {
+                                if (LLJoint* root_jointp = avatarp->getRootJoint())
+                                {
+                                    root_jointp->touch();
+                                    root_jointp->updateWorldMatrixChildren();
+                                    const LLVector3 hud_name_pos =
+                                        avatarp->idleCalcNameTagPosition(root_jointp->getWorldPosition());
+                                    avatarp->idleUpdateNameTag(hud_name_pos);
+                                    avatarp->idleUpdateVoiceVisualizerPosition(hud_name_pos);
+                                }
+
+                                // ... and also the avatar's unrigged attachments (those are not
+                                // children of the vehicle, they are children of the avatar)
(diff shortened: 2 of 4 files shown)


indra/newview/llvoavatar.cpp
```diff
@@ -3015,10 +3015,18 @@ void LLVOAvatar::idleUpdateVoiceVisualizer(bool voice_enabled, const LLVector3 &
                 }
             }
         }
-        mVoiceVisualizer->setPositionAgent(position);
+        idleUpdateVoiceVisualizerPosition(position);
```
akleshchev, line 3018: This looks like it is from a different bug.

trish-sl: That was changed/refactored after the addition of `idleUpdateVoiceVisualizerPosition` since it does the same task. I could revert just that line though. mM goal was to mimic the pattern of the nametag by having a helper function to call for when you want to update the visual indicator.

akleshchev (approved):

Viscerous:
I already had a local fix for this issue so I figured I'd give some input on the differing approach. The existing movement update in `calcFocusPositionTargetGlobal()` occurs after the normal movement pass, so updating the focused drawable there can desynchronise render state and leave `EARLY_MOVE` state affecting subsequent frames. Cascading that update to the rest of the linkset improves focus, but the camera path becomes responsible for updating linksets, seated avatars, attachments, name tags, and voice indicators.

For my fork's approach, removing the late camera-owned movement update kept the linkset stable without needing the separate cascade updating states. I also found that children may retain drawable generation `-1`, `getRenderPosition()` then returns their logical object position rather than the current composed drawable position, which explains the remaining child-focus discrepancy. I have a smaller replacement patch and runtime test results available at https://github.com/secondlife/viewer/pull/6265 if you’d be open to comparing implementations before this merges.

trish-sl:
> I already had a local fix for this issue so I figured I'd give some input on the differing approach. The existing movement update in `calcFocusPositionTargetGlobal()` occurs after the normal movement pass, so updating the focused drawable there can desynchronise render state and leave `EARLY_MOVE` state affecting subsequent frames. Cascading that update to the rest of the linkset improves focus, but the camera path becomes responsible for updating linksets, seated avatars, attachments, name tags, and voice indicators.
> 
> For my fork's approach, removing the late camera-owned movement update kept the linkset stable without needing the separate cascade updating states. I also found that children may retain drawable generation `-1`, `getRenderPosition()` then returns their logical object position rather than the current composed drawable position, which explains the remaining child-focus discrepancy. I have a smaller replacement patch and runtime test results available at #6265 if you’d be open to comparing implementations before this merges.

The main reason I went this way with my implementation is to try to avoid changing the logic in hopes of minimizing the chance of breaking changes to existing content which had apparently happened in the past with other attempts.

Will give your version of the fix a try to compare, if there's no perceptible side effect with it, it would be simpler to take #6265 instead.

trish-sl:
Tested #6265 isolated from any other fixes and I couldn't find any side effect so far. 

I also refactored my implementation to be smarter in regards to the EARLY_MOVE state.
For getRenderPosition, I gated a bit more conservatively on the off chance it might cause side effects we are not expecting. 

Both fixes look good to me: on vehicles, seated avatars and their rigged/unrigged attachments, nametags and voice icons are properly smoothed, moving children of the linksets as well. There is no more stuttering/hiccups/dislocation of objects to be observed. I could not find any regression with other objects that have movement (doors, physical prims like socces balls, and other animated or physical objects)

QA might want to test both thoroughly and pick whichever they're most comfortable with.

akleshchev:
Not moving root objects outside out of pipeline seems like a better approach, but any idea why was that movement added in first place?

Viscerous:
It appears to be very old code that was seemingly already present in the 2007 import. I'm guessing the idea was to update the object's drawable before reading `getRenderPosition()` so the camera follows the latest rendered position. A later update exempted avatars from this so I suspect developers already recognised that those were on their own update path.

`gPipeline.updateMove()` runs before `gAgentCamera.updateCamera()` however, so the camera's extra update happens *after* the linkset's normal movement pass. Since it sets `EARLY_MOVE`, the next normal pass sees and clears that and skips the drawable's move, introducing the out of phase movement.

trish-sl:
Investigated further this week, and I am coming to a similar conclusion.

By calling `gPipeline.updateMoveDampedAsync`, the function `calcFocusPositionTargetGlobal` updates the focused root and causes the camera code to have rendering side effects which then forces the camera code to do all the defensive code to fix the rendering side-effects.

I've tried by daily-driving a version of the viewer without the whole movement code tied to the camera and I didn't manage to find a side effect caused by not updating focused objects during normal use. 
The only reason why I think the camera focus-object-update code would be useful is if something somehow needs `gPipeline.updateMove()` to happen after `gAgentCamera.updateCamera()` because objects were moved in-between calls .

Digging for such cases, I found exactly 1. It looks like in the idle loop, the joystick moves objects between rendering and camera updates which _could_ be one of the reason the camera focus tracking code was needed; see here: 
https://github.com/secondlife/viewer/blob/95942d7abea7d39f1faa5c623486f998850eac26/indra/newview/llappviewer.cpp#L5592
https://github.com/secondlife/viewer/blob/95942d7abea7d39f1faa5c623486f998850eac26/indra/newview/llappviewer.cpp#L5609
https://github.com/secondlife/viewer/blob/95942d7abea7d39f1faa5c623486f998850eac26/indra/newview/llappviewer.cpp#L5612

akleshchev:
> Digging for such cases, I found exactly 1. It looks like in the idle loop, the joystick moves objects between rendering and camera updates which could be one of the reason the camera focus tracking code was needed; see here:

Thank you, I will add joystick in build mode' to QA test plan, but code wise seems like it only sends position to server, doesn't actually move anything (manip tools will be responsible for local movement).

PR #6235, opened 2026-09-02
FIX - Closing minimized search floater should not reopen script floater

### LINKED ISSUE #6248: Inworld Notices randomly removed
### Description

Persisted notifications are trimmed wrongly, which causes the latest notifications to be dropped on relog instead of trimming starting from the oldest notifications.

Before logging out (4 notifs): https://gyazo.com/60b8ab529e447f7ba32623cc6372116a
After logging back in (2, the older ones): https://gyazo.com/44084cf468f849abd08059d2f95fbe47

For an easy test, change MaxPersistentNotifications to 2.
Note that texture picker does not permit minimization and simply hides.

Same ticket on firestorm: https://jira.firestormviewer.org/browse/FIRE-36502

trish-sl:
This is a minor UX issue found while editing several scripts at once.
Repro steps:
- Open the script editor
- Open the search floater (CTRL+F) for that script floater
- Minimize both floaters
- Close the search&replace floater
- Notice how closing the floater has re-maximized your script floater.

I could not find an issue/canny on this

### DIFF
--- a/indra/newview/llpreviewscript.cpp
+++ b/indra/newview/llpreviewscript.cpp
@@ -188,7 +188,6 @@ public:
 
     virtual bool hasAccelerators() const;
     virtual bool handleKeyHere(KEY key, MASK mask);
-    void closeFloater(bool app_quitting = false) override;
 
 private:
 
@@ -210,6 +209,7 @@ LLFloaterScriptSearch::LLFloaterScriptSearch(LLScriptEdCore* editor_core)
     mEditorCore(editor_core)
 {
     buildFromFile("floater_script_search.xml");
+    setCanMinimize(false);
 
     sInstance = this;
 
@@ -269,21 +269,6 @@ void LLFloaterScriptSearch::show(LLScriptEdCore* editor_core)
 LLFloaterScriptSearch::~LLFloaterScriptSearch()
 {
     sInstance = NULL;
-}
-
-void LLFloaterScriptSearch::closeFloater(bool app_quitting)
-{
-    // If this floater is minimized while dependent on another window,
-    // closing it should not restore the dependee.
-    if (isMinimized())
-    {
-        if (LLFloater* dependee = getDependee())
-        {
-            dependee->removeDependentFloater(this);
-        }
-    }
-
-    LLFloater::closeFloater(app_quitting);
 }
 
 // static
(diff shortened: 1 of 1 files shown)


akleshchev:
I'm not sure it's a good idea. And this new behavior needs a review from product.

 I checked some other 'dependee floaters':
1. Texture pickers can not be collapsed, instead it hides fully when parent colapses. That feels like the best behavior.
2. Avatar picker is very similar to other pickers, but it can be collapsed (if bugged) and is single instance, so it's easy to lose track of. If you have a bunch of floaters open and only one picker I would expect at least some indication of what floater picker belonged to. 

Suggestion:
A. Follow texture pickers example and hide search instead of collapsing. Bringing child up brings parent up either way, no point to show both.
B. At least make the collapsed floater's parent floater flash briefly after a child was closed.

trish-sl:
> I'm not sure it's a good idea. And this new behavior needs a review from product.
> 
> I checked some other 'dependee floaters':
> 
> 1. Texture pickers can not be collapsed, instead it hides fully when parent colapses. That feels like the best behavior.
> 2. Avatar picker is very similar to other pickers, but it can be collapsed (if bugged) and is single instance, so it's easy to lose track of. If you have a bunch of floaters open and only one picker I would expect at least some indication of what floater picker belonged to.
> 
> Suggestion: A. Follow texture pickers example and hide search instead of collapsing. Bringing child up brings parent up either way, no point to show both. B. At least make the collapsed floater's parent floater flash briefly after a child was closed.

I like suggestion A, I think it is the most sensible and least invasive solution -> Hide the search floater when a script is minimized so it doesnt result in 2 minimized floaters stacked on top of each other (seeing the minimized search floater is unhelpful and it is just ui clutter, and it indirectly fixes the targeted UX issue).
I'll rework this fix with suggestion A and push an update within the coming day(s).

If anyone else has input before I get to it, feel free to let me know.

akleshchev (approved):

indra/newview/llpreviewscript.cpp
```diff
@@ -209,6 +209,7 @@ LLFloaterScriptSearch::LLFloaterScriptSearch(LLScriptEdCore* editor_core)
     mEditorCore(editor_core)
 {
     buildFromFile("floater_script_search.xml");
+    setCanMinimize(false);
```
akleshchev, line 212: Probably better to specify this in xml, but since, for some odd reason, texture picker does it the same way, merging as is.
Thank you!

PR #6236, opened 2026-09-02
Change/Fix - Retain newest notifications when trimming persisted notifications

### LINKED ISSUE #6248: Inworld Notices randomly removed

trish-sl:
Originally reported on Firestorm's Jira as [FIRE-36502](https://jira.firestormviewer.org/browse/FIRE-36502), I could not find a related canny.

Persisted notifications are trimmed naively, which causes the latest notifications to be dropped on relog instead of trimming starting from the oldest notifications.

These screenshots were taken on firestorm, but the behavior is identical on the current LL viewer.

Testing with a `MaxPersistentNotifications `of `2` persisted notifications:

Before logging out (4 notifs): https://gyazo.com/60b8ab529e447f7ba32623cc6372116a
After logging back in (2, the older ones): https://gyazo.com/44084cf468f849abd08059d2f95fbe47 

With the fix:
Before logging out (3 notifs):  https://gyazo.com/3f89f7352ab10e171226cb63f1f45b7b 
After logging in (2, the newer ones): https://gyazo.com/75e3ce3b038253209692a9d52bce7ac8 

This fix is a change in behavior, but I would argue it is the desired one.

### DIFF
--- a/indra/newview/llpersistentnotificationstorage.cpp
+++ b/indra/newview/llpersistentnotificationstorage.cpp
@@ -61,9 +61,9 @@ void LLPersistentNotificationStorage::saveNotifications()
         return;
     }
 
-    const S32 max_to_save = llmax(1, gSavedSettings.getS32("MaxPersistentNotifications"));
-    std::vector<LLNotificationPtr> selected_notifications;
-    selected_notifications.reserve(max_to_save);
+const S32 max_to_save = llmax(1, gSavedSettings.getS32("MaxPersistentNotifications"));
+std::vector<LLNotificationPtr> selected_notifications;
+selected_notifications.reserve(std::min<size_t>(static_cast<size_t>(max_to_save), history_channel->size()));
 
     auto history_begin = history_channel->beginHistory();
     auto history_it = history_channel->endHistory();
--- a/indra/newview/llpersistentnotificationstorage.cpp
+++ b/indra/newview/llpersistentnotificationstorage.cpp
@@ -65,29 +65,30 @@ void LLPersistentNotificationStorage::saveNotifications()
     std::vector<LLNotificationPtr> selected_notifications;
     selected_notifications.reserve(std::min<size_t>(static_cast<size_t>(max_to_save), history_channel->size()));
 
-    auto history_begin = history_channel->beginHistory();
-    auto history_it = history_channel->endHistory();
-    while (history_it != history_begin)
+    for (auto history_it = history_channel->rBeginHistory(), history_end = history_channel->rEndHistory();
+        history_it != history_end;
+        ++history_it)
     {
-        --history_it;
         LLNotificationPtr notification = *history_it;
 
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

akleshchev (approved):

indra/newview/llpersistentnotificationstorage.cpp
```diff
@@ -51,40 +54,54 @@ void LLPersistentNotificationStorage::saveNotifications()
...
-    LLSD output = LLSD::emptyMap();
-    LLSD& data = output["data"];
+    const S32 max_to_save = llmax(1, gSavedSettings.getS32("MaxPersistentNotifications"));
+    std::vector<LLNotificationPtr> selected_notifications;
+    selected_notifications.reserve(std::min<size_t>(static_cast<size_t>(max_to_save), history_channel->size()));
 
-    for ( std::vector<LLNotificationPtr>::iterator it = history_channel->beginHistory(), end_it = history_channel->endHistory();
-        it != end_it;
-        ++it)
+    auto history_begin = history_channel->beginHistory();
```
akleshchev: I think introducing LLPersistentNotificationChannel::rBeginHistory() { sortHistory(); return mHistory.rbegin(); } with a 'for' loop would have been much cleaner. While --history_it would work thanks to this being a vector, the solution and overal structure is confusing at first glance, it is not easy to read.

indra/newview/llpersistentnotificationstorage.cpp
```diff
@@ -51,40 +54,54 @@ void LLPersistentNotificationStorage::saveNotifications()
...
-        LLNotificationPtr notification = *it;
+        --history_it;
+        LLNotificationPtr notification = *history_it;
 
         // After a notification was placed in Persist channel, it can become
-        // responded, expired or canceled - in this case we are should not save it
-        if(notification->isRespondedTo() || notification->isCancelled()
-            || notification->isExpired())
+        // responded, expired or canceled - in this case we should not save it
+        if (!(notification->isRespondedTo() || notification->isCancelled() || notification->isExpired()))
```
akleshchev: Perhaps use &&?
!( A || B || C ) == !A && !B && !C
I feel like old if-continue was simpler thus easier to read.

maxim-productengine (approved):

akleshchev:
Thank you, merged!

PR #6238, opened 2026-09-02
FIX - Follow Camera stays stuck when unexpectedly unsat

### LINKED ISSUE #6254: When avatar on vehicle cannot enter a parcel, avatar is left in a stuck state.
Strange but repeatable behavior when an avatar cannot enter a parcel but a vehicle can. Avatar gets knocked off the vehicle and is left sitting in midair, unable to walk or stand.

https://secondlife.canny.io/admin/feedback/server-bugs/p/when-avatar-on-vehicle-cannot-enter-a-parcel-avatar-is-left-in-a-stuck-state

trish-sl:
When you cross a region and a vehicle gets deleted due to "The region you are trying to enter is currently experiencing issues" (and possibly in other cases as well), your avatar becomes forcibly unseated and may remain connected to the original sim.
In those situations, the viewer does not properly clear the active camera, making it impossible to turn your camera at all. The only way to fix this broken state is to sit on a seat with a scripted camera and unsit. (But most people end up relogging). Friends and I have experienced this bug in other situations as well, all of which were due to forced unsits.

Due to the nature of the bug, it is very difficult for me to test it, I've been sitting on this fix for half a year at this point. However, I've recently come across the region issue again and my camera was not trapped unlike the other passengers of the busted vehicle, I've also not had a camera stuck problem since including this fix which seems to prove it was a success. 

The fix comes in two parts -> 
1) avoid calling getOffObject() directly (my theory is that when it gets called, mDrawable is null so none of the logic the function does is performed) and instead setParent to NULL so it calls getOffObject and resets camera explicitly.  
2) recover from a missing object you got up from if you are not seated while getOffObject is called  (likely a bad state).

There may be more than one canny on the topic, but I have found this related canny: https://feedback.secondlife.com/server-bugs/p/when-avatar-on-vehicle-cannot-enter-a-parcel-avatar-is-left-in-a-stuck-state

### DIFF
--- a/indra/newview/llfollowcam.cpp
+++ b/indra/newview/llfollowcam.cpp
@@ -845,6 +845,11 @@ void LLFollowCamMgr::removeFollowCamParams(const LLUUID& source)
     delete params;
 }
 
+void LLFollowCamMgr::clearActiveFollowCamParams()
+{
+    mParamStack.clear();
+}
+
 bool LLFollowCamMgr::isScriptedCameraSource(const LLUUID& source)
 {
     param_map_t::iterator found_it = mParamMap.find(source);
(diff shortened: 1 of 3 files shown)


akleshchev (approved):

akleshchev:
Merged, thank you!

Issue #6255, opened 2026-09-03
Skinned Mesh Shader Optimizations


naitro2010:
Some processing that used to run on the GPU for every skinned vertex can be moved into a one time precomputed step on the CPU to hopefully improve frame rates with lots of visible avatars.
I'm making a draft pull request with my changes now for review and testing.

## Test Plan

A/B testing with multiple different GPU vendors and computers will be needed as well as verifying that rendering doesn't break on different platforms.

PR #6256, opened 2026-09-03
Skinned Mesh Shader Optimizations and C++ changes to support

naitro2010:
## Description

This PR moves some processing that used to run on the GPU for every skinned vertex into a precomputed step on the CPU to hopefully improve frame rates with lots of visible avatars.

Here is the linked issue:  https://github.com/secondlife/viewer/issues/6255

---

## Checklist

- [Y] I have provided a clear title and detailed description for this pull request.
- [Y] If useful, I have included media such as screenshots and video to show off my changes.
- [Y] The PR is linked to a relevant issue with sufficient context.
- [M] I have tested the changes locally and verified they work as intended.  (I cherrypicked the changes from a local branch with my other modifications to Alchemy Viewer.  They work on Alchemy Viewer with my local build.)
- [?] All new and existing tests pass.
- [?] Code follows the project's style guidelines.
- [?] Documentation has been updated if needed.
- [Y] Any dependent changes have been merged and published in downstream modules
- [Y] I have reviewed the [contributing guidelines](https://github.com/secondlife/viewer/blob/develop/CONTRIBUTING.md).

---

## Additional Notes
This is my first time creating a pull request for Second Life.
Please let me know if I can do anything to improve the next pull requests I make or if there are any code quality issues that I can fix.
I mostly wrote these changes to improve the frame rate for my new version of the viewer with Stereo 3D Virtual Reality rendering that I've been working on but I think they might help with normal rendering too.
I don't have a large number of different GPUs to test on so I'm leaving this pull request as a draft until more people are able to test it on their hardware.

### DIFF
--- a/indra/newview/app_settings/shaders/class1/avatar/objectSkinV.glsl
+++ b/indra/newview/app_settings/shaders/class1/avatar/objectSkinV.glsl
@@ -24,49 +24,30 @@
 
 in vec4 weight4;
 
-uniform mat3x4 matrixPalette[MAX_JOINTS_PER_MESH_OBJECT];
+uniform mat4 matrixPalette[MAX_JOINTS_PER_MESH_OBJECT];
 
 mat4 getObjectSkinnedTransform()
 {
-    int i;
 
-    vec4 w = fract(weight4);
     vec4 index = floor(weight4);
-
-    index = min(index, vec4(MAX_JOINTS_PER_MESH_OBJECT-1));
-    index = max(index, vec4( 0.0));
-
-    w *= 1.0/(w.x+w.y+w.z+w.w);
+    vec4 w   = fract(weight4)*2.0;
 
     int i1 = int(index.x);
     int i2 = int(index.y);
     int i3 = int(index.z);
     int i4 = int(index.w);
-
-    mat3 mat = mat3(matrixPalette[i1])*w.x;
-         mat += mat3(matrixPalette[i2])*w.y;
-         mat += mat3(matrixPalette[i3])*w.z;
-         mat += mat3(matrixPalette[i4])*w.w;
-
-    vec3 trans = vec3(matrixPalette[i1][0].w,matrixPalette[i1][1].w,matrixPalette[i1][2].w)*w.x;
-         trans += vec3(matrixPalette[i2][0].w,matrixPalette[i2][1].w,matrixPalette[i2][2].w)*w.y;
-         trans += vec3(matrixPalette[i3][0].w,matrixPalette[i3][1].w,matrixPalette[i3][2].w)*w.z;
-         trans += vec3(matrixPalette[i4][0].w,matrixPalette[i4][1].w,matrixPalette[i4][2].w)*w.w;
-
+     
     mat4 ret;
-
-    ret[0] = vec4(mat[0], 0);
-    ret[1] = vec4(mat[1], 0);
-    ret[2] = vec4(mat[2], 0);
-    ret[3] = vec4(trans, 1.0);
-
-    return ret;
-
+    ret = ((matrixPalette[i1] * w.x)
+          + (matrixPalette[i2] * w.y)
+          + (matrixPalette[i3] * w.z)
+          + (matrixPalette[i4] * w.w));
+     return ret;
 #ifdef IS_AMD_CARD
-   // If it's AMD make sure the GLSL compiler sees the arrays referenced once by static index. Otherwise it seems to optimise the storage awawy which leads to unfun crashes and artifacts.
-   mat3x4 dummy1 = matrixPalette[0];
-   mat3x4 dummy2 = matrixPalette[MAX_JOINTS_PER_MESH_OBJECT-1];
-#endif
-
+     // If it's AMD make sure the GLSL compiler sees the arrays referenced once by static index. Otherwise it seems to optimise the storage away which leads to unfun crashes and artifacts.
+     mat4 dummy1 = matrixPalette[0];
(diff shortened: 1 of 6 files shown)


naitro2010:

naitro2010:
Is there a way to make a release build for this pull request with GitHub actions that I can download and test?  I've only tested with a local build so far.

Thanks :)

akleshchev:
> Is there a way to make a release build for this pull request with GitHub actions that I can download and test?

There is a full, downloadable build at the bottom?

naitro2010:
I just merged the latest changes from develop because the newer LLSD parsing integration test failed to pass.

naitro2010:
Apparently GitHub is having internal server errors when downloading dependencies now 😢

akleshchev:
@naitro2010 there is a green checkbox near the commit, click it, it will show a list of 'checks', that's the build, open any OS specifc one and go to summary.

Artifacts are at the bottom of the summary. https://github.com/secondlife/viewer/actions/runs/34837962086

naitro2010:
@akleshchev Thanks :)

I'll test with the release build on Windows now.

akleshchev:
Note that it's ReleaseOS, not Release, but the difference should be negligible for shaders.

naitro2010:
I just tested and it seems like avatar and skinned mesh rendering works on Windows without any issues.
I'm not sure how to profile for A/B testing though.
I think I did see a way to record and replay a connection in the Advanced or Developer menu but I'm not sure how to use it or if it will work for profiling the same scene multiple times.

akleshchev:
Basically you moved some repeating math from gpu to make it one-time on cpu?

Sadly shaders are not in my wheelhouse and I have no idea how to test skinning either.

akleshchev (approved):

akleshchev:
@Geenz Please take a look.

cosmic-linden (commented):
Just starting to review this.

As a general note, math calculations done in C++ in lieu of a shader should be done via glm when possible, but I'm not yet sure if that's applicable here.

Have you tried running `Develop > Render Tests > Frame Profile` and checking the log for objectSkinV run time? Intuitively, I would think this could make an impact assuming the code is correct (as avatars can have many vertices), but it could be interesting to know.

indra/newview/app_settings/shaders/class1/avatar/objectSkinV.glsl
```diff
@@ -24,49 +24,30 @@
...
-    int i;
 
-    vec4 w = fract(weight4);
     vec4 index = floor(weight4);
-
-    index = min(index, vec4(MAX_JOINTS_PER_MESH_OBJECT-1));
-    index = max(index, vec4( 0.0));
-
-    w *= 1.0/(w.x+w.y+w.z+w.w);
+    vec4 w   = fract(weight4)*2.0;
```
cosmic-linden, line 33: I don't really understand why this factor of two is multiplied here. Looks like a new factor of 0.5 is applied in `LLFace::getGeometryVolume` and `Primitive::upload` which this is trying to compensate for. See also `pbrmetallicroughnessV.glsl`, which is another shader which makes use of `weight4`.

cosmic-linden (changes requested):

indra/newview/app_settings/shaders/class1/avatar/objectSkinV.glsl
naitro2010: Thanks for reviewing my changes. :)

I think the scaling I added in primitive.cpp might be a bug.

weight4 in getGLTFTransform is using a different format for the actual weights.
objectSkinV uses the integer portion for the index and the fractional portion for the weight.
getGLTFTransform splits the two into joints and weights.
It is a bit confusing to have the same input used in two different shaders with different data formats.

Reverting the changes in primitive.cpp seemed to work with normal meshes.  I'm going to test with a GLTF PBR mesh next.

naitro2010: I just reverted the changes to primitive.cpp and everything seems to still work ok with a local build.
Please let me know if everything else looks good and when another ReleaseOS build is ready.
Thanks :)

naitro2010: Also, the factor of 0.5 is for floating point precision with weights close to 1.0 on the GPU in case it rounds up.  I don't know if it is needed on most GPUs though.

cosmic-linden: > in case it rounds up

Why would it round up? And why would that fix only be relevant to certain GPUs?

naitro2010: It could round up if the precision on the GPU is different than fp32 and the weight is really close to 1.0.
There might be some other cases where it could also have rounding errors accumulate.
If it works without the scaling factor on all CPUs and GPUs it probably isn't needed.
I have run into issues with floating point precision before which is why I added the scaling factor for the fractional portion.

It might be better to have the joints be integers like in pbrmetallicroughnessV.glsl instead of using the integer portion of the floating point weights.  It would also make it so the scaling is no longer potentially required.

I'll make two local builds and use the Frame Profiler to compare.

Thanks :)

cosmic-linden: If the GPU precision is lower than on CPU, I doubt a factor of two would fix it. Based on what I'm reading, floating point precision on the GPU can differ by multiple digits. And even if it only differed by one digit, I don't think your factor of two workaround is going to address that (best case, it will decrement the floating-point exponent and have no impact on precision at all). I think it would be best to not include that alleged workaround unless we have a bug report which it fixes.

PR #6265, opened 2026-09-05
Fix moving linksets lagging when alt-cammed

### LINKED ISSUE #6253: Child prims and avatars of a moving linkset can appear dislocated from the root prim if the camera is focused on the root.

Viscerous:
Stops camera focus calculation from advancing drawable movement outside the normal pipeline pass, and uses initialised drawable render transforms for child prims.

This fixes both root- and child-focused cases without manually cascading camera updates through the linkset and avatar hierarchy. Alternative implementation to #6234. Fixes #6253.

This fix was implemented on my fork initially and tries to remedy the issue at the root cause sites, but it could have different knock-ons than the alternative fix, so QA might want to see which behaves better.

### DIFF
--- a/indra/newview/llagentcamera.cpp
+++ b/indra/newview/llagentcamera.cpp
@@ -1628,33 +1628,8 @@ LLVector3d LLAgentCamera::calcFocusPositionTargetGlobal()
     {
         if (mFocusObject.notNull() && !mFocusObject->isDead() && mFocusObject->mDrawable.notNull())
         {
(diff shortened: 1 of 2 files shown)


akleshchev (approved):

akleshchev:
@Geenz If you have no issues with this, I'm going to merge. Camera should not be respondible for movement updates, especially not in such a conditional manner.

akleshchev:
Thank you, merged!

PR #6266, opened 2026-09-05
Smooth tracked cameras relative to their focus

### LINKED ISSUE #6253: Child prims and avatars of a moving linkset can appear dislocated from the root prim if the camera is focused on the root.

Viscerous:
Smooths tracked-object cameras in focus-relative space rather than global space. Moving-focus translation is applied directly while orbit and zoom changes remain smoothed, matching the existing avatar-relative camera path and preventing the focused object from drifting or wobbling in view.

Follow-up to #6265 and related to #6253. Note that this fixes remaining correctness/mild wobble issues remaining that neither #6234 or #6265 are designed to fix.


akleshchev:
Thank you, merging!

PR #6271, opened 2026-09-07
Harden LLUIColor traversal, LLButton state color fallbacks, and color table node stability

Shadowolf7:
## Description

Addresses a crash (`SIGSEGV`) in `LLButton::draw()` triggered when a button enters the `disabled + selected` state (such as when toggling "Edit linked" in the build floater, which disables active tab buttons or toggles). Inlined tail-call optimization in `LLUIColor::get()` turned unbounded recursion into an unchecked `while (mColorPtr != NULL)` loop, which dereferenced invalid or unmapped memory if `mColorPtr` was wild or pointed to an extracted node.

### Changes:
1. **Bounded Indirection in `LLUIColor::get()`**:
   - Replaced recursion with an iterative traversal bounded by `MAX_COLOR_DEPTH = 4`.
   - Prevents runaway pointer chasing, crashes from unmapped memory, and potential infinite loops from circular color definitions.
2. **Logical State Fallbacks in `LLButton`**:
   - In `LLButton::LLButton(const Params& p)`: Initialize `mSelectedLabelColor`, `mDisabledLabelColor`, and `mDisabledSelectedLabelColor` from their logical parents if not explicitly provided (`disabled_selected` -> `disabled` -> `unselected`).
   - In `LLButton::draw()`: When drawing `disabled + selected`, fall back to `mDisabledLabelColor.get()` if `mDisabledSelectedLabelColor.get()` resolves to `LLColor4::magenta` (unset / missing in skin).
3. **Preserve Node Stability in `LLUIColorTable::setColor()`**:
   - Removed node extraction (`mLoadedColors.extract()`) when storing user-customized colors.
   - Preserves permanent address stability for all nodes in `mLoadedColors`, preventing borrowed pointers held by existing widgets from dangling.

## Related Issues
- Resolves crashes when toggling button/control states that place widgets in rare state combinations (`disabled + selected`).

## Checklist
- [x] I have reviewed the contributing guidelines https://github.com/secondlife/viewer/blob/develop/CONTRIBUTING.md.

### DIFF
--- a/indra/llui/llbutton.cpp
+++ b/indra/llui/llbutton.cpp
@@ -212,19 +212,6 @@ LLButton::LLButton(const LLButton::Params& p)
         mSelectedLabel = mUnselectedLabel;
     }
 
-    if (!p.label_color_selected.isProvided())
-    {
-        mSelectedLabelColor = mUnselectedLabelColor;
-    }
-    if (!p.label_color_disabled.isProvided())
-    {
-        mDisabledLabelColor = mUnselectedLabelColor;
-    }
-    if (!p.label_color_disabled_selected.isProvided())
-    {
-        mDisabledSelectedLabelColor = mDisabledLabelColor;
-    }
-
     // Hack to make sure there is space for at least one character
     if (getRect().mRight >= 0 && getRect().getWidth() > 0 &&
         getRect().getWidth() - (mRightHPad + mLeftHPad) < mGLFont->getWidth(std::string(" ")))
@@ -847,10 +834,6 @@ void LLButton::draw()
         if ( getToggleState() )
         {
             label_color = mDisabledSelectedLabelColor.get();
-            if (label_color == LLColor4::magenta)
-            {
-                label_color = mDisabledLabelColor.get();
-            }
         }
         else
         {
--- a/indra/llui/llbutton.cpp
+++ b/indra/llui/llbutton.cpp
@@ -212,6 +212,19 @@ LLButton::LLButton(const LLButton::Params& p)
         mSelectedLabel = mUnselectedLabel;
     }
 
+    if (!p.label_color_selected.isProvided())
+    {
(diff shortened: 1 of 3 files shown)


indra/llui/llbutton.cpp
```diff
@@ -834,6 +847,10 @@ void LLButton::draw()
         if ( getToggleState() )
         {
             label_color = mDisabledSelectedLabelColor.get();
+            if (label_color == LLColor4::magenta)
```
Ansariel, line 850: I don't think relying on `LLColor4::magenta` being the color for undefined is a good idea.

Shadowolf7: Agreed. Removed the button constructor fallback cascading and the `LLColor4::magenta` check in `LLButton::draw()` so button state colors continue to rely on theme/skin definitions. Also restored node extraction in `LLUIColorTable::setColor()` to maintain pointer stability for existing `LLUIColor` handles.

akleshchev (approved):

indra/llui/lluicolor.cpp
```diff
@@ -58,7 +58,17 @@ void LLUIColor::set(const LLUIColor* color)
...
-    return (mColorPtr == NULL ? mColor : mColorPtr->get());
+    const LLUIColor* cur = this;
+    size_t depth = 0;
+    constexpr size_t MAX_COLOR_DEPTH = 4;
+
+    while (cur->mColorPtr != nullptr && depth < MAX_COLOR_DEPTH)
+    {
+        cur = cur->mColorPtr;
+        depth++;
+    }
```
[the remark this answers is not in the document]
akleshchev: @Shadowolf7 How deep can this actually be? Old code had no depth check.

The '4' might be a bit too cautiocious if it's the one from colors.xml. I don't expect things to go that deep, but to be on the safe side I would use something like '10'.

akleshchev (approved):

PR #6272, opened 2026-09-08
Fix SLPlugin error when it unloads libcef

### LINKED ISSUE #1015: SLPlugin - libcef.dll - Common crash in plugin infra probably during quit/TP/region crossings.
### Environment

Second Life Release 7.1.2.7215179142 (64bit)
Release Notes

You are at 159.9, 77.6, 47.4 in LindenWorld B located at simhost-019b8bea6b2cc58b3.agni
SLURL: http://maps.secondlife.com/secondlife/LindenWorld%20B/160/78/47
(global coordinates 237,984.0, 244,302.0, 47.4)
Second Life Server 2024-02-21.7995320426
Release Notes

CPU: 12th Gen Intel(R) Core(TM) i9-12900HK (2918.4 MHz)
Memory: 65208 MB
OS Version: Microsoft Windows 10/11 64-bit (Build 22000.2538)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 3050 Ti Laptop GPU/PCIe/SSE2

Windows Graphics Driver Version: 31.0.15.4609
OpenGL Version: 4.6.0 NVIDIA 546.09

Window size: 1506x1020
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 3000kbit/s
LOD factor: 1.5
Render quality: 4
Advanced Lighting Model: Enabled
Texture memory: 4095MB
Disk cache: Max size 3993.6 MB (23.1% used)

J2C Decoder Version: KDU v7.10.4
Audio Driver Version: FMOD Studio 2.02.13
Dullahan: 1.14.0.202310131404
  CEF: 118.4.1+g3dd6078+chromium-118.0.5993.54
  Chromium: 118.0.5993.54
LibVLC Version: 3.0.16
Voice Server Version: Vivox 4.10.0000.32327.5fc3fe7c.399bd0e

Packets Lost: 16/12,357 (0.1%)
March 18 2024 14:50:55

### Description

While diagnosing an unrelated problem in an SL and Firestorm environment, I noticed a bunch of crash reports on SLPlugin.exe coming out of dxdiag.  I wondered if these were common and I was immediately able to reproduce this on my dev system.  The crash gets reported via standard Windows Event Viewer and there will be a pair of events (Event IDs 1000/1001):

```
Faulting application name: SLPlugin.exe, version: 0.0.0.0, time stamp: 0x657b81b8
Faulting module name: libcef.dll_unloaded, version: 118.4.1.0, time stamp: 0x651cb3d9
Exception code: 0xc0000005
Fault offset: 0x00000000046433f0
Faulting process id: 0x1d6c
Faulting application start time: 0x01da797d35a55042
Faulting application path: C:\Program Files\SecondLifeViewer\SLPlugin.exe
Faulting module path: libcef.dll
Report Id: 86eda01a-dcc8-4278-8595-f3eeafee2090
Faulting package full name: 
Faulting package-relative application ID: 
```

The faulting offset is consistent over runs so at least this isn't a Heisenbug.  Both SL and FS show identical offsets for the same libcef build so I suspect this is a DLL unmap bug possibly in our use of CEF, possibly elsewhere.  But it will be reproducible.  [sp?]  

I don't expect this to be critical but one question is:  are we deliberately ignoring this in bugsplat?

### Reproduction steps

1. Windows SL session
2. Go to slcorps meeting area.
3. TP to Boston platform:  http://maps.secondlife.com/secondlife/LindenWorld%20B/160/78/47
4. Walk around
5. TP back to slcorps meeting area.
6. You should have some crashes in Event Viewer

Hecklezz:
Issue Fixed: https://github.com/secondlife/viewer/issues/1015

For years, the Windows Error reporter logs an SLPlugin.exe error regarding unloading libcef.dll, and on other platforms there is likely a similar issue happening but just silently ignored/handled.

Based off the information mentioned in the issues comment here- https://github.com/secondlife/viewer/issues/1015#issuecomment-5315373316 
I looked into it further and indeed the call to delete mInstance, which ends up calling apr_dso_unload() in the LLPluginInstance destructor, is the root cause of the fault.

However, their proposed fix of calling TerminateProcess where they suggested to, and only on Windows, is unnecessary and instead it would just be better to not call delete mInstance for all OS (as others may just silently be handling the fault), let the process gracefully exit, which then ends up calling exit(0) in the LLPluginProcessChild destructor instead due to a different issue noted about apr_dso_unload() here- https://github.com/secondlife/viewer/blob/3ac6222c27a24737b21d73a12337d39155ffc445/indra/llplugin/llpluginprocesschild.cpp#L57-L60
The OS will reclaim any memory used once it exits, and the memory footprint of not manually deleting mInstance when unloading is absolutely negligible and will only last a handful of frames at most before reclaimed by the OS after the process actually exits.

**NOTE- PLEASE READ/ANSWER:**
If instead you believe simply removing the call to apr_dso_unload() in the LLPluginInstance destructor altogether would be better, which then removes the necessary exit(0) mentioned above to fix the other deadlock problem, I can change the PR accordingly, keep the delete mInstance's, and remove the exit(0).

### DIFF
--- a/indra/llplugin/llpluginprocesschild.cpp
+++ b/indra/llplugin/llpluginprocesschild.cpp
@@ -241,8 +241,10 @@ void LLPluginProcessChild::idle(void)
 
         case STATE_UNLOADED:
             killSockets();
-            delete mInstance;
-            mInstance = NULL;
+            // Deleting mInstance here causes apr_dso_unload() to trigger a fault on SLPlugin when it unloads libcef
+            // Instead, let the process exit on its own and allow the OS reclaim the memory to prevent the error
+            //delete mInstance;
+            //mInstance = NULL;
             setState(STATE_DONE);
             break;
 


callumlinden (approved):
LGTM.

akleshchev:
Thank you, merging.

PR #6286, opened 2026-09-08
Fix possible crash getting energy attribute for keywords

### LINKED ISSUE #6285: Possible crash getting energy attribute for keywords
### Environment

Second Life Release 26.4.0.33894281402 (64bit)
<details><summary>Rest of environment-</summary>
Release Notes

CPU: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz (3600.01 MHz)
Memory: 16302 MB
OS Version: Microsoft Windows 10 64-bit (Build 19045.7663)
Graphics Card Vendor: NVIDIA Corporation
Graphics Card: NVIDIA GeForce RTX 2080/PCIe/SSE2

Windows Graphics Driver Version: 32.0.16.1088
OpenGL Version: 4.6.0 NVIDIA 610.88

Window size: 1920x1080
Font Size Adjustment: 96pt
UI Scaling: 1
Draw distance: 128m
Bandwidth: 10000kbit/s
LOD factor: 1.5
Render quality: 4
Texture memory: 8192MB
Disk cache: Max size 11468.8 MB (0.4% used)

J2C Decoder Version: KDU v8.4.1
Audio Driver Version: OpenAL, version 1.1 ALSOFT 1.24.2 / OpenAL Community / OpenAL Soft: OpenAL Soft
Dullahan: 1.26.0.202510161628
  CEF: 139.0.40+g465474a+chromium-139.0.7258.139
  Chromium: 139.0.7258.139
LibVLC Version: 3.0.21
Voice Server Version:  

September 08 2026 14:12:07
</details>

### Description

PR with possible fix: https://github.com/secondlife/viewer/pull/6286

There is a potential crash here- https://github.com/secondlife/viewer/blob/c37498bb4017e89654c0687a3f84792ecb445c47/indra/llui/llkeywords.cpp#L367-L371
std::stod will throw an exception if the string is empty, which based on the code, is a possible outcome and is supposed to default to 0.0 if it is empty. But right now instead, it will just crash before it can default the value.

Prior to 26.4, it didn't use std::stod at all so this is a new potential crash point. 

Note: I noticed this being a possible crash when logging into an OpenSim simulator on Firestorm with 26.4 merged in that didn't send an energy attribute.

### Reproduction steps

Unable to forcefully reproduce unless if you create a test sim where the energy attribute is empty/missing.

Hecklezz:
Issue Fixed: https://github.com/secondlife/viewer/issues/6285

This PR simply checks if the energy attribute is empty before calling std::stod on the value to prevent an exception being thrown on an empty string.

Additionally, added the missing break that was causing a [ES.78 warning](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es78-dont-rely-on-implicit-fallthrough-in-switch-statements).

**Note:** If you believe it would be best to just drop the empty and std::stod check entirely, which returns it to how it used to be before 26.4, just let me know and I'll change it.

### DIFF
--- a/indra/llui/llkeywords.cpp
+++ b/indra/llui/llkeywords.cpp
@@ -364,7 +364,7 @@ void LLKeywords::processTokensGroup(const LLSD& tokens, std::string_view group)
                         break;
                     case LLKeywordToken::TT_FUNCTION:
                         tooltip = getAttribute("return") + " " + outer_itr->first + "(" + getArguments(arguments) + ");";
-                        if (std::stod(getAttribute("energy")) >= 0)
+                        if (getAttribute("energy").empty() || std::stod(getAttribute("energy")) >= 0)
                         {
                             tooltip.append("\nEnergy: ");
                             tooltip.append(getAttribute("energy").empty() ? "0.0" : getAttribute("energy"));
@@ -373,6 +373,7 @@ void LLKeywords::processTokensGroup(const LLSD& tokens, std::string_view group)
                         {
                             tooltip += ", Sleep: " + getAttribute("sleep");
                         }
+                        break;
                     default:
                         break;
                 }


akleshchev (approved):

marchcat (approved):
Thank you!

PR #6296, opened 2026-09-10
Fix LLWorld SpaceTime updating

FelixWolf:
## Description
Another tiny fix.

When attempting to make a animation sync, I noticed that there is a bug with LLWorld's `getSpaceTimeUSec()` function.
Presently, SpaceTime is stored as a U64, while LLUnit is F32. Because of the way C++ is compiling the code, it converts the U64 to a F32 before doing the addition, and ends up storing a invalid number for SpaceTime.
We could increase LLUnit to a F64, but I'm unsure of issues that may cause, or if we would still lose precision eventually, so doing it this way would potentially the better option.

Looking around the code, I don't see anywhere that this is used, as far as I am aware, it is completely unused, but I'm fixing it because I need to use it for feature development, and also it'd be nice if someone else comes around needing to use it doesn't spend time debugging why `getSpaceTimeUSec()` doesn't update properly.

## Related Issues

- [ ] **Please link to a relevant GitHub issue for additional context.**
  - **Bug Fix:** Link to an issue that includes reproduction steps and testing guidance.
  - **Feature/Enhancement:** Link to an issue with a write-up, rationale, and requirements.

Issue Link: 

---

## Checklist

---

## Additional Notes

Not really a way to test this without a debug UI / logger, but the problem is it won't demonstrate the before and after.
I give my word that it was broken before hand, and this fixes it though.

### DIFF
--- a/indra/llmessage/CMakeLists.txt
+++ b/indra/llmessage/CMakeLists.txt
@@ -38,6 +38,7 @@ set(llmessage_SOURCE_FILES
     llioutil.cpp
     llmessagebuilder.cpp
     llmessageconfig.cpp
+    llmessagejsonbuilder.cpp
     llmessagereader.cpp
     llmessagetemplate.cpp
     llmessagetemplateparser.cpp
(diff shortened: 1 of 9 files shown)


FelixWolf:
Oops, I included debug code for testing stuff.

PR #6329, opened 2026-09-15
Retry HTTP 403 Forbidden texture fetches on region crossing

Shadowolf7:
### Problem
When an avatar transitions between regions or teleports, capability URLs and CDN authentication signatures change. If a texture fetch request was in-flight during the transition or presented with an expired capability token, AWS CloudFront / Agni asset CDNs return `HTTP 403 Forbidden`.

In `LLTextureFetchWorker::doWork`, retry logic exists for `http_not_found` (404) and `http_service_unavail` (503), but `HTTP_FORBIDDEN` (403) was omitted. Unhandled 403 responses fall through to `"abort: fail harder"`, causing `LLViewerFetchedTexture::processFetchResults` to call `setIsMissingAsset(true)`. This permanently blacklists the textures as missing for the remainder of the session, leaving surfaces transparent or black.

### Solution
1. In `indra/newview/lltexturefetch.cpp`, declare `http_forbidden(HTTP_FORBIDDEN)` and add an `else if (http_forbidden == mGetStatus)` block in `LLTextureFetchWorker::doWork`.
2. When a 403 occurs across region transitions (`!region || mLastRegionId != region->getRegionID()`) or on an initial attempt (`mRetryAttempt == 0`), clear `mUrl`, increment `mRetryAttempt`, reset `mLastRegionId`, and transition back to `INIT` to request a fresh capability URL from the current region.
3. In `indra/newview/llviewertexture.cpp`, guard `setIsMissingAsset()` in `processFetchResults()` to ensure transient HTTP 403 authorization failures do not permanently poison textures in the session cache.


akleshchev:
Isn't this basically https://github.com/secondlife/viewer/pull/6169 but from a different angle?

PR #5730, opened 2026-04-28
Fix picks pointing to the wrong location if the region moved

trish-sl:
This fixes #5519

A pick pointing to an old region will use the wrong global coordinate. 
You can test the fix using Dan Linden's "bug island" pick that is pointing to the old global coordinates.

This change makes it so the global coordinate used are the latest received instead of the potentially stale ones

PR #6356, opened 2026-09-21
Fix picks saving the wrong location

### LINKED ISSUE #6354: [26.4] Profile Picks Saving Incorrect Co-ordinates
1. Log in with 26.4 viewer (https://releasenotes.secondlife.com/viewer/26.4.0.34502396632.html)
2. Create a new Profile Pick (on a parcel that does not have a Landing Point set - just to rule that out)
3. Look at the coordinates of the Pick.

Observed: The coordinates says 128,128,0

Image

4. Log out. Clear cache
5. Log in with a 26.3 viewer 

Observed: The correct coordinate is show for that Pick. 

Expected: The 26.4 viewer should show the correct coordinates of the Pick.

----
Original report follows:
In profile picks, when creating a new pick or on set location for an existing pick the co-ordinates are now incorrect. Sometimes the X and Y are correct, sometimes they are incorrect. The Z co-ordinate shows as 0. The correct co-ordinates show initially, but when  selecting create pick, they change to incorrect ones. The worse example of incorrect co-ordinates can be found at this location https://maps.secondlife.com/secondlife/Fog%20Bank/106/209/3000

create a new pick here, correct co-ordinates shown, then save pick and it has always changed to 128, 128, 0. More information can be found in this SL Forums topic https://community.secondlife.com/forums/topic/534771-the-new-mystery-of-location-height/

https://secondlife.canny.io/admin/feedback/bug-reports/p/profile-picks-saving-incorrect-co-ordinates

Hecklezz:
Issue Fixed: https://github.com/secondlife/viewer/issues/6354

The fix from this PR- https://github.com/secondlife/viewer/pull/5730 fixed an issue where picks contains the wrong global data, so to try fix it, it set the global position to the parcels data.
However, this causes a different issue when you save new picks it overwrites the location you want to save to the parcels origin location.

This PR changes the previous fix to check if the region handle has changed first before overwriting the global position, and if it has changed, it calculates the new origin position and sets the global position to that instead. If the region handles match, it doesn't update the global position.

**Before merging:** Someone please double check my changes is the appropriate fix and something else wouldn't be better. Also confirm on your end the previous fix still works (test Dan's Bug island Pick), and the new issue is fixed (saving picks use the correct location).


trish-sl:
Looks good to me, original bug is fixed and this solves the reported side effect

akleshchev (approved):

akleshchev:
`old_region_handle != new_region_handle`

Don't see a better way to handle this. Not unless landmarks or server side change.

PR #6358, opened 2026-09-21
Threaded udp packet ring

akleshchev:
Draft for now (pending tracy investigation). Move packet ring into a separate thread to not hold render thread.


akleshchev:
@AndrewMeadows Does this make sense to you?

indra/llmessage/lludpreceiverthread.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "lludpreceiverthread.h"
+#include "llproxy.h"
+#include "net.h"
+
+LLUDPReceiverThread::LLUDPReceiverThread(S32 hSocket, std::shared_ptr<PacketQueue> queue)
+:   LLThread("UDP Receiver"),
+    mSocket(hSocket),
+    mQueue(std::move(queue))
+{
+}
```
AndrewMeadows: The `std::move()` works but is not helpful to the reader.  When I see something like that I think: "Either a relatively big container is being passed or this is called in a hot-spot and someone has a good reason to optimize," which is why I was reading the code as if `queue` wasn't a pointer but an actual container by reference and if that were true... then the code  wouldn't work.  Eventually I realized: "Oh `queue` is actually just a smart pointer and... the `std::move()` is just being used as a fancy copy.  Since the ctor is probably only called once in the entire lifetime of the process I think a `std::move()` here is overkill and misleading.

Not a blocker.  Just me having a ["wtf?!" moment](https://muhammad-rahmatullah.medium.com/wtf-per-minute-an-actual-measurement-for-code-quality-780914bf9d4b).

AndrewMeadows (approved):
This looks like it would work, and it is nice to get the actual packet collection off of the main thread --> it probably helps.  However, it would be even better to move all of `LLMessageSystem::drainUdpSocket()` work (with the double-queue prioritization sort) onto the new thread.  That would remove one whole buffering step which is happening on the main thread.

indra/llmessage/lludpreceiverthread.cpp
akleshchev: I will add comment there.

akleshchev:
> LLMessageSystem::drainUdpSocket()

Will check that, thank you. I'm treating this as a first step. And will try to move other smallest parts of network handling into this new thread one step at a time as tackling it whole is not a reasonable approach.

akleshchev:
> it would be even better to move all of LLMessageSystem::drainUdpSocket()

Done, it runs, but not profiled yet.

indra/newview/llappviewer.cpp
```diff
@@ -6064,59 +6064,47 @@ void LLAppViewer::idleNetwork()
...
+                size_t num_buffered = gMessageSystem->getNumDecodedPending();
+                if (num_buffered > (size_t)total_decoded)
                 {
-                    // Grow CheckMessagesMaxTime until we process more packets each frame than arrive.
-                    // This might spiral out of control on very slow computers on fast networks when
-                    // the bandwidth settings are too high.  There is a mechanism for providing backpressure
-                    // to network bandwidth but it may be inadequate for the task
-                    // (see LLViewerThrottle::updateDynamicThrottle() for more details).
-                    CheckMessagesMaxTime *= 1.035f; // 3.5% ~= 2x in 20 frames, ~8x in 60 frames
+                    CheckMessagesMaxTime *= 1.035f;
```
AndrewMeadows: I'd prefer to keep most of that lost comment.

